PR #2, opened 2016-03-16
Auto Update

jmprusi:
* Added Auto update feature: Looks for changes in the config and deploys it.. 
* Added Signal support: USR1-> reload openresty, USR2-> Forces download of config and redeploy

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,35 +1,40 @@
-#!/bin/sh -u
+#!/usr/local/bin/dumb-init /bin/sh
 # 3scale (operations@3scale.net)
+set -u
 
-mkdir /opt/openresty/nginx/lua/
+if [ ${AUTO_UPDATE_INTERVAL} != 0 ] && [ ${AUTO_UPDATE_INTERVAL} -lt 60 ]; then
+	echo "AUTO_UPDATE_INTERVAL should be 60 or greater"
+	exit 1
+fi
 
 reload_openresty() {
   echo "Reloading Openresty"
-  pkill -USR1 -o nginx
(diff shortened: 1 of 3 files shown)


entrypoint.sh
```diff
@@ -1,19 +1,71 @@
...
+reload_openresty() {
+  echo "Reloading Openresty"
+  pkill -USR1 -o nginx
+}
 
-mkdir /opt/openresty/nginx/lua/
+download_threescale_config() {
+  echo "Downloading threescale configuration, using endpoint: ${THREESCALE_ENDPOINT}"
+  cd /tmp
+  rm -rf /tmp/${THREESCALE_PROVIDER_KEY}/
```
mikz: Imo this is task for `mktemp`. Also, provider key should be treated at least a bit securely, so it should not be part of any path.

mikz:
I'd:

* extract the autoupdate to different .sh file
* use https://github.com/Yelp/dumb-init#session-behavior to start the auto update in background, but nginx in foreground
* I'd clean it up a bit, try to remove some duplication etc.

but great job!

entrypoint.sh
andrewdavidmackenzie: Tokens FTW :-)

Let's see what needed for us to release use of tokens to all...

Andrew

jmprusi: Yes, let me clean that part a little bit!

mikz:
I think it is good for now. We definitely could improve and polish few things (extracting the auto update to different file, nginx in foreground, ...), but it works and we should get it into hands of people.

jmprusi:
great!

PR #9, opened 2016-05-17
replaced alpine docker image with centos

jmprusi:


andrewdavidmackenzie:
When we merge, we should have a couple of people test tutorial from scratch.

jmprusi:
I tested from scratch.. I will ping @unleashed to test it.

Issue #8, opened 2016-05-17
Make placeholder in template match convention in Support site


andrewdavidmackenzie:
Make the THREESCALE_ADMIN_URL placeholder that appears in the UI (in the new app form from the template) consistent with the support site "standard" (i.e https://MYDOMAIN-admin.3scale.net/).

PR #10, opened 2016-05-17
Make the THREESCALE_ADMIN_URL placeholder that appears in the UI cons…

jmprusi:
…istent with the support site


jmprusi:
resolve #8

PR #14, opened 2016-07-14
fixing DNS issues in openshift

jmprusi:
- Detect if we are in openshift, use default route
- If we are not in openshift, use /etc/resolv.conf
- Let user set the resolver


mikz:
Would be cool to link some knowledge base, but otherwise 👍

PR #15, opened 2016-07-15
Jmprusi/fix dns

jmprusi:


jmprusi:
Sorry about merging it without review. Need the patch.

PR #24, opened 2016-08-30
[WIP] Data driven configuration

mikz:
Drops code generation and downloading code in favour
of having all the code configured by JSON configuration.


mikz:
Made this PR to point to [`v2` branch](https://github.com/3scale/docker-gateway/tree/v2). Will merge it and then issue PRs pointing to that branch.

nginx.conf
```diff
@@ -0,0 +1,130 @@
...
+      set $backend_endpoint 'http://127.0.0.1:3001';
+
+      proxy_pass_request_headers off;
+      proxy_pass $backend_endpoint/transactions/authrep.xml?$backend_authentication_type=$backend_authentication_value&service_id=$service_id&$usage&$credentials; # &log%5Bcode%5D=$arg_code&log%5Brequest%5D=$arg_req&log%5Bresponse%5D=$arg_resp;
+      proxy_set_header  Host  "$backend_host";
+      proxy_set_header  X-3scale-User-Agent "nginx$deployment";
+      proxy_set_header  X-3scale-Version "2016-08-29T08:09:01Z";
+
+      log_by_lua_block {
+        ngx.log(ngx.INFO, '[authrep] ' .. ngx.var.request_uri .. ' ' .. ngx.var.status)
```
mpguerra: nice one!

nginx.conf
```diff
@@ -0,0 +1,130 @@
...
+        local configuration = require('configuration')
+        local provider = require('provider')
+
+        local config = configuration.boot()
+
+        provider.init(config)
+        provider.access()
+      }
+
+      body_filter_by_lua_block {
```
mpguerra: Is there some way to make these optional so we're not always logging requests? For some clients this can increase their latency quite a bit and is a common bit of code that we have had to rip out in the past

src/provider.lua
```diff
@@ -0,0 +1,306 @@
...
+local http = {
+  get = function(url)
+    ngx.log(ngx.INFO, '[http] requesting ' .. url)
+    local res = ngx.location.capture(assert(url), { share_all_vars = true })
+
+    ngx.log(ngx.INFO, '[http] status: ' .. tostring(res.status))
+    return res
+  end
+}
+local function oauth(params, service)
```
mpguerra: We should make this more like authrep function, they should work the same apart from endpoint.

Other alternative would be to pass in the auth strategy to the function and use that to choose the right endpoint to target

src/provider.lua
```diff
@@ -0,0 +1,306 @@
...
+    ngx.var.cached_key = table.concat({service.id, params.user_key}, ':')
+
+  elseif backend_version == '2' then
+    params.app_id = parameters[credentials.app_id]
+    params.app_key = parameters[credentials.app_key] -- or ""  -- Uncoment the first part if you want to allow not passing app_key
+
+    ngx.var.cached_key = table.concat({service.id, params.app_id, params.app_key}, ':')
+
+  elseif backend_version == 'oauth' then
+    error('oauth unsupported')
```
mpguerra: 😮  I guess this is coming in a later revision... ;)

travis/install-openresty.sh
```diff
@@ -0,0 +1,19 @@
...
+
+PREFIX=${1:-$HOME/openresty}
+
+if [ ! -d "${PREFIX}/bin" ]; then
+  cd /tmp/
+  wget -T 60 -q -c http://openresty.org/download/${OPENRESTY_VERSION}.tar.gz
+  tar -xzf ${OPENRESTY_VERSION}.tar.gz
+  rm -rf ${OPENRESTY_VERSION}.tar.gz
+  cd ${OPENRESTY_VERSION}
+  ./configure --prefix="${PREFIX}" --with-luajit-xcflags=-DLUAJIT_ENABLE_LUA52COMPAT --with-debug
```
mpguerra: what happens if a customer needs other modules to be compiled? can this be modified to add additional modules?

t/001-management.t
```diff
@@ -0,0 +1,29 @@
...
+repeat_each(2);
+no_root_location();
+run_tests();
+
+__DATA__
+
+=== TEST 1: management
+This is just a simple demonstration of the
+echo directive provided by ngx_http_echo_module.
+--- http_config eval: $::HttpConfig
```
mpguerra: Would be good to explain what is happening here during a training, specifically with the eval filter

nginx.conf
mikz: but it does not really work :) will need a test to prove that

nginx.conf
mikz: Yep. Will make it optional. Via some flag set before in the access block.

src/provider.lua
mikz: Yeah, there are plenty of improvements. So far it was just a copy from the generated configuration.

travis/install-openresty.sh
mikz: Then they can use dynamic modules or compile own openresty. This script is for installing it on travis. 
We will ship and support just the docker image. So if they would want custom modules they should fork the image and use theirs. Or compile dynamic module and inject it in.

t/001-management.t
mikz: Some info on the eval in the book: https://openresty.gitbooks.io/programming-openresty/content/testing/preparing-tests.html
Or in the documentation http://search.cpan.org/dist/Test-Nginx/lib/Test/Nginx/Socket.pm#Filters

It just uses that variable.

PR #28, opened 2016-09-04
[docker] cleanup & test

mikz:
* local install of luarocks
* test nginx config
* compose to build & run it
* remove auto update & config downloading

points to https://github.com/3scale/docker-gateway/pull/27

maybe we should try https://circleci.com/docs/docker/#caching-docker-layers to cache the image


jmprusi:
👍

PR #31, opened 2016-09-05
[management] basic stub of management API

mikz:
offers:

* liveness probe
* readyness probe
* config dump

It will be accessible on port 8090.


jmprusi:
<3

PR #34, opened 2016-09-05
[s2i] use source to image

mikz:
- [x] use s2i
- [ ] download s2i on travis

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ dependencies:
 	luarocks make --local rockspec
 
 build:
-	$(S2I) build . openresty-builder docker-gateway-test
+	$(S2I) build . quay.io/3scale/openresty-builder docker-gateway-test
 
 bash:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -18,12 +19,12 @@ dependencies:
 	luarocks make --local rockspec
 
 build:
-	$(DOCKER_COMPOSE) build
+	$(S2I) build . openresty-builder docker-gateway-test
```
jmprusi: should we download the openresty-builder and build it locally or depend on the quay.io/3scale/openresty-builder:SOME_VERSION ? 

I would go for the second option.

mikz: I'd much rather depend on some published version. At least for now it would make things much easier.

jmprusi: Ok, we can publish the openresty-builder manually.. then look for a way to automate it.. just to unblock this.

mikz: I'll automate it if I can find 30 minutes between meetings and a pub later.

jmprusi:
👍

PR #39, opened 2016-09-07
[readme] improve readme for osx development

mikz:
just use APItools luarocks formula
and mention use of Test::Nginx

/cc @jcechace


jmprusi:
👍

PR #40, opened 2016-09-07
Fix for permissions denied when creating /opt/app/client_body_temp

jmprusi:


mikz:
Looks like this is not necessary now.

jmprusi:
@mikz I see, this was necessary before,  because of the incorrect chown 1001:1001.. 
closing this.

PR #46, opened 2016-09-07
entrypoint was failing on nslookup failure, this fixes the issue by shelling out the var

jmprusi:
nslookup call was crashing the entrypoint,
so shell out to prevent the crash and test it works with invalid dns


mikz:
👍

PR #49, opened 2016-09-07
Fix for incorrect http.d import causing resolver.conf not to be imported

jmprusi:
Nginx.conf is not including the contents of the http.d because of bad path


mikz:
This fails when trying to make request? I noticed yesterday, that the config is wrong. But all the docker tests are passing. So I'd like to have a test that this works. Maybe adding it to one of the probes?

PR #53, opened 2016-09-08
load configuration at boot time

mikz:
that should make readiness probe to mark instance as live
without requiring to make any request to it


jmprusi:
👍

PR #55, opened 2016-09-09
fallback to curl when internal lua client does not work

mikz:
which can happen on badly configured openshift (almost all of them)
as they have dns server that is not reachable and nginx bails out when resolving

https://github.com/openshift/origin/issues/3820


jmprusi:
:+1:

PR #57, opened 2016-09-09
fix Host header passed to the api backend

mikz:
* do not use empty host override
* extract host from the api backend


jmprusi:
👍

PR #71, opened 2016-09-13
experimental path based routing

mikz:
implements #22

* try to match the service host and mapping rules to determine which service to use

@tnscorcoran after merging this you'll be able to start the gateway with `APICAST_PATH_ROUTING_ENABLED` environment variable set to whatever to enable this feature.

It will allow you to have two services with the same host, but still route to the one that matches the rules.

So if you have services A and B both with host 'api.example.com' where
A matches GET /a
and B matches GET /b

you'll be able to make requests:

```
GET /a
Host: api.example.com
```

```
GET /b
Host: api.example.com
```

just fine and they'll be routed to the appropriate service.

/cc @vramosp correct me if I'm not explaining it clearly

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -7,6 +7,7 @@ env THREESCALE_DEPLOYMENT_ENV;
 env THREESCALE_PORTAL_ENDPOINT;
 env THREESCALE_CONFIG_FILE;
 env APICAST_CUSTOM_CONFIG;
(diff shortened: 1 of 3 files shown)


vramosp:
👍 

cc @VinayBhalerao

VinayBhalerao:
Forget to comment here earlier. Tested and looks good to me.

Issue #69, opened 2016-09-13
/dev/std* not accessible by non-root on docker < 1.9.0


jayenashar:
nginx fails to read the conf file with `access_log` set to `/dev/stdout`
##### Steps To Reproduce
1. install centos 6
2. install docker (1.7 is the latest supported by the old kernel)
3. create a container using `:latest`
##### Current Result

nginx starts but does not forward any requests to the private URL
##### Expected Result

nginx forwards requests
##### Additional Information
- command to run docker: `docker run --detach --publish=127.0.0.1:319:8080 --name=3scale --env=THREESCALE_PROVIDER_KEY=$THREESCALE_PROVIDER_KEY --env=THREESCALE_ADMIN_URL=https://tripgo-admin.3scale.net quay.io/3scale/gateway`
- relevant nginx output: `2016/09/13 05:48:17 [emerg] 7#0: open() "/dev/stdout" failed (13: Permission denied)`
##### Workaround

maybe this one in /entrypoint.sh: https://github.com/docker/docker/issues/6880#issuecomment-220637337 then point nginx to log to `/tmp/logpipe`

mikz:
Not sure how we resolve that in V2 because we dropped dumb-init and want to run just nginx.

jmprusi:
We can fix this for the master branch only, I will take a look at the possible workarounds.

jayenashar:
I don't totally understand how your docker setup is/will-be, but I see https://github.com/3scale/docker-gateway/blob/v2/entrypoint.sh is still around.  Are you dropping that as well?

mikz:
All you can see in v2 branch will ne there. Entrypoint for sure.

jayenashar:
can the workaround go in entrypoint.sh?

mikz:
The problem is we want want another process like cat reading from fifo because that can die and will not be respawn and will become zombie and nginx logging to fifo can mean performance hit because if the fifo is not consumed fast enough it will stop reviewing traffic.
So we could just log to a file but then we would need log rotation. Really don't know what is the best option but don't see anything obvious.

jayenashar:
while i agree that `cat` dying would cause problems, would `cat` die in practice?

is patching nginx an option?  a special `access_log` value that means stdout?

mikz:
I have no idea if cat would die in practice and I don't really want to find out.
But, considering the options. If we would opt in for some workaround solution only when we detect /dev/std* is not writeable.
[...]

PR #74, opened 2016-09-14
removed error_log output to stderr because docker bug in old releases

jmprusi:
related to https://github.com/3scale/docker-gateway/issues/69

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -83,7 +83,7 @@ NAMESERVER=$(pick_dns_server)
 export RESOLVER=${RESOLVER:-${NAMESERVER}}
 
 sed -E -i "s/listen\s+80;/listen 8080;/g" /opt/openresty/nginx/conf/nginx.conf
-nginx -g "daemon off; error_log stderr info;" &
+nginx -g "daemon off;" &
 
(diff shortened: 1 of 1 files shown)


mikz:
I'd not say it closes the issue, because it will still remain in v2. 
And there, to fix it properly we should try to write to the logs and if it does not succeed symlink them to dev null.

jmprusi:
Do you plan on support old docker version in v2?

mikz:
I don't know yet. Something to decide.

jayenashar:
sorry i don't think this is related to #69.  #69 is about using `/dev/stdout` as a file, not about using stderr as a file descriptor.  error_log with stderr works fine.  it's access_log (which doesn't support stdout/stderr) that doesn't work because `/dev/stdout` gives `permission denied`

jmprusi:
@jayenashar I see the issue, we removed the access_log.. but we didn't update the docker image, sorry about that.

Can you try building the image from this repo, and running it? 

I will update the docker image in the quay repo with the latest code.

jmprusi:
I will close this PR, and open another adding access_log again but with the simple fallback method.

PR #76, opened 2016-09-15
Added fallback in case /dev/stdout is not writable, due old docker version

jmprusi:
This is only for master branch not v2.

related to #69

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -4,14 +4,12 @@ set -u
 
 #Old docker version have a bug that prevents a non-root user to write
 #directly to /dev/stdout. (https://github.com/docker/docker/issues/6880) 
-export STDOUT_WRITABLE
-
 if [ -w "/dev/stdout" ]; then
   echo "/dev/stdout is writable, enabling access_log output"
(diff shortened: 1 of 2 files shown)


jayenashar:
quay.io/3scale/gateway seems to be working for me!  thanks!

entrypoint.sh
```diff
@@ -38,8 +51,12 @@ download_threescale_config() {
   unzip nginx.zip
   # Most docker PaaS doesn't allow docker to run as root
   # lets use the 8080 port instead of 80
+  if [ $STDOUT_WRITABLE = "true" ]; then
+    sed -E -i "/server\s+\{/a access_log /dev/stdout combined;" nginx_*.conf
```
jmprusi: By default there's no ```access_log``` directive... 

Ok I will change the code to do the following:

If /dev/stdout is writable  -> symlink log/access.log to /dev/stdout
If /dev/stdout is no writable -> symlink log/access.log to /dev/null
And always add the access_log directive to the log pointing to log/access.log 

This should be a more "consistent" solution.

PR #81, opened 2016-09-22
run prove inside build docker

mikz:
so we can test that it works the same inside

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,10 +43,11 @@ install:
 before_script:
 - mkdir -p $(dirname ${TEST_NGINX_ERROR_LOG})
 script:
-- make test
-- make test-nginx
-- make test-docker
+- make busted
+- make prove
 - make build
+- make test-docker
+- make prove-docker
 after_success:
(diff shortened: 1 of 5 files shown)


mikz:
Tests are failing because for this query:

```perl
 "\x{88}3\x{01}\x{00}\x{00}\x{01}\x{00}\x{00}\x{00}\x{00}\x{00}\x{00}\x{09}localhost\x{00}\x{00}\x{01}\x{00}\x{01}"
```

dns responds with:
```perl
"\x{88}3\x{81}\x{80}\x{00}\x{01}\x{00}\x{01}\x{00}\x{00}\x{00}\x{00}\x{09}localhost\x{00}\x{00}\x{01}\x{00}\x{01}\x{09}localhost\x{00}\x{00}\x{01}\x{00}\x{01}\x{00}\x{00}\x{00}\x{01}\x{00}\x{04}\x{00}"
```

and nginx considers it invalid

mikz:
It is kind of ridiculous we have to require nginx with ipv6 just to be able to disable it.

PR #84, opened 2016-09-26
Create build.md

mikz:
* document components
* document build process
* document continuous release

### DIFF
--- a/doc/build.md
+++ b/doc/build.md
@@ -21,6 +21,7 @@ Lua Dependencies are defined in [`apicast-VERSION.rockspec`](https://github.com/
 ## Components
 
 APIcast is using [source-to-image](https://github.com/openshift/source-to-image) to build the final Docker image.
+You'll need to have source-to-image installed and available on your system.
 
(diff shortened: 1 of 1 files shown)


doc/build.md
```diff
@@ -0,0 +1,43 @@
...
+
+Lua Dependencies are defined in [`apicast-VERSION.rockspec`](https://github.com/3scale/docker-gateway/blob/50daf279b3cf2da80b20ad473ec820d7a364b688/apicast-0.1-0.rockspec) file.
+
+* `lua-resty-http` cosocket based http client to be used instead of the `ngx.location.capture`
+* `inspect` library to pretty print data structures
+* `router` used to implement internal APIs
+
+## Components
+
+APIcast is using [source-to-image](https://github.com/openshift/source-to-image) to build the final Docker image.
```
jmprusi: Perhaps we should point that source-to-image should be installed before, as is a hard req or even provide a Make task for installing it..

mikz: Good point!

PR #89, opened 2016-10-03
[travis] fix image push after build

mikz:
POSIX compliant condition


unleashed:
Complainant?

unleashed:
Also, I don't think `[` is POSIX compliant, you should use `test`.

mikz:
Yeah. Typo, fixed :)

The == is bash only and this works in zsh too.

PR #107, opened 2016-10-19
balancer_by_lua

### LINKED ISSUE #90: Use balancer_by_lua to use upstream with keepalive
Using `proxy_pass $var` will not use upstream and keep-alive connection.

Either we have to use cosockets and lua-resty-http to stream API request and responses, or use balancer_by_lua in upstream.

mikz:
closes #90

introduces `lua-resty-balancer` that will later be extracted into own package.

it is a library for implementing load balancers

# TODO

- [x] benchmark ([gist](https://gist.github.com/mikz/bafc0e25e3809cef2761b70a1df3d67a) with flamegraphs) 
- [x] implement 2 balancing algorithms (round-robin, random)

### DIFF
--- a/apicast/conf.d/apicast.conf
+++ b/apicast/conf.d/apicast.conf
@@ -60,9 +60,10 @@ location / {
     local configuration = require('configuration')
     local provider = require('provider')
 
-    local config = configuration.boot()
-
-    provider.init(config)
+    if not provider.configured then
+      local config = configuration.boot()
+      provider.init(config)
+    end
   }
   access_by_lua_block {
     local provider = require('provider')
--- a/apicast/conf.d/apicast.conf
+++ b/apicast/conf.d/apicast.conf
@@ -4,10 +4,12 @@ location = /threescale_authrep {
   internal;
 
   proxy_pass_request_headers off;
+  proxy_http_version 1.1;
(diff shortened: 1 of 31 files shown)


apicast/conf.d/apicast.conf
```diff
@@ -4,10 +4,12 @@ location = /threescale_authrep {
   internal;
 
   proxy_pass_request_headers off;
+  proxy_http_version 1.1;
   proxy_pass $backend_endpoint/transactions/authrep.xml?$backend_authentication_type=$backend_authentication_value&service_id=$service_id&$usage&$credentials; # &log%5Bcode%5D=$arg_code&log%5Brequest%5D=$arg_req&log%5Bresponse%5D=$arg_resp;
   proxy_set_header  Host  "$backend_host";
   proxy_set_header  X-3scale-User-Agent "nginx$deployment";
   proxy_set_header  X-3scale-Version "$version";
+  proxy_set_header  Connection "";
```
hallelujah: Ah right it is keep-alive :)

apicast/conf.d/apicast.conf
```diff
@@ -43,19 +52,22 @@ location / {
...
-    provider.init(config)
+    if not provider.configured then
+      local config = configuration.boot()
+      provider.init(config)
+    end
   }
   access_by_lua_block {
     local provider = require('provider')
-    provider.call()
+    provider.call()()
```
hallelujah: That is quite ugly but ok as it returns a function.

mikz: Yep. I plan to change this in the future. It is pretty mind boggling :)

PR #123, opened 2016-11-08
Avoid unnecessary calls to require() in module.lua

### LINKED ISSUE #122: Performance hit in module.lua because of unnecessary calls to require()
I'm writing a module for Apicast v2. I used `stapxx` to profile it, and realized that there was something wrong. Around 75-80% of the time was consumed on the method `prequire()` of `module.lua`.

My module defines `rewrite()`, `access()`, etc. instead of `call()`. The problem is that in the `load()` method of `module.lua` there's this:

```lua
  local files = {
    [ name .. '.' .. phase ] = 'call', -- like apicast.init exposing .call
    [ name ] = phase -- like apicast exposing .init
  }

  for file, method in pairs(files) do
    local ok, ret = prequire(file)
```

and `prequire()` only caches a result when it's successful. In other words, as I haven't defined `call()` in my module, `prequire()` fails, and it does not cache that it failed, so it retries in every single request. This is a huge performance hit. When I fixed this, I noticed that the time spent `prequire()` went down from 75-80% to almost now.

davidor:
Closes #122

### DIFF
--- a/apicast/src/module.lua
+++ b/apicast/src/module.lua
@@ -56,6 +56,15 @@ local prequire = function(file)
 
   if ok then
     cache[file] = mod
+  else
+    -- We need to cache that we tried requiring and failed so we do not try
+    -- again. If we don't, we'll call require() on each request. This greatly
+    -- affects performance.
+    -- We can't set cache[file] to nil. That's the same as saying that it is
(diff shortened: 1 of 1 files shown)


mikz:
Looks good! At some point I'd like to revisit the log levels so those messages would not be shown in the default configuration.

PR #125, opened 2016-11-09
Sort directives by execution order in apicast.conf

davidor:
The content phase is executed before the log one, but in the config `log_by_lua` appeared before `content_by_lua`. There are some operations performed to calculate run times and to understand them it's necessary to know the order of execution of the phases. I think that sorting them in the config helps.


mikz:
👍 

I was thinking about renaming the content phase to `post_action`. So it is more obvious. Wdyt?

davidor:
I agree. I'll change that.

mikz:
The `module.lua` allows explicit phase to be called. So maybe just use that?

PR #131, opened 2016-11-10
auto update interval

mikz:
set either `AUTO_UPDATE_INTERVAL` or `-i N` when starting APIcast and see it update every N seconds!

No idea how to test this tough.

### DIFF
--- a/apicast/conf/nginx.conf
+++ b/apicast/conf/nginx.conf
@@ -16,6 +16,7 @@ env APICAST_MODULE;
 env APICAST_MISSING_CONFIGURATION;
 env APICAST_REQUEST_LOGS;
 env APICAST_RESPONSE_CODES;
+env AUTO_UPDATE_INTERVAL;
 
 # error_log stderr notice;
 # error_log logs/error.log warn;
--- a/apicast/src/apicast.lua
+++ b/apicast/src/apicast.lua
@@ -2,6 +2,7 @@ local configuration = require('configuration')
 local provider = require('provider')
 local balancer = require('balancer')
 local util = require('util')
+local pcall = pcall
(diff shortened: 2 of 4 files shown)


apicast/conf/nginx.conf
```diff
@@ -14,6 +14,8 @@ env REDIS_PORT;
 env RESOLVER;
 env APICAST_MODULE;
 env APICAST_MISSING_CONFIGURATION;
+env AUTO_UPDATE_INTERVAL;
+env PATH;
```
davidor: Where is this used? It's not related to the auto update feature, is it?

apicast/src/apicast.lua
```diff
@@ -30,6 +31,41 @@ function _M.init()
   end
 end
 
+function _M.init_worker()
+  local interval = tonumber(os.getenv('AUTO_UPDATE_INTERVAL'), 10)
```
davidor: I'd extract that 10 to a constant: `DEFAULT_UPDATE_INTERVAL`. Also, we are using 10 as the default interval here and 60 in the `.env` file. I'd rather see the same value in both places for consistency.

mikz:
This is not a default value, but integer base.

apicast/src/apicast.lua
mikz: This is not a default value, but integer base.

davidor: Right.
Do we need to specify that 10? It's the default value and what everyone would expect. Am I missing something?

mikz: For example in javascript it does not work like that. So honestly I don't know what to expect :)

PR #133, opened 2016-11-10
Add generic plugin to avoid duplicating code in custom plugins

### LINKED ISSUE #132: Provide a generic plugin to avoid code duplication across custom plugins
To add a custom plugin, we only need to define a lua module that implements a few methods like rewrite(), access(), etc. that are called in the appropriate nginx phases.

The problem is that as it is defined right now, the plugin system provided by Apicast forces custom plugins to define all those methods, even if they do not have anything to do with the functionality that the custom plugin adds. We should implement some changes so methods that are not likely to change from plugin to plugin do not need to be duplicated across all the custom plugins.

davidor:
Closes #132

@mikz Let me know if you think this is a good idea. I'll add tests for the new generic plugin in this PR later.
Also, for now, I've just extracted init(), rewrite(), and balancer(). I think that at least parts of access() should be extracted too, but there's a lot of logic in there. That needs a more careful analysis.

### DIFF
--- a/apicast/src/plugin.lua
+++ b/apicast/src/plugin.lua
@@ -1,0 +1,57 @@
+-- Apicast provides a plugin system. To add a custom plugin, we only need to
+-- define a module that implements a few methods like rewrite(), access(), etc.
+-- that are called in the appropriate nginx phases.
+-- The goal of this lua module is to implement the methods that are not likely
+-- to be changed from plugin to plugin so we do not have to repeated code
+-- across all of them.
+
+local balancer = require 'balancer'
+
+-- Plugins need access to provider and configuration, so we need to expose them
+local _M = { provider = require 'provider',
+             configuration = require 'configuration' }
+
+local missing_configuration = os.getenv('APICAST_MISSING_CONFIGURATION') or 'log'
+
+local function handle_missing_configuration(err)
+  if missing_configuration == 'log' then
+    ngx.log(ngx.ERR, 'failed to load configuration, continuing: ', err)
+  elseif missing_configuration == 'exit' then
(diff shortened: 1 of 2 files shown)


apicast/src/plugin.lua
```diff
@@ -0,0 +1,57 @@
...
+-- define a module that implements a few methods like rewrite(), access(), etc.
+-- that are called in the appropriate nginx phases.
+-- The goal of this lua module is to implement the methods that are not likely
+-- to be changed from plugin to plugin so we do not have to repeated code
+-- across all of them.
+
+local balancer = require 'balancer'
+
+-- Plugins need access to provider and configuration, so we need to expose them
+local _M = { provider = require 'provider',
```
mikz: I don't know. I think both provider and configuration are badly defined and scoped.
Exposing them makes me feel dirty. The way provider stores configuration is specific to that and I'd not like much to expose it.

mikz: Wouldn't it just make more sense for xc to inherit from apicast itself?

davidor: `configuration` does not need to be exposed. I realized it wasn't needed, but forgot to delete that part.

Regarding inheriting from Apicast itself, the problem is that xc does not need post_action(), body_filter(), nor header_filter(). Also, there are parts of access() that should be common to both of them, but not all.

mikz: Then post_action, body_filter and header_filter can be set to nil. And access() can be just a bit duplicated. Or we can try how to extract similar parts of access somewhere else like public table methods.

mikz:
We don't need this after merging #135 right?

davidor:
No, we don't. Closing.

PR #134, opened 2016-11-10
add options to ease development

### LINKED ISSUE #124: add option to disable lua code cache and enable code/configuration reloading
We need to have an option to disable lua code cache and enable code reloading.
That will also make configuration reloading work. Later we might need two different options.
One for code and second for configuration.

##### Version

[provide output of the `nginx -V` or `openresty -V` command from openshift/local terminal]
[provide timestamp of the docker image from `docker inspect --format='{{.Created}}' quay.io/3scale/apicast:v2` ]

##### Steps To Reproduce

1. `bin/apicast -C`
2. do a request 
3. change code and configuration
3. see configuration and code reloaded

##### Current Result

does not reload the configuration or code

##### Expected Result

should reload the configuration and code from filesystem/api

##### Additional Information

the `-C` option is just suggestion. can be something else.

mikz:
closes #124

- [x] reloading config on every request

so the lua code cache is annoying. there is no way, to set it up globally and has to be set in the http block.
that means, it can't be passed from the command line (that sets only main block).
I tried if in location, but did not work well. so the only option left is having a file that gets overridden by the bin/apicast every time you execute it.


andrewdavidmackenzie:
We probably need to document clearly that this should not be put into production (or used in serious volume load tests in development or staging), as it will result in a hit to MT endpoint to fetch the config for every API hit? (if I understand it correctly).

PR #135, opened 2016-11-11
Execute functionality that can be used in other modules in the rewrite phase

davidor:
This moves some of the logic that runs in the access phase to the rewrite one. Only 2 things have been moved, saving the service and the data needed to resolve upstream to ngx.ctx. This greatly simplifies writing a new module that shares a lot of functionality with the Apicast one and inherits from it.

### DIFF
--- a/apicast/src/configuration.lua
+++ b/apicast/src/configuration.lua
@@ -180,6 +180,28 @@ function _M.parse_service(service)
         -- if there was no match, usage is set to nil and it will respond a 404, this behavior can be changed
         return usage_t, concat(matched_rules, ", ")
       end,
+      -- Given a request, extracts from its params the credentials of the
+      -- service according to its backend version.
+      -- This method returns a table that contains:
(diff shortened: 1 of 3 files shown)


apicast/src/configuration.lua
```diff
@@ -180,6 +180,28 @@ function _M.parse_service(service)
...
         return usage_t, concat(matched_rules, ", ")
       end,
+      -- Given a request, extracts from its params the credentials of the
+      -- service according to its backend version.
+      -- This method returns a table that contains:
+      --     user_key when backend version == 1
+      --     app_id and app_key when backend version == 2
+      --     access_token when backen version == oauth
+      --     empty when backend version is unknown
+      extract_credentials = function(_, request)
```
mikz: where was this initially? maybe it can be removed?

davidor: It was not anywhere else, but I need it for my module.

PR #139, opened 2016-11-16
various changes to increase resiliency when network is down

mikz:
do not crash, rather try to handle errors gracefully

- network down
- missing nameservers
- ipv6
- html error messages


apicast/src/provider.lua
```diff
@@ -39,20 +39,32 @@ local _M = {
...
 
   _M.contents = configuration.encode(contents)
-  _M.configured = true
-  _M.configuration = config
-  _M.services = config.services or {} -- for compatibility reasons
+
+  if config then
+    _M.configured = true
+    _M.configuration = config
+    _M.services = config.services or {} -- for compatibility reasons
```
davidor: Would be good to know what those compatibility reasons are ... :)

apicast/src/configuration.lua
```diff
@@ -435,7 +439,8 @@ function _M.download(endpoint)
   if body and res.status == 200 then
     ngx.log(ngx.DEBUG, 'configuration response received:' .. body)
 
-    local ok, err = _M.decode(body)
+    local ok
```
mikz: luacheck complaining about shadowing `err`

apicast/src/provider.lua
mikz: yeah, I don't really know :) I guess when testing with system or for liveness/readiness probes of openshift.

apicast/src/configuration.lua
davidor: I see. If we had the code more divided into methods or used more explicit var names this wouldn't be needed, but I'm OK with that for now.

mikz: Yeah. I'm not sure how to solve this. If you have several possible error responses in one method, should they have different names? Or reuse them? Guess more reading of official packages needed :)

PR #141, opened 2016-11-16
[apicast] include custom configuration for apicast

mikz:
# apicast.d/ssl.conf

```
listen 8443 ssl;
ssl_certificate cert.key;
```

@tnscorcoran included example how to start with SSL certificate: 9e4648b


apicast/src/configuration.lua
```diff
@@ -483,13 +483,14 @@ function _M.curl(endpoint)
...
 
+  local timeout = getenv('CURL_TIMEOUT') or 3
   local scheme, user, pass, host, port, path = unpack(url)
 
   if port then host = concat({host, port}, ':') end
 
   url = concat({ scheme, '://', concat({user or '', pass or ''}, ':'), '@', host, path or '/admin/api/nginx/spec.json' }, '')
 
-  local config, exit, code = util.system('curl --silent --show-error --fail --max-time 3 --location ' .. url)
+  local config, exit, code = util.system('curl --silent --show-error --fail --max-time ' .. timeout .. ' --location ' .. url)
```
andrewdavidmackenzie: This change (adding timeout to fetching of config) could be functionally separated from the example, and done in a separate PR?

mikz: It totally could be different PR. But it is already a different commit, so just did not see the benefit.

PR #147, opened 2016-11-17
DNS domain search

mikz:
`/etc/resolv.conf` can specify `search` section that is used for domains without TLD.

For example on openshift, there is service `foobar`. If you try `dig foobar` you won't get an answer.
The proper host is `foobar.something.svc.cluster.local`. That domain is defined in the `search` section of `/etc/resolv.conf`.


unleashed:
This is crazy mad. :+1:

PR #148, opened 2016-11-18
New Resource Owner Template Oauth Flow files - generalised

tnscorcoran:
These are breaking changes - expecting feedback on refined way to do this.

### DIFF
--- a/apicast/src/get_token.lua
+++ b/apicast/src/get_token.lua
@@ -1,9 +1,9 @@
 local cjson = require 'cjson'
 local ts = require 'threescale_utils'
 
--- As per RFC for Authorization Code flow: extract params from Authorization header and body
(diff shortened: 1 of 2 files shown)


VinayBhalerao:
This would not support and break authorization code flow, right?

I think we should have a clean way to plug and execute all four flows 1) Authorization code flow (default) 2) client credential 3) resource owner 4) Implicit flow just by separating the ``grant_type`` parameter.

andrewdavidmackenzie:
Vinay, that is the intention....but to come up with code that enables all 4
flows we want to see what it takes to do each one.

This PR would break other flow and will not be merged....but we still want
to see it....

VinayBhalerao:
I understand now

mikz:
Closing. Will be kept on github as a reference.

PR #176, opened 2016-12-06
Rename parameters name

VinayBhalerao:
It's nice to have consistent naming ``APICAST_*`` while creating new application using parameters

```oc new-app -f...-p APICAST_RELOAD_CONFIG=1 -p APICAST_PATH_ROUTING=1 -p APICAST_RESPONSE_CODES=1 ``` 

instead of 

```oc new-app -f...-p APICAST_RELOAD_CONFIG=1 -p PATH_ROUTING=1 -p RESPONSE_CODES =1```

### DIFF
--- a/openshift/apicast-template.yml
+++ b/openshift/apicast-template.yml
@@ -44,15 +44,15 @@ objects:
           - name: APICAST_SERVICES
             value: ${APICAST_SERVICES}
           - name: APICAST_MISSING_CONFIGURATION
-            value: ${MISSING_CONFIGURATION}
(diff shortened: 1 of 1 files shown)


mikz:
Yes. But I'd probably like to have it the other way around. Not sure if prefixing everything with APICAST is a good idea. It is apicast template, so obviously, it is all related to apicast.

andrewdavidmackenzie:
We should review documents that mention those parameter names and if there are any, rename them in the same PR.

mikz:
Opened #207 to track this. Feel free to continue discussion there.

PR #177, opened 2016-12-06
Explain to use make help

andrewdavidmackenzie:
Explain that you can find other test targets (docker etc) by using make help

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,13 +49,13 @@ script:
 - make check
 - make busted
 - make prove
-- make build
-- make test-docker
+- make builder-image
+- make test-builder-image
 - make prove-docker
-- make test-docker-release
+- make test-runtime-image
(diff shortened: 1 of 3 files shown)


mikz:
Pushed small clarification. Here is `make help` output:

```
test                           Run all tests
busted                         Test Lua.
check                          Run luacheck to lint lua files
prove                          Test nginx
prove-docker                   Test nginx inside docker
builder-image                  Build builder image
runtime-image                  Build runtime image
push                           Push image to the registry
bash                           Run bash inside the builder image
test-builder-image             Smoke test the builder image. Pass any docker image in IMAGE_NAME parameter.
test-runtime-image             Smoke test the runtime image. Pass any docker image in IMAGE_NAME parameter.
clean                          Remove all running docker containers
help                           Print this help
```

PR #202, opened 2017-01-04
[config] extract credentials handling

mikz:
* into new object: Service
* in the future can be extracted further in to Credentials
* uses new resty-http_authorization library
* now is a metamethod on new Service object
* remove old get_credentials
* overall unify credentials retrieval and handling

extracted from #162

This is a breaking change as it totally removes `:get_credentials` and changes `:extract_credentials` method. Imo it should be easier for modules to work with this, but would be good to get some feedback from @davidor.

the next steps would be to unify parameter extraction to fix https://github.com/3scale/apicast/issues/163
it should be prepared to parse json too and be lazy, as the payloads can be huge

### DIFF
--- a/apicast/src/configuration/service.lua
+++ b/apicast/src/configuration/service.lua
@@ -67,85 +67,85 @@ local credentials_mt = {
   end
 }
 
-local backend_version_credentials = {
-  ['1'] = function(credentials)
-    local name = (credentials.user_key or 'user_key')
-    local user_key
+local backend_version_credentials = { }
 
-    if credentials.location == 'query' then
-      user_key = ngx.var['arg_' .. name] or read_body_args(name)[1]
+function backend_version_credentials.version_1(config)
+  local name = (config.user_key or 'user_key')
+  local user_key
 
-    elseif credentials.location == 'headers' then
-      user_key = read_http_header(name)
+  if config.location == 'query' then
+    user_key = ngx.var['arg_' .. name] or read_body_args(name)[1]
 
-    elseif credentials.location == 'authorization' then
-      local auth = http_authorization.new(ngx.var.http_authorization)
+  elseif config.location == 'headers' then
+    user_key = read_http_header(name)
 
-      user_key = auth.userid or auth.password or auth.token
-    else
-      return nil, 'invalid credentials location'
-    end
+  elseif config.location == 'authorization' then
+    local auth = http_authorization.new(ngx.var.http_authorization)
 
-    return { user_key = user_key }
-  end,
-
-  ['2'] = function(credentials)
-    local app_id_name = (credentials.app_id or 'app_id')
-    local app_key_name = (credentials.app_key or 'app_key')
-
-    local app_id, app_key
-
-    if credentials.location == 'query' then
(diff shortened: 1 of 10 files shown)


apicast/src/configuration/service.lua
```diff
@@ -0,0 +1,167 @@
...
+      local v
+      repeat
+        k, v = next(t, k)
+      until k == nil or type(k) ~= 'number'
+      return k, v
+    end, t, nil
+  end
+}
+
+local backend_version_credentials = {
```
davidor: I think that extracting one method for each of the versions would help readability. What do you think @mikz ?

apicast/src/configuration/service.lua
```diff
@@ -0,0 +1,167 @@
...
+
+    elseif credentials.location == 'authorization' then
+      local auth = http_authorization.new(ngx.var.http_authorization)
+
+      user_key = auth.userid or auth.password or auth.token
+    else
+      return nil, 'invalid credentials location'
+    end
+
+    return { user_key, user_key = user_key }
```
davidor: Why not just { user_key = user_key } ?

mikz: because `table.concat` merges just integer keys, so if this object is meant to be used for both creating the cache key and as  a hash to make a query string from it needs to be this weird mixture of array/hash

apicast/src/configuration/service.lua
mikz: Like:
```lua

local function backend_version_1_credentials(config)
  ...
end

backend_version_credentials['1'] = backend_version_1_credentials
```
?

davidor: That's what I would do.

apicast/src/configuration/service.lua
davidor: Thanks for the clarification :+1:

apicast/src/configuration/service.lua
```diff
@@ -0,0 +1,167 @@
...
+
+    -- https://tools.ietf.org/html/rfc6750#section-2.1 says:
+    -- Resource servers MUST support this method. [Bearer]
+    access_token = access_token or authorization.token
+
+    return { access_token, access_token = access_token }
+  end
+}
+
+function _M.extract_credentials(self)
```
davidor: I would document what this method returns because it's a public method that can be used from third-party modules.

apicast/src/configuration/service.lua
mikz: Will describe it there. It is worth a comment.

apicast/src/configuration/service.lua
mikz: Yep. Will integrate ldoc as some other code is using it already: https://github.com/3scale/apicast/pull/193

apicast/src/configuration/service.lua
```diff
@@ -0,0 +1,209 @@
+-----------------
+--- Service configuratin object.
```
davidor: typo

mikz: fixed & rebased

PR #213, opened 2017-01-10
[config] print warning when multiple services share host

mikz:
as it can lead to hard to debug errors and using wrong service


mikz:
👍

PR #220, opened 2017-01-13
Add state to client_data for returning back to client

### LINKED ISSUE #219: [OAuth] Differentiate state value sent by client from state value generated by APIcast
There is an issue with Postman no longer working with APIcast OAuth due to the fact that Postman sends a state value which is then overwritten by APIcast for its own purpose. 

Ideally, we would differentiate between the state value the client sends and the state value that is used internally by APIcast as the client expects to receive back the same state value it sent and we're returning the APIcast state value instead. 

This causes the Authorization response step to fail in Postman and hence no access token is requested from APIcast.

mpguerra:
closes #219

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Cache all calls to `os.getenv` via custom module [PR #231](https://github.com/3scale/apicast/pull/231)
 
 ### Fixed
+- [OAuth] Return correct state value back to client
+
 ### Removed
 
 ## [3.0.0-alpha1] - 2017-01-16
--- a/apicast/src/authorized_callback.lua
+++ b/apicast/src/authorized_callback.lua
@@ -108,9 +108,9 @@ local function store_code(client_data, params, code)
 end
 
 -- Returns the code to the client
-local function send_code(client_data, params, code)
+local function send_code(client_data, code)
   ngx.header.content_type = "application/x-www-form-urlencoded"
-  return ngx.redirect( client_data.redirect_uri .. "?code="..code.."&state=" .. (params.state or ""))
+  return ngx.redirect( client_data.redirect_uri .. "?code="..code.."&state=" .. (client_data.state or ""))
 end
 
 -- Get Authorization Code
@@ -121,7 +121,7 @@ local function get_code(service_id, params)
   local stored = store_code(client_data, params, code)
 
   if stored then
-    send_code(client_data, params, code)
+    send_code(client_data, code)
   end
 end
 
--- a/t/005-apicast-oauth.t
+++ b/t/005-apicast-oauth.t
@@ -76,14 +76,12 @@ Location: http://example.com/redirect?error=invalid_client
     }
   }
 --- request
-GET /authorize?client_id=id&redirect_uri=otheruri&response_type=code&scope=whatever
+GET /authorize?client_id=id&redirect_uri=otheruri&response_type=code&scope=whatever&state=123456
 --- error_code: 302
 --- response_headers_like
-Location: http://example.com/redirect\?scope=whatever&response_type=code&state=\w+&tok=\w+&redirect_uri=otheruri&client_id=id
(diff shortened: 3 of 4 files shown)


MarkCheshire:
Isn't the oauth RFC clear on which behavior meet the spec?

mpguerra:
Yes, see: https://tools.ietf.org/html/rfc6749#section-4.1.1

Postman sends a state value, so we need to make sure we return that same state value back to Postman. 

However, we also use state within the gateway to keep track of the requests for a particular customer, but the client itself is sending a state value which is overwritten by the one we generate within the gateway to keep track of the request. The RFC does not deal with the case where you have a Gateway in the middle, and we probably don't want to use the same state value the client sends for the gateway state value, so we just need to make sure we keep them separate.

This does not really change the RFC behaviour.

apicast/src/authorized_callback.lua
```diff
@@ -108,7 +108,7 @@ local function store_code(client_data, params, code)
 end
 
 -- Returns the code to the client
-local function send_code(client_data, params, code)
+local function send_code(client_data, code)
```
mikz: You are changing the method signature, but not changing the caller. The caller still passes `params` parameter which would get used instead of `code`.

t/005-apicast-oauth.t
```diff
@@ -361,3 +362,38 @@ GET /?access_token=foobar
 --- error_code: 200
 --- response_body
 yay, upstream
+
+=== TEST 12: calling /authorize with state returns same value back on redirect_uri
```
mpguerra: Not sure whether I should add this new test or just update TEST 9, any suggestions @mikz ?

mikz: I'd say test 2 and 3 are quite similar to what you want, no? Just not checking the state parameter.

mpguerra: Well, we might want to add some additional ones sending the state parameter also, but since it's actually optional (but recommended) I think it makes sense to have separate test for it i.e test where client doesn't send state parameter vs tests where client sends state parameter

mpguerra: Yes, I can copy test 2 and 3, sending an additional state parameter, and they currently pass, but that won't test that the state that the client sends will be sent back to it with the code, since we need to invoke the callback to confirm this. 

That's what I'm struggling with, since to invoke callback I need to know what the apicast state value is, hence what test 12 is doing.

So my questions here would be whether test 12 is valid and whether we additionally need copies of tests 2 and 3 to validate that sending the state parameter doesn't break anything. 

Alternatively should test 12 be replaced with something else?

Where I'm struggling is in captur

mikz: I guess it is fine now. It is really confusing, because we are reusing `params.state` for the nonce and that is so evil. I pushed some clarifications on the state length to indicate it is one or the other when we reuse the field.

mikz: Imo we should rename our `state` to `nonce` or something to not reuse the field as it is really confusing. But that can be done in another PR.

mikz:
@mpguerra could you add a CHANGELOG entry and we can merge this?

t/005-apicast-oauth.t
MarkCheshire: Agree with @mikz this is what I was getting at when we discussed the other day

mpguerra:
ready for final review/merge

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Cache all calls to `os.getenv` via custom module [PR #231](https://github.com/3scale/apicast/pull/231)
 
 ### Fixed
+- [OAuth] Return correct state value back to client
```
mikz: sorry, there needs to be a space after, you can see the error in the GitHub visual diff 
I guess we should keep them with spaces already in the template.

mpguerra: better? :)

mikz:
Thank you 🥇

PR #221, opened 2017-01-13
Fix test numbering

mpguerra:


mikz:
There is a script to do that. I guess we should run it as part of the CI and fail the build if there are any changes.

PR #253, opened 2017-02-07
performance optimizations

mikz:
# JIT JIT JIT JIT JIT JIT JIT

LuaJIT can compile commonly used code paths into machine and yield massive speedups.

To do that those paths have to be fairly simple, use supported APIs and be used a lot.

## NYI - using supported APIs

LuaJIT does not compile everything. Just some supported subset. You can see what is (not) supported on  http://wiki.luajit.org/NYI. Those not supported methods should be avoided at all cost on hot code paths.

This means no: unpack, pairs, ipairs, new functions, next, ...

## Fairly simple

> Reduce number of unbiased/unpredictable 
> Heavily biased branches (>95% in one direction) are fine

So don't have some flip flop conditions that can go both ways.

That also means: never ever use global variables. Always alias them to local ones closest as possible.

# Profiling

This PR adds 3 lines to nginx.conf that can be uncommented to do some interesting stuff.

* `jit.v` prints debug messages why something was not compiled
* `jit.p` shows stats of interpreted/jited code on exit
* `jit.dump` shows luajit bytecode instructions and reasons of not compiling

# GC

Everything allocated has to also be cleaned. So imagine following code:

```lua
ngx.log(ngx.DEBUG, 'some error: ' .. value)
```
It does allocate two strings and calls `__tostring` metamethod on `value` because it concatenates strings.

Better to just let it concatenate only when the log level exists:

```lua
ngx.log(ngx.DEBUG, 'some error: ', value)
```

# Comparison

https://gist.github.com/mikz/89bfde40b43ac03dc0bd5c58527c5ad1

# Resources

* http://wiki.luajit.org/NYI
* https://en.blog.nic.cz/2015/08/12/embedding-luajit-in-30-minutes-or-so/
* http://wiki.luajit.org/Numerical-Computing-Performance-Guide
* http://stackoverflow.com/questions/7167566/luajit-2-optimization-guide
* https://luapower.com/luajit-notes


Vagrantfile
```diff
@@ -59,5 +56,6 @@ Vagrant.configure("2") do |config|
      ln -sf /usr/local/stapxx/samples/*.sxx /usr/local/bin/
      ln -sf /usr/local/openresty-systemtap-toolkit/fix-lua-bt /usr/local/bin/
      usermod -a -G stapusr,stapdev vagrant
+     echo -e 'vagrant\t\t\t-\tnofile\t\t1000000' > /etc/security/limits.d/90-nofile.conf
```
davidor: Could you add a comment explaining what this does?

mikz: Sure!

PR #257, opened 2017-02-08
[WIP]Feature custom metric report

kevprice83:
Use case: SOAP API has mapping rules defined in the SOAP object and it is expensive to parse the body to extract operation value. Using the custom header value which is the same as some metric/method system_name allows us to report to that metric in 3scale. This also covers the scenario where there are multiple SOAP operation but only one end point and therefore one mapping rule.


apicast/src/configuration.lua
```diff
@@ -40,10 +42,25 @@ local function regexpify(path)
   return path:gsub('?.*', ''):gsub("{.-}", '([\\w_.-]+)'):gsub("%.", "\\.")
 end
 
+local function check_custom_metric(custom_metric, req)
+  local metric = ngx.req.get_headers()["X-Custom-Metric"]
```
andrewdavidmackenzie: IIRC the RFC allows multiple headers with the same name.
That would allow reporting of multiple metrics....

Not supported (i.e. might break?) this code?

andrewdavidmackenzie:
a test?
:-)

mikz:
Do I understand it correctly, that some client supplied HTTP header value is later used as a metric name to report to backend? 
Can't clients just supply whatever ?

apicast/src/configuration.lua
kevprice83: Always take the value of the first maybe?

kevprice83:
The client will need to supply something that is supported, i.e. defined within 3scale account of the provider. This is then made clear in the documentation describing that API. I haven't written the response code and message in the case that it is an unsupported value - TODO!

kevprice83:
Closing this as again it is very specific but I will revisit this as it might be a useful snippet for SOAP integrations.

PR #263, opened 2017-02-15
performance tuning

mikz:
* keep more connections opened
* allow configuring workers via env variable

we were able to pus through up to 5k requests per second

### DIFF
--- a/apicast/bin/apicast
+++ b/apicast/bin/apicast
@@ -46,6 +46,7 @@ log_level="${!log}"
 mkdir -p "${apicast_dir}/logs"
 
 daemon=off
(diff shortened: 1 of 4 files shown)


apicast/bin/apicast
```diff
@@ -46,6 +46,7 @@ log_level="${!log}"
 mkdir -p "${apicast_dir}/logs"
 
 daemon=off
+worker_processes=${APICAST_WORKERS:-1}
```
jmprusi: This ENV var is really useful, we should document it.

mikz: true, will add it to `.env`

PR #285, opened 2017-03-07
Add /healthz endpoint for health checks

rnavarro:
Add a /healthz endpoint for load balancer health checks when running in kubernetes.

https://github.com/kubernetes/ingress/tree/master/examples/health-checks/gce


mikz:
Hi, thanks for the contribution!

We probably should do better job documenting it, but already have health check endpoints.

See https://github.com/3scale/apicast/blob/96e5c5e988d29141d444992d951caae5d04fc302/openshift/apicast-template.yml#L67-L78
It is accessing the [Management API](https://github.com/3scale/apicast/blob/master/doc/management-api.md) and has some logic for the liveness/readiness.

Readiness is determined by having a proper configuration and liveness is just that the Management API booted.

rnavarro:
Thanks for the info!

Those look great, the only problem I see is that for kubernetes ingress' the health check mechanism is very basic.

As such, the health check URLs need to be on the same port as the serving port.

In you examples the health checks are running on the management port.

I tried many different things to try to get these health checks to work off the management port but I couldn't get it to work.

The only way I got this to work was to out the health check on the serving port.

Is there a clean way to serve those health checks off the management and regular ports?

mikz:
Interesting. Which version of Kubernetes are you using? 

We are using OpenShift since 3.2 (that would be using Kubernetes 1.2) and it works just fine. 

The Kubernetes docs also mention `port` which is what we use: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#using-a-named-port

rnavarro:
We're running on GCE using version 1.5.3 for this cluster right now.

What you linked to is the monitoring for containers in a pod.

While an **ingress** _references_ the readiness probe from a pod definition, it's merely used as a loose template for generating the actual health check in GCE.

If you look here:
https://github.com/kubernetes/ingress/tree/master/examples/health-checks/gce

At the bottom under limitations we see:

> The readiness probe must be exposed on the port matching the servicePort specified in the Ingress

That's the limitation that we're running up against when trying to use apicast with an ingress.

mikz:
Ah! I see now. Haven't really worked with raw Kubernetes. 

I think it would be too invasive to have a path defined in APIcast because in theory you could proxy to another API that has `/healthz` endpoint.

Also, wouldn't you want to use this to monitor the underlying API and not APIcast?

Also what is the benefit of using health checks in Ingress ? It can't really do anything with it no ? Like route to different backend.

rnavarro:
So in Google land a kubernetes ingress is modeled as a GCE HTTP load balancer (https://cloud.google.com/compute/docs/load-balancing/http/)

This offers us great performance, additional monitoring, and SSL termination at the very edge of the network.

I'll attempt to represent this in text but it probably won't work out too well, our stack looks like this:

(Internet) --> GCE HTTP Load Balancer (A) --> multiple apicast pods --> GCE HTTP Load Balancer (B) --> multiple API application pods

In our stack the ingress/http load balancer (B) health checks monitor our actual API instances (which is what you were eluding to)

What I'm looking for is monitoring for our apicast pods from the ingress/http load balancer (A)

Unfortunately with GCE HTTP Load balancers you're **required** to have an active health check for all backends. If the health check fails no traffic will be routed (this is unlike their Network Load Balancers)

I'm certainly happy to change the _path_ of this health check URL, I see we already have things such as "/_threescale/oauth_store_token" or "/_threescale/check_credentials" in the same configuration.

Modifying the HC url to be "/_threescale/healthz" works for me if that's a better middle ground.

With that the healthz endpoint is under the existing _threescale namespace.

mikz:
Makes perfect sense, thanks for the explanation. Sounds like quite a limitation in those health checks.

Anyway, adding it to the `/_threescale/` namespace would be fine for now I guess. We plan to remove it in the future so there is nothing defined in the path of the proxy traffic. But that will take time.

So should this readiness probe do the same as the one in the management api ? I opened a PR trying to describe it a bit more: https://github.com/3scale/apicast/pull/286

Basically it returns error if there is no configuration.

rnavarro:
@mikz I think that would make the most sense operate the same way and just expose it on an additional endpoint.

mikz:
@rnavarro any updates regarding this?

rnavarro:
Sorry, I had forgotten about this after we committed locally.

I've updated the PR for review.

mikz:
@rnavarro thank you very much!

PR #290, opened 2017-03-09
option to drop post_action

mikz:
https://github.com/3scale/apicast/pull/289 revealed limitations of post_action

this PR offers new experimental flag to turn off post_action 
and use nginx timers instead

setting `APICAST_REPORTING_THREADS` to a value greater than 0 is going to enable asynchronous reporting to backend.

------

doing synchronous call in the post_action will
push the latency to the client on the same connection

using timers we can make that fully asynchronous
not being tied to the request phase at all

setting APICAST_REPORTING_THREADS environment variable
to a number greater than 0 will enable reporting in asynchronous
timers instead of synchronous call

the number defines a maximum number of parallel calls
to backend per worker

if there is new report coming but no connections is available
APIcast will wait for up to 10 seconds before falling back to the
synchronous call

please note this is **experimental** feature

# TODO
- [x] write a test case for https://github.com/3scale/apicast/pull/302 (ca45731)


apicast/src/management.lua
```diff
@@ -113,13 +113,28 @@ function _M.disabled()
   ngx.exit(ngx.HTTP_FORBIDDEN)
 end
 
+function _M.info()
```
andrewdavidmackenzie: This looks separate from no post_action changes, adding an /info endpoint, no?

mikz: Yes it is separate commit. Not really useful without the post_action changes because there where no timers before.

andrewdavidmackenzie:
I'd remove this one from ER5 Michal and work on it in an experimental branch....

mikz:
The whole feature is opt-in and does not affect the normal flow. Will do benchmark tomorrow to verify nothing changed and it can be merged. This is done just waiting.

davidor:
I ran some tests to make sure that setting `APICAST_REPORTING_THREADS=0` does not introduce a performance regression in the normal Apicast flow.

I performed 2 tests, each of them comparing Apicast master branch with this PR:
1) Deployed an Apicast, an Apicast in echo mode (as api backend), and a 3scale backend. All of them in different machines. In this scenario Apicast is limited by the 3scale backend.
2) Deployed just an Apicast and another one in echo mode (as api backend) in separate machines.

The tests were performed using JMeter from an external machine. There are many factors that can affect the throughput, latencies, etc. So absolute numbers are not really meaningful. I just want to focus on the comparison between the master branch and this PR. I used c4.xlarge aws machines, configured Apicast with 4 workers, and run JMeter for 5 minutes.

**Results for test 1:**
Apicast master
1_master

PR
1_pr

**Results for test 2:**
Apicast master
2_master


PR
2_pr

The results are similar in both cases.

andrewdavidmackenzie:
So, focussing on the second of the two comparisons:

In the Pr version:

Thruput is 97.7% of Master
95%ile latency is 107% of Master

PR #307, opened 2017-03-16
Fix the custom module example

ErwinSteffens:
I was  trying to extend the default apicast module by using [this example](https://github.com/3scale/apicast/blob/master/examples/custom-module/verbose.lua), but we ran into some issues. I think this merge request should fix the example, but I'am not much of an lua expert.

The `module.lua` file loads the `apicast` or the overriden module, but calls the `new()` method when it is available. When the new method is not overridden it just returns the `apicast` module and not the module override.

The new code works for us.

### DIFF
--- a/examples/custom-module/verbose.lua
+++ b/examples/custom-module/verbose.lua
@@ -3,10 +3,14 @@ local apicast = require('apicast')
 local _M = { _VERSION = '0.0' }
 local mt = { __index = apicast }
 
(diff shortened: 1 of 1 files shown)


mikz:
Thank you for reporting this. Will try to get it merged today.

mikz:
@ErwinSteffens I made a PR that tries to describe module inheritance and adds an example - #362

PR #317, opened 2017-03-20
update readme with steps for openshift deployment option

kevprice83:
This PR will update the READme as well as add the redis template necessary for the OpenShift deployment.

### DIFF
--- a/doc/oauth-guide.md
+++ b/doc/oauth-guide.md
@@ -33,6 +33,32 @@ Here we have added the `-v` flag to run APIcast in verbose mode, in order to all
 
 Our Authorization Server should also be up and running and ready to receive requests.
 
+## Running APIcast with OAuth (on OpenShift)
+
+APIcast in OAuth mode as mentioned previously has a Redis dependency that needs to be configured. In this case it will be necessary to deploy the Redis database separately using the template provided [here](https://github.com/3scale/apicast/tree/master/openshift).
+
+It's also worth mentioning that if you are using the provided Ruby client and Auth server then they must be exposed on a publicly available **DNS** to correctly route the requests from the OpenShift cluster. A free service like [Ngrok](https://ngrok.com/) is a useful tool for testing.
+
+## Pre-requisites
+
+- Cluster is up and running
+- New project has been created
+
+Proceed to the following steps if you have fulfilled the 2 previous requirements and then continue the set up as described in our [OpenShift tutorial](https://support.3scale.net/docs/deployment-options/apicast-openshift)
+
+### Steps
+
+The following commands are based on the **oc cluster up** installation.
+
+1. Log in as admin and create a persistent volume using `oc login -u system:admin` and `redis-persistent.json`
+  - mkdir -p /var/lib/docker/pv/01
+  - chmod g+w /var/lib/docker/pv/01
+  - chcon -Rt svirt_sandbox_file_t /var/lib/docker/pv/
+  - oc new-app --param PV=01 -f redis-persistent.json
+
+2. Log in as developer using `oc login -u developer:developer`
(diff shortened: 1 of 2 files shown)


doc/oauth-guide.md
```diff
@@ -33,6 +33,32 @@ Here we have added the `-v` flag to run APIcast in verbose mode, in order to all
...
+
+### Steps
+
+The following commands are based on the **oc cluster up** installation.
+
+1. Log in as admin and create a persistent volume using `oc login -u system:admin` and `redis-persistent.json`
+  - mkdir -p /var/lib/docker/pv/01
+  - chmod g+w /var/lib/docker/pv/01
+  - chcon -Rt svirt_sandbox_file_t /var/lib/docker/pv/
+  - oc new-app --param PV=01 -f redis-persistent.json
```
mikz: This is not going to work. There are no persistent volumes set up.

kevprice83: This is the persistent volume

mikz: No. This is template that uses Persistent Volumes by Persistent Volume Claims. No Persistent Volumes are created and have to be created before creating Persistent Volume Claims.

kevprice83: Okay I see what you mean, I pushed the wrong version of the template I was using locally. My bad...

openshift/redis-persistent.json
```diff
@@ -65,6 +65,25 @@
...
+      "spec": {
+        "accessModes": [
+          "ReadWriteOnce",
+          "ReadWriteMany"
+        ],
+        "capacity": {
+          "storage": "10Gi"
+        },
+        "persistentVolumeReclaimPolicy": "Recycle",
+        "hostPath": {
```
mikz: This makes the template not really reusable between installations. PersistentVolumes are usually not created in application templates. Using `hostPath` will lock the deployment to one node and is not production ready.

doc/oauth-guide.md
```diff
@@ -33,6 +33,32 @@ Here we have added the `-v` flag to run APIcast in verbose mode, in order to all
...
+- Cluster is up and running
+- New project has been created
+
+Proceed to the following steps if you have fulfilled the 2 previous requirements and then continue the set up as described in our [OpenShift tutorial](https://support.3scale.net/docs/deployment-options/apicast-openshift)
+
+### Steps
+
+The following commands are based on the **oc cluster up** installation.
+
+1. Log in as admin and create a persistent volume using `oc login -u system:admin` and `redis-persistent.json`
```
mikz: Are you sure persistent is needed? From what I remember it is just used for few seconds to hold some state and ephemeral would be just fine.

kevprice83: I guess so. I tested ephemeral and then implemented persistent based on your advice :) I can certainly re-do this commit to only use the ephemeral Redis as it is only storing keys with state values for the OAuth handshake. I'm really not that up-to-speed with OpenShift and Redis so all advice is appreciated here.

mikz: I suggested using ephemeral first and then going for persistence - if you need persistence. 
Do you need persistence? I don't think so. So just use ephemeral.

kevprice83: Yeah it seems Redis is used for the OAuth handshake only in this case, anyway, I saw it mentioned somewhere that the Redis dependency will be removed from OAuth altogether, is that correct?

mikz: Yes. https://github.com/3scale/apicast/issues/174

mpguerra:
@kevprice83 do you still need my review on this? Or do you need to update based on comments by @mikz 

Would be good to update the support site with these additional details once ready

kevprice83:
Closing due to rewrite of the OAuth integration pending.

PR #325, opened 2017-03-21
[remote_v2] fix boot loading configuration

mikz:
fixes a bug introduced in https://github.com/3scale/apicast/pull/323
which prevented apicast in production to boot


mikz:
Ignoring the changelog entry because it was introduced earlier today and has no value for the changelog.

Issue #322, opened 2017-03-21
Keycloak host ssl not verified


ErwinSteffens:
See [keycloak.lua](https://github.com/3scale/apicast/blob/master/apicast/src/oauth/keycloak.lua). Should the `verify` not be set to `true`?

```
local http_client = http_ng.new{
    backend = client,
    options = {
        ssl = { verify = false }
    }
}
```

I know there are some implications with the root certificates in the image, but I think it is a little dangerous to validate the JWT with a certificate from a source that is not verified.

mikz:
We can't really set it to true in all cases because plenty of development set-ups will have invalid certificates. Plenty of other setups have custom CA which is not included / trusted by the image, so checks fail.

We should make this configurable, but probably will have to be opt-in even though I know it should be opt-out.

mikz:
WIP https://github.com/3scale/apicast/pull/332

I'll aim to have it on by default (docker, native) but off by default on openshift (where you'll probably have invalid SSL).

mikz:
There is also much fundamental issue with ssl verification in openresty. 
See the discussion in the mailing list: https://groups.google.com/forum/#!topic/openresty-en/SuqORBK9ys0

OpenResty can't use system certificates and always needs to define all certificates it is going to use. 
To make this work nginx and openresty have to be patched. 

The suggested approach is to use 
  https://github.com/openresty/nginx-devel-utils/blob/master/export-builtin-trusted-certs  to generate the certificate chain and  https://github.com/openresty/lua-nginx-module#lua_ssl_trusted_certificate  with 
    https://github.com/openresty/lua-nginx-module#lua_ssl_verify_depth .

We might do some automation of this in the future, but for now. If you are interested in making ssl verification to work you'll probably have to do those steps.

mikz:
So in the end I'll be probably able to use system ca-bundle in https://github.com/3scale/apicast/pull/332
so it might work! Hope to deliver it tomorrow.

PR #332, opened 2017-03-22
[env] one way to control openssl peer verification

mikz:
still is off by default, but at least can be turned on
closes https://github.com/3scale/apicast/issues/322

# APIcast <> 3scale AMP

Connection can be verified by setting the `OPENSSL_VERIFY` env variable to `true`.
It is **off** by default.

# APIcast <> upstream

Is also **off** by default. Can be enabled by injecting custom config `apicast.d/proxy_ssl.conf` with following content:

```nginx
proxy_ssl_verify on;
```

# CA Bundle

Both verifications are set up to use the operating system's ca bundle.


apicast/libexec/boot.lua
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env resty
```
mikz: this should be removed and same changes applied to the keycloak configuration loader

PR #341, opened 2017-03-28
Store JWT in Auth cache with ttl

### LINKED ISSUE #309: [OAuth] Take JWT expiry into account in authorizations response cache
Currently, for OpendID Connect, the JWT token is used as the key against which we check authorizations in the APIcast authorization responses cache. Since the JWT has a TTL, it should ideally also be automatically removed from the cache when it expires.  https://github.com/3scale/apicast/pull/283#issuecomment-287057444

##### Version

ER5

mpguerra:
In theory the expiry is checked already when validating the token and the call should be rejected before it even gets to the point where we are checking the auth response cache.

However for sake of correctness, JWT cached_key should be stored with ttl

To do: 

- [x] Store jwt in cache with ttl
- [x] Tests
- [x] Fix nginx tests
- [x] CHANGELOG

Closes #309

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,12 +19,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Improve error logging when failing to download configuration [PR #335](https://github.com/3scale/apicast/pull/325)
 - Service hostnames are normalized to lower case [PR #336](https://github.com/3scale/apicast/pull/326)
 - Don't attempt to perform post\_action when request was handled without authentication [PR #343](https://github.com/3scale/apicast/pull/343)
+- Store authorization responses with a ttl, if sent [PR #341](https://github.com/3scale/apicast/pull/341)
 
 ### Fixed
 
 - Do not return stale service configuration when new one is available [PR #333](https://github.com/3scale/apicast/pull/333)
 - Memory leak in every request [PR #339](https://github.com/3scale/apicast/pull/339)
 - Remove unnecessary code and comments [PR #344](https://github.com/3scale/apicast/pull/344)
+- JWT expiry not taken into account in authorization response cache [PR #283](https://github.com/3scale/apicast/pull/283) / [Issue #309](https://github.com/3scale/apicast/issues/309) / Fixed by [PR #341](https://github.com/3scale/apicast/pull/341)
 
 ## [3.0.0-beta3] - 2017-03-20
 
--- a/apicast/src/oauth/keycloak.lua
+++ b/apicast/src/oauth/keycloak.lua
@@ -28,6 +28,12 @@ _M.params = {
   }
 }
 
+local function timestamp_to_seconds_from_now(expiry)
+  local time_now = ngx.now()
+  local ttl = expiry and (expiry - time_now) or nil
+  return ttl
+end
+
 -- Formats the realm public key string into Public Key File (PKCS#8) format
 local function format_public_key(key)
   local formatted_key = "-----BEGIN PUBLIC KEY-----\n"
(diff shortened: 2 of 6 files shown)


apicast/src/oauth/keycloak.lua
```diff
@@ -27,6 +27,12 @@ _M.params = {
   }
 }
 
+local function timestamp_to_seconds_from_now(expiry)
+  local time_now = os.time()
```
mikz: `ngx.now()` should be used as it is more performant and does not include a syscall


also the case when expiry is nil should be handled

apicast/src/proxy.lua
```diff
@@ -190,8 +190,8 @@ function _M:authorize(service, usage, credentials)
 
     if res.status == 200 then
       if api_keys then
-        ngx.log(ngx.INFO, 'apicast cache write key: ', cached_key)
-        api_keys:set(cached_key, 200)
+        ngx.log(ngx.INFO, 'apicast cache write key: ', cached_key, ', ttl: ', ttl or 'none')
```
mikz: I think the `none` is not really useful as it prints `nil` otherwise which is already pretty explicit.

apicast/src/oauth/keycloak.lua
```diff
@@ -157,17 +163,21 @@ function _M:transform_credentials(credentials)
...
 
     if err then
-      return nil, err
+      return nil, nil, err
     end
 
     if jwt_obj.payload then
       local app_id = jwt_obj.payload.aud
+      local ttl
+      if jwt_obj.payload.exp then
```
mikz: I think the condition is not necessary. All this can be handled in `timestamp_to_seconds_from_now`.

spec/ngx_helper.lua
```diff
@@ -2,15 +2,18 @@ local busted = require('busted')
...
   ngx.ctx = ngx_ctx
+  ngx.shared = ngx_shared
 end
 
 busted.after_each(cleanup)
 busted.teardown(cleanup)
 
 busted.before_each(function()
   ngx.ctx = { }
+  ngx.shared = { }
```
mikz: I'd not go as far as this. Imo each test can reset this the same they reset `ngx.var`.

spec/proxy_spec.lua
```diff
@@ -80,4 +80,36 @@ describe('Proxy', function()
...
+    local authorize = proxy.authorize
+    local service = { backend_authentication = { value = 'not_baz' } }
+    local usage = 'foo'
+    local credentials = 'client_id=blah'
+
+    it('takes ttl value if sent', function()
+      local ttl = 80
+      ngx.var = { cached_key = credentials, usage=usage, credentials=credentials, http_x_3scale_debug='baz', real_url='blah' }
+      ngx.ctx = { backend_upstream = ''}
+      ngx.shared.api_keys = { cached_key = '', get = function () return {} end }
```
mikz: I think here it should do the same as for `ngx.var`.

```lua
ngx.shared = { api_keys = ... }
```

Because otherwise you are mutating the `shared` table and leaking it to another tests .

Also I don't see why there is `cached_key = ''`

CHANGELOG.md
```diff
@@ -55,7 +57,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Known Issues
 
-- JWT expiry not taken into account in authorization response cache [PR #283](https://github.com/3scale/apicast/pull/283) / [Issue #309](https://github.com/3scale/apicast/issues/309)
```
mikz: I think this can stay in the release notes. Because it was a known issue for that version.

PR #357, opened 2017-04-27
Add HTTP Cache support to http_ng

mikz:
Using `Cache-Control: private` the http client 
should be able to cache responses and use the cache
on repeated calls

Good read: https://tools.ietf.org/html/rfc7234

You can use it like:
```lua

local cache_store = require('resty.http_ng.cache_store')
local resty_backend = require('resty.http_ng.backend.resty')
local cache_backend = require('resty.http_ng.backend.cache')

-- cache_backend.new(resty_backend) would use default cache store, but you can customise its size
local backend = cache_backend.new(resty_backend, { cache_store = cache_store.new(100) })
local http_client = http_ng.new({ backend = backend })
```

### DIFF
--- a/apicast/src/resty/http_ng.lua
+++ b/apicast/src/resty/http_ng.lua
@@ -90,7 +90,7 @@ http.method = function(method, client)
     local req_params = get_request_params(method, client, url, options)
     local req = http.request.new(req_params)
 
-    return client.backend.send(req)
+    return client.backend:send(req)
   end
 end
 
@@ -111,7 +111,7 @@ http.method_with_body = function(method, client)
     req_params.body = body
     local req = http.request.new(req_params)
 
-    return client.backend.send(req)
+    return client.backend:send(req)
   end
 end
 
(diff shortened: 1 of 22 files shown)


apicast/src/resty/http_ng.lua
```diff
@@ -87,7 +87,7 @@ http.method = function(method, client)
     local req_params = get_request_params(method, client, url, options)
     local req = http.request.new(req_params)
 
-    return client.backend.send(req)
+    return client.backend:send(req)
```
macejmic: what's the difference between `:` and `.` ?

mikz: Hah! `:` is syntactic sugar for passing self.

So these are the same code:

```lua
object:method(param)
object.method(object, param)
```

This works in the function definition too:

```lua

function _M.method(self, param)
end

function _M:method(param)
end
```

Passing `self` It is the only way how a function can know about the object being called on.

PR #358, opened 2017-04-28
Oauth2 extensions pili

mpguerra:


apicast/src/oauth/apicast_oauth.lua
```diff
@@ -34,25 +34,23 @@ end
 
 function _M.extract_params()
   local params = {}
-  local header_params = ngx.req.get_headers()
+  local auth
 
-  params.authorization = {}
-
-  if header_params['Authorization'] then
-    params.authorization = re.split(ngx.decode_base64(re.split(header_params['Authorization']," ", 'oj')[2]),":", 'oj')
```
kevprice83: Why is the base64 decode function removed here? http_authorization module does this?

apicast/src/oauth/apicast_oauth.lua
```diff
@@ -78,13 +76,13 @@ function _M.respond_and_exit(status, body, headers)
   ngx.exit(ngx.HTTP_OK)
 end
 
-function _M.respond_with_error(status, message)
+function _M.respond_with_error(status, message, description)
```
kevprice83: difference between message and description?

apicast/src/oauth/apicast_oauth.lua
```diff
@@ -412,27 +392,26 @@ function _M.callback()
...
   end
-  
+
   code = ok
   ngx.header.content_type = "application/x-www-form-urlencoded"
   return ngx.redirect( client_data.redirect_uri .. "?code="..code.."&state=" .. (client_data.state or ""))
-
 end
 
 return _M
```
kevprice83: newline necessary at the end of each file?

apicast/src/oauth/apicast_oauth.lua
mpguerra: correct

apicast/src/oauth/apicast_oauth.lua
mpguerra: see rfc: https://tools.ietf.org/html/rfc6749#section-5.2 message should really be called error as per the rfc and then you can also have error_description but this latter one is optional

apicast/src/oauth/apicast_oauth.lua
mpguerra: dunno, luacheck didn't complain :)

apicast/conf/nginx.conf
```diff
@@ -4,6 +4,7 @@ env REDIS_URL;
 env RESOLVER;
 env BACKEND_ENDPOINT_OVERRIDE;
 env RHSSO_ENDPOINT;
+env APICAST_OAUTH_ACCESS_TOKEN_TTL;
```
mikz: All variables starting with `APICAST_` all forwarded  automatically.

PR #363, opened 2017-05-09
Fixes from review

mpguerra:


apicast/src/oauth/apicast_oauth.lua
```diff
@@ -145,7 +145,7 @@ end
 local function generate_access_token(client_id)
   local token = ts.sha1_digest(tostring(random.bytes(20, true)) .. client_id)
 
-  return { ["access_token"] = token, ["token_type"] = "bearer", ["expires_in"] = env.get('APICAST_OAUTH_ACCESS_TOKEN_TTL') or 604800 }
+  return { [access_token] = token, [token_type] = "bearer", [expires_in] = env.get('APICAST_OAUTH_ACCESS_TOKEN_TTL') or 604800 }
```
mikz: This is using `access_token` variable which is global and undefined. It should be `{ access_token = token }`.

mikz:
I guess we should skip the changelog entries between branches and require them just for master.

apicast/src/oauth/apicast_oauth.lua
mpguerra: yes, sorry, clearly still asleep... :/

PR #375, opened 2017-06-12
fixes typo to extract Redis port from REDIS_URL

### LINKED ISSUE #374: Redis port cannot be extracted from REDIS_URL environment variable
Small typo in the code [here](https://github.com/3scale/apicast/blob/3.0-stable/apicast/src/threescale_utils.lua#L147) means that the following error occurs when setting the Redis instance to connect to APIcast via `REDIS_URL`

```
[error] 29#29: *39 lua entry thread aborted: runtime error: /opt/app-root/src/src/threescale_utils.lua:153: bad argument #2 to 'connect' (number expected, got string)
```

kevprice83:
A typo in the code meant the Redis port number could not be extracted from `REDIS_URL` fixes #374

### DIFF
--- a/apicast/src/threescale_utils.lua
+++ b/apicast/src/threescale_utils.lua
@@ -144,7 +144,7 @@ function _M.connect_redis(options)
   opts.timeout = options and options.timeout or redis_conf.timeout
 
   local host = opts.host or env.get('REDIS_HOST') or "127.0.0.1"
-  local port = opts.prot or env.get('REDIS_PORT') or 6379
+  local port = opts.port or env.get('REDIS_PORT') or 6379
 
   local red = redis:new()
 


kevprice83:
Probably we should include some tests for this also...

kevprice83:
@hallelujah yes I can write some tests for that soon

mikz:
Yep. This definitely needs a test case as it was horribly broken. But the test case will also probably require refactor to split the options & connect into two different phases so we don't need redis on different port in the tests.

mikz:
@kevprice83 plan to finish this soon?

kevprice83:
@mikz I can take a look this week and start writing some tests. Will aim for end of next week if that sounds reasonable to you?

mikz:
Release scheduled for 21.8. So the end of the next week sounds good.

PR #402, opened 2017-08-21
Fixing Fixing headline formatting.

spier:
Not sure what was up with those hashtags before but the "Guideline Summary" did not get displayed correct as a headline in github.


spier:
I only changed a single line in a .md file so I don't think that any of these failing checks are actually related to my change. Maybe octobot just got stricter since it evaluated this file last :)

PR #404, opened 2017-08-24
[example] adds example for collecting response metrics

nburkley:
We had a requirement to collect metrics on responses being returned to a user. 

We couldn't find any existing solution to guide us so we build a custom module and config. We've created an example from our solution in the hopes that it helps someone else with a similar need.

I used [examples/custom-config](https://github.com/3scale/apicast/tree/master/examples/custom-config) and [examples/cors](https://github.com/3scale/apicast/tree/master/examples/cors) as templates when putting this together. 

Any feedback is welcome!

### DIFF
--- a/examples/response-metrics/apicast_response_metrics.lua
+++ b/examples/response-metrics/apicast_response_metrics.lua
@@ -1,0 +1,66 @@
+--- Custom APICAST_MODULE which overrides post_action from original
+--- https://github.com/3scale/apicast/blob/3.0-stable/apicast/src/apicast.lua
+
+-- load and initialize the parent module
+local apicast = require('apicast').new()
+
+local _M = { _VERSION = '3.0.0', _NAME = 'APIcast with response metrics' }
+local mt = { __index = setmetatable(_M, { __index = apicast }) }
+
+function _M.new()
+  return setmetatable({}, mt)
+end
+
+local function send_count_metrics()
+  -- get the document count from the custom response header
+  local document_count = ngx.resp.get_headers()['x-document-count'];
+  local user_key = ngx.req.get_headers()['user_key'];
+
+  -- only report metrics if the response was a success and the custom header exists
+  if ngx.status == ngx.HTTP_OK and document_count then
+    ngx.log(ngx.INFO, '[3scale-metrics] sending metric to 3scale document-count: ', document_count)
+    local report = ngx.location.capture("/report_metric",
+      {
+        args = {
+          user_key    = user_key,
+          metric_name = 'document_count',
+          count       = document_count
+        },
+        copy_all_vars = true
+      }
+    );
+
+    if report.status == ngx.HTTP_ACCEPTED  then
+      ngx.log(ngx.INFO, '[3scale-metrics] document_count metric succeeded ', report.status)
+    else
+      ngx.log(ngx.WARN, '[3scale-metrics] document_count metric update failed. status: ', report.status)
+    end
+  end
+end
+
+function _M:post_action()
+  local request_id = ngx.var.original_request_id
+  local post_action_proxy = self.post_action_proxy
+
+  if not post_action_proxy then
+    return nil, 'not initialized'
+  end
+
+  -- send custom metrics if this is a document search
+  if ngx.var.request_uri:match '^/v1/documents' then
+    send_count_metrics()
+  end
+
+  local p = ngx.ctx.proxy or post_action_proxy[request_id]
+
+  post_action_proxy[request_id] = nil
+
+  if p then
+    return p:post_action()
+  else
+    ngx.log(ngx.INFO, 'could not find proxy for request id: ', request_id)
+    return nil, 'no proxy for request'
+  end
+end
+
+return _M
--- a/examples/response-metrics/response_metrics.conf
+++ b/examples/response-metrics/response_metrics.conf
@@ -1,0 +1,24 @@
+location = /report_metric {
+  internal;
+  resolver 8.8.8.8;
+
+  set $user_key $arg_user_key;
+  set $metric_name $arg_metric_name;
+  set $document_count $arg_count;
+  set $user_credentials transactions[0][user_key]=$user_key;
+  set $metrics transactions[0][usage][$metric_name]=$document_count;
+  set $reporting_url https://su1.3scale.net:443;
+  set $path /transactions.xml?$backend_authentication_type=$backend_authentication_value&service_id=$service_id&$metrics&$user_credentials;
+
+  proxy_pass_request_headers off;
+  proxy_http_version 1.1;
+  proxy_set_header  Host  "$backend_host";
+  proxy_set_header  User-Agent "$user_agent";
+  proxy_set_header  X-3scale-User-Agent "$deployment";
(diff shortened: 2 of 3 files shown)


examples/response-metrics/response_metrics.conf
```diff
@@ -0,0 +1,24 @@
+location = /report_metric {
+  internal;
+  resolver 8.8.8.8;
```
nburkley: Ideally, we would not have to define a resolver here, but unlike the other conf blocks, the default resolver doesn't appear to be picked up here.

examples/response-metrics/response_metrics.conf
```diff
@@ -0,0 +1,24 @@
...
+
+  set $user_key $arg_user_key;
+  set $metric_name $arg_metric_name;
+  set $document_count $arg_count;
+  set $user_credentials transactions[0][user_key]=$user_key;
+  set $metrics transactions[0][usage][$metric_name]=$document_count;
+  set $reporting_url https://su1.3scale.net:443;
```
nburkley: We tried using `$backend_endpoint` here, but it's value was `"backend_upstream"` which resulted in a peering lookup error `failed to set current backend peer: missing peers while connecting to upstream`.

nburkley:
Hi @mikz or other maintainers - any feedback on this PR?

mikz:
I plan to take a look and suggest some improvements. Surely this week. Sorry for taking so long. 

5. 9. 2017 v 16:00, Niall Burkley <notifications@github.com>:

nburkley:
No worries, thanks @mikz!

mikz:
@nburkley I think this is very important use case and we could try to build it inside the gateway.

The plan would be following:

1. define header(s) that APIcast would read to see metrics to report
  For example `X-APIcast-Report-Metrics: Documents, Size, Time`.
  And send headers `X-APIcast-Metric-Documents: 14`, `X-APIcast-Metric-Size: 10024`, `X-APIcast-Metric-Time: 234 `.

2. define header_filter phase that will read those headers and add them to the context to be reported by the post_action

3. extend https://github.com/3scale/apicast/pull/290 to support report call

4. change the post action to check if there are any reports and push them to the 3scale backend


But back to the example.
I'd suggest to try to touch the nginx configuration as little as possible and do everything in lua. There is already backend client that does authrep. Adding there a report method is trivial. Then this example could be simplified to just use the [backend_client](https://github.com/3scale/apicast/blob/master/apicast/src/backend_client.lua) with some `ngx.timer.at(0)` to run it in the background.

I think this is good use case and we will have to improve the way of writing such customizations by improving the foundations.


I can either merge the example as it is and improve it later when those new APIs are available or try to work on this PR as I create those APIs and merge it as complete example. 
I think the second approach could be more useful if you'd like to clean up your custom module.

With the backend_client all the nginx customization could go away and just leave the custom lua module.

nburkley:
@mikz Thanks for the helpful feedback. 

I'm very happy to hear you're considering integrating logic like this directly into the gateway. Ideally, we'd prefer to customize our 3scale gateway a little as possible and handle all our business logic in our application layer or the 3scale admin, so enhancements like this would be very welcome.

> There is already backend client that does authrep.

We did look at using this initially, but we would like our API limits to be "soft limits" where API access is not disrupted when the limit is hit. The authrep call returns a `409` response for metrics that have reached a limit. While we can do this call in the background and ignore the response, it doesn't log metrics once the limit is reached - something that the Report endpoint does do.

From what I can see we would either need to modify the [backend_clint](https://github.com/3scale/apicast/blob/master/apicast/src/backend_client.lua) to include a reporting call or have a custom nginx conf as we're currently doing.

> I can either merge the example as it is and improve it later when those new APIs are available or try to work on this PR as I create those APIs and merge it as a complete example.

I don't mind waiting to include this as a complete example. As I mentioned, I created this PR because we were having trouble finding the 'correct' way to do this ourselves and thought it might be helpful for someone else in a similar situation. If you think this example will make more sense once the other API changes have been made, then I'm happy to wait for a more complete example. 

In the meantime, if there are any other changes or improvements I can make to this, I'm happy to help.

examples/response-metrics/apicast_response_metrics.lua
```diff
@@ -0,0 +1,66 @@
...
+
+local function send_count_metrics()
+  -- get the document count from the custom response header
+  local document_count = ngx.resp.get_headers()['x-document-count'];
+  local user_key = ngx.req.get_headers()['user_key'];
+
+  -- only report metrics if the response was a success and the custom header exists
+  if ngx.status == ngx.HTTP_OK and document_count then
+    ngx.log(ngx.INFO, '[3scale-metrics] sending metric to 3scale document-count: ', document_count)
+    local report = ngx.location.capture("/report_metric",
```
kevprice83: @nburkley as @mikz pointed out it would be better to do this in lua. You could make 2 requests here to 3scale, first an authorise and then a report. This is done in the post_action phase and doesn't impact the latency on the client request. If the authorize returns a 409 you can allow the call through but if it returns a 403 then the call should be rejected, then you can report based on the authorize response. The authorize already exists in the backend_client so you would just need to add the report call there.

examples/response-metrics/apicast_response_metrics.lua
```diff
@@ -0,0 +1,66 @@
...
+function _M:post_action()
+  local request_id = ngx.var.original_request_id
+  local post_action_proxy = self.post_action_proxy
+
+  if not post_action_proxy then
+    return nil, 'not initialized'
+  end
+
+  -- send custom metrics if this is a document search
+  if ngx.var.request_uri:match '^/v1/documents' then
```
kevprice83: Better if this regex is not hardcoded. One suggestion could be to define another environment variable that consists of the regex pattern(s) to match against. `APICAST_REPORT_METRIC_MATCHER` for example... or at the very least define a variable with the regex pattern outside the scope of the function which can then be used anywhere within this custom module.

examples/response-metrics/apicast_response_metrics.lua
nburkley: Sure, I can do that - I was trying to avoid modifying the [backend_clint](https://github.com/3scale/apicast/blob/master/apicast/src/backend_client.lua) code - hence the call to an nginx conf.

I didn't have any auth-checks as I as going with the assumption that if the request made it thorugh our stack successfully it was already authorized - I can add this though.

examples/response-metrics/apicast_response_metrics.lua
nburkley: Sure, make sense, I'll tidy this up.

nburkley:
thanks for the feedback @kevprice83  - I'll ping you when I've updated the PR.

mikz:
This is now doable by [a policy](https://github.com/3scale/apicast/blob/3.2-stable/doc/policies.md).
Our backend client supports `:report` method (https://github.com/3scale/apicast/commit/f388d17afed29263babc560708be70ac19400e3f) so a policy can call it in `post_action` phase for example.

nburkley:
Thanks @mikz Closing this in favour of https://github.com/3scale/apicast/commit/f388d17afed29263babc560708be70ac19400e3f

PR #418, opened 2017-09-01
[bin] improve busted runner

mikz:
* it is in Perl instead of Shell
* it is aware of current working directory
* can run tests in different folders
* can load code in different folders

example:

```shell
bin/busted -C examples/cors
```

or from within different directory:

```shell
../../bin/busted
```

### DIFF
--- a/bin/busted
+++ b/bin/busted
@@ -1,3 +1,21 @@
-#!/usr/bin/env sh
+#!/usr/bin/env perl
 
-exec resty --http-include spec/fixtures/echo.conf bin/busted.lua "$@"
+use strict;
+use warnings FATAL => 'all';
+
+use File::Basename;
+use Cwd;
+
+my $dirname = dirname(__FILE__);
+my $cwd = getcwd();
+
+chdir "$dirname/..";
+
+my $apicast = getcwd();
+
+exec '/usr/bin/env', 'resty',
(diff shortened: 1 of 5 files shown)


bin/busted
```diff
@@ -1,3 +1,18 @@
-#!/usr/bin/env sh
+#!/usr/bin/env perl
```
maneta: Is always good to:

```
use warnings;
use strict;
```

It protects you of a lot of stuff (https://perlmaven.com/always-use-warnings)

mikz: Thx! Push'd.

bin/busted
```diff
@@ -1,3 +1,21 @@
...
+use Cwd;
+
+my $dirname = dirname(__FILE__);
+my $cwd = getcwd();
+
+chdir "$dirname/..";
+
+my $apicast = getcwd();
+
+exec '/usr/bin/env', 'resty',
```
maneta: You are sure you want just exec Right?
 
The big difference is that system() creates a fork process and waits to see if the command succeeds or fails - returning a value. exec() does not return anything, it simply executes the command.

mikz: Well exec replaces the current process and that is exactly what I want.

mikz: This is just a wrapper to prepare some flags and execute different command. So to not have to handle error codes, input, output etc it is just nice to exec.

PR #422, opened 2017-09-14
[brew] add Brewfile

mikz:
* defines dependencies
* can be installed by `brew bundle`


ddcesare:
Oh, this Brewfile you meant. I thought you were saying something about RH brew hehe.

PR #425, opened 2017-09-15
[ci] profiling

mikz:


mikz:
Closing for now because we'd need whole VM instead of just docker to run profiling.

PR #429, opened 2017-09-19
Apicast Example : dynamic router

nmasse-itix:
Hi,

I had a request to write a custom module for Apicast that dynamically routes API requests to the right backend, based on an HTTP Header of the request. 

The use case is for example:
- a Load Balancer in front of Apicast sets an HTTP Header based on the location of the request (internal / external or prod / dev) 
- based on that information, the API Gateway has to route the request to the correct backend (internal / external, etc.).

Of course, to route the request to right backend means having a sort of convention for urls in each environment or having a service catalog. I chose the later approach. 

So I implemented a sample service catalog using plain nginx configuration.

I think it could be a nice sample code for users having the same requirement. 

Looking forward to hear your feedback on this idea. :)

Best regards

### DIFF
--- a/examples/dynamic-router/README.md
+++ b/examples/dynamic-router/README.md
@@ -1,0 +1,47 @@
+# Dynamic Routing for Apicast
+
+## Introduction
+
+This project provides a dynamic routing module for Apicast. It routes the
+API request to the appropriate backend, based on an HTTP Header of the request.
+
+A sample use case could be:
+ - A Load Balancer in front of Apicast identifies the source of the request (`internal`/`external` or `dev`/`prod`)
+ - The LB add the corresponding header (`x-env: dev` or `x-env: prod` for instance)
+ - Based on this header, the Apicast routes the API request to the corresponding backend
+
+The API Backends are discovered by querying a Service Catalog. A sample service
+catalog is given with this project.
+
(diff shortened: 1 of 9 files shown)


nmasse-itix:
I updated my code to comply with code climate

mikz:
@nmasse-itix it should be possible to implement this as pure policy without any nginx configuration changes. 
You can take a look at the upstream policy that does routing by the Request URI: https://github.com/3scale/apicast/tree/541b579ec170d8166b5476193118f963e6351983/gateway/src/apicast/policy/upstream


Also it is possible this might be in the end solved by https://github.com/3scale/apicast/issues/744 as all the matching functionality would be shared by all policies.

mikz:
No rush. It has been opened for 9 months, so it can wait a few more :) 
Maybe we will solve #744 by then and it will be basically done by the upstream policy already.

davidor:
@nmasse-itix I believe that this can be achieved by combining the conditional policy with the upstream one as pointed out by @mikz .

I included an integration test that shows how it works in https://github.com/3scale/apicast/pull/819

Can you confirm that this PR can be closed ?

eloycoto:
Spoke with @nmasse-itix and this is already implemented by the routing policy.

PR #431, opened 2017-09-26
Add dockerized development environment

davidor:
This PR adds a Dockerfile based on the builder image that installs all the dependencies required to run both the busted and the Test::Nginx tests.

The docker-compose file mounts a volume so the project source code is available in the image and sync'ed  with the local apicast directory. This way we can develop locally using our preferred environment and still be able to run the tests inside the docker container.

This PR also adds a helper bash script to run the tests.

### DIFF
--- a/Dockerfile-development
+++ b/Dockerfile-development
@@ -1,0 +1,50 @@
+FROM quay.io/3scale/s2i-openresty-centos7:master
+LABEL maintainer="dortiz@redhat.com"
+
+USER root
+
+# 'default' user is defined in the base image
+ARG USER_NAME=default
+ARG USER_HOME=/home/centos
+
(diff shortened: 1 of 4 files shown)


Dockerfile-development
```diff
@@ -0,0 +1,50 @@
+FROM quay.io/3scale/s2i-openresty-centos7:master
+MAINTAINER David Ortiz <dortiz@redhat.com>
```
mikz: I think we can get rid of the `MAINTAINER` in favour of `LABEL maintainer`

https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

davidor: I didn't know it was deprecated. 
Fixed :+1:

PR #432, opened 2017-09-26
[THREESCALE-360] fix loading configuration with null values

mikz:
OIDC configuration can be null, that does not work like nil

### DIFF
--- a/spec/configuration_store_spec.lua
+++ b/spec/configuration_store_spec.lua
@@ -11,6 +11,14 @@ describe('Configuration Store', function()
 
       assert.equal(service, store:find_by_id('42'))
     end)
+
+    it('stores config with OIDC', function()
+      local store = configuration.new()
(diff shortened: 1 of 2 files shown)


spec/configuration_store_spec.lua
```diff
@@ -11,6 +11,14 @@ describe('Configuration Store', function()
 
       assert.equal(service, store:find_by_id('42'))
     end)
+
+    it('stores config with OIDC', function()
```
davidor: Shouldn't we add a test for the opposite case too? I don't see any tests for this class with `oidc[i] ~= ngx.null`.

mikz: Those are tested in the integration test that actually verifies it is used correctly. 

https://github.com/3scale/apicast/blob/master/t/019-apicast-oidc.t

This is very scoped regression test.

davidor: :+1:

PR #444, opened 2017-10-06
[ci] enable junit

mikz:
* output busted tests as junit
* output Test::Nginx as junit
* upload them as artifacts
* process them on circleci

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,7 @@ jobs:
       - run: make prove-docker
   build:
     docker:
-      - image: quay.io/3scale/s2i-openresty-centos7:1.11.2.5-1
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -37,19 +37,26 @@ jobs:
       - run: make prove-docker
   build:
     docker:
-      - image: quay.io/3scale/s2i-openresty-centos7:1.11.2.5-1
+      - image: quay.io/3scale/s2i-openresty-centos7:junit
```
mikz: This needs to be changed to new released version.

mikz: Will do a release after https://github.com/3scale/s2i-openresty/pull/37 gets merged.

PR #448, opened 2017-10-09
Make oauth tokens TTL configurable

### LINKED ISSUE #190: [OAuth] Make access token ttl configurable
Currently, the access token ttl is hardcode at 7 days (604800 seconds.)

Ideally we should be able to override this without changing the lua code.

davidor:
Fixes #190

### DIFF
--- a/t/005-apicast-oauth.t
+++ b/t/005-apicast-oauth.t
@@ -308,6 +308,7 @@ Location: http://example.com/redirect\?code=\w+&state=clientstate
 --- main_config
   env REDIS_HOST=$TEST_NGINX_REDIS_HOST;
   env RESOLVER=$TEST_NGINX_RESOLVER;
+  env APICAST_OAUTH_TOKENS_TTL=123;
 --- http_config
   lua_package_path "$TEST_NGINX_LUA_PATH";
   init_by_lua_block {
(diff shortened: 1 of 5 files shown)


t/005-apicast-oauth.t
```diff
@@ -791,6 +805,86 @@ GET /t
 Content-Type: application/json
 --- error_code: 200
 --- response_body
+{"token_type":"bearer","expires_in":123,"access_token":"token"}
+--- no_error_log
+[error]
+
+=== TEST 18: default token TTL
+When a default TTL is not specified, it applies a default of 7 days (604800 s)
```
mikz: Can you try with empty value? That is common OpenShift fuckup.

davidor: Good point. I added a new test (`TEST 19`) which is basically a copy-pasta of `TEST 18`. The only difference is an env. Any ideas to avoid this kind of duplication in Test::Nginx ?

mikz: No idea :( It really sucks to have this kind of duplication.

mikz: Maybe would be useful to test it on lower level - in busted.

davidor: Agreed. I thought about that, but then I saw that there are _0_ unit tests for oauth, so I thought that it would be better to just complete the existing integration tests for now. Maybe this is something we could address in a future PR.

mikz: Definitely. 👍

PR #449, opened 2017-10-10
template nginx configuration with liquid

mikz:
Replace bash script with a proper Lua CLI and template the nginx configuration by liquid.

This wouldn't be possible without awesome liquid interpreter https://github.com/chenxianyu2015/liquid-lua. Thanks.

## Before `bin/apicast -h`
```
Usage bin/apicast
  -h Show this help
  -c <file> Path to custom config file (JSON).
  -d Daemonize
  -v Increase verbosity (can be repeated)
  -i Cache configuration for N seconds. Using 0 will reload on every request (not for production).
  -w <workers> Number of worker processes to start.
  -m <on|off> Whether to start worker processes. Only for development.
  -s <signal> Send signal to a master process: stop, quit, reopen, reload
  -p <pid> Path to the PID file.
  -b Load configuration on boot.
  -e Deployment environment. Can be staging or production.
```

## After `bin/apicast -h`

```
Usage: apicast-cli start [OPTIONS]

Start APIcast

Options:
   --template <template> Nginx config template. (default: conf/nginx.conf.liquid)
   -e <environment>, --environment <environment>
                         Deployment to start.
   --dev                 Start in development environment
   -t, --test            Test the nginx config
   --debug               Debug mode. Prints more information.
   -c <configuration>, --configuration <configuration>
                         Path to custom config file (JSON)
   -d, --daemon          Daemonize.
   -w <workers>, --workers <workers>
                         Number of worker processes to start.
   -p <pid>, --pid <pid> Path to the PID file.
   -b, --boot            Load configuration on boot.
   -l, --lazy            Load configuration on demand.
   -i <refresh_interval>, --refresh-interval <refresh_interval>
                         Cache configuration for N seconds. Using 0 will reload on every request (not for production).
   -v, --verbose         Increase logging verbosity (can be repeated).
   -q, --quiet           Decrease logging verbosity.
   --log-level <log_level>
                         Set log level (default: warn)
   --log-file <log_file> Set log file (default: stderr)
   -h, --help            Show this help message and exit.

      Example: apicast start --dev
        This will start APIcast in development mode.
```


apicast/Roverfile.lock
```diff
@@ -1,8 +1,12 @@
+ansicolors 1.0.2-3
+argparse 0.5.0-1
 busted 2.0.rc12-1
 dkjson 2.5-2
 inspect 3.1.0-1
 ldoc 1.4.6-2
+liquid scm-1
```
mikz: We will need to work around this before https://github.com/luarocks/luarocks/issues/739 is fixed. Probably by doing the whole clone and copy in rover.

mikz: Fixed by https://github.com/3scale/lua-rover/pull/2

apicast/Roverfile.lock
```diff
@@ -1,20 +1,23 @@
...
+dkjson 2.5-2|
+inspect 3.1.0-1|
+ldoc 1.4.6-2|
+liquid scm-1|811a73e38fdd9fdea116be4baf310ca326b96c77
+lua-resty-env 0.4.0-1|
+lua-resty-execvp 0.1.0-1|
+lua-resty-http 0.11-0|
+lua-resty-iputils 0.3.0-1|
+lua-resty-jwt 0.1.11-0|
+lua-resty-repl 0.0.6-0|3878f41b7e8f97b1c96919db19dbee9496569dda
```
davidor: Why the hash in this one? It has a version.

mikz: Yes, but version can still point to a scm to get the source. Depends if it is a rock or a rockspec.

rock has a source packaged in the bundle but rockspec has to download it from somewhere
in this case repl has only rockspec and that points to a git tag https://luarocks.org/modules/saksmlz/lua-resty-repl/0.0.6-0

apicast/conf/nginx.conf.liquid
```diff
@@ -0,0 +1,92 @@
+env REDIS_HOST;
```
davidor: Is `apicast/conf/nginx.conf` still needed?

davidor: There's a couple of settings present in the original config but not on this template:
```
  lua_max_running_timers 2048;
  lua_socket_pool_size 512;
```

mikz: aha! they probably lost during the rebase, good catch!

the plain nginx.conf should not be needed

mikz: Fixed both (removed the old file and added those missing directives).

PR #450, opened 2017-10-10
Policy chain - first version

mikz:
Early prototype of how we could chain policies.

Policies can export data that is stored in read only structure and resolved in order they are defined in.
So values can be overlaid, but not overridden and only the latest one is being used.
This would allow overriding some internal structures being passed between policies.

Context object is passed to each policy phase invocation method. This context object has last layer mutable so policies can pass data.

Policy chain is itself a policy. All phase methods are compatible. So policy chain can contain policy chains.
This is used to split the chain into two: global and local. Global chain is responsible for loading configuration and local can be created based on downloaded configuration. 

# TODO

- [x] ~ figure out how installed policies can change the nginx template ~ **https://github.com/3scale/apicast/issues/477**
  - the list of policies enabled globally can be defined by some global config
  - the important piece is how would the policy be able to extend the nginx template
  - consider how policies would interact with each other and blocks that could extend
  - decide if it is template manipulation (like something like https://github.com/spree/deface ) or some dsl compiled to nginx config

### DIFF
--- a/apicast/src/configuration_loader.lua
+++ b/apicast/src/configuration_loader.lua
@@ -55,7 +55,7 @@ local function ttl()
 end
 
 function _M.global(contents)
-  local module = require('module')
+  local module = require('policy_chain')
 
   return _M.configure(module.configuration, contents)
 end
--- a/apicast/src/configuration_loader.lua
+++ b/apicast/src/configuration_loader.lua
@@ -55,9 +55,9 @@ local function ttl()
 end
 
 function _M.global(contents)
-  local module = require('policy_chain')
+  local context = require('executor'):context()
 
-  return _M.configure(module.configuration, contents)
+  return _M.configure(context.configuration, contents)
 end
 
 function _M.configure(configuration, contents)
--- a/apicast/src/executor.lua
+++ b/apicast/src/executor.lua
@@ -1,0 +1,35 @@
+local setmetatable = setmetatable
+
+local policy_chain = require('policy_chain')
+local linked_list = require('linked_list')
+
+local _M = { }
+
+local mt = { __index = _M }
+
+-- forward all policy methods to the policy chain
+for i=1, #(policy_chain.PHASES) do
+    local phase_name = policy_chain.PHASES[i]
+
+    _M[phase_name] = function(self, ...)
+        return self.policy_chain[phase_name](self.policy_chain, self:context(), ...)
+    end
+end
+
+function _M.new()
+    local local_chain = policy_chain.build()
+
+    local load_configuration = policy_chain.load('policy.load_configuration', local_chain)
+
+    local global_chain = policy_chain.build({ load_configuration, local_chain })
+
+    return setmetatable({ policy_chain = global_chain }, mt)
+end
+
+function _M:context()
+    local config = self.policy_chain:export()
+
+    return linked_list.readwrite({}, config)
+end
+
+return _M.new()
--- a/apicast/src/executor.lua
+++ b/apicast/src/executor.lua
@@ -1,18 +1,23 @@
-local setmetatable = setmetatable
+local resty_env = require('resty.env')
+
+if resty_env.get('APICAST_MODULE') then
+  return require('module')
+end
 
 local policy_chain = require('policy_chain')
+local policy = require('policy')
 local linked_list = require('linked_list')
+
+local setmetatable = setmetatable
 
 local _M = { }
 
 local mt = { __index = _M }
 
 -- forward all policy methods to the policy chain
-for i=1, #(policy_chain.PHASES) do
-    local phase_name = policy_chain.PHASES[i]
-
-    _M[phase_name] = function(self, ...)
-        return self.policy_chain[phase_name](self.policy_chain, self:context(phase_name), ...)
+for _,phase in policy:phases() do
+    _M[phase] = function(self, ...)
+        return self.policy_chain[phase](self.policy_chain, self:context(phase), ...)
     end
 end
 
--- a/apicast/src/executor.lua
+++ b/apicast/src/executor.lua
@@ -17,8 +17,7 @@ for i=1, #(policy_chain.PHASES) do
 end
 
 function _M.new()
-    local local_chain = policy_chain.build()
-
+    local local_chain = policy_chain.load('policy.local_chain')
     local load_configuration = policy_chain.load('policy.load_configuration', local_chain)
 
     local global_chain = policy_chain.build({ load_configuration, local_chain })
--- a/apicast/src/executor.lua
+++ b/apicast/src/executor.lua
@@ -12,23 +12,43 @@ for i=1, #(policy_chain.PHASES) do
     local phase_name = policy_chain.PHASES[i]
 
     _M[phase_name] = function(self, ...)
-        return self.policy_chain[phase_name](self.policy_chain, self:context(), ...)
+        return self.policy_chain[phase_name](self.policy_chain, self:context(phase_name), ...)
     end
 end
 
 function _M.new()
-    local local_chain = policy_chain.load('policy.local_chain')
-    local load_configuration = policy_chain.load('policy.load_configuration', local_chain)
-
-    local global_chain = policy_chain.build({ load_configuration, local_chain })
+    local global_chain = policy_chain.build({
+        'policy.load_configuration', 'policy.find_service', 'policy.local_chain'
+    })
 
     return setmetatable({ policy_chain = global_chain }, mt)
 end
 
-function _M:context()
-    local config = self.policy_chain:export()
+local function build_context(executor)
+    local config = executor.policy_chain:export()
 
     return linked_list.readwrite({}, config)
+end
+
+local function shared_build_context(executor)
+    local ctx = ngx.ctx or {}
+    local context = ctx.context
+
+    if not context then
+        context = build_context(executor)
+
+        ctx.context = context
+    end
+
+    return context
+end
+
+function _M:context(phase)
+    if phase == 'init' then
+        return build_context(self)
+    end
+
+    return shared_build_context(self)
 end
 
 return _M.new()
--- a/apicast/src/executor.lua
+++ b/apicast/src/executor.lua
@@ -4,12 +4,6 @@
 -- when calling the policy chain methods. This 'context' contains information
 -- shared among policies.
 
-local resty_env = require('resty.env')
-
-if resty_env.get('APICAST_MODULE') then
-  return require('module')
-end
-
 local policy_chain = require('policy_chain')
 local policy = require('policy')
 local linked_list = require('linked_list')
--- a/apicast/src/executor.lua
+++ b/apicast/src/executor.lua
@@ -1,3 +1,9 @@
+--- Executor module
+-- The executor has a policy chain and will simply forward the calls it
+-- receives to that policy chain. It also manages the 'context' that is passed
+-- when calling the policy chain methods. This 'context' contains information
+-- shared among policies.
+
 local resty_env = require('resty.env')
 
 if resty_env.get('APICAST_MODULE') then
@@ -54,6 +60,9 @@ local function shared_build_context(executor)
     return context
 end
 
+--- Shared context among policies
+-- @tparam string phase Nginx phase
+-- @treturn linked_list The context. Note: The list returned is 'read-write'.
 function _M:context(phase)
     if phase == 'init' then
         return build_context(self)
--- a/apicast/src/executor.lua
+++ b/apicast/src/executor.lua
@@ -12,6 +12,12 @@ local setmetatable = setmetatable
 
 local _M = { }
 
+local DEFAULT_POLICIES = {
+    'policy.load_configuration',
+    'policy.find_service',
+    'policy.local_chain'
+}
+
 local mt = { __index = _M }
 
 -- forward all policy methods to the policy chain
@@ -21,12 +27,12 @@ for _,phase in policy.phases() do
     end
 end
 
-function _M.new()
-    local global_chain = policy_chain.build({
-        'policy.load_configuration', 'policy.find_service', 'policy.local_chain'
-    })
+local function global_chain()
+    return policy_chain.build(DEFAULT_POLICIES)
+end
 
-    return setmetatable({ policy_chain = global_chain }, mt)
+function _M.new()
+    return setmetatable({ policy_chain = global_chain() }, mt)
 end
 
 local function build_context(executor)
--- a/apicast/src/executor.lua
+++ b/apicast/src/executor.lua
@@ -15,7 +15,7 @@ local _M = { }
 local mt = { __index = _M }
 
 -- forward all policy methods to the policy chain
-for _,phase in policy:phases() do
+for _,phase in policy.phases() do
     _M[phase] = function(self, ...)
         return self.policy_chain[phase](self.policy_chain, self:context(phase), ...)
     end
--- a/apicast/src/linked_list.lua
+++ b/apicast/src/linked_list.lua
@@ -1,0 +1,45 @@
+local setmetatable = setmetatable
+
+local _M = {
+
+}
+
+local noop = function() end
+
+
+local empty_t = setmetatable({}, { __newindex = noop })
+local __index = function(t,k)
+    return t.current[k] or t.next[k]
+end
+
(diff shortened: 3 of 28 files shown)


apicast/src/policy.lua
```diff
@@ -0,0 +1,26 @@
+local _M = {}
+
+local setmetatable = setmetatable
+local policy_chain = require('policy_chain')
```
davidor: How about requiring only what is needed? `require('policy_chain').PHASES`

apicast/src/policy.lua
```diff
@@ -0,0 +1,26 @@
...
+    local mt = { __index = policy }
+
+    function policy.new()
+        return setmetatable({}, mt)
+    end
+
+    for i=1,#(policy_chain.PHASES) do
+        policy[policy_chain.PHASES[i]] = noop
+    end
+
```
davidor: Should we also define `policy['export']`?

apicast/src/policy.lua
```diff
@@ -0,0 +1,26 @@
...
+    return policy, mt
```
davidor: I'd document what this method returns as I think returning the `mt` might be a bit counter-intuitive. I think we should document that it's returned so `policy.new` can be overriden. Any other reason?

apicast/src/policy/load_configuration.lua
```diff
@@ -0,0 +1,33 @@
+local setmetatable = setmetatable
+
+local _M, mt = require('policy').new()
```
davidor: We should probably call `new()` with a name. Otherwise, it will use the default one.

apicast/src/policy.lua
mikz: That would not really save anything. The whole file is executed and evaluated anyway. 
And it would require unrolling it back when we need second property from that module.

apicast/src/policy.lua
mikz: That is the only reason so far. I could also store it in `_M` like `_M.class` or `_M.mt`.  Not sure what is better though. 

But for sure has to be documented as this is public API for new policies.

mikz: I'll try to set the `mt` as metatable of the returned module. Then you could get it by `getmetatable`. Maybe it won't cause infinite loop :)

apicast/src/policy/load_configuration.lua
mikz: Good point!

apicast/src/policy.lua
mikz: In the end went with not returning it at all and letting the user to decide what to do with it. They should upvalue the new function to a variable and call it later.

apicast/src/configuration_loader.lua
```diff
@@ -55,9 +55,9 @@ local function ttl()
 end
 
 function _M.global(contents)
-  local module = require('module')
+  local context = require('executor'):context()
```
mikz: This will not work when `APICAST_MODULE` is set.

apicast/src/management.lua
```diff
@@ -1,7 +1,7 @@
 local _M = {}
 
 local cjson = require('cjson')
-local module = require('module')
+local context = require('executor'):context()
```
mikz: This won't work when `APICAST_MODULE` env is set.

apicast/src/executor.lua
```diff
@@ -0,0 +1,59 @@
+local resty_env = require('resty.env')
+
+if resty_env.get('APICAST_MODULE') then
+  return require('module')
```
mikz: Maybe we should do this in the apicast module itself? To keep the compatibility with the old module system. Basically allow people to replace the APIcast module. But we need to figure out how to allow them to require it but to outside return their module. 
Maybe this belongs to the policy_chain.build where it defaults to the apicast module?

spec/proxy_spec.lua
```diff
@@ -66,7 +66,7 @@ describe('Proxy', function()
     assert.same('function', type(proxy.post_action))
   end)
 
-  it('finds service by host', function()
+  pending('finds service by host', function()
```
mikz: This needs to be moved to the find service policy spec.

spec/proxy_spec.lua
```diff
@@ -75,7 +75,7 @@ describe('Proxy', function()
     assert.falsy(proxy:find_service('unknown'))
   end)
 
-  it('does not return old configuration when new one is available', function()
+  pending('does not return old configuration when new one is available', function()
```
mikz: This needs to be moved to the find service policy spec.

davidor:
There are some commits here already included in other PRs (curl in benchmarks, DEBUG=1, etc.)

apicast/src/policy.lua
```diff
@@ -0,0 +1,42 @@
+local _M = {
+    PHASES = {
```
davidor: Should this table be exposed? Users of this module only call `.phases()`.

apicast/src/policy.lua
```diff
@@ -0,0 +1,42 @@
...
+
+    for _, phase in _M:phases() do
+        policy[phase] = noop
+    end
+
+    return policy
+end
+
+function _M:phases()
+    return ipairs(self.PHASES)
```
davidor: Why not return `self.PHASES` directly?

apicast/src/policy.lua
mikz: True, no need for it.

apicast/src/policy.lua
mikz: Because it is always used as iterator I thought that it would be nice to return it. Then the caller for example could not modify this table and just iterate.

apicast/src/policy.lua
```diff
@@ -0,0 +1,42 @@
...
+
+local setmetatable = setmetatable
+local ipairs = ipairs
+
+local noop = function() end
+
+--- Initialize new policy
+-- Returns a new policy that you can extend however you want.
+-- @tparam string name Name of the new policy.
+-- @tparam string version Version of the new policy. Default value is 0.0
```
davidor: Currently, the code does not set a default value.
Do you see any reason to set 0.0 instead of nil?

apicast/src/policy.lua
davidor: Makes sense. Nice trick to make the table read-only.

apicast/src/policy.lua
mikz: I think it would be nice to be able to assume there is always a version.

apicast/src/linked_list.lua
```diff
@@ -0,0 +1,45 @@
...
+}
+
+local function linked_list(item, next, mt)
+    return setmetatable({
+        current = item or empty_t,
+        next = next or empty_t
+    }, mt)
+end
+
+local function readonly_linked_list(item, next)
```
davidor: Can't we receive a table and set __newindex = noop in the metatable instead?

apicast/src/policy_chain.lua
```diff
@@ -0,0 +1,89 @@
...
+local error = error
+local rawset = rawset
+local type = type
+local require = require
+local noop = function() end
+
+local linked_list = require('linked_list')
+local policy = require('policy')
+
+local _M = {
```
davidor: I think that this class is exposing lots of methods that it shouldn't expose. I think that only `build`, `export` and all the phases (access, rewrite, etc.) should be public. What do you think?

mikz: Yep, good idea.

davidor: Also, it'd be good to document the public methods. Specially, I think that the purpose of `export` is not obvious looking at this module in isolation.

apicast/src/policy/local_chain.lua
```diff
@@ -0,0 +1,36 @@
+local policy = require('policy')
+local Proxy = require('proxy')
+local _M = policy.new('Local Policy Chain')
+
+local default_chain = require('policy_chain').build()
```
davidor: There's no need to call `.build()` here. It's already done in the `policy_chain` module.

apicast/src/linked_list.lua
```diff
@@ -0,0 +1,45 @@
...
+
+
+local empty_t = setmetatable({}, { __newindex = noop })
+local __index = function(t,k)
+    return t.current[k] or t.next[k]
+end
+
+local ro_mt = {
+    __index = __index,
+    __newindex = noop,
```
davidor: I think this should raise instead of fail silently. I think it can be confusing for users of this module to run something like `t[k] = '123'` and see that it does not have any effect and does not raise any error.

mikz: Good point.

apicast/src/policy/local_chain.lua
mikz: Yeah I was hoping for that too, but this fails somehow. My guess is that the apicast module has some state and needs one copy in the memory to be used in both access and post_action. 
You can see the failing test when this is nil. But probably there are better ways to solve this. Just something to keep in mind.

apicast/src/executor.lua
```diff
@@ -0,0 +1,59 @@
...
+-- forward all policy methods to the policy chain
+for _,phase in policy:phases() do
+    _M[phase] = function(self, ...)
+        return self.policy_chain[phase](self.policy_chain, self:context(phase), ...)
+    end
+end
+
+function _M.new()
+    local global_chain = policy_chain.build({
+        'policy.load_configuration', 'policy.find_service', 'policy.local_chain'
```
davidor: How about extracting this table to a constant ?

mikz: We can. This would also be controlled by some global configuration file.

apicast/src/policy/find_service.lua
```diff
@@ -0,0 +1,73 @@
...
+    end
+    if found then break end
+  end
+
+  return found or ngx.log(ngx.WARN, 'service not found for host ', host)
+end
+
+local function find_service_cascade(configuration, host)
+  local found
+  local request = ngx.var.request
```
davidor: Should this module receive the request via `context` ?

mikz: Interesting idea. Guess that depends if changes like `req.set_uri` or `set_method` are propagated and are changing this variable. If yes, then policies can change this variable indirectly and it is all fine. But if this is immutable then it should be in the context, so other policies can override it.
And maybe it is just better to get rid of this variable and use more granular nginx apis. Because it just splits the string into method, path, etc. anyway https://github.com/3scale/apicast/blob/master/apicast/src/configuration.lua#L201-L204 so it can use req.get_method etc.

spec/policy/find_service_spec.lua
```diff
@@ -0,0 +1,108 @@
+local configuration = require('configuration')
+
+describe('find_service', function()
+  describe('.rewrite', function()
+    describe('when path routing is enabled', function()
```
davidor: I realized that I didn't test the case where no mapping rules are matched and find_service falls back to 'strict' mode.

spec/policy/find_service_spec.lua
```diff
@@ -0,0 +1,108 @@
...
+              pattern = '/def',
+              http_method = 'GET',
+              metric_system_name = 'hits',
+              delta = 2
+            } }
+          }
+        })
+
+        local configuration_store = require('configuration_store').new()
+        configuration_store:store({ services = { service_1, service_2 } })
```
davidor: I think it would be better to test with 3 services. Otherwise, we won't know if this works by coincidence because it just returns the first or the last inserted.

davidor: Done.

spec/policy/find_service_spec.lua
davidor: Done.

apicast/src/policy/find_service.lua
```diff
@@ -0,0 +1,73 @@
...
+    if found then break end
+  end
+
+  return found or find_service_strict(configuration, host)
+end
+
+function _M.new(...)
+  local self = new(...)
+
+  if configuration_store.path_routing then
```
davidor: How about receiving the configuration or even `path_routing_enabled` via context in the params? It would allow us to reduce coupling and would also simplify the tests.

mikz: Sounds good. But maybe we need something else to pass around to the new method. But maybe using the context would not be so bad.

apicast/src/policy/load_configuration.lua
```diff
@@ -0,0 +1,33 @@
+local _M = require('policy').new('Load Configuration')
+
+local configuration_loader = require('configuration_loader').new()
+local configuration_store = require('configuration_store')
+
+local new = _M.new
+
+function _M.new(...)
+    local policy = new(...)
+    policy.configuration = configuration_store.new()
```
davidor: As I commented above for a different module, I think it'd be good to inject the configuration.

mikz: Agree. We have no facility how to inject stuff to the `new` method. But definitely would be useful to pass something around.

apicast/src/policy/local_chain.lua
davidor: I'd call `.build()` with `{ 'apicast' }`. I know it's the default, but I think making it explicit makes the code easier to follow.
What do you think?

mikz: Yep. Possibly even removing the `{ 'apicast' }` default and doing it explicitly everywhere (hopefully just one place).

davidor:
I think now we have a good first version for supporting policy chains :+1:  
To complete it, this is a list of things I think we need. Although I'm not sure we should tackle all of them in this PR:
- Document all the public methods.
- Document the purpose of each of the classes involved (executor, policy, etc.). I think this would make the code easier to follow.
- For each of the phases of the policies defined, document what the policy expects to have available in the given 'context' and also, how the policy modifies the 'context'.
- Think about what we can pass store in the 'context' in the first links of the chain to avoid couplings like the ones I pointed out in previous comments.

mikz:
@davidor thanks. Great work! I think we should definitely write the docs in this PR. The rest can wait for another iteration.
I'd wait with documenting what is needed/exported from/to the context as possibly we would like to have some runtime verification. And decoupling can wait too. We explore that approach when writing more policies.

apicast/src/policy.lua
davidor: :+1:
`policy` no longer depends on `policy_chain`, so this no longer applies.

spec/proxy_spec.lua
davidor: Done.

t/apicast-policy-chains.t
```diff
@@ -0,0 +1,65 @@
...
+
+__DATA__
+
+=== TEST 1: custom policy chain
+This test uses the phase logger policy to verify that all of its phases are run
+when we use a policy chain that contains it. The policy chain also contains the
+normal apicast policy, so we can check that the authorize flow continues
+working. Notice that 'post_action' and 'log' are not ran. There's no
+post_action defined and log only runs when post_action runs.
+Also, init and init_worker are run before these tests, so what's logged in them
```
davidor: I didn't explain this well. init() and init_worker() do not run because I defined the custom policy as a local/service one instead of a global one. The service is not available during the init() and init_worker() phases, and that's why local/service policies cannot run in init() and init_worker().

apicast/src/policy/load_configuration.lua
davidor: All the policies are now initialized with names :+1:

spec/proxy_spec.lua
davidor: I think this one belongs to `configuration_store`. In fact, the same thing it's already tested there. I think we can just delete this test.

mikz: Yeah. 👍

apicast/src/executor.lua
davidor: How about doing this in the `local_chain` module?

When building the local chain, the default is `require('policy_chain').build({ 'apicast' })`. I think we could check the env 'APICAST_MODULE' and do `require('policy_chain').build({ 'module' })` instead if it's set.

mikz: Hmmmm. That sounds good 👍

davidor:
@mikz I addressed all the comments you and me wrote. I think this is ready.

PR #453, opened 2017-10-12
Make limits exceeded error configurable

### LINKED ISSUE #368: Return a 429 to the client when limits are exceeded

davidor:
Closes #368


apicast/conf.d/apicast.conf
```diff
@@ -26,6 +26,10 @@ location = /threescale_authrep {
   proxy_set_header  X-3scale-Version "$version";
   proxy_set_header  Connection "";
 
+  # Enable the option to receive the rejection reason in the
+  # '3scale-rejection-reason' header of the response.
+  proxy_set_header  3scale-options "rejection_reason_header=1";
```
mikz: Could we make this read from a variable? So it can be customized?

apicast/src/backend/cache_handler.lua
```diff
@@ -17,6 +17,14 @@ local mt = {
   end,
 }
 
+-- Returns the rejection reason from the headers of a 3scale backend response.
+-- The header is set only when the authrep call to backend enables the option
+-- to get the rejection reason. This is specified in the '3scale-options'
+-- header of the request.
+local function rejection_reason(response_headers)
+  return response_headers and response_headers['3scale-rejection-reason']
```
mikz: Should it return `'not authorized'` when there is no response for whatever reason?

apicast/src/configuration.lua
```diff
@@ -143,13 +143,16 @@ function _M.parse_service(service)
       hosts = proxy.hosts or { 'localhost' }, -- TODO: verify localhost is good default
       api_backend = proxy.api_backend,
       error_auth_failed = proxy.error_auth_failed,
+      error_limits_exceeded = proxy.error_limits_exceeded,
```
mikz: We probably should start providing default values for this. Especially when system is behind with the UI.

apicast/src/proxy.lua
```diff
@@ -213,8 +222,13 @@ function _M:authorize(service, usage, credentials, ttl)
 
     local res = http.get(internal_location)
 
-    if not self:handle_backend_response(cached_key, res, ttl) then
-      error_authorization_failed(service)
+    local authorized, rejection_reason = self:handle_backend_response(cached_key, res, ttl)
+    if not authorized then
+      if rejection_reason == 'limits_exceeded' then
+        error_limits_exceeded(service)
```
mikz: Can you add `return` to both branches, so it is obvious that nothing is executed after (because it calls `ngx.exit`)?

davidor: Good idea to make it more explicit. Changed :+1:

apicast/src/configuration.lua
davidor: Agreed. I added defaults based on the ones shown in the UI.
This is included in the latest 3 commits: implementation + integration tests + unit tests.

apicast/src/configuration.lua
```diff
@@ -142,14 +142,17 @@ function _M.parse_service(service)
...
-      error_auth_failed = proxy.error_auth_failed,
-      error_auth_missing = proxy.error_auth_missing,
-      auth_failed_headers = proxy.error_headers_auth_failed,
-      auth_missing_headers = proxy.error_headers_auth_missing,
-      error_no_match = proxy.error_no_match,
-      no_match_headers = proxy.error_headers_no_match,
+      error_auth_failed = proxy.error_auth_failed or 'Authentication failed',
+      error_limits_exceeded = proxy.error_limits_exceeded or 'Limits exceeded',
+      error_auth_missing = proxy.error_auth_missing or 'Authentication parameters missing',
+      auth_failed_headers = proxy.error_headers_auth_failed or 'application/json; charset=utf-8',
```
mikz: This looks wrong. Because the default value is not a json. So the content type should be text/plain.

davidor: :man_facepalming: right.

davidor: Fixed.

mikz:
Excellent work! 🥇

apicast/conf.d/apicast.conf
davidor: I missed this comment. Will address in a different PR.

PR #456, opened 2017-10-19
use backend http client

mikz:
* use one endpoint for backend communication
* simplify test suite

# TODO
- [x]  cover changes in the backend client by unit tests
- [x] performance test

### DIFF
--- a/apicast/src/backend_client.lua
+++ b/apicast/src/backend_client.lua
@@ -13,6 +13,7 @@ local setmetatable = setmetatable
 local concat = table.concat
 local insert = table.insert
 local len = string.len
+local format = string.format
 
 local http_ng = require('resty.http_ng')
 local user_agent = require('user_agent')
@@ -20,6 +21,7 @@ local resty_url = require('resty.url')
 local resty_env = require('resty.env')
 
 local _M = {
+  endpoint = resty_env.get("BACKEND_ENDPOINT_OVERRIDE")
 
 }
 
@@ -29,8 +31,8 @@ local mt = { __index = _M }
 -- @tparam Service service object with service definition
 -- @tparam http_ng.backend http_client async/test/custom http backend
 -- @treturn backend_client
-function _M.new(_, service, http_client)
-  local endpoint = service.backend.endpoint or ngx.var.backend_endpoint
+function _M:new(service, http_client)
+  local endpoint = self.endpoint or service.backend.endpoint or error('missing endpoint')
   local service_id = service.id
 
   if not endpoint then
@@ -39,7 +41,7 @@ function _M.new(_, service, http_client)
 
   local authentication = { service_id = service_id }
 
-  if service.backend_authentication.type then
+  if service.backend_authentication and service.backend_authentication.type then
     authentication[service.backend_authentication.type] = service.backend_authentication.value
   end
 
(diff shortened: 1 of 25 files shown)


apicast/src/backend_client.lua
```diff
@@ -13,13 +13,15 @@ local setmetatable = setmetatable
...
 local len = string.len
+local format = string.format
 
 local http_ng = require('resty.http_ng')
 local user_agent = require('user_agent')
 local resty_url = require('resty.url')
 local resty_env = require('resty.env')
 
 local _M = {
+  endpoint = resty_env.get("BACKEND_ENDPOINT_OVERRIDE")
```
mikz: This should be ignoring empty string.

apicast/src/backend_client.lua
```diff
@@ -94,15 +89,30 @@ local function call_backend_transaction(self, path, ...)
...
+  local res = http_client.get(url, options)
 
   ngx.log(ngx.INFO, 'backend client uri: ', url, ' ok: ', res.ok, ' status: ', res.status, ' body: ', res.body)
 
   return res
 end
 
+local authorize_options = {
+  headers = {
+    ['3scale-options'] =  'rejection_reason_header=1'
```
mikz: @davidor this should be more dynamic, right? So we can have `no_body` for normal authrep but not for oauth calls.

davidor: Yes, we need that.

davidor: Let's leave this for a future PR when we actually add more options to avoid adding more complexity to this PR.

t/003-apicast.t
```diff
@@ -662,11 +647,11 @@ status code (429).
 --- config
   include $TEST_NGINX_APICAST_CONFIG;
 
-  set $backend_endpoint 'http://127.0.0.1:$TEST_NGINX_SERVER_PORT';
-
   location /transactions/authrep.xml {
     content_by_lua_block {
-      ngx.header['3scale-rejection-reason'] = 'limits_exceeded';
+      if ngx.var['http_3scale_options'] == 'rejection_reason_header=1' then
```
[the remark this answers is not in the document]
mikz: @davidor `$http_` headers are mapped to the request headers: http://nginx.org/en/docs/http/ngx_http_core_module.html#var_http_

davidor:
@mikz I added a few commits on top of yours. Please let me know if you think that there's anything else we should do before benchmarking this against master.

mikz:
👍 good job @davidor 

I think this should be good to go for the performance test.

davidor:
I benchmarked this. I Deployed just an Apicast (authorizing without going to 3scale's backend) and another one in echo mode (as api backend) in separate machines.

The tests were performed using JMeter from an external machine. Absolute numbers are not really meaningful, as there are many factors that can affect them. I just want to focus on the comparison between this branch and its base. I used c4.xlarge aws machines, configured Apicast with 4 workers, and run JMeter for 10 minutes.

### Without this PR
no


### With this PR
yes


Unfortunately, this PR introduces an important performance regression.

davidor:
Update: the last 3 commits increase the performance, but it's still not on par with the performance of the master branch. It's important to note that this benchmark is only testing cpu as Apicast is acting both as auth backend, and api backend. This means that the difference might not be noticeable in a real-case scenario where the api and the auth backend are in different machines.

PR:
pr

Master:
master

mikz:
👍

mikz:
Added changelog entry in https://github.com/3scale/apicast/commit/0108fb21754348c18ec2b8a5c1e49003a8ac4dcf

PR #458, opened 2017-10-19
[t] use one base class

mikz:
unify all tests to use one base
so we can share setting between tests


apicast/src/configuration.lua
```diff
@@ -129,12 +128,34 @@ end
...
 local Service = require 'configuration.service'
 
+local noop = function() end
+local function readonly_table(table)
+    return setmetatable({}, { __newindex = noop, __index = table })
+end
+
+local empty_t = readonly_table()
+
+local function backend_endpoint(proxy)
```
davidor: The indentation here is not consistent with the rest of the codebase.

mikz: Indeed. Haven't configured my IDE and used the default 4 spaces indentation. 
Must admit that I quite like it. Somehow it makes it more readable. Also the OpenResty code is indented with 4 spaces. 
I'd consider switching to 4 spaces. Wdyt?

davidor: I don't mind as long as it's spaces :) and we are consistent across the codebase.
http://lua-users.org/wiki/LuaStyleGuide

t/001-management.t
```diff
@@ -1,17 +1,7 @@
-use Test::Nginx::Socket::Lua 'no_plan';
-use Cwd qw(cwd);
+use lib 't';
 
-my $pwd = cwd();
-my $apicast = $ENV{TEST_NGINX_APICAST_PATH} || "$pwd/apicast";
+use TestAPIcast 'no_plan';
```
davidor: What is `no_plan` ?

t/016-balancer.t
```diff
@@ -1,22 +1,14 @@
...
+$ENV{TEST_NGINX_HTTP_CONFIG} = "$TestAPIcast::path/http.d/*.conf";
 $ENV{RESOLVER} = '127.0.1.1:5353';
 
-
 env_to_nginx(
     'RESOLVER'
 );
 master_on();
-log_level('debug');
 repeat_each(1);
```
davidor: Should we define `TEST_NGINX_REPEAT_EACH` instead?
From TestAPIcast: `repeat_each($ENV{TEST_NGINX_REPEAT_EACH} || 2);`

t/001-management.t
mikz: I don't really know. I guess it defines in what order to execute the tests. Search for `plan` on http://search.cpan.org/~agent/Test-Nginx/lib/Test/Nginx/Socket.pm

https://openresty.gitbooks.io/programming-openresty/content/ also have few mentions of it.

t/016-balancer.t
mikz: Some tests force one iteration because 2nd one would hit cache and have different result.

Because the repeat_each is in the base this is an override. So all tests default to 2 (or the env var) except few ones that are forced to one.

t/TestAPIcast.pm
```diff
@@ -0,0 +1,74 @@
+package TestAPIcast;
```
davidor: Nice DRY :+1:

.circleci/config.yml
```diff
@@ -45,6 +45,7 @@ jobs:
     working_directory: /opt/app-root/apicast
     steps:
       - checkout
+      - run: rm -rf lua_modules
```
davidor: Why is this needed now, but not before? :thinking:

mikz: It is needed for local run. There is no way to ignore folders on local build.

PR #459, opened 2017-10-21
Add missing unit tests for configuration_store

davidor:
This PR adds some tests for the configuration_store module. It also fixes a couple of the existing tests.

### DIFF
--- a/spec/configuration_store_spec.lua
+++ b/spec/configuration_store_spec.lua
@@ -144,7 +144,7 @@ describe('Configuration Store', function()
 
         store:reset()
 
-        assert.equal(0, #store.cache)
+        assert.same({}, store.cache.hasht)
       end)
 
       it('deletes all services', function()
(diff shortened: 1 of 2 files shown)


spec/configuration_store_spec.lua
```diff
@@ -11,9 +12,56 @@ describe('Configuration Store', function()
...
+    it('sets the configured flag to true', function()
+      local store = configuration.new()
+
+      store:store({ services = { id = '42', hosts = { 'example.com' } } })
+      assert.truthy(store.configured)
+    end)
+
+    describe('when path routing is enabled', function()
+      setup(function() -- Wouldn't be needed if we injected path_routing_enabled
+        package.loaded.configuration_store.path_routing = true;
```
mikz: What about using `configuration` instead? What require returns should be the same object.

spec/configuration_store_spec.lua
```diff
@@ -11,9 +12,56 @@ describe('Configuration Store', function()
...
+      end)
+
+      teardown(function()
+        package.loaded.configuration_store.path_routing =
```
mikz: What about using local variable for the previous value and setting it back? It would not need to duplicate the code reading from env.

davidor:
I addressed your two comments @mikz

PR #460, opened 2017-10-23
use new s2i-openresty with rover integration

mikz:
Extracted from https://github.com/3scale/apicast/pull/449

Use rover provided by https://github.com/3scale/s2i-openresty/pull/38 

This change requires APIcast to start with a different prefix than the current source code because the directory is not writeable anymore.


apicast/Roverfile.lock
```diff
@@ -0,0 +1,24 @@
+ansicolors 1.0.2-3
+argparse 0.5.0-1
+busted 2.0.rc12-1
+dkjson 2.5-2
+inspect 3.1.0-1
+ldoc 1.4.6-2
+liquid scm-1
+lua-resty-env 0.4.0-1
+lua-resty-execvp 0.1.0-1
```
mikz: This needs to be updated. Those dependencies are not there yet.

.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     environment:
       S2I_VERSION: "1.1.7-226afa1"
       DOCKER_COMPOSE_VERSION: "1.16.1"
-      OPENRESTY_VERSION: "1.11.2.5-1"
+      OPENRESTY_VERSION: "rover"
```
davidor: A few lines below there's: `quay.io/3scale/s2i-openresty-centos7:1.11.2.5-2`. (Notice that the openresty version doesn't match :scream: )
Should it be adpated to use `rover` as well?

apicast/config-5.1.lua
```diff
@@ -0,0 +1,4 @@
+-- vim: set ft=lua:
+rocks_trees = {
```
davidor: Can you add a comment explaining this?

.circleci/config.yml
davidor: Also, with this change, we'll no longer have documented the openresty version we are using.
Maybe we should use something like "rover-1.11.2.5-1" ?

mikz: Yeah. I don't intend to merge it pointing to rover. After this is ready I'll merge the rover branch and do a new release of the s2i image with a correct version.

apicast/config-5.1.lua
mikz: Yep. Will do.

PR #464, opened 2017-11-02
Run benchmark on CircleCI

mikz:
Run wrk benchmark on CircleCI, so we get some ballpark performance numbers to stop massive performance regression.


davidor:
I'll rebase #456 on top of master to see if this would have detected the perf regression that we found.

PR #471, opened 2017-11-08
[cli] resolve files local to the current folder

mikz:
so for example `-c examples/configuration/echo.json` works

the cli changes directory to apicast and then all relative paths would
stop working

looks like this broke by #449


davidor:
I think that breaking the `read` method into smaller ones would help readability. Possible ones: `absolute_path`, `is_valid_path`, `strip_trailing_slash`, etc. What do you think?

mikz:
Yep. Good idea. Will do.

PR #472, opened 2017-11-08
[t] add DEBUG=1 to enable logging to stderr

mikz:
It is nice to not have to remember the full env variable. Just setting `DEBUG=1` will run in debug mode with more information.


davidor:
The PR and commit titles say this logs to stdout, but the code says it logs to stderr.

mikz:
Well spotted, will change.

PR #474, opened 2017-11-08
blackbox tests using Test::Nginx 

mikz:
This implements a base test class to provide blackbox testing of APIcast using the Test::Nginx framework.

Blackbox tests should define just JSON config.
They definitely should not hook into nginx configuration by setting or changing variables.
This makes the blackbox tests much more maintainable solution for the future.

Depends on https://github.com/3scale/apicast/pull/478

@3scale/qe this might be poking a bit into your turf, but still has no interaction with rest of the platform. It is meant for just static configuration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ apicast/logs/
 *.rock
 lua_modules/
 tmp/benchmark/
+apicast/config/test.lua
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -8,3 +8,4 @@ apicast/logs/
 *.rock
 lua_modules/
 tmp/benchmark/
+apicast/config/test.lua
```
mikz: We should allow loading any lua file as a config and make the test runner use a temporary file instead of this one.

PR #478, opened 2017-11-14
[resty.resolver] try to resolve upstream peers

mikz:
* defining an upstream can be used as a shortcut to skip DNS resolution
* if an upstream is defined with the same name then use the peers

extracted from https://github.com/3scale/apicast/pull/474

### DIFF
--- a/apicast/src/resty/resolver.lua
+++ b/apicast/src/resty/resolver.lua
@@ -14,6 +14,7 @@ local io_type = io.type
 local re_match = ngx.re.match
 local resolver_cache = require 'resty.resolver.cache'
 local dns_client = require 'resty.resolver.dns_client'
+local upstream = require 'ngx.upstream'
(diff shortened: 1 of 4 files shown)


apicast/src/resty/resolver.lua
```diff
@@ -251,6 +252,22 @@ local function search_dns(self, qname, stale)
...
 end
 
+local function resolve_upstream(qname)
+  local peers, err = upstream.get_primary_peers(qname)
+
+  if not peers and err then
+    return nil, err
+  end
+
+  for i=1, #peers do
```
davidor: According to the `if` above, here we could have `peers = nil`, and then `#peers` would raise an error.

mikz: True that is possible. Will change the condition to return when there are no peers.

PR #483, opened 2017-11-15
Enable the 'no_body' option when authorizing against 3scale backend

davidor:
When `no_body` is enabled via the `3scale-options-header`, backend does not return a response body. The body normally has lots of detail: metrics, limits, plans, etc. In the normal auth flow, all this information is not required. We only need to check the status code of the response. By enabling the `no_body` option, we will save some work to the 3scale backend and also reduce network traffic.

no_body cannot be enabled in the oauth flow, as there we need to parse the response body.

### DIFF
--- a/apicast/src/backend_client.lua
+++ b/apicast/src/backend_client.lua
@@ -124,11 +124,23 @@ local function create_token_path(service_id)
   return format('/services/%s/oauth_access_tokens.xml', service_id)
 end
 
-local authorize_options = {
-  headers = {
-    ['3scale-options'] =  'rejection_reason_header=1'
-  }
-}
+-- Returns the authorize options that 3scale backend accepts. Those options
+-- are specified via headers. Right now there are 2:
+--  - rejection_reason_header: asks backend to return why a call is denied
+--    (limits exceeded, application key invalid, etc.)
+--  - no_nody: when enabled, backend will not return a response body. The
+--    body has many information like metrics, limits, etc. This information is
+--    parsed only when using oauth. By enabling this option will save some work
+--    to the 3scale backend and reduce network traffic.
+local function authorize_options(using_oauth)
+  local headers = { ['3scale-options'] = 'rejection_reason_header=1' }
+
+  if not using_oauth then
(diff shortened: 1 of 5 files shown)


apicast/src/backend_client.lua
```diff
@@ -124,11 +124,23 @@ local function create_token_path(service_id)
...
-}
+-- Returns the authorize options that 3scale backend accepts. Those options
+-- are specified via headers. Right now there are 2:
+--  - rejection_reason_header: asks backend to return why a call is denied
+--    (limits exceeded, application key invalid, etc.)
+--  - no_nody: when enabled, backend will not return a response body. The
+--    body has many information like metrics, limits, etc. This information is
+--    parsed only when using oauth. By enabling this option will save some work
+--    to the 3scale backend and reduce network traffic.
+local function authorize_options(using_oauth)
```
mikz: Wouldn't be nicer to pass `self` instead of `using_oauth` ? 

Because if the condition changes we would have to change only one place - this method.

mikz: I see that it would also require changing the `authrep_path` method to expect self instead of just boolean.

Tough choice. Not so sure now. I guess it is fine as it is then!

davidor: I like this option because it makes explicit that the only thing the method needs to know about is if we are running the oauth flow or not. Although as you said, we might need to change the params this method receives and the callers in the future.

davidor:
Added a changelog entry and rebased on top of master so it doesn't give conflicts.

apicast/src/backend_client.lua
```diff
@@ -124,11 +124,23 @@ local function create_token_path(service_id)
...
+  local headers = { ['3scale-options'] = 'rejection_reason_header=1' }
+
+  if not using_oauth then
+    headers['3scale-options'] = headers['3scale-options'] .. '&no_body=1'
```
unleashed: @davidor why don't you first construct the string and then create the `headers` table... just look at the amount of accesses you are performing.

mikz: it is all going to be JITed ! :trollface:

PR #486, opened 2017-11-20
namespace all apicast source code

mikz:
# tl;dr

From now instead of `require('policy')` use `require('apicast.policy')`.  All APIcast source code is in the `apicast` namespace.  

------

Right now APIcast is polluting global namespace with its source files.
As more dependencies are installed there is higher chance of a clash.
So we need to namespace all APIcast source code in `apicast` folder inside 
the Lua load path.

To prevent weird paths like `apicast/apicast/src/apicast` this also renames the 
folder that contains all APIcast distributed source from `apicast` to `gateway`.

Print deprecation warnings when old code refers to new paths:

```
2017/11/20 17:11:13 [warn] 91413#25180842: [lua] loader.lua:18: DEPRECATION: when loading apicast code use correct prefix: require("apicast.policy_chain")
2017/11/20 17:11:13 [warn] 91413#25180842: [lua] loader.lua:18: DEPRECATION: when loading apicast code use correct prefix: require("apicast.policy")
2017/11/20 17:11:13 [warn] 91413#25180842: [lua] loader.lua:18: DEPRECATION: when loading apicast code use correct prefix: require("apicast.linked_list")
2017/11/20 17:11:13 [warn] 91413#25180842: [lua] loader.lua:29: DEPRECATION: file renamed, change: require("apicast") to: require("apicast.policy.apicast")
```

Please note the benchmark does not work because the new configuration is not compatible with the old one. But the old one should be compatible with the new one.

Also note that if someone was overriding a path of APIcast then it will stop working after this change. 
Requiring old files continue to work with a deprecation warning, but overriding those files on old paths 
will not be taken into effect in the APIcast codebase. This is potential BREAKING CHANGE for some customizations.

Another interesting change is the echo policy. Now it short circuits the upstream api and responds directly in the rewrite phase stopping everything else.

/cc @3scale/support


gateway/src/apicast/loader.lua
```diff
@@ -0,0 +1,51 @@
+local loadfile = loadfile
```
davidor: I think It would be good to document the purpose of this module.

mikz: True, good point :)

mikz: Should be addressed in https://github.com/3scale/apicast/pull/486/commits/b1d677a043724e3b6069b35f36ae85b3e721f08d.

t/deprecation-warnings.t
```diff
@@ -0,0 +1,35 @@
+use lib 't';
+use TestAPIcastBlackbox 'no_plan';
+
+repeat_each(1);
+run_tests();
+
+__DATA__
+
+=== TEST 1: authentication credentials missing
```
davidor: I'd change the title and description of the test so they make reference to what is actually being tested here.

mikz: hah, yeah my copy paste issue
well spotted

davidor: :eyes:

PR #487, opened 2017-11-20
Convert existing CORS module into a policy

### LINKED ISSUE #484: Adapt CORS policy to new policy chain system
We need to adapt the current CORS module https://github.com/3scale/apicast/blob/3.1-stable/examples/cors/cors.lua

davidor:
Closes #484 

This PR simply converts the existing CORS module into a policy without changing its behavior.
I've also added integration tests based on the new `TestAPIcastBlackbox` module.

@mikz  I have some questions about the current behavior of this CORS module. Please check my comments and let me know what you think.

### DIFF
--- a/apicast/src/policy/cors.lua
+++ b/apicast/src/policy/cors.lua
@@ -1,19 +1,65 @@
+--- CORS policy
+-- This policy enables CORS (Cross Origin Resource Sharing) request handling.
+-- The policy is configurable. Users can specify the values for the following
+-- headers in the response:
+--   - Access-Control-Allow-Headers
+--   - Access-Control-Allow-Methods
+--   - Access-Control-Allow-Origin
+--   - Access-Control-Allow-Credentials
+-- By default, those headers are set so all the requests are allowed. For
+-- example, if the request contains the 'Origin' header set to 'example.com',
+-- by default, 'Access-Control-Allow-Origin' in the response will be set to
+-- 'example.com' too.
+
 local policy = require('policy')
 local _M = policy.new('CORS Policy')
 
-local function set_cors_headers()
-  local origin = ngx.var.http_origin
+local new = _M.new
 
-  if not origin then return end
-
-  ngx.header['Access-Control-Allow-Headers'] = ngx.var.http_access_control_request_headers
-  ngx.header['Access-Control-Allow-Methods'] = ngx.var.http_access_control_request_method
-  ngx.header['Access-Control-Allow-Origin'] = origin
-  ngx.header['Access-Control-Allow-Credentials'] = 'true'
+--- Initialize a CORS policy
+-- @tparam[opt] table config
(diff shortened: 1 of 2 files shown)


apicast/src/policy/cors.lua
```diff
@@ -0,0 +1,37 @@
+local policy = require('policy')
+local _M = policy.new('CORS Policy')
+
+local function set_cors_headers()
+  local origin = ngx.var.http_origin
+
+  if not origin then return end
+
+  ngx.header['Access-Control-Allow-Headers'] = ngx.var.http_access_control_request_headers
```
davidor: I'm not sure about these 2. Should we return the headers and methods that apicast supports? Or should we just accept everything because we don't know that the upstream api accepts?

mikz: I think this module should accept a configuration that would whitelist what HTTP methods / Headers are supported. And possibly what origins too and if to pass credentials or not.

But until we have that configuration we can't do much other than just allow all. And even with the configuration there should be a way to configure it as accept all that browser requests.

davidor: Makes sense. We can use the "configuration" field of a policy to pass this information for now: https://github.com/3scale/apicast/blob/master/apicast/src/configuration.lua#L167

mikz: Yep. That sounds good! Forgot it is there :)

apicast/src/policy/cors.lua
```diff
@@ -0,0 +1,95 @@
...
+-- @field[opt] allow_origin Allowed origins (e.g. 'http://example.com', '*')
+-- @field[opt] allow_credentials Boolean
+function _M.new(config)
+  local self = new()
+  self.config = config or {}
+  return self
+end
+
+local function set_access_control_allow_headers(allow_headers)
+  local value = allow_headers and concat(allow_headers, ',')
```
mikz: I think it should be possible to assign a table to the header and nginx should take care of the serialization. That could simplify the code a bit.

apicast/src/policy/cors.lua
```diff
@@ -0,0 +1,95 @@
...
+--   - Access-Control-Allow-Origin
+--   - Access-Control-Allow-Credentials
+-- By default, those headers are set so all the requests are allowed. For
+-- example, if the request contains the 'Origin' header set to 'example.com',
+-- by default, 'Access-Control-Allow-Origin' in the response will be set to
+-- 'example.com' too.
+
+local concat = table.concat
+
+local policy = require('policy')
```
mikz: This will produce a deprecation warning once https://github.com/3scale/apicast/pull/486 is merged. No action needed here, but that PR is going to have to fix this.

apicast/src/policy/cors.lua
davidor: You're right. I didn't know that.
Changed.

Issue #481, opened 2017-11-14
Provide config to set global policies


mikz:
https://github.com/3scale/apicast/pull/450 implements policy chains that can be per service/global.

Service ones can be configured in service configuration. 
Global ones have no configuration yet.

Ideally we could use [lua-toml](https://github.com/jonstoler/lua-toml/) and provide nice human readable config in [TOML](https://github.com/toml-lang/toml). ~~However due to a licensing issue (https://github.com/jonstoler/lua-toml/issues/15) we can't use it yet~~. It is released under MIT license, so is compatible.

As an alternative we could try to extend the current JSON config with another global key for global configuration. But it is possible that would work only when loading configuration on boot.

It is possible to download configuration from network because we control the CLI and can make network calls. So the CLI start command can actually download the configuration to a temporary file.

Other proposals welcome.

andrewdavidmackenzie:
Any lua-yaml as a next-best option?

mikz:
The yaml ones I found require extra library which seems unfortunate and would complicate the build process and manual installation.

mikz:
So there is already a concept of an environment.

* development.lua:
  https://github.com/3scale/apicast/blob/153a2badfe6bedb56cbe8565cf34c587625557f5/apicast/config/development.lua
* production.lua:
  https://github.com/3scale/apicast/blob/153a2badfe6bedb56cbe8565cf34c587625557f5/apicast/config/production.lua
* test.lua
  https://github.com/3scale/apicast/blob/153a2badfe6bedb56cbe8565cf34c587625557f5/t/TestAPIcastBlackbox.pm#L89-L102

Those can be passed by CLI flag `-e` or `THREESCALE_DEPLOYMENT_ENV` env variable. The file can be overriden by `APICAST_ENVIRONMENT_CONFIG` env variable. This file is lua for now so you can do complicated stuff. This file has to return a table that is passed to the liquid template and used to render the nginx configuration. That means you can have the same nginx config and just some differences between the environments - like lua code cache, listening port, ...
The idea was that for example the production lua file could automatically detect number of cores from cgroups and use correct number of workers.

We could extend it on the policy list and add it there too. And start treating the environment as the global config file. We also could offer TOML version of that file that would be simple and more obvious than Lua.

PR #489, opened 2017-11-21
Delete old CORS module example

davidor:
This PR deletes the old CORS module example. Now we have a CORS policy (#487) that uses the new policy chain system, so this example is no longer relevant.


andrewdavidmackenzie:
It's not a big deal for me....but

 if this (and other) examples cannot be made to work with the api that's in the head of master, I would just delete them....

If people want them as examples to evolve/learn etc - they can do that with what's in the last release....

mikz:
Also we could replace this by how to enable the CORS policy. Because right now there is not much documentation about that. But that should be tackled in https://github.com/3scale/apicast/issues/481

PR #492, opened 2017-11-21
[cors] move cors policy after namespacing sources

mikz:
https://github.com/3scale/apicast/pull/486 broke https://github.com/3scale/apicast/pull/487


davidor:
@mikz It doesn't appear here, but looks like #487 also broke the uploading of artifacts in Travis: https://travis-ci.org/3scale/apicast/builds/305318479

mikz:
@davidor looks like that was broken forever (since the CircleCI move) https://travis-ci.org/3scale/apicast/builds/304785869 since it does not fail the build.

PR #493, opened 2017-11-21
Adapt Docker development env to namespace changes

davidor:
This PR simply adapts the paths used by the development Docker to the new structure introduced in https://github.com/3scale/apicast/pull/486


davidor:
Right. I'll add that in a future PR.

PR #496, opened 2017-11-22
ability to configure policy chains globally

mikz:
closes https://github.com/3scale/apicast/issues/481

It is possible to return a table with field `policy_chain` that is going to be used as the policy chain instead of the default one.

Example:

```lua
local policy_chain = require('apicast.policy_chain').default()

policy_chain:insert(mypolicy)

return {
  policy_chain = policy_chain
}
```

### DIFF
--- a/gateway/src/apicast/cli/environment.lua
+++ b/gateway/src/apicast/cli/environment.lua
@@ -16,6 +16,7 @@ local print = print
 local pairs = pairs
 local ipairs = ipairs
 local tostring = tostring
+local tonumber = tonumber
 local insert = table.insert
 local concat = table.concat
 local re = require('ngx.re')
@@ -127,6 +128,7 @@ function _M:add(env)
 
     local config = loadfile(path, 't', {
         print = print, inspect = require('inspect'), context = self._context,
+        tonumber = tonumber, tostring = tostring,
         pcall = pcall, require = require, assert = assert, error = error,
     })
 
--- a/gateway/src/apicast/cli/environment.lua
+++ b/gateway/src/apicast/cli/environment.lua
@@ -48,10 +48,12 @@ _M.default_environment = 'production'
 
 --- Default configuration.
 -- @tfield ?string ca_bundle path to CA store file
+-- @tfield ?policy_chain policy_chain @{policy_chain} instance
 -- @tfield ?{string,...} nameservers list of nameservers
 -- @table environment.default_config default configuration
 _M.default_config = {
     ca_bundle = resty_env.value('SSL_CERT_FILE'),
+    policy_chain = require('apicast.policy_chain').default(),
     nameservers = parse_nameservers(),
 }
 
--- a/gateway/src/apicast/policy/echo.lua
+++ b/gateway/src/apicast/policy/echo.lua
@@ -1,9 +1,36 @@
-local policy = require('apicast.policy')
-local _M = policy.new('Echo Policy')
+--- Echo policy
+-- Print the request back to the client and optionally set a status code.
+-- Also can interrupt the execution and skip the current phase or
+-- the whole processing of the request.
 
-function _M.rewrite()
+local _M = require('apicast.policy').new('Echo Policy')
+
+local tonumber = tonumber
+local new = _M.new
+
+function _M.new(configuration)
+  local policy = new(configuration)
+
+  if configuration then
+    policy.status = tonumber(configuration.status)
+    policy.exit = configuration.exit
+  end
+
+  return policy
+end
+
+function _M:rewrite()
+  if self.status then
+    ngx.status = self.status
+  end
+
   ngx.say(ngx.var.request)
-  ngx.exit(0)
+
+  if self.exit == 'request' then
+    return ngx.exit(ngx.status)
+  elseif self.exit == 'phase' then
+    return ngx.exit(0)
+  end
 end
 
 return _M
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Calls 3scale backend with the 'no_body' option enabled. This reduces network traffic in cases where APIcast does not need to parse the response body [PR #483](https://github.com/3scale/apicast/pull/483)
 - Methods to modify policy chains [PR #505](https://github.com/3scale/apicast/pull/505)
 - Ability to load several environment configurations [PR #504](https://github.com/3scale/apicast/pull/504)
+- Ability to configure policy chain from the environment configuration [PR #496](https://github.com/3scale/apicast/pull/496)
 
 ## Changed
 
--- a/examples/policy_chain/README.md
+++ b/examples/policy_chain/README.md
@@ -1,0 +1,13 @@
+# Configure Policy Chain
+
+Environment configuration can define the global policy chain. You can provide custom chain or insert policies to the default one.
+
+## Using Echo Policy
+
+[Echo policy](../../gateway/src/apicast/policy/echo.lua) accepts configuration option to terminate the request phase. See the example in [`configuration.lua`](./configuration.lua).
+
+You can start it as:
+
+```shell
+ECHO_STATUS=202 bin/apicast --environment examples/policy_chain/configuration.lua
+```
--- a/examples/policy_chain/configuration.lua
+++ b/examples/policy_chain/configuration.lua
@@ -1,0 +1,13 @@
+local resty_env = require('resty.env')
+local policy_chain = require('apicast.policy_chain').default()
+local echo_policy = require('apicast.policy.echo').new({
+  status = tonumber(resty_env.value('ECHO_STATUS') or 201),
+  exit = 'request'
+})
+
+-- add Echo policy to the chain on the 1st place
+policy_chain:insert(echo_policy, 1)
+
+return {
+  policy_chain = policy_chain
+}
--- a/gateway/http.d/init.conf
+++ b/gateway/http.d/init.conf
@@ -6,11 +6,18 @@ init_by_lua_block {
     require("resty.core")
     require('resty.resolver').init()
 
+    local env = require('apicast.cli.environment').load()
+    local context = env:context()
     local module = require('apicast.executor')
 
     if not module then
       ngx.log(ngx.EMERG, 'fatal error when loading the root module')
       os.exit(1)
+    end
+
+    if context.policy_chain then
+        module = module.new(context.policy_chain)
+        package.loaded['apicast.executor'] = module
     end
 
     module:init()


examples/configuration/echo.lua
```diff
@@ -0,0 +1,7 @@
+local policy_chain = require('apicast.policy_chain').build({ })
+
+policy_chain:insert(policy_chain.load('apicast.policy.echo'))
+
+return {
+	policy_chain = policy_chain
```
mikz: We will have to start documenting the environment config format. It is Lua, so no JSON schema.

spec/policy_chain_spec.lua
```diff
@@ -92,6 +92,38 @@ describe('policy_chain', function()
     end, 'readonly table')
   end)
 
+  describe('.insert', function()
+
+    it('adds policy to the chain', function()
```
davidor: I'd modify this test or add a new one. For `.insert` we need to test that it adds the policy to the end of the list. If we do not initialize the chain with a few elements, we won't be able to test that. This test would pass even if the code added the policy to the beginning of the list or in a random position.

mikz: Good point 👍 Will fix.

spec/policy_chain_spec.lua
```diff
@@ -92,6 +92,38 @@ describe('policy_chain', function()
...
+      local chain = _M.new({}):freeze()
+
+      local ok, err = chain:insert(policy, 1)
+
+      assert.is_nil(ok)
+      assert.equal(err, 'frozen chain')
+      assert.equal(0, #chain)
+    end)
+  end)
+
```
davidor: Would be good to test `.default` for completeness. The other two methods introduced in this PR for that module are tested.

mikz: Good point. Fixing 👍

gateway/src/apicast/policy_chain.lua
```diff
@@ -96,6 +109,25 @@ function _M:export()
...
+    self.frozen = true
+    return self
+end
+
+--- Insert a policy into the chain
+-- @tparam Policy policy the policy to be added to the chain
+-- @tparam[opt] int position the position to add the policy to, defaults to last one
+function _M:insert(policy, position)
+    if self.frozen then
+        return nil, 'frozen chain'
```
davidor: Maybe we should be more specific in the error message. Not sure this will be obvious to contributors.

gateway/src/apicast/policy_chain.lua
```diff
@@ -96,6 +109,25 @@ function _M:export()
     return chain
 end
 
+--- Freeze the policy chain to prevent modifications.
+-- After calling this method it won't be possible to insert more policies.
+-- @treturn self
```
davidor: nitpick: missing type

gateway/src/apicast/policy_chain.lua
mikz: I think the error messages in openresty are pretty simple because people can treat them like error code and match them. I'd probably even go for just `'frozen'` so it is easier to match.

Error codes from [lua-resty-redis](https://github.com/openresty/lua-resty-redis/blob/master/lib/resty/redis.lua):

* not initialized
* not subscribed
* closed
* timeout

If "frozen" is not understandable I could rename it to: immutable, finalized, locked (other suggestions welcome).

gateway/src/apicast/policy_chain.lua
mikz: 👍 

This file is probably not generating the doc as we had a whitelist in config.ld. 

Will add it there and see the generated documentation.

examples/configuration/echo.lua
```diff
@@ -0,0 +1,7 @@
+local policy_chain = require('apicast.policy_chain').build({ })
```
davidor: I think these 2 lines could be simplified to:
`local policy_chain = require('apicast.policy_chain').build({'apicast.policy.echo'})`

mikz: Yes. I wanted to show the `:insert` as an example. But that would be better with the default chain.

examples/configuration/echo.lua
```diff
@@ -0,0 +1,7 @@
...
+
+policy_chain:insert(policy_chain.load('apicast.policy.echo'))
```
davidor: This insert + load combination looks a bit confusing to me. I wonder if we should make `.load` private.

t/configuration-loading-boot-with-config.t
```diff
@@ -1,14 +1,11 @@
 use lib 't';
-use TestAPIcast 'no_plan';
+use TestAPIcastBlackbox 'no_plan';
 
-$ENV{TEST_NGINX_HTTP_CONFIG} = "$TestAPIcast::path/http.d/init.conf";
 $ENV{APICAST_CONFIGURATION_LOADER} = 'boot';
-$ENV{THREESCALE_CONFIG_FILE} = 't/servroot/html/config.json';
+#$ENV{TEST_NGINX_NO_CLEAN} = 1;
```
davidor: Commented.

examples/configuration/echo.lua
mikz: And just rely on `require(...).new()` instead of `.load` ?

t/configuration-loading-boot-with-config.t
mikz: Well spotted. 👍

gateway/src/apicast/cli/command/start.lua
```diff
@@ -44,10 +44,13 @@ local function update_env(env)
     end
 end
 
-local function nginx_config(context, dir, path, env)
-    update_env(env)
 
-    local template = Template:new(context, dir, true)
+local function apicast_root()
+    return resty_env.get('APICAST_DIR') or pl.path.abspath('.')
```
davidor: Should we call `.value` instead?

mikz: Yes. Good point!

gateway/src/apicast/cli/command/start.lua
```diff
@@ -75,40 +78,73 @@ local function get_log_level(self, options)
...
+
+    config:save()
+
+    return config
+end
+
+local function openresty_binary(candidates)
+    return resty_env.value('APICAST_OPENRESTY_BINARY') or
+        resty_env.value('TEST_NGINX_BINARY') or
+        pick_openesty(candidates)
```
davidor: typo, also, I find naming here a bit confusing. This method is choosing the openresty binary but it's calling one named `pick_openresty`.

gateway/src/apicast/cli/environment.lua
```diff
@@ -0,0 +1,129 @@
...
+
+local function parse_nameservers()
+    local resolver = require('resty.resolver')
+    local nameservers = {}
+
+    for _,nameserver in ipairs(resolver.init_nameservers()) do
+        insert(nameservers, tostring(nameserver))
+    end
+
+    if #nameservers > 0 then
```
davidor: Why return nil instead of {} ?

mikz: Because then it is easier to verify that there are no nameservers. `resolver` directive in the nginx conf has to have at least one server. So if there are none we can't set it.

gateway/src/apicast/cli/environment.lua
```diff
@@ -0,0 +1,129 @@
+local pl_path = require('pl.path')
```
davidor: Would be good to document the purpose of this module.

mikz: Yep, Opened https://github.com/3scale/apicast/pull/500 so we can see the documentation artifacts and verify it looks correctly. Will write it after that one is merged and I can see the documetation.

gateway/src/apicast/cli/command/start.lua
mikz: Yes. I was too lazy to rename this function. Will do 👍

davidor:
I think it would be good to add an integration test that shows how to configure policy chains via environment.

mikz:
@davidor yep that is a good idea 👍  Will do.

gateway/src/apicast/policy_chain.lua
davidor: Fair enough. I was thinking more about clarifying what the error is about to possible contributors rather than making sure it's easy to match.

mikz: That probably should be addressed by the method documentation describing the error messages. Will do 👍

examples/configuration/echo.lua
davidor: What about `.insert` calling `.load` like `.build` does and making `.load` private?

I think that `policy_chain:insert('apicast.policy.echo')` or `policy_chain:insert(echo.new(something))` looks better than `policy_chain:insert(policy_chain.load(...))`.

gateway/src/apicast/cli/environment.lua
davidor: Would be good to document this.

examples/configuration/echo.lua
mikz: I'd like `insert` to be an API that does one thing only. So I'd prefer it does not initialize the module because then it would also accept the configuration.

I agree `.load` is kind of an useless abstraction and it would be better to use plain require. The only reason for is existence was that I wanted to have an abstraction so the internals can be replaced without much hassle. 
But when replacing internal and for example changing `:new` to `.new` the policies would have to change anyway and it would leak this concept to the policy. But at least all the uses of the policy would not have to change.

So we should agree if we want to expose the policy initialization to users and deal with the consequences if we need to ever change it or expose some method to load/initialize policies.

examples/configuration/echo.lua
mikz: What about:

```lua
local policy_chain = require('apicast.policy_chain').default()
local echo = require('apicast.policy.echo').new({ exit = 0 })

policy_chain:insert(echo, 1)
```

 ? 

It shows off the insert - how to extend the default chain and how to pass options to policy.

davidor: But `apicast.policy.echo` doesn't have a constructor that can receive that table.

mikz: Well I'd change it to accept a configuration for sake of the example.

davidor: OK. In that case, I think that the latest example looks better than the original one :+1:

spec/policy_chain_spec.lua
mikz: Fixed in https://github.com/3scale/apicast/pull/496/commits/1fbfff80566e9728a6a77b5cab1db2b6d00a3a5b and will merge it in a rebase.

gateway/src/apicast/cli/environment.lua
mikz: Hopefully https://github.com/3scale/apicast/pull/496/commits/5aa2e0c985aed253c5b29dba54a8a112d240e5c2 is clear enough and describes that the type of this can be a table of strings or nil. Would that be enough?

gateway/src/apicast/cli/environment.lua
mikz: Documented in https://github.com/3scale/apicast/commit/5aa2e0c985aed253c5b29dba54a8a112d240e5c2

gateway/src/apicast/cli/environment.lua
```diff
@@ -0,0 +1,155 @@
...
+    if #nameservers > 0 then
+        return nameservers
+    end
+end
+
+
+local _M = {}
+---
+-- @field default_environment Default environment name.
+-- @table self
```
mikz: Couldn't figure out a better way to document this. This actually means there would be a property `self` of the module returned by this file. 
But there is no way I could find to have properties of the module just plain values with description. Only tables can be properties of the module and only tables can have fields.

spec/policy_chain_spec.lua
mikz: Fixed in https://github.com/3scale/apicast/pull/496/commits/d340bf1b423fd68189419873e5e098607930abcd but going to be squashed.

gateway/src/apicast/policy_chain.lua
mikz: Fixed in https://github.com/3scale/apicast/pull/496/commits/f4be8fb4f04f5588db901b00b97c1965d40b5078 and going to be squashed.

gateway/src/apicast/policy_chain.lua
mikz: Fixed https://github.com/3scale/apicast/pull/496/commits/f4be8fb4f04f5588db901b00b97c1965d40b5078 and going to be squashed.

examples/configuration/echo.lua
mikz: Fixed in https://github.com/3scale/apicast/pull/496/commits/02ad4ab34e95041f9b288c73b27565d845aae123 and going to be squashed.

examples/configuration/echo.lua
mikz: The new example looks like https://github.com/3scale/apicast/pull/496/commits/02ad4ab34e95041f9b288c73b27565d845aae123 so it uses plain `require`. Guess that is ok for now.

t/configuration-loading-boot-with-config.t
mikz: Fixed in https://github.com/3scale/apicast/pull/496/commits/19a3bb85984ef0441779018d6bf4151eb2110aa7 and will be squashed.

gateway/src/apicast/cli/command/start.lua
mikz: Did mass change in https://github.com/3scale/apicast/pull/496/commits/569c52061498256d549b2ad9f9d47673c8cabfc5 and will squash it.

gateway/src/apicast/cli/command/start.lua
mikz: Fixed in https://github.com/3scale/apicast/pull/496/commits/d5a7ca002f60bca98e1dba734238ead55b583050 and will squash it.

examples/configuration/echo.lua
```diff
@@ -0,0 +1,8 @@
+local policy_chain = require('apicast.policy_chain').default()
```
davidor: `examples/configuration`contains many examples for different things. I think it'd be good from now on to explain what each one of them is about. A sentence at the begginning of the file should be enough.

gateway/src/apicast/policy/echo.lua
```diff
@@ -1,9 +1,22 @@
-local policy = require('apicast.policy')
-local _M = policy.new('Echo Policy')
+local _M = require('apicast.policy').new('Echo Policy')
```
davidor: Similar to the comment above. This policy is trivial, but thinking about future contributors, I think that a description of 1 or 2 lines would help.

examples/configuration/echo.lua
mikz: Good point. This one probably should be in own folder.

PR #497, opened 2017-11-22
Add headers policy

### LINKED ISSUE #490: Add policy to add, modify and delete request and response headers
The request is to add a policy that allows the user to add, modify, and delete headers. Both response headers and request ones.

davidor:
Closes #490

### DIFF
--- a/gateway/src/apicast/policy/headers.lua
+++ b/gateway/src/apicast/policy/headers.lua
@@ -1,0 +1,125 @@
+--- Headers policy
+-- This policy allows to include custom headers that will be sent to the
+-- upstream as well as modify or delete the ones included in the original
+-- request.
+-- Similarly, this policy also allows to add, modify, and delete the headers
+-- included in the response.
+
+local ipairs = ipairs
+local type = type
+local insert = table.insert
+
+local policy = require('apicast.policy')
+local _M = policy.new('Headers policy')
+
+local new = _M.new
+
+local function new_header_value(current_value, value_to_add)
+  local new_value = current_value or {}
+
+  if type(new_value) == 'string' then
+    new_value = { new_value }
+  end
+
+  insert(new_value, value_to_add)
+  return new_value
+end
+
+local function push_request_header(header_name, value, req_headers)
+  local new_value = new_header_value(req_headers[header_name], value)
+  ngx.req.set_header(header_name, new_value)
+end
+
+local function set_request_header(header_name, value)
+  ngx.req.set_header(header_name, value)
+end
+
+local function add_request_header(header_name, value, req_headers)
+  if req_headers[header_name] then
+    push_request_header(header_name, value, req_headers)
+  end
+end
+
+local function push_resp_header(header_name, value)
+  local new_value = new_header_value(ngx.header[header_name], value)
+  ngx.header[header_name] = new_value
+end
+
+local function set_resp_header(header_name, value)
+  ngx.header[header_name] = value
+end
+
+local function add_resp_header(header_name, value)
+  if ngx.header[header_name] then
+    push_resp_header(header_name, value)
+  end
+end
+
+local command_functions = {
+  request = {
+    push = push_request_header,
+    add = add_request_header,
+    set = set_request_header
+  },
+  response = {
+    push = push_resp_header,
+    add = add_resp_header,
+    set = set_resp_header
+  }
+}
+
+-- header_type can be 'request' or 'response'.
+local function run_commands(commands, header_type, ...)
+  for _, command in ipairs(commands) do
+    local command_func = command_functions[header_type][command.op]
+    command_func(command.header, command.value, ...)
+  end
+end
+
+-- Initialize the config so we do not have to check for nulls in the rest of
+-- the code.
+local function init_config(config)
+  local res = config or {}
+  res.request = res.request or {}
+  res.response = res.response or {}
+  return res
+end
+
+--- Initialize a Headers policy
+-- @tparam[opt] table config
+-- @field[opt] request Table with the operations to apply to the request headers
+-- @field[opt] response Table with the operations to apply to the response headers
+-- Each operation is a table with three elements:
+--   1) op: can be 'add', 'set' or 'push'.
+--   2) header
+--   3) value
+-- The push operation:
+--   1) When the header is not set, creates it with the given value.
+--   2) When the header is set, it creates a new header with the same name and
+--      the given value.
+-- The set operation:
+--   1) When the header is not set, creates it with the given value.
+--   2) When the header is set, replaces its value with the given one.
+--   3) Deletes a header when the value is "".
+-- The add operation:
+--   1) When the header is not set, it does nothing.
+--   2) When the header is set, it creates a new header with the same name and
+--      the given value.
+function _M.new(config)
+  local self = new()
+  self.config = init_config(config)
+  return self
+end
+
+function _M:rewrite()
+  -- This is here to avoid calling ngx.req.get_headers() in every command
+  -- applied to the request headers.
+  local req_headers = ngx.req.get_headers() or {}
+  run_commands(self.config.request, 'request', req_headers)
+end
+
+function _M:header_filter()
+  run_commands(self.config.response, 'response')
+end
+
+return _M
--- a/t/apicast-policy-headers.t
+++ b/t/apicast-policy-headers.t
@@ -1,0 +1,496 @@
+use lib 't';
+use TestAPIcastBlackbox 'no_plan';
+
+repeat_each(1);
+run_tests();
+
+__DATA__
+
+=== TEST 1: 'set' operation in request headers
+We test 4 things:
+1) Set op with a header that does not exist creates it with the given value.
+2) Set op with a header that exists, clears it and sets the given value.
+3) Set op with an empty value deletes the header.
+4) Check that the headers are received in the upstream but not in the response
+   of the original request.
+--- backend
+  location /transactions/authrep.xml {
+    content_by_lua_block {
+      local expected = "service_token=token-value&service_id=42&usage%5Bhits%5D=2&user_key=value"
+      local args = ngx.var.args
+      if args == expected then
+        ngx.exit(200)
+      else
+        ngx.log(ngx.ERR, expected, ' did not match: ', args)
+        ngx.exit(403)
+      end
+    }
+  }
+--- configuration
+{
+  "services": [
+    {
+      "id": 42,
+      "backend_version":  1,
+      "backend_authentication_type": "service_token",
+      "backend_authentication_value": "token-value",
+      "proxy": {
+        "api_backend": "http://test:$TEST_NGINX_SERVER_PORT/",
+        "proxy_rules": [
+          { "pattern": "/", "http_method": "GET", "metric_system_name": "hits", "delta": 2 }
+        ],
+        "policy_chain": [
+          { "name": "apicast.policy.apicast" },
+          {
+            "name": "apicast.policy.headers",
+            "configuration":
+              {
+                "request":
+                  [
+                    { "op": "set", "header": "New-Header", "value": "config_value_nh" },
+                    { "op": "set", "header": "Existing-Header", "value": "config_value_eh" },
+                    { "op": "set", "header": "Header-To-Delete", "value": "" }
+                  ]
+              }
+          }
+        ]
+      }
+    }
+  ]
+}
+--- upstream
+  location / {
+     content_by_lua_block {
+       local assert = require('luassert')
+       assert.same('config_value_nh', ngx.req.get_headers()['New-Header'])
+       assert.same('config_value_eh', ngx.req.get_headers()['Existing-Header'])
+       assert.is_nil(ngx.req.get_headers()['Header-To-Delete'])
+       ngx.say('yay, api backend');
+     }
+  }
+--- request
+GET /?user_key=value
+--- more_headers
+Existing-Header: request_value_eh1
+Existing-Header: request_value_eh2
+Header-To-Delete: request_value_htd
+--- response_body
+yay, api backend
+--- response_headers
+New-Header:
+Existing-Header:
+Header-To-Delete:
+--- error_code: 200
+--- no_error_log
+[error]
+
+=== TEST 2: 'push' operation in request headers
+We test 3 things:
+1) Push op with a header that does not exist creates it with the given value.
+2) Push op with a header that exists, creates a new header with the same name
+   and the given value.
+3) Check that the headers are received in the upstream but not in the response
+   of the original request.
+--- backend
+  location /transactions/authrep.xml {
+    content_by_lua_block {
+      local expected = "service_token=token-value&service_id=42&usage%5Bhits%5D=2&user_key=value"
+      local args = ngx.var.args
+      if args == expected then
+        ngx.exit(200)
+      else
+        ngx.log(ngx.ERR, expected, ' did not match: ', args)
+        ngx.exit(403)
+      end
+    }
+  }
+--- configuration
+{
+  "services": [
+    {
+      "id": 42,
+      "backend_version":  1,
+      "backend_authentication_type": "service_token",
+      "backend_authentication_value": "token-value",
+      "proxy": {
+        "api_backend": "http://test:$TEST_NGINX_SERVER_PORT/",
+        "proxy_rules": [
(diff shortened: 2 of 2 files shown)


gateway/src/apicast/policy/headers.lua
```diff
@@ -0,0 +1,89 @@
...
+end
+
+local function delete_req_headers(headers)
+  for _, header in ipairs(headers) do
+    ngx.req.clear_header(header)
+  end
+end
+
+local function add_resp_headers(headers)
+  for header, value in pairs(headers) do
```
mikz: `pairs` can't be JITed. So we should either use Luafun or make it a numerically indexed table.
I know this is pain, but that is a limitation of LuaJIT.

davidor: I'd rather evaluate Luafun later.
Do you see any other option apart from having two separate arrays in the config? Something like this:
```
          {
            "name": "apicast.policy.headers",
            "configuration":
              {
                "response":
                  {
                    "headers_to_add": { "Header1", "Header2" }
                    "values_of_headers_to_add": { "value1", "value2" }
                  }
              }
          }
```

mikz: I was thinking about the configuration a bit and it really depends how the UI will look like.
But something like this seemed flexible:


```json
 {
            "name": "apicast.policy.headers",
            "configuration":
              {
                "response":
                  [
                    { "name": "Header1", "op": "set", "value": "foobar" },
                    { "name": "Header1", "op": "add", "value": "foobar1" },
                    { "name": "Header2", "op": "remove" },
                  ]
              }
          }
```

davidor: That could work :+1: 
Also, I see "add" there. What would that do? Add only if the header has not been set already?

mikz: @davidor `add` is the array operation (in my mind). Because all headers that have `,` are technically arrays and can be sent as individual headers:

```
Set-Cookie: foo=bar, one=two
```

Is the same as:

```
Set-Cookie: foo=bar
Set-Cookie: one=two
```

davidor: I see. I didn't get that you were refering to 'add' as the array operation :+1:

davidor: I'm thinking about the difference between `add` and `set` in you're example.

Do you think that `add` should fail if there's no a `set` op for the same header before or done by Apicast? If that's not the case, having just `add` or `set` would be enough.

I don't see a clear use case where we'd like to add a value separated by ',' in a header only if that header has been already defined.

mikz: I think `add` should not care if the value exists. It just should add to it. Its purpose is adding header values without the risk of overriding them. 

Now I wonder if `remove` can be implemented in terms of `set` (empty value). Not sure if nginx is going to send the empty header or not.

andrewdavidmackenzie: How would 'set' work in the case of repeated headers like your cookie example?
Blows away the entire array of values and sets a new value (or an array of values if that is allowed?).

I assume 'remove' removes all values if the header was an array?

Is it possible to 'add' an array of values?

davidor: @andrewdavidmackenzie 
If I understood correctly:
1) Yes. I think that to be coherent with @mikz definition, we should delete the exiting array of values and initialize it with the value passed to the set op.
2) Yes.
3) That's something we need to define.

mikz: Ok, I'll try to spec it out in more detail.

# set

Ability to override any value (including non existing) to a specific value. Removes all previous values and sets the one desired.

```json
{ "op": "set", "name": "X-Header", "value": "foo" }
```

Examples:


```
GET /
X-Header: bar, daz
```

will be transformed to:

```
GET /
X-Header: foo
```

Request:
```
GET /
X-Foo: sometihing
```

will be transformed to:
```
GET /
X-Foo: something
X-Header: foo
```

# remove

Can be implemented as `set` with value of empty string. Nginx will not send out empty headers.


```json
{ "op": "set", "name": "X-Header", "value": "" }
```

# add


```json
{ "op": "add", "name": "X-Header", "value": "foo" }
```

Will append `, value` to the existing value of the header or set a new `value` if there is no previous value. Duplicates allowed.

request

```
GET /
```

transformed to

```
GET / 
X-Header: foo
```

request

```
GET /
X-Header: foo
```

transformed to

```
GET
X-Header: foo, foo
```

davidor: Perfect :ok_hand: 

There's only one question left. @andrewdavidmackenzie wrote it above. Should we allow:
```
{ "op": "add", "name": "X-Header", "value": "abc, def" }
```

To transform the request
```
GET /
X-Header: foo
```

Into:
```
GET /
X-Header: foo, abc, def
```

?

mikz: @davidor I think that is fine and within the specs. There is no escaping of `,`. Less rules - better :)

davidor:
Ready for review @mikz

gateway/src/apicast/policy/headers.lua
```diff
@@ -0,0 +1,106 @@
...
+local function set_request_header(header, value)
+  ngx.req.clear_header(header)
+
+  if value and value ~= '' then
+    ngx.req.set_header(header, value)
+  end
+end
+
+local function add_request_header(header, value)
+  local current_value = ngx.req.get_headers()[header]
```
mikz: We should check the performance of this. `get_headers` will read all the headers. That might not be needed.
You can access individual headers as ngx var `http_header_name` (`ngx.var.http_x_request_id` for example).

gateway/src/apicast/policy/headers.lua
```diff
@@ -0,0 +1,106 @@
...
+  ngx.req.set_header(header, new_header_value(current_value, value))
+end
+
+local function set_resp_header(header, value)
+  ngx.header[header] = nil
+
+  if value and value ~= "" then
```
mikz: Why the condition? We want to set empty string / nil to remove the header no?

gateway/src/apicast/policy/headers.lua
mikz: Ok. Just did a little experiment.

For a request:

```
GET /
Foo: bar
Foo: foo
```

`ngx.req.get_headers()` returns: `foo = { "bar", "foo" }`
`ngx.var.http_foo` returns `"bar"`

That is very unfortunate.

If we need to use get_headers then it should be done once and not in a loop.

gateway/src/apicast/policy/headers.lua
```diff
@@ -0,0 +1,106 @@
...
+local ipairs = ipairs
+
+local policy = require('apicast.policy')
+local _M = policy.new('Headers policy')
+
+local new = _M.new
+
+local function new_header_value(current_value, value)
+  if current_value then
+    return current_value .. ', ' .. value
```
mikz: So I read https://tools.ietf.org/html/rfc7230#section-3.2.2 and it looks like this is more complex than I though because there is an exception: set-cookie header.

Also this is different on request/response. Request can't have duplicated headers and they have to be sent in one field with commas, response can (and they can be concatenated with comma with some exception).

Also the A.2 of describes Cookie in the request has the same issue. https://arxiv.org/pdf/cs/0105018.pdf?

As it is concatenated with semicolon instead of a comma.

mikz: >   A sender MUST NOT generate multiple header fields with the same field
   name in a message unless either the entire field value for that
   header field is defined as a comma-separated list [i.e., #(values)]
   or the header field is a well-known exception (as noted below).

>  A recipient MAY combine multiple header fields with the same field
   name into one "field-name: field-value" pair, without changing the
   semantics of the message, by appending each subsequent field value to
   the combined field value in order, separated by a comma.  The order
   in which header fields with the same field name are received is
   therefore significant to the interpretation of the combined field
   value; a proxy MUST NOT change the order of these field values when
   forwarding a message.

A "sender" is the one sending the message (in HTTP request that would be the client, in HTTP response the server). A "recipient" is the party reading ht HTTP message (server in the request, client in the response).

So my understanding is that the message header can be split into several headers with the same name if the value is a comma separated list and it is not on a blacklist (Cookie, Set-Cookie, ...). 


So I think it should be safe to "add" as a second header with the same name. 
Because it is allowed for the sender to send it split as several headers if the header is array-like. We can't maintain a list of array-like headers but we can tell users to figure it out. 

So I think we would be following the RFC if we would "add" a header as:

```
GET /
Header: val
Header: other
```

And then the recipient can concatenate them with comma. And we leave the blacklist on the user. If they do this operation on some unsupported header (like Host) then it is their problem.

gateway/src/apicast/policy/headers.lua
mikz: From https://github.com/openresty/lua-nginx-module/issues/161#issuecomment-9116263 it looks like both the http_HEADER var and get_headers should have same performance, so better to use get_headers once.

t/apicast-policy-headers.t
```diff
@@ -0,0 +1,355 @@
...
+        "api_backend": "http://test:$TEST_NGINX_SERVER_PORT/",
+        "proxy_rules": [
+          { "pattern": "/", "http_method": "GET", "metric_system_name": "hits", "delta": 2 }
+        ]
+      }
+    }
+  ]
+}
+--- backend
+  location /transactions/authrep.xml {
```
mikz: Do we want to have this in each policy test? I'd say it might be too much and actually testing the apicast policy.

davidor: I was thinking the same. Would be good to have a way of declaring a 'default backend' so we don't copy paste the same thing over and over :)

davidor: I agree that this is too verbose, but at the same time, I think it's important to check that adding this policy does not break the normal auth flow.

mikz: Yes, but we could verify it with the phase logger policy? Just add it after and see it works.

And we can provide default backend in the blackbox test base. If no backend is defined we can add there whatever.

davidor:
@mikz I made some changes as we agreed based on https://tools.ietf.org/html/rfc7230#section-3.2.2

Now the 'add' operation adds a new header with the same name if it already exists, instead of concatenating values separated by commas in the value of the existing header.

I also added a 'push' operation.

t/apicast-policy-headers.t
davidor: I'd rather address this in a separate PR, as this is a problem that we have in multiple test files.

gateway/src/apicast/policy/headers.lua
```diff
@@ -0,0 +1,129 @@
...
+-- The add operation:
+--   1) When the header is not set, creates it with the given value.
+--   2) When the header is set, it creates a new header with the same name and
+--      the given value.
+-- The set operation:
+--   1) When the header is not set, creates it with the given value.
+--   2) When the header is set, replaces its value with the given one.
+--   3) Deletes a header when the value is "".
+-- The push operation:
+--   1) When the header is not set, it does nothing.
```
mikz: I thought `push` would always create new header. Thats why it is **push** no ? Like the array operation (and assumption that no header is an empty array).

gateway/src/apicast/policy/headers.lua
```diff
@@ -0,0 +1,129 @@
...
+--- Initialize a Headers policy
+-- @tparam[opt] table config
+-- @field[opt] request Table with the operations to apply to the request headers
+-- @field[opt] response Table with the operations to apply to the response headers
+-- Each operation is a table with three elements:
+--   1) op: can be 'add', 'set' or 'push'.
+--   2) header
+--   3) value
+-- The add operation:
+--   1) When the header is not set, creates it with the given value.
```
mikz: I thought the **add** would be set only when the header is missing. Like a default value.
Like the shdict operation `add` https://github.com/openresty/lua-nginx-module/blob/master/README.markdown#ngxshareddictadd

gateway/src/apicast/policy/headers.lua
```diff
@@ -0,0 +1,129 @@
...
+end
+
+local function push_request_header(header, value, req_headers)
+  if req_headers[header] then
+    add_request_header(header, value, req_headers)
+  end
+end
+
+local function add_resp_header(header, value)
+  local new_value = new_header_value(ngx.header[header], value)
```
mikz: I'd consider renaming `header` to `name` or `header_name`. Because it looks like there is too much use of `header` (because of the `ngx.header`).

t/apicast-policy-headers.t
mikz: I'd love get rid of those. Or at least have it in just one test. Because this is making plenty of noise that is not helping the test readability.

t/apicast-policy-headers.t
```diff
@@ -0,0 +1,510 @@
...
+    }
+  }
+--- upstream
+  location / {
+     content_by_lua_block {
+       local existing_header_values = ngx.req.get_headers()['Existing-Header']
+
+       if ngx.req.get_headers()['New-Header'] ~= 'config_value_nh' then
+         ngx.log(ngx.ERR, "The header 'New-Header' does not have the correct value")
+       elseif existing_header_values[1] ~= 'request_value_eh1' or
```
mikz: Should use use `assert.same` here ? 

It should be easy as `assert = require("luassert")`.

t/apicast-policy-headers.t
```diff
@@ -0,0 +1,510 @@
...
+              {
+                "request":
+                  [
+                    { "op": "push", "header": "New-Header", "value": "config_value_nh" },
+                    { "op": "push", "header": "Existing-Header", "value": "config_value_eh" }
+                  ]
+              }
+          }
+        ],
+        "api_backend": "http://test:$TEST_NGINX_SERVER_PORT/",
```
mikz: I think it would be nice to reorganise the configuration in a way the most useful options fit the screen with the upstream/request definition. So the backend block could be gone (or moved to the top), the "api_backend" could be on the top of the config and on the bottom would be the policy_chain configuration. The next would be upstream block and the request+response. That should fit one screen so you could read the test and understand all important parts without scrolling...
Just my 2ç.

t/apicast-policy-headers.t
mikz: @davidor missed your previous comment. Yes we can address it in different PR. 👍

gateway/src/apicast/policy/headers.lua
davidor: Good idea :+1:

t/apicast-policy-headers.t
davidor: Yes. Good idea :+1:

t/apicast-policy-headers.t
davidor: Good point :+1:

t/apicast-policy-headers.t
mikz: Also when dealing with this we could consider making api_backend and backend endpoint optional - that could disable the call and result in smaller more focused configuration file. 
I guess when writing these files by hand we want them to be as small and focused as possible - as our tests.

gateway/src/apicast/policy/headers.lua
davidor: Ah. So there has been a misunderstanding here. I got 'push' and 'add' the opposite way.
The way you describe is coherent with what openresty does, so I think it would be good to follow that.

gateway/src/apicast/policy/headers.lua
```diff
@@ -0,0 +1,129 @@
...
+  -- This is here to avoid calling ngx.req.get_headers() in every command
+  -- applied to the request headers.
+  local req_headers = {}
+  if header_type == 'request' then
+    req_headers = ngx.req.get_headers() or {}
+  end
+
+  for _, command in ipairs(commands) do
+    local command_func = command_functions[header_type][command.op]
+    command_func(command.header, command.value, req_headers)
```
mikz: So this is passing the request headers to the response commands too, right?
That might be confusing a bit.
What about calling it `cache` and pass it from the outside (rewrite handler)?

davidor: Yes. This is passing rq headers to response commands too.
Not sure I follow. what would you call `cache`?

mikz: Cache is a bad name. Probably just better to use `...` varargs and pass the `get_headers()` from the `rewrite` function.

PR #498, opened 2017-11-22
Delete 'ONLY' in cors test

davidor:
This was probably set up for debugging in the past and I forgot to delete it.


mikz:
We had a target in the makefile that made it crash when it found ONLY. Looks like it is `make prove` but it is not used on CircleCI because carton is not there. Maybe we could just remove carton support from the make target and use that instead to have the safety net again.

PR #502, opened 2017-11-28
improve documentation

mikz:
* fix make target not updating when deep nested files changed
* break luadoc compat mode for nicer tag syntax


doc/config.ld
```diff
@@ -7,6 +7,7 @@ project = 'APIcast'
 title = 'NGINX based API gateway used to integrate your internal and external API services with 3scale’s API Management Platform'
 merge = true
 format = 'markdown'
+not_luadoc = true
```
mikz: This allows `@field name description`. Instead of:

```
--- description
-- @field name
```

PR #504, opened 2017-11-28
allow passing multiple environment configuration files

mikz:
* the `-e` or `--environment` CLI flag can be repeated
 and it can be either an environment name or a path to a file
* properly handle empty environment values


gateway/src/apicast/cli/environment.lua
```diff
@@ -0,0 +1,135 @@
...
+local pairs = pairs
+local insert = table.insert
+local concat = table.concat
+local re = require('ngx.re')
+
+local _M = {}
+---
+-- @field default_environment Default environment name.
+-- @table self
+_M.default_environment = 'production' -- default environment name
```
davidor: nitpick: redundant comment

mikz: looks like bad rebase, thanks!

gateway/src/apicast/cli/command/start.lua
```diff
@@ -44,10 +44,13 @@ local function update_env(env)
     end
 end
 
-local function nginx_config(context, dir, path, env)
-    update_env(env)
 
-    local template = Template:new(context, dir, true)
+local function apicast_root()
+    return resty_env.get('APICAST_DIR') or pl.path.abspath('.')
```
davidor: Should be `.value`

davidor: Ah didn't see that it's fixed in the second commit :)

gateway/src/apicast/cli/environment.lua
mikz: Fixed!

PR #505, opened 2017-11-28
methods for policy chain manipulation

mikz:
It should be possible to load a policy chain and 
add policies to it.
After this chain is added to the executor it should not be modified.

This is going to be a foundation of how to configure policy chain 
by the environment configuration: https://github.com/3scale/apicast/pull/504

See the generated documentation: https://2580-53965666-gh.circle-artifacts.com/0/doc/modules/policy_chain.html


davidor:
Just for reference, these commits were part of https://github.com/3scale/apicast/pull/496
I read our comments and discussions in that PR and everything looks good :+1:

PR #506, opened 2017-11-28
[circleci] parallelize Test::Nginx suite

mikz:
Reduces CircleCI `prove` from 53 to 13s.
Automatically runs prove in parallel when `nproc` is available. And `make prove` will run tests in randomized directories and ports.

To achieve this all tests need to allocate a random port to listen on. That includes the TCP/UDP servers provided by Test::Nginx to stub DNS server.

The environment configuration can now configure echo/management/backend ports:

```lua
return {
  port = {
      apicast = '8080',
      management = '8090',
      backend = '8081',
      echo = '8081',
    },
}
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,7 +58,11 @@ jobs:
             - lua_modules
       - run: mkdir -p tmp/junit
       - run: $(make rover) exec make busted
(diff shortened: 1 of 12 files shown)


davidor:
So maybe we should wait until we solve the problem before merging this, right?

mikz:
Sure, this can't be merged right now.

Issue #488, opened 2017-11-21
Define resolver in nginx config


mikz:
It is possible that custom policies will want to use cosockets with internal nginx name resolution like https://github.com/nmasse-itix/apicast-logger/issues/2 and using our wrappers would be too verbose.

So adding detected nameservers to the generated nginx config should provide sane default for contributors writing policies as they could use `ngx.socket.tcp:connect` with a name instead of just IP address or using our wrapper.

PR #508, opened 2017-11-28
[nginx] define resolver directive in the config

mikz:
so cosockets use internal resolver and don't have to
go through APIcast resolver.

Closes https://github.com/3scale/apicast/issues/488

### DIFF
--- a/gateway/src/apicast/cli/environment.lua
+++ b/gateway/src/apicast/cli/environment.lua
@@ -14,9 +14,31 @@ local assert = assert
 local error = error
 local print = print
 local pairs = pairs
+local ipairs = ipairs
+local tostring = tostring
 local insert = table.insert
 local concat = table.concat
 local re = require('ngx.re')
+
+local function parse_nameservers()
+    local resolver = require('resty.resolver')
+    local nameservers = {}
+
+    for _,nameserver in ipairs(resolver.init_nameservers()) do
+        -- resty.resolver returns nameservers as tables with __tostring metamethod
+        -- unfortunately those objects can't be joined with table.concat
+        -- and have to be converted to strings first
+        insert(nameservers, tostring(nameserver))
+    end
+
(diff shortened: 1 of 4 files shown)


gateway/src/apicast/cli/environment.lua
```diff
@@ -14,10 +14,26 @@ local assert = assert
 local error = error
 local print = print
 local pairs = pairs
+local ipairs = ipairs
+local tostring = tostring
 local insert = table.insert
 local concat = table.concat
 local re = require('ngx.re')
 
+local function parse_nameservers()
```
davidor: Shouldn't this method be exposed in resty.resolver?

mikz: resty.resolver exposes the nameservers as tables with some metamethods. Those objects can be joined with table.join so they have to be serialized first into `ip:port`. I guess it is worth a comment in the code. Will fix it.

mikz: Fixed.

gateway/src/apicast/cli/environment.lua
```diff
@@ -14,10 +14,32 @@ local assert = assert
...
+    local resolver = require('resty.resolver')
+    local nameservers = {}
+
+    for _,nameserver in ipairs(resolver.init_nameservers()) do
+        -- resty.resolver returns nameservers as tables with __tostring metamethod
+        -- unfrotunately those objects can't be joined with table.concat
```
davidor: typo :eyes:

mikz: ⭕️ Ⓜ️ :godmode:

gateway/src/apicast/cli/environment.lua
davidor: That helps :+1:

PR #512, opened 2017-11-29
Add document that explains the new policies system

### LINKED ISSUE #509: Document how policy chains work
We should document how the new policy chains system work. The main goal is to provide developers with everything they need in order to program and integrate their own policies.

davidor:
Closes #509

### DIFF
--- a/doc/policies.md
+++ b/doc/policies.md
@@ -1,0 +1,181 @@
+# APIcast policies
+
+The behaviour of APIcast is customizable via policies. A policy basically tells
+APIcast what it should do in each of the nginx phases. This document details
+how policies work and how you can write your own and integrate them into
+APIcast.
+
+- [**Policies**](#policies)
+- [**Policy chains**](#policy-chains)
+- [**APIcast default policies**](#apicast-default-policies)
+- [**Write your own policy**](#write-your-own-policy)
+- [**Integrate your policies**](#integrate-your-policies)
+
+
+## Policies
+
+A policy tells APIcast what it should do in each of the nginx phases: `init`,
+`init_worker`, `rewrite`, `access`, `balancer`, `header_filter`, `body_filter`,
+`post_action`, and `log`.
+
+Policies can share data between them. They do that through what we call the
+`context`. Policies can read from and modify that context in every phase.
+
+
+## Policy chains
+
+Policies can be combined using a policy chain. A policy chain is simply a
+sorted list of policies.
+
+The way policy chains work is as follows: suppose that we have a policy A that
+describes what to do in the `rewrite` and `header_filter` phases and a policy B
+that describes what to run in `access` and `header_filter`. Assume also that
+when describing the chain, we indicate that policy A should be run before
+policy B. When APIcast receives a request, it will check the policy chain
+described to see what it should run on each phase:
+- rewrite: execute the function policy A provides for this phase.
+- access: execute the function policy B provides for this phase.
+- balancer: do nothing. Neither policy A nor B describe what to do.
+- header_filter: execute first the function policy A provides for this phase
+  and then the function policy B provides for this phase. Remember that policy
+  chains define an order, and we specified that policy A comes before policy B.
+- body_filter: do nothing. Neither policy A nor B describe what to do.
+- post_action: do nothing. Neither policy A nor B describe what to do.
+- log: do nothing. Neither policy A nor B describe what to do.
+
+Notice that we did not indicate what APIcast does in the `init` and the
+`init_worker` phases. The reason is that those two are not executed in every
+request. `init` is executed when APIcast boots, and `init_worker` when each
+of each of its workers start.
+
+### Types
+
+There are two types of policy chains in APIcast: per-service chains and a
+global chain. Both of them can be configured.
+
+As the name indicates, per-service chains allow us to define a specific chain
+for a given service. This means that we can apply different policies to our
+services, or the same policies but configured differently, or in a different
+order. On the other hand, there's only one global chain, and the behavior it
+defines applies to all the services.
+
+
+## APIcast default policies
+
+By default, APIcast applies the `apicast` policy to all the services. This
+policy includes all the functionality offered by APIcast (mapping rules
+matching, authorization and reporting against 3scale backend, etc.). In the
+future, this policy will be split and each of the resulting policies will be
+replaceable with custom ones.
+
+
+## Write your own policy
+
+To write your own policy you need to write a Lua module that instantiates a
+[Policy](../gateway/src/apicast/policy.lua) and defines a method for each of
+the phases where it needs to execute something.
+
+Suppose that we wanted to run a policy that logged a message in the `rewrite`
+and the `header_filter` phases. This is how our module would look
+like:
+```lua
+local policy = require('apicast.policy')
+local _M = policy.new('My custom policy')
+
+function _M:rewrite()
+  ngx.log(ngx.INFO, 'Passing through the rewrite phase.')
+end
+
+function _M:header_filter()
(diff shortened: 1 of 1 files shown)


mikz:
@nmasse-itix this might be interesting to you

doc/policies.md
```diff
@@ -0,0 +1,180 @@
...
+- [**Integrate your policies**](#integrate-your-policies)
+
+
+## Policies
+
+A policy tells APIcast what it should do in each of the nginx phases: `init`,
+`init_worker`, `rewrite`, `access`, `balancer`, `header_filter`, `body_filter`,
+`post_action`, and `log`.
+
+Policies can share data between them. They do that through what we call the
```
andrewdavidmackenzie: maybe say "share data between phases" to be even more explicit (or people could understand sharing between policies)

doc/policies.md
```diff
@@ -0,0 +1,180 @@
...
+
+Policies can be combined using a policy chain. A policy chain is simply a
+sorted list of policies.
+
+The way policy chains work is as follows: suppose that we have a policy A that
+describes what to do in the `rewrite` and `header_filter` phases and a policy B
+that describes what to run in `access` and `header_filter`. Assume also that
+when describing the chain, we indicate that policy A should be run before
+policy B. When APIcast receives a request, it will check the policy chain
+described to see what it should run on each phase:
```
andrewdavidmackenzie: what would the term/name to use here? "policy chain definition" (sustantivo)

doc/policies.md
```diff
@@ -0,0 +1,180 @@
...
+- rewrite: execute what policy A specifies for this phase.
```
andrewdavidmackenzie: "execute the method/function policy A provides for this phase" or similar?

doc/policies.md
```diff
@@ -0,0 +1,180 @@
...
+### Types
+
+There are two types of policy chains in APIcast: per-service chains and a
+global chain. Both of them can be configured.
+
+As the name indicates, per-service chains allow us to define a specific chain
+for a given service. This means that we can apply different policies to our
+services, or the same policies but configured differently, or in a different
+order. On the other hand, there's only one global chain, and the behavior it
+defines applies to all the services.
```
andrewdavidmackenzie: so, policies defined by both chains will be run?
How is order defined between policies in the global chain and policies in the service chain....

doc/policies.md
```diff
@@ -0,0 +1,180 @@
...
+replaceable with custom ones.
+
+
+## Write your own policy
+
+To write your own policy you need to write a Lua module that instantiates a
+[Policy](../gateway/src/apicast/policy.lua) and defines a method for each of
+the phases where it needs to execute something.
+
+Suppose that we wanted to run a policy that logged some message in the
```
andrewdavidmackenzie: "logged a message"

doc/policies.md
davidor: Yes, per-service are included as part of the global one.

doc/policies.md
```diff
@@ -0,0 +1,180 @@
...
+end
+```
+
+Policies can also have a configuration:
+```lua
+local policy = require('apicast.policy')
+local _M = policy.new('My custom policy')
+local new = _M.new
+
+function _M.new(config)
```
andrewdavidmackenzie: Maybe explain when new() is called?
How to store values from the config it receives for use in the per-request methods later?

doc/policies.md
```diff
@@ -0,0 +1,180 @@
...
+  -- we want to access them from other phases.
+  self.option_a = config.option_a
+  self.option_b = config.option_b
+
+  return self
+end
+```
+
+In the [policy folder](../gateway/src/apicast/policy) you can find several
+policies. These ones are quite simple and can be used as examples to write your
```
andrewdavidmackenzie: "They are quite..."

doc/policies.md
```diff
@@ -0,0 +1,180 @@
...
+own:
+- [Echo](../gateway/src/apicast/policy/echo.lua)
+- [Phase logger](../gateway/src/apicast/policy/phase_logger.lua)
+- [CORS](../gateway/src/apicast/policy/cors.lua)
+- [Headers](../gateway/src/apicast/policy/headers.lua)
+
+
+## Integrate your policies
+
+At some point, it will be possible to configure policies through the 3scale UI,
```
andrewdavidmackenzie: "configure policies and policy chains" ?

doc/policies.md
```diff
@@ -0,0 +1,180 @@
...
+but you can also integrate them in APIcast using a configuration file. Remember
```
andrewdavidmackenzie: use the same verb: "configure"?

doc/policies.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+Policies are specified in the `proxy` object of each service. A policy chain is
+a JSON array where each object represents a policy with a name and an optional
+configuration.
+
+Here's an example that shows how to define a policy chain with the CORS policy
+and the APIcast one. Please note that the configuration of services is more
+complex than shown here. This is a very simplified version to illustrate how
+policies can be integrated:
```
andrewdavidmackenzie: "configured"?

doc/policies.md
```diff
@@ -0,0 +1,180 @@
...
+            "name":"apicast.policy.apicast"
+          }
+        ]
+      }
+    }
+  ]
+}
+```
+
+If instead of configuring the policy chain of a specific service, you need to
```
andrewdavidmackenzie: "you want to"...

doc/policies.md
davidor: I don't think that's correct, because I'm not listing all of them, just some examples.

doc/policies.md
davidor: :white_check_mark:

doc/policies.md
```diff
@@ -0,0 +1,180 @@
...
+return _M
+```
+
+If we wanted to read or write in the `context`:
+```lua
+function _M:rewrite(context)
+  -- Read something from the context.
+  local smth = context.something
+
+  -- Write 'b' in the context so other policies or later phases can read it.
```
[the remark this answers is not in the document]
davidor: :white_check_mark:

doc/policies.md
davidor: :white_check_mark:

doc/policies.md
davidor: :white_check_mark:

doc/policies.md
davidor: :white_check_mark:

doc/policies.md
davidor: :white_check_mark:

doc/policies.md
davidor: The order of policies in the global chain is configurable. The order of policies in per-service chains is also configurable.

doc/policies.md
davidor: To clarify, whatever you write in the context will be available in later phases of the same policy and in other policies as well. Do you think that's clear in the document?

doc/policies.md
davidor: :white_check_mark:

doc/policies.md
davidor: Not sure what you mean here @andrewdavidmackenzie

doc/policies.md
davidor: I think that documenting when it's instantiated it's too much detail for this document.
Your second question is exactly what I wanted to show with this example. See `self.option_a` and `self.option_b` below.

davidor:
Thanks for the review @andrewdavidmackenzie . I addressed all your comments.

PR #513, opened 2017-11-29
[circleci] fail on --ONLY flags in tests

mikz:
* improve make target we had
  * run `rover exec`
  * drop carton
  * support setting the harness


.circleci/config.yml
```diff
@@ -58,7 +58,7 @@ jobs:
             - lua_modules
       - run: mkdir -p tmp/junit
       - run: $(make rover) exec make busted
-      - run: JUNIT_OUTPUT_FILE=tmp/junit/prove.xml $(make rover) exec prove --harness=TAP::Harness::JUnit # no carton on centos
+      - run: JUNIT_OUTPUT_FILE=tmp/junit/prove.xml make prove HARNESS=TAP::Harness::JUnit
```
davidor: Why has this changed from `--harness` to `HARNESS`?

davidor: OK. Got it in the next commit :+1:

mikz: becase this needs to pass the harness param to make target 

so the make target has `HARNESS ?= TAP::Harness` and it can be overridden from the outside to a different harness

before it was executing the prove command directly which accepts it as CLI option

davidor: Why is `HARNESS` defined after the command to run, though?

mikz: make accepts params like that, or by ENV so it works both before and after (but they can have different semantics, like make can be set to not fetch params from ENV)

davidor:
I think we can delete the `carton` target from the Makefile.

Makefile
```diff
@@ -45,7 +45,7 @@ busted: dependencies $(ROVER) ## Test Lua.
 	@- luacov
 
 nginx:
-	@ ($(NGINX) -V 2>&1 | grep -e '--with-ipv6' > /dev/null) || (>&2 echo "$(NGINX) `$(NGINX) -v 2>&1` does not have ipv6 support" && exit 1)
+	@ ($(NGINX) -V 2>&1) > /dev/null
```
davidor: I'm a bit confused by this @mikz 
We were checking that unnecessarily before, then?

mikz: It was necessary in nginx <= 1.12. Because it did not offer ipv6 support by default. 1.13 does.

davidor: :+1:

PR #518, opened 2017-11-30
Update base image of development Dockerfile to use latest s2i-openresty

### LINKED ISSUE #514: Adapt development Dockerfile to latest s2i OpenResty image
The development Dockerfile is using an old version of the s2i OpenResty image: https://github.com/3scale/apicast/blob/285776600e6b79f0f7f80ace88b0f4eee672b911/Dockerfile-development#L1

It also relies on luarocks to install dependencies. We should switch to rover.

davidor:
Closes #514

### DIFF
--- a/Dockerfile-development
+++ b/Dockerfile-development
@@ -1,4 +1,4 @@
-FROM quay.io/3scale/s2i-openresty-centos7:1.11.2.5-1
+FROM quay.io/3scale/s2i-openresty-centos7:1.13.6.1-rover2
 LABEL maintainer="dortiz@redhat.com"
 
 USER root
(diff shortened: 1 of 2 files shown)


Dockerfile-development
```diff
@@ -8,43 +8,23 @@ ARG USER_NAME=default
...
-COPY cpanfile ${USER_HOME}
-COPY gateway/apicast-0.1-0.rockspec ${USER_HOME}/apicast/
-COPY rockspec ${USER_HOME}
+COPY gateway/Roverfile ${USER_HOME}/gateway
+COPY gateway/Roverfile.lock ${USER_HOME}/gateway
 
 WORKDIR ${USER_HOME}
 
-RUN cpan install
+RUN rover install --roverfile=${USER_HOME}/gateway/Roverfile
```
mikz: If you'd copy the Makefile then you can just do `make dependencies`

davidor: Right. I'm going to change this.

PR #522, opened 2017-12-05
Add JSON schemas for the policy configurations

davidor:
This PR adds JSON schemas for the configurations of the policies that have one. This schemas will be used by the UI to know which parameters it needs to ask the user when configuring a policy.

This PR also changes the organization a bit. Now every policy has its own directory with that contains the policy itself and its schema.

Also, in order to be able to require `apicast.policy.a_policy` instead of `apicast.policy.a_policy.policy`, I needed to change lua paths in several places. Ideas on how to make this simpler welcome :)


gateway/conf/nginx.conf.liquid
```diff
@@ -42,7 +42,8 @@ http {
   log_format time '[$time_local] $host:$server_port $remote_addr:$remote_port "$request" $status $body_bytes_sent ($request_time) $post_action_impact';
   access_log off;
 
-  lua_package_path ";;{{prefix}}/?.lua;{{prefix}}/src/?.lua";
+  # src/?/policy.lua allows us to require apicast.policy.apolicy
```
mikz: I think it is fine now. 👍  Later we can solve it with custom loader.

gateway/src/apicast/policy/cors/schema.json
```diff
@@ -0,0 +1,26 @@
...
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
+    "allow_methods": {
+      "type": "array",
+      "items": {
+        "type": "string",
+        "enum": ["GET", "HEAD", "POST", "PUT", "DELETE", "PATCH"]
```
mikz: To be future-proff we definitely should include `OPTIONS` and when we are at it I think we can include `TRACE` and `CONNECT` too. https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods

davidor: Right.

PR #523, opened 2017-12-05
[rockspec] fix luarocks rockspec

mikz:
so apicast can be installed from luarocks

broken since #486


davidor:
Travis build fails because there's a broken link, but it's unrelated to this PR.

davidor:
Fixed.

PR #526, opened 2017-12-11
[rockspec] consolidate rockspecs

mikz:
now just one rockspec is used - the one that is uploaded to luarocks.org

So we don't have to maintain two lists of dependencies.


gateway/apicast-scm-1.rockspec
```diff
@@ -10,11 +10,16 @@ description = {
...
+   'lua-resty-http',
+   'inspect',
+   'router',
+   'lua-resty-jwt',
+   'lua-resty-url',
+   'lua-resty-env',
+   'lua-resty-execvp',
+   'liquid',
+   'argparse',
+   'penlight',
```
davidor: `datafile` disappeared.

mikz: Yep. It is not needed.

mikz: Removing the script that was using it in https://github.com/3scale/apicast/pull/527

PR #528, opened 2017-12-12
[t] use Test::APIcast from CPAN

mikz:
* extract the testing framework to own repository https://github.com/3scale/Test-APIcast
* use s2i image that can install CPAN dependencies

be aware this increases the CI build time (s2i builder+runtime) significantly as there are dependencies that have to be pulled


t/apicast-async-reporting.t
```diff
@@ -1,5 +1,5 @@
 use lib 't';
-use TestAPIcast 'no_plan';
+use Test::APIcast 'no_plan';
```
davidor: This namespace looks a bit strange to me. Is it common in Perl to start testing libraries with `Test::`?
I guess that's it, because we also have `Test::Nginx`.

davidor:
:+1: 
remember to fix the changelog conflict before merging :)

t/apicast-async-reporting.t
mikz: Yep. And there is Test::Simple and Test::More.

davidor: :+1:

Issue #495, opened 2017-11-21
Policies should be able to extend credentials 


mikz:
Some policies will need to access/modify credentials the client authenticated with.

Unfortunately they are extracted and sent to backend in one method:
https://github.com/3scale/apicast/blob/632d0f38a436b95c864be0c901713ff0cd3dbc5c/gateway/src/apicast/proxy.lua#L238-L293

That makes it very hard to hook into this process. 

For example adding support for End User Plans needs to add `user_id` to the credentials: https://github.com/3scale/apicast/commit/3b2a0da509b5d14375f4042c185f58df51fc0555

One way to solve this would be to split credentials extraction into own module. 
Another way would be to have some "credentials extractor" object that is exported into the shared context by the apicast module, but could be exported by another module and overridden.

Another one would be to do this in rewrite phase and leave access phase just for the actual call. 
Then any policy could manipulate the credentials between the rewrite and access phase.

/cc @kevprice83

PR #529, opened 2017-12-12
Introduce URL rewriting policy

davidor:
This is a first version of the URL rewriting policy. I'd rather agree on the basics first and add more functionality later.

Currently, this policy works well when placed after the apicast policy in the policy chain. However, it does not when placed before the apicast policy. In that case, the URL would be rewritten, and the matching rules would apply over that rewritten URL. To make that work, we need to make a few changes in the code before. For example, we need to stop relying on `ngx.var.request` when matching mapping rules, because it does not change after rewriting the url with `ngx.req.set_uri`.

### DIFF
--- a/gateway/src/apicast/policy/url_rewriting/policy.lua
+++ b/gateway/src/apicast/policy/url_rewriting/policy.lua
@@ -1,0 +1,75 @@
+--- URL rewriting policy
+-- This policy allows to modify the path of a request.
+
+local ipairs = ipairs
+local sub = ngx.re.sub
+local gsub = ngx.re.gsub
+
+local policy = require('apicast.policy')
+local _M = policy.new('URL rewriting policy')
+
+local new = _M.new
+
+local substitute_functions = { sub = sub, gsub = gsub }
+
+-- func needs to be ngx.re.sub or ngx.re.gsub.
+-- This method simply calls one of those 2. They have the same interface.
+local function substitute(func, subject, regex, replace, options)
+  local new_uri, num_changes, err = func(subject, regex, replace, options)
+
+  if not new_uri then
+    ngx.log(ngx.WARN, 'There was an error applying the regex: ', err)
+  end
+
+  return new_uri, num_changes > 0
+end
+
+-- Returns true when the URL was rewritten and false otherwise
+local function apply_rewrite_command(command)
+  local func = substitute_functions[command.op]
+
+  if not func then
+    ngx.log(ngx.WARN, "Unknown URL rewrite operation: ", command.op)
+  end
+
+  local new_uri, changed = substitute(
+    func, ngx.var.uri, command.regex, command.replace, command.options)
+
+  if changed then
+    ngx.req.set_uri(new_uri)
+  end
+
+  return changed
+end
+
+--- Initialize a URL rewriting policy
+-- @tparam[opt] table config Contains the rewrite commands.
+-- The rewrite commands are based on the 'ngx.re.sub' and 'ngx.re.gsub'
+-- functions provided by OpenResty. Please check
+-- https://github.com/openresty/lua-nginx-module for more details.
+-- Each rewrite command is a table with the following fields:
+--
+--   - op: can be 'sub' or 'gsub'.
+--   - regex: regular expression to be matched.
+--   - replace: string that will replace whatever is matched by the regex.
+--   - options[opt]: options to control how the regex match will be done.
+--     Accepted options are the ones in 'ngx.re.sub' and 'ngx.re.gsub'.
+--   - break[opt]: defaults to false. When set to true, if the command rewrote
+--     the URL, it will be the last command applied.
+function _M.new(config)
+  local self = new()
+  self.config = config or {}
+  return self
+end
+
+function _M:rewrite()
+  for _, command in ipairs(self.config) do
+    local rewritten = apply_rewrite_command(command)
+
+    if rewritten and command['break'] then
+      break
+    end
+  end
+end
+
+return _M
--- a/gateway/src/apicast/policy/url_rewriting/schema.json
+++ b/gateway/src/apicast/policy/url_rewriting/schema.json
@@ -1,0 +1,32 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "URL rewriting policy configuration",
+  "type": "object",
+  "properties": {
+    "commands": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "op": {
+            "type": "string",
+            "enum": ["sub", "gsub"]
+          },
+          "regex": {
+            "type": "string"
+          },
+          "replace": {
+            "type": "string"
+          },
+          "options": {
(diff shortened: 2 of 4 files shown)


gateway/src/apicast/policy/url_rewriting/policy.lua
```diff
@@ -0,0 +1,80 @@
...
+    ngx.log(ngx.WARN, 'There was an error applying the regex: ', err)
+  end
+
+  return new_uri, num_changes > 0
+end
+
+-- Returns the substitute function to be applied according to the rewrite
+-- operation. The result can be ngx.re.sub or ngx.re.gsub.
+local function substitute_func(rewrite_operation)
+  if rewrite_operation == 'sub' then
```
mikz: Maybe better to have a dispatch table?

gateway/src/apicast/policy/url_rewriting/schema.json
```diff
@@ -0,0 +1,32 @@
...
+          "regex": {
+            "type": "string"
+          },
+          "replace": {
+            "type": "string"
+          },
+          "options": {
+            "type": "string"
+          },
+          "break": {
```
mikz: We are considering also other mutually exclusive values like: `last`, `redirect`, `persistent` right?

IMO might be nicer to model it in a way they are truly mutually exclusive by being a value of one key: `"type": "break"`. 

But since we are not releasing this yet, It could wait until we have some more options.

t/apicast-policy-url-rewriting.t
```diff
@@ -0,0 +1,236 @@
+use lib 't';
+use TestAPIcastBlackbox 'no_plan';
+
+repeat_each(1);
```
mikz: Any reason why we can't have default 2 runs ?

davidor: No. My bad. Copied it from the wrong place.

gateway/src/apicast/policy/url_rewriting/policy.lua
davidor: OK. I don't have a strong preference about this one. I applied that pattern in the headers policy https://github.com/3scale/apicast/blob/d8ddb6a8313e22637d7b0c3bca35a3c645d106ad/gateway/src/apicast/policy/headers/policy.lua#L58

mikz: I have a feeling it could have better performance, but can't really say without a benchmark/profile/jit trace.

davidor:
I changed the couple of things that you mentioned in your comments @mikz

gateway/src/apicast/policy/url_rewriting/policy.lua
```diff
@@ -0,0 +1,74 @@
...
+
+  if changed then
+    ngx.req.set_uri(new_uri)
+  end
+
+  return changed
+end
+
+--- Initialize a URL rewriting policy
+-- @tparam[opt] table config. Contains the rewrite commands.
```
mikz: the dot after config looks like extra. You can see it at https://3244-53965666-gh.circle-artifacts.com/0/doc/modules/policy.url_rewriting.policy.html

gateway/src/apicast/policy/url_rewriting/policy.lua
```diff
@@ -0,0 +1,74 @@
...
+-- The rewrite commands are based on the 'ngx.re.sub' and 'ngx.re.gsub'
+-- functions provided by OpenResty. Please check
+-- https://github.com/openresty/lua-nginx-module for more details.
+-- Each rewrite command is a table with the following fields:
+--   - op: can be 'sub' or 'gsub'.
+--   - regex: regular expression to be matched.
+--   - replace: string that will replace whatever is matched by the regex.
+--   - options[opt]: options to control how the regex match will be done.
+--     Accepted options are the ones in 'ngx.re.sub' and 'ngx.re.gsub'.
+--   - break[opt]: defaults to false. When set to true, if the command rewrote
```
mikz: This is not recognized by the ldoc and looks kind of ugly in the generated doc. Not sure how better describe it though.

gateway/src/apicast/policy/url_rewriting/policy.lua
davidor: Well spotted. Fixed.

gateway/src/apicast/policy/url_rewriting/policy.lua
davidor: Fixed :)

MarkCheshire:
I don't think it is a valid use case to apply the rewrite policy before the Apicast policy, so I recommend to avoid adding that complexity.

andrewdavidmackenzie:
I agree - we should document that in some customer visible place so that people looking at using this policy see that they should place it after.

andrewdavidmackenzie:
Talking with @davidor I understand some more and so expand the explanation.....

I think this is the first policy we add that has a coupling with an existing policy (apicast policy)?
i.e. Mapping rules matches paths of the request and increments metrics etc, and the URL rewrite policy can change request paths.

Longer-term I think it's valid to consider mapping rules "one more policy", but at the moment it's pretty baked-in to both apicast (via the apicast policy) and System UI.
*I think* in System for AMP 2.2 we are not planning to extract this as an optional policy that can be placed anywhere in the pipeline (please correct me if I'm wrong).

So, for AMP 2.2 - the fact we have it all baked-in kind of avoids the problem.

However, if in the future we do break it out Mapping Rules as an optional, re-orderable policy - then there would be a dependency between the policy configs that the user would have to be aware of, and reordering of policies would change the overall behaviour, (or put another way - Mapping rules would have to be written knowing the urls produced by a previous url rewrite policy).

Does that make sense?

mikz:
It does make sense. But it is all possible now without extracting mapping rules into own policy.
It is enough to move extracting mapping rules into different phase - from access to rewrite.

We have a need to do something between mapping rules / credentials extraction and calling 3scale backend (https://github.com/3scale/apicast/issues/495). Customers override internal methods to add custom metrics.

By moving the mapping rules and credentials extraction into rewrite phase we allow custom policies to hook into the process and modify matched metrics and credentials before they are sent to backend.

That gives us enough flexibility in 2.2 release without extracting Mapping Rules into own policy. 

And rewriting URLs policy should work both when it is before/after APIcast policy.
Simply because you indicate the rewriting should project in the mapped metrics or not.
When used before you indicate that you want mapping rules to be applied on rewritten URL.
When used after you want to apply mapping rules on what client sent not considering the rewritten URL.

The change is very small and the use case is valid. But the code in question is frequently overridden by customers: https://github.com/3scale/apicast/blob/632d0f38a436b95c864be0c901713ff0cd3dbc5c/gateway/src/apicast/configuration.lua#L215
So we would like to also give them means to accomplish what they want without overriding internal methods. Hence the extraction of mapping rules / credentials to the rewrite phase.
That is needed for other policies to work anyway (like SOAP that needs to add custom metric based on a header).

I appreciate the concern, but we are going to make it work and make the code better and faster in the process. Simply because we need it for other things too.

PR #530, opened 2017-12-13
Fix ldocs formatting in the CORS policy

davidor:
Introduce extra empty lines so the list shows correctly with one item per line.


davidor:
Added a changelog entry to make the probot-changelog check pass and be able to merge.

PR #532, opened 2017-12-18
use apicast.loader to load policies

mikz:
instead of using a package.path


davidor:
I'd add a comment in the existing description of the loader module explaining that now it's also responsible for loading policies.

davidor:
Travis build fails because some links in docs.docker.com are broken.

PR #535, opened 2017-12-22
introduce content phase

mikz:
Content phase is for generating actual content. APIcast policy is going to use `ngx.exec` to  do internal redirect to the `@upstream` location which handles fetching the upstream response by post_action.

Policies can hook into this process by implementing own `content` handler that will respond with headers and body. For example some policy could fetch the upstream response by cosocket API instead of using nginx proxy_pass.

# Performance

## Before

```
wrk --connections 10 --threads 10 --duration 60 'http://localhost:8080/?user_key=foo'
Running 1m test @ http://localhost:8080/?user_key=foo
  10 threads and 10 connections
    Latency     3.46ms    1.02ms  15.42ms   82.60%
    Req/Sec   291.47     31.81   626.00     79.82%
  174339 requests in 1.00m, 63.34MB read
Requests/sec:   2901.39
Transfer/sec:      1.05MB
```

## After

```
wrk --connections 10 --threads 10 --duration 60 'http://localhost:8080/?user_key=foo'
Running 1m test @ http://localhost:8080/?user_key=foo
  10 threads and 10 connections
    Latency     3.59ms    1.01ms  16.29ms   82.44%
    Req/Sec   280.77     30.10   404.00     78.13%
  167966 requests in 1.00m, 61.02MB read
Requests/sec:   2795.01
Transfer/sec:      1.02MB
```

### DIFF
--- a/t/apicast-policy-chains.t
+++ b/t/apicast-policy-chains.t
@@ -8,11 +8,8 @@ __DATA__
 === TEST 1: custom policy chain
 This test uses the phase logger policy to verify that all of its phases are run
 when we use a policy chain that contains it. The policy chain also contains the
-normal apicast policy, so we can check that the authorize flow continues
-working. Notice that 'post_action' and 'log' are not ran. There's no
-post_action defined and log only runs when post_action runs.
-init and init_worker are not run either. They are not executed in policies
-defined at the service level.
+normal apicast policy, so we can check that the authorize flow continues working.
+Phases init and init_worker are not executed for policies defined at the service level.
 
 --- http_config
   include $TEST_NGINX_UPSTREAM_CONFIG;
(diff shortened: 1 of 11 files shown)


t/apicast-policy-chains.t
```diff
@@ -8,11 +8,7 @@ __DATA__
 === TEST 1: custom policy chain
 This test uses the phase logger policy to verify that all of its phases are run
 when we use a policy chain that contains it. The policy chain also contains the
-normal apicast policy, so we can check that the authorize flow continues
-working. Notice that 'post_action' and 'log' are not ran. There's no
-post_action defined and log only runs when post_action runs.
-init and init_worker are not run either. They are not executed in policies
-defined at the service level.
+normal apicast policy, so we can check that the authorize flow continues working.
```
davidor: The part about `init` and `init_worker` still applies, but has been deleted.

mikz: right!

PR #540, opened 2018-01-05
[vagrant] use fedora 26

mikz:
Change Vagrant box from CentOS to Fedora so we can use the latest tooling.
For development it makes sense to use latest profiling tooling like eBPF, perftools, SystemTap, ...
In centos all those tools are quite ancient because of the 3.11 kernel.

* support for running/testing with different openresty binaries
* install cpan dependencies inside vagrant
* run redis inside vagrant
* respect APICAST_DIR
* performance tune nginx config
* link all tools from openresty systemtap toolkit

### DIFF
--- a/gateway/bin/apicast
+++ b/gateway/bin/apicast
@@ -82,7 +82,13 @@ _LUA_
     $bin = $lua_file;
 }
 
+
 my @resty_args = ();
+
+my $nginx = $ENV{APICAST_OPENRESTY_BINARY} || $ENV{TEST_NGINX_BINARY};
+if (defined $nginx) {
+    push @resty_args, '--nginx', $nginx;
+}
 
 # Add directories to the lua load path.
 # APIcast source and a local src directory.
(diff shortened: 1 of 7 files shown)


gateway/bin/apicast
```diff
@@ -82,7 +82,9 @@ _LUA_
     $bin = $lua_file;
 }
 
-my @resty_args = ();
+my $nginx = $ENV{TEST_NGINX_BINARY} || $ENV{APICAST_OPENRESTY_BINARY} || 'openresty';
```
davidor: Any reason why `TEST_NGINX_BINARY` has preference over `APICAST_OPENRESTY_BINARY` ?

We are doing the opposite in https://github.com/3scale/apicast/blob/25009f6e3f6029931066aa1c0737d70014a00570/gateway/src/apicast/cli/command/start.lua#L100

Also, it might be worth it to extract the method that selects the binary to a place where the cli and this bin can call it.

davidor:
What's the reason to change from centos to fedora? Would be good to include in the commit or the PR description.

gateway/bin/apicast
mikz: Good point, will make it the same.

mikz:
@davidor updated the description. Main reason is outdated profiling toolchain on CentOS.

PR #541, opened 2018-01-05
Extract rejection reason in Proxy instead of CacheHandler

davidor:
Extracting the rejection reason and checking if the request has been authorized has nothing to do with the cache. We only need to check the response status and a header. This is why I'd rather not have this in the cache_handler module.

Apart from that, this change will simplify the implementation of the caching policy.


davidor:
@mikz I added a changelog and fixed a minor codeclimate issue.

PR #543, opened 2018-01-08
[nginx] performance tuning nginx config

mikz:
* enable PCRE JIT
* decrease timer resolution
  

# Before

```
Running 5m test @ http://localhost:8080/?user_key=foo
  10 threads and 100 connections
    Latency    35.62ms   43.91ms   1.17s    96.23%
    Req/Sec   330.45    119.79     1.43k    60.45%
  978844 requests in 5.00m, 350.02MB read
Requests/sec:   3262.07
Transfer/sec:      1.17MB
```

# After

```
Running 5m test @ http://localhost:8080/?user_key=foo
  10 threads and 100 connections
    Latency    30.21ms   28.69ms   1.22s    96.09%
    Req/Sec   360.57    111.95     0.94k    60.84%
  1071946 requests in 5.00m, 383.31MB read
Requests/sec:   3571.97
Transfer/sec:      1.28MB
```

---

Looks like 10% increase.


davidor:
It would be interesting to measure the performance improvement.
The benchmark task says 1.8krps vs 4.8krps. I don't think it can be trusted :)

PR #546, opened 2018-01-09
Cache policy

davidor:
First version of the caching policy.

The functionality is almost the same as the behavior that can be achieved using `APICAST_BACKEND_CACHE_HANDLER`. Apart from the two modes supported (strict, resilient), this policy adds the possibility of configure Apicast to not use a cache and make all the calls synchronously.

This does not break compatibility. If this policy is not included, users can still use `APICAST_BACKEND_CACHE_HANDLER`.

This PR duplicates some of the code in `cache_handler.lua`. I think that's fine because we'll get rid of the duplication once we split Apicast into smaller policies.

### DIFF
--- a/gateway/src/apicast/backend/cache_handler.lua
+++ b/gateway/src/apicast/backend/cache_handler.lua
@@ -40,7 +40,7 @@ function _M.handlers.strict(cache, cached_key, response, ttl)
     -- so to not write the cache twice lets write it just in authorize
 
     if fetch_cached_key(cached_key) ~= cached_key then
-      ngx.log(ngx.INFO, 'apicast cache write key: ', cached_key, ', ttl: ', ttl, ' sub: ')
+      ngx.log(ngx.INFO, 'apicast cache write key: ', cached_key, ', ttl: ', ttl)
       cache:set(cached_key, 200, ttl or 0)
     end
   else
--- a/gateway/src/apicast/policy/caching/policy.lua
+++ b/gateway/src/apicast/policy/caching/policy.lua
@@ -2,13 +2,14 @@
 -- Configures a cache for the authentication calls against the 3scale backend.
 -- The 3scale backend can authorize (status code = 200) and deny (status code =
 -- 4xx) calls. When it fails, it returns a 5xx code.
--- This policy support two kinds of caching:
+-- This policy support three kinds of caching:
 --   - Strict: it only caches authorized calls. Denied and failed calls
 --     invalidate the cache entry.
 --   - Resilient: caches authorized and denied calls. Failed calls do not
 --     invalidate the cache. This allows us to authorize and deny calls
 --     according to the result of the last request made even when backend is
 --     down.
+--   - None: disables caching.
 
 local policy = require('apicast.policy')
 local _M = policy.new('Caching policy')
@@ -37,22 +38,27 @@ local function resilient_handler(cache, cached_key, response, ttl)
   end
 end
 
+local function disabled_cache_handler()
+  ngx.log(ngx.DEBUG, 'Caching is disabled. Skipping cache handler.')
+end
+
 local handlers = {
   resilient = resilient_handler,
-  strict = strict_handler
+  strict = strict_handler,
+  none = disabled_cache_handler
 }
 
 local function handler(config)
   if not config.caching_type then
-    ngx.log(ngx.ERR, 'Caching type not specified, falling back to strict.')
-    return handlers.strict
+    ngx.log(ngx.ERR, 'Caching type not specified. Disabling cache.')
+    return handlers.none
   end
 
   local res = handlers[config.caching_type]
 
   if not res then
-    ngx.log(ngx.ERR, 'Invalid caching type, falling back to strict.')
-    res = handlers.strict
+    ngx.log(ngx.ERR, 'Invalid caching type. Disabling cache.')
+    res = handlers.none
   end
 
   return res
--- a/gateway/src/apicast/policy/caching/policy.lua
+++ b/gateway/src/apicast/policy/caching/policy.lua
@@ -1,0 +1,74 @@
+--- Caching policy
+-- Configures a cache for the authentication calls against the 3scale backend.
+-- The 3scale backend can authorize (status code = 200) and deny (status code =
+-- 4xx) calls. When it fails, it returns a 5xx code.
+-- This policy support two kinds of caching:
+--   - Strict: it only caches authorized calls. Denied and failed calls
+--     invalidate the cache entry.
+--   - Resilient: caches authorized and denied calls. Failed calls do not
+--     invalidate the cache. This allows us to authorize and deny calls
+--     according to the result of the last request made even when backend is
+--     down.
+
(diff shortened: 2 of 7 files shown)


gateway/src/apicast/policy/caching/schema.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Caching policy configuration",
+  "type": "object",
+  "properties": {
+    "exit": {
+      "type": "caching_type",
+      "enum": ["resilient", "strict"]
```
mikz: We could offer `none` as suggested in https://issues.jboss.org/browse/THREESCALE-587.

davidor: :+1:

gateway/src/apicast/backend/cache_handler.lua
```diff
@@ -34,28 +34,29 @@ local function fetch_cached_key()
...
-  if response.status == 200 then
+function _M.handlers.strict(cache, cached_key, status_code, ttl)
+  if status_code == 200 then
     -- cached_key is set in post_action and it is in in authorize
     -- so to not write the cache twice lets write it just in authorize
 
     if fetch_cached_key(cached_key) ~= cached_key then
-      ngx.log(ngx.INFO, 'apicast cache write key: ', cached_key, ', ttl: ', ttl, ' sub: ')
+      ngx.log(ngx.INFO, 'apicast cache write key: ', cached_key,
+                        ', ttl: ', ttl, ' sub: ')
```
mikz: Hm. The `sub: ` there does not really say anything. I guess 3e can remove it as it is some copy paste error.

gateway/src/apicast/policy/caching/policy.lua
```diff
@@ -0,0 +1,94 @@
...
+
+local policy = require('apicast.policy')
+local _M = policy.new('Caching policy')
+
+local new = _M.new
+
+local function strict_handler(cache, cached_key, status_code, ttl)
+  if status_code == 200 then
+    ngx.log(ngx.INFO, 'apicast cache write key: ', cached_key,
+                      ', ttl: ', ttl, ' sub: ')
```
mikz: Here is also `sub: ` without anything after it.

gateway/src/apicast/policy/caching/policy.lua
```diff
@@ -0,0 +1,94 @@
...
+  if status_code and status_code < 500 then
+    ngx.log(ngx.INFO, 'apicast cache write key: ', cached_key,
+                      ' status: ', status_code, ', ttl: ', ttl)
+
+    cache:set(cached_key, status_code, ttl or 0)
+  end
+end
+
+local function disabled_cache_handler()
+  return function() end
```
mikz: This is going to allocate function every time this is executed. It is not necessary no?

gateway/src/apicast/policy/caching/policy.lua
```diff
@@ -0,0 +1,94 @@
...
+  end
+end
+
+local function init_config(config)
+  local res = {}
+
+  if not config.caching_type then
+    ngx.log(ngx.ERR, 'Caching type not specified. Cache disabled.')
+    res.caching_type = 'none'
+  elseif config.caching_type == 'resilient' or
```
mikz: This would be nicer by checking a key in a table no?

gateway/src/apicast/policy/caching/policy.lua
```diff
@@ -0,0 +1,94 @@
...
+-- @tparam table context
+-- Overrides the cache_handler method of the proxy object included in the
+-- context. If no proxy object is received, it raises an error.
+function _M:access(context)
+  if not context.proxy then
+    ngx.log(ngx.ERR, 'Proxy not defined in the context')
+    return
+  end
+
+  context.proxy.cache_handler = handler(self.config.caching_type)
```
mikz: What if this policy would just export the correct key in `:export()` method?
Then the APIcast policy could set the `cache_handler` to the proxy if it is in the context.

Then the "ownership"  of the proxy object would still be in the APIcast module. 
I'd be a bit worried that this is pretty deep access into internals of a module from some unrelated part of code. APIcast module initializes the Proxy object, so it could pass the cache handler as an argument for example.

gateway/src/apicast/backend/cache_handler.lua
```diff
@@ -34,28 +34,29 @@ local function fetch_cached_key()
   return ok and stored
 end
 
-function _M.handlers.strict(cache, cached_key, response, ttl)
-  if response.status == 200 then
+function _M.handlers.strict(cache, cached_key, status_code, ttl)
```
mikz: This is breaking change for someone who written a custom handler, right?

gateway/src/apicast/backend/cache_handler.lua
davidor: Ah your right. I didn't pay attention to this line and then copied it to the new policy.

gateway/src/apicast/policy/caching/policy.lua
davidor: Right. We can choose one in the initializer.

gateway/src/apicast/backend/cache_handler.lua
davidor: Yes. Didn't think about users overridding this method.

gateway/src/apicast/policy/caching/policy.lua
davidor: Definitely. I was thinking about the phases (`rewrite`, `access`, etc.) and forgot about `export()`.

davidor: It turns out that `export()` is not working correctly for policies included in the local policy chain.

I suggest we address that problem on a separate PR. I've been investigating a bit and didn't find a straightforward solution.

To address your concern, in the rewrite phase of the cache policy, I included the cache_handler in the context. Later, the Apicast policy checks if it has been included, and if so, overrides the cache_handler method of the instance of Proxy that it owns.

davidor:
@mikz I addressed your comments. Ready for review.

gateway/src/apicast/policy/caching/policy.lua
mikz: This line can be removed, right? 
Or maybe we could add a debug log line saying that it skipped cache.

davidor: :+1:

gateway/src/apicast/policy/caching/policy.lua
```diff
@@ -0,0 +1,80 @@
...
+end
+
+local function disabled_cache_handler()
+  ngx.log(ngx.DEBUG, 'Caching is disabled. Skipping cache handler.')
+end
+
+local handlers = {
+  resilient = resilient_handler,
+  strict = strict_handler,
+  none = disabled_cache_handler
```
mikz: @davidor we also should have "allow" handler that will allow the the request when backend is down.

PR #549, opened 2018-01-10
updates specific env var names to be more consistent with OS template…

### LINKED ISSUE #548: Inconsistent env var names with OS template names
I think we should have more consistency between the APIcast env var names and the openshift template param names.

For example: [APICAST_SERVICES](https://github.com/3scale/apicast/blob/master/doc/parameters.md#apicast_services) should be `APICAST_SERVICES_LIST` to be consistent with the [template equivalent](https://github.com/3scale/apicast/blob/master/openshift/apicast-template.yml#L40)

Do we want to do this for other parameters? All the parameters? At least wherever it makes sense we should.

kevprice83:
… param names fixes #548 

@3scale/documentation please take this into account when preparing docs for the next release


gateway/src/apicast/configuration_loader/remote_v2.lua
```diff
@@ -183,7 +183,7 @@ function _M:call(environment)
 end
 
 local services_subset = function()
-  local services = resty_env.get('APICAST_SERVICES')
+  local services = resty_env.get('APICAST_SERVICES_LIST') or resty_env.get('APICAST_SERVICES')
```
mikz: This would fail if `APICAST_SERVICES_LIST` is empty string but `APICAST_SERVICES` is something.

Would be good to use `resty_env.value` instead that will ignore empty value and return `nil`.

kevprice83: ah okay good point

kevprice83:
@mikz shall I do the same for the `APICAST_PATH_ROUTING_ENABLED` parameter or we just keep this to one change in this PR?

mikz:
@kevprice83 sure

kevprice83:
@jmprusi a couple of env vars have changed names here, not sure if that will have to be taken into account for the 2.1 => 2.2 migration. I've at least tested my own image built from this PR and the new template and all works well. Not tried doing a migration though.

PR #551, opened 2018-01-11
[http_ng] improve performance

mikz:
Try to apply some smaller and bigger optimizations to improve overral performance of the client.

see profiles: https://gist.github.com/mikz/e8d29fcb60374cff1174930658da694c

# Performance

## before

```
wrk --connections 10 --threads 10 --duration 60 'http://localhost:8080/?user_key=foo'
Running 1m test @ http://localhost:8080/?user_key=foo
  10 threads and 10 connections
    Latency     6.15ms    1.86ms  28.68ms   82.61%
    Req/Sec   163.82     22.43   373.00     71.98%
  98004 requests in 1.00m, 35.61MB read
Requests/sec:   1630.88
Transfer/sec:    606.72KB
```

## after

```
wrk --connections 10 --threads 10 --duration 60 'http://localhost:8080/?user_key=foo'
Running 1m test @ http://localhost:8080/?user_key=foo
  10 threads and 10 connections
    Latency     5.71ms    1.45ms  40.24ms   81.81%
    Req/Sec   176.55     24.01   346.00     78.54%
  105626 requests in 1.00m, 38.37MB read
Requests/sec:   1757.55
Transfer/sec:    653.85KB
```

### DIFF
--- a/gateway/src/resty/http_ng/headers.lua
+++ b/gateway/src/resty/http_ng/headers.lua
@@ -9,8 +9,8 @@ local getmetatable = getmetatable
 local type = type
 local lower = string.lower
 local upper = string.upper
-local sub = string.sub
 local ngx_re = ngx.re
+local re_gsub = ngx_re.gsub
 
 local normalize_exceptions = {
   etag = 'ETag'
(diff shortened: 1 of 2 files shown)


davidor:
This is great :+1: 

In the future, I think it'd be good to perform a benchmark for each optimization so we can see the impact of each of them separately.

andrewdavidmackenzie:
A classic optimization trade-off result I see: averages go down, and maximums go up.

PR #552, opened 2018-01-15
[wip][remote] fallback to loading all proxy configs

mikz:
needed for multi-tenancy


mikz:
Not necessary. Works on master already.

PR #554, opened 2018-01-18
[cli] detect number of CPU cores

mikz:
* use nproc to detect number of cores
  - nproc correctly detects limits set by docker

/cc @jmprusi

### DIFF
--- a/gateway/src/apicast/cli/command/start.lua
+++ b/gateway/src/apicast/cli/command/start.lua
@@ -179,7 +179,7 @@ local function configure(cmd)
     cmd:flag("-d --daemon", "Daemonize.")
     cmd:option("-w --workers",
         "Number of worker processes to start.",
-        resty_env.value('APICAST_WORKERS') or 1)
+        resty_env.value('APICAST_WORKERS') or Environment.default_config.worker_processes)
     cmd:option("-p --pid", "Path to the PID file.")
     cmd:mutex(
         cmd:flag('-b --boot',
--- a/gateway/src/apicast/cli/environment.lua
+++ b/gateway/src/apicast/cli/environment.lua
@@ -6,6 +6,7 @@
 local pl_path = require('pl.path')
 local resty_env = require('resty.env')
 local linked_list = require('apicast.linked_list')
+local util = require('apicast.util')
 local setmetatable = setmetatable
 local loadfile = loadfile
 local pcall = pcall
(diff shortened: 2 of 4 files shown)


gateway/src/apicast/cli/command/start.lua
```diff
@@ -179,7 +179,7 @@ local function configure(cmd)
     cmd:flag("-d --daemon", "Daemonize.")
     cmd:option("-w --workers",
         "Number of worker processes to start.",
-        resty_env.value('APICAST_WORKERS') or 1)
+        resty_env.value('APICAST_WORKERS') or Environment.default_config.worker_processes)
```
davidor: Should we call `tonumber` on the first part of the or? I guess it would work anyway, but I think it would be clearer.

gateway/src/apicast/cli/environment.lua
```diff
@@ -40,6 +41,16 @@ local function parse_nameservers()
     end
 end
 
+local function cpus()
+    local cores = tonumber(resty_env.get('NUMBER_OF_CORES'))
```
davidor: Do we need this? Is there any reason someone would use `NUMBER_OF_CORES` instead of the existing `APICAST_WORKERS`?

davidor: Also, if we start accepting this ENV, we should document it https://github.com/3scale/apicast/blob/master/doc/parameters.md

Just realized that `APICAST_WORKERS` is not there.

mikz: All software collection images use NUMBER_OF_CORES, so I thought it would make sense to follow that. 

But probably it does not make sense because the SCL images override `NUMBER_OF_CORES` with own value.

Will remove it.

gateway/src/apicast/cli/command/start.lua
mikz: But what if you want to set `APICAST_WORKERS=auto` ?

davidor: :+1:

davidor: I think it's not obvious that APICAST_WORKERS accepts 'auto'. People familiar with nginx might guess it, but I think we should document it.

mikz: `APICAST_WORKERS` is currently undocumented and I think it is pretty low level and stay like that for now. 

* in production we auto detect correct number of cores when deployed by our docker image
* development starts on 1 core

So people really should not need tuning the number of workers.

PR #556, opened 2018-01-18
[proxy] split proxy:call() into rewrite and access phase

mikz:
fixes https://github.com/3scale/apicast/issues/495

This allows policy authors to modify `usage` in the rewrite phase as it is being used later in the `access` phase. That can be used to for example report custom metrics matched by the policy.

### DIFF
--- a/gateway/src/apicast/proxy.lua
+++ b/gateway/src/apicast/proxy.lua
@@ -126,6 +126,8 @@ local function output_debug_headers(service, usage, credentials)
 end
 
 function _M:authorize(service, usage, credentials, ttl)
+  if not usage or not credentials then return nil, 'missing usage or credentials' end
+
   local encoded_usage = encode_args(usage)
   if encoded_usage == '' then
     return error_no_match(service)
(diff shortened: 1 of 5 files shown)


gateway/src/apicast/proxy.lua
```diff
@@ -206,36 +208,29 @@ function _M.set_upstream(service)
...
--- call the proxy and return a handler function
--- that will perform an action based on the path and backend version
--- @tparam service service service object
--- @treturn nil|function access function (when the request needs to be authenticated with this)
--- @treturn nil|function handler function (when the request is not authenticated and has some own action)
-function _M:call(service)
-  service = _M.set_service(service or ngx.ctx.service)
-
-  self.oauth = service:oauth()
+function _M:handle_oauth(service)
```
davidor: I think we should make this `local`?

mikz: Yeah, probably hood idea.

gateway/src/apicast/proxy.lua
```diff
@@ -206,36 +208,29 @@ function _M.set_upstream(service)
...
+  local oauth = service:oauth()
+  self.oauth = oauth
```
mikz: Maybe this assignment is not needed then.

davidor:
Looks good :+1:

gateway/src/apicast/proxy.lua
davidor: It is, unless we change something.
There's `self.oauth:transform_credentials` at the end of `rewrite()`.
We should check if it's possible to move that block of code to `handle_oauth`.

mikz: Yep. It was needed for transform_credentials. Now it returns `oauth` and is assigned in the `:rewrite`.

spec/proxy_spec.lua
```diff
@@ -61,7 +29,7 @@ describe('Proxy', function()
       service.credentials = { location = 'headers' }
       service.backend_version = 2
       ngx.var.http_app_key = 'key'
-      assert.falsy(proxy:access(service))
+      assert.falsy(proxy:rewrite(service))
```
davidor: kind of strange to find `rewrite()` in `describe(':access')`

gateway/src/apicast/proxy.lua
davidor: Better :+1:

PR #561, opened 2018-01-23
[wip] Add maintenance mode policy

alanmoran:
First attempt at a PR to apicast. Really basic policy just to introduce myself to Lua and writing some tests. 
Wasn't sure what was the best way to test my request didn't hit upstream. 

Any comments, tips or suggestions?

### DIFF
--- a/examples/policies/maintenance-mode/1.0.0/maintenance_mode.lua
+++ b/examples/policies/maintenance-mode/1.0.0/maintenance_mode.lua
@@ -1,0 +1,37 @@
+-- This is a simple policy. It allows you reject incoming requests with a
+-- specified status code and message.
+-- It's useful for maintance periods or to temporarily block an API
+
+local _M = require('apicast.policy').new('Maintenance mode', '1.0.0')
+
+local tonumber = tonumber
+local new = _M.new
+
+
+function _M.new(configuration)
+  local policy = new(configuration)
+
+  -- Set some default values
+  policy.status_code = 503
+  policy.message = "503 Service Unavailable - Maintenance"
+
+  if configuration then
+    policy.status_code = tonumber(configuration.status) or policy.status_code
+    policy.message = configuration.message or policy.message
+  end
+
+  return policy
+end
+
+function _M:access()
+
+  local status = self.status_code
+  local msg = self.message
+
+  ngx.status = status
+  ngx.say(msg)
+
+  return ngx.exit(ngx.status)
+end
+
+return _M
--- a/examples/policies/t/apicast-policy-maintenance-mode.t
+++ b/examples/policies/t/apicast-policy-maintenance-mode.t
@@ -1,0 +1,107 @@
+use lib 't';
+use Test::APIcast::Blackbox 'no_plan';
+
+repeat_each(1);
+
+env_to_apicast(
+    'APICAST_POLICY_LOAD_PATH' => "$ENV{PWD}/examples/policies"
+);
+
+run_tests();
+
+__DATA__
+
+=== TEST 1: Use maintenance mode using default values
+Testing 3 things:
+1) Check default status code
+2) Check default response message
+3) Validates upstream doesn't get the request
+
+--- configuration
+{
+  "services": [
+    {
+      "id": 42,
+      "backend_version":  1,
+      "backend_authentication_type": "service_token",
+      "backend_authentication_value": "token-value",
+      "proxy": {
+        "policy_chain": [
+          { "name": "maintenance-mode", "version": "1.0.0" },
+          { "name": "apicast.policy.apicast" }
+        ],
+        "api_backend": "http://test:$TEST_NGINX_SERVER_PORT/",
+        "proxy_rules": [
+          { "pattern": "/", "http_method": "GET", "metric_system_name": "hits", "delta": 2 }
+        ]
+      }
+    }
+  ]
+}
+--- upstream
+  location / {
+    content_by_lua_block {
+      local assert = require('luassert')
+      assert.is_nil(ngx.req.get_uri_args())
+      ngx.say('No response from me');
+    }
+  }
+--- request
+GET /?test
+--- response_body 
+503 Service Unavailable - Maintenance
+--- error_code: 503
+--- no_error_log
+[error]
+
+
+=== TEST 2: Use maintenance mode using custom values
+Testing 3 things:
+1) Check custom status code
+2) Check custom response message
+3) Validates upstream doesn't get request
+
+--- configuration
+{
+  "services": [{
+    "id": 42,
+    "backend_version": 1,
+    "backend_authentication_type": "service_token",
+    "backend_authentication_value": "token-value",
+    "proxy": {
+      "policy_chain": [{
+          "name": "maintenance-mode", "version": "1.0.0",
+          "configuration": {
+            "message": "Be back soon",
+            "status": 501
+          }
+        },
+        {
+          "name": "apicast.policy.apicast"
+        }
+      ],
+      "api_backend": "http://test:$TEST_NGINX_SERVER_PORT/",
+      "proxy_rules": [{
+        "pattern": "/",
+        "http_method": "GET",
+        "metric_system_name": "hits",
+        "delta": 2
+      }]
+    }
+  }]
+}
+--- upstream
+  location / {
+    content_by_lua_block {
+      local assert = require('luassert')
+      assert.is_nil(ngx.req.get_uri_args())
+      ngx.say('No response from me');
+    }
+  }
+--- request
+GET /?test
+--- response_body 
+Be back soon
+--- error_code: 501
+--- no_error_log
+[error]
--- a/examples/policies/maintenance-mode/1.0.0/README.md
+++ b/examples/policies/maintenance-mode/1.0.0/README.md
@@ -1,0 +1,23 @@
+# Maintenance Mode
+
+A policy which allows you reject incoming requests with a specified status code and message. It's useful for maintenance periods or to temporarily block an API.
+
+## Properties
+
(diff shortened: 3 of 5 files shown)


gateway/src/apicast/policy/maintenance_mode/README.md
```diff
@@ -0,0 +1,25 @@
+#Maintenance Mode
```
davidor: Github fails to render this correctly. You need to add a space after '#'.

gateway/src/apicast/policy/maintenance_mode/README.md
```diff
@@ -0,0 +1,25 @@
...
+The following is a list of possible properties and default values.
+
+Property name         | Default           | Description     |
+--------------------|------------------|-----------------------|
+status (integer, _optional_)				| 503   | Response code    |
+message (string, _optional_)      		| 503 Service Unavailable - Maintenance   |  Response message  |
+
+##Example Configuration
+```json
+"policy_chain": [{
```
davidor: I would wrap this between `{` and `}` so it becomes a valid json. Also, I think that indenting this would help readability.

gateway/src/apicast/policy/maintenance_mode/README.md
davidor: Same with all the other titles in this doc.

gateway/src/apicast/policy/maintenance_mode/policy.lua
```diff
@@ -0,0 +1,37 @@
...
+-- It's useful for maintance periods or to temporarily block an API
+
+local _M = require('apicast.policy').new('Maintenance mode')
+
+local tonumber = tonumber
+local new = _M.new
+
+
+function _M.new(configuration)
+  local policy = new(configuration)
```
davidor: There's no need to pass `configuration` as a param here.

gateway/src/apicast/policy/maintenance_mode/policy.lua
```diff
@@ -0,0 +1,37 @@
...
+    policy.status_code = tonumber(configuration.status) or policy.status_code
+    policy.message = configuration.message or policy.message
+  end
+
+  return policy
+end
+
+function _M:access()
+
+local status = self.status_code
```
davidor: These 2 lines are not properly indented.

gateway/src/apicast/policy/maintenance_mode/policy.lua
```diff
@@ -0,0 +1,37 @@
...
+local msg = self.message
+
+  ngx.status = status
+  ngx.say(msg)
+
+  return ngx.exit(ngx.status)
+end
+
+return _M
```
davidor: Github is complaining because there isn't a newline here.
Same with the rest of files included in this PR.

gateway/src/apicast/policy/maintenance_mode/policy.lua
davidor: This made me realize that we are doing the same in the echo policy. We'll need to fix that.

t/apicast-policy-maintenance-mode.t
```diff
@@ -0,0 +1,91 @@
...
+  "services": [
+    {
+      "id": 42,
+      "backend_version":  1,
+      "backend_authentication_type": "service_token",
+      "backend_authentication_value": "token-value",
+      "proxy": {
+        "policy_chain": [
+        { "name": "apicast.policy.maintenance_mode",
+          	"configuration": {
```
davidor: Check indentation here.

t/apicast-policy-maintenance-mode.t
```diff
@@ -0,0 +1,91 @@
...
+repeat_each(1);
+run_tests();
+
+__DATA__
+
+=== TEST 1: Use maintenance mode using default values
+Testing 3 things:
+1) Check default status code
+2) Check default response message
+3) TODO - validate upstream doesn't get the request
```
davidor: There are several options. You could print something in `upstream` and then check that it does not appear in the output, or you could put and assertion using `luassert`.

davidor:
Great job @alanmoran :+1: 

I added some comments about minor issues I found, but this is a good example of how to write and test a policy. I would be interested to hear if you found the documentation to be good enough and the problems you had while writing the policy.

I don't think we should include this in the `gateway/src/apicast/policy`, but rather in the `examples` folder.

gateway/src/apicast/policy/maintenance_mode/policy.lua
mikz: @davidor I think we might want to keep this as this is the signature of the initializer. That the underlying policy does not do anything with it is an implementation detail that can change in the future. What if the policy would actually store it in `self.configuration` ? We would have to change all policies to pass it again.

davidor: @mikz I see what you mean, but `new()` here is calling `function policy.new()` in `policy.lua` which does not accept any parameters. Maybe we should add it then to make it less confusing?

gateway/src/apicast/policy/maintenance_mode/policy.lua
alanmoran: Mmm is this still complaining? I'm fairly sure my git removes the last newline of a file. (I need to check this)

alanmoran:
@davidor Thanks for the review. I really appreciate it. I've made most of the changes you've recommended and put my feedback into @vramosp feedback doc. 

Agree on not adding it as one of the default OOTB polices. I'll wait until I hear back about where community policies might live. Could be a good first candidate.

gateway/src/apicast/policy/maintenance_mode/policy.lua
alanmoran: @davidor Ah damn. I actually followed the echo as my baseline. Are you both happy for me to keep this as the signature or change it for now?

gateway/src/apicast/policy/maintenance_mode/policy.lua
kevprice83: @alanmoran I think there *needs* to be a new line.

andrewdavidmackenzie:
consider adding as an example in a sub-folder for them?

gateway/src/apicast/policy/maintenance_mode/policy.lua
alanmoran: @kevprice83 I added it and its there locally but I think git or github removing it.

alanmoran:
@mikz I've rebased and moved policy to the examples folder. Updated structure to use init.lua also.

You should be able to test by updating `examples/policies/example.json`:

davidor:
@alanmoran good job :+1: 

Unfortunately, as you can see, the tests failed on CircleCI.

```
nginx: [error] init_by_lua error: module 'init' not found:
no file '/opt/app-root/src/gateway/policies/maintenance_mode/builtin/init.lua'
no file '/opt/app-root/src/gateway/src/apicast/policy/maintenance_mode/init.lua'
```

This is because by default, Apicast does not look for policies in the `examples` directory where your policy is. We need to configure Apicast to look in that directory. This can be done using the `APICAST_POLICY_LOAD_PATH` https://github.com/3scale/apicast/blob/master/doc/parameters.md#apicast_policy_load_path or the `--policy-load-path` option. In the tests, this is the way we pass ENVs:

```
env_to_apicast(
    'APICAST_POLICY_LOAD_PATH' => "$ENV{PWD}/examples/policies"
);
```

That needs to be specified before `run_tests()`. You can find several examples in our existing tests.

There's another problem. All the policies that are not the built-in ones, need to have a version in the configuration so Apicast knows how to load them. So you need something like this in the configuration:

```
{
    "name": "maintenance-mode", "version": "1.0.0"
}
```

Also, notice that you specified `maintenance-mode`, but the path is called `maintenance_mode`. They need to match.

alanmoran:
@davidor I've fixed up the PR. Biggest change really is that I moved the tests under `examples/policies/t`. I believe we need to do further work to get the CI testing those. 

In terms of `maintenance-mode` and `maintenance_mode`, I actually followed ngx-example but happy to change. Doesn't seem to break?

mikz:
@alanmoran after https://github.com/3scale/apicast/pull/632 is merged CI will run integration tests in `examples` folder too.

examples/policies/t/apicast-policy-maintenance-mode.t
```diff
@@ -0,0 +1,107 @@
...
+--- configuration
+{
+  "services": [{
+    "id": 42,
+    "backend_version": 1,
+    "backend_authentication_type": "service_token",
+    "backend_authentication_value": "token-value",
+    "proxy": {
+      "policy_chain": [{
+          "name": "maintenance-mode", "version": "1.0.0"
```
mikz: This is invalid JSON and the test fails. There is missing `,` at the and of the line.

examples/policies/maintenance-mode/1.0.0/maintenance_mode.lua
```diff
@@ -0,0 +1,37 @@
...
+
+  if configuration then
+    policy.status_code = tonumber(configuration.status) or policy.status_code
+    policy.message = configuration.message or policy.message
+  end
+
+  return policy
+end
+
+function _M:access()
```
mikz: Both tests fail because this is supposed to be `rewrite`.

APIcast policy rejects the requests in rewrite phase because they don't have authentication. 
So you either want to first require authentication and then show a maintenance page or show maintenance regardless of passed authentication parameters.

davidor:
@alanmoran There's a new requirement to include this policy as a builtin one. I used your commits in https://github.com/3scale/APIcast/pull/1105

PR #562, opened 2018-01-23
Add upstream policy

davidor:
This is a first version of the upstream policy. For now, it only allows to modify the host of a request based on a set of rules that act on its path, but we can add more functionality later.

### DIFF
--- a/gateway/src/apicast/policy/upstream/policy.lua
+++ b/gateway/src/apicast/policy/upstream/policy.lua
@@ -68,7 +68,10 @@ function _M:content()
 
   for _, rule in ipairs(self.rules) do
     if match(req_uri, rule.regex) then
+      ngx.log(ngx.DEBUG, 'upstream policy uri: ', req_uri, ' regex: ', rule.regex, ' match: true')
       return change_upstream(rule.url)
+    elseif ngx.config.debug then
+      ngx.log(ngx.DEBUG, 'upstream policy uri: ', req_uri, ' regex: ', rule.regex, ' match: false')
     end
   end
 end
--- a/gateway/src/apicast/policy/upstream/policy.lua
+++ b/gateway/src/apicast/policy/upstream/policy.lua
@@ -1,0 +1,76 @@
+--- Upstream policy
+-- This policy allows to modify the host of a request based on its path.
+
+local resty_resolver = require('resty.resolver')
+local resty_url = require('resty.url')
+local format = string.format
+local ipairs = ipairs
+local match = ngx.re.match
+local table = table
+
+local _M = require('apicast.policy').new('Upstream policy')
+
+local new = _M.new
+
+local function proxy_pass(url)
+  local res = format('%s://upstream%s', url.scheme, url.path or '')
+
+  local args = ngx.var.args
+  if url.path and args then
+    res = format('%s?%s', res, args)
+  end
+
+  return res
+end
+
+local function change_upstream(url)
+  ngx.ctx.upstream = resty_resolver:instance():get_servers(
+    url.host, { port = url.port })
+
+  ngx.var.proxy_pass = proxy_pass(url)
+  ngx.req.set_header('Host', url.host)
+
+  -- We need to check that the headers have not already been sent. They could
+  -- have been sent in a different policy, for example.
+  if not ngx.headers_sent then
+    ngx.exec("@upstream")
+  end
+end
+
+-- Parses the urls in the config so we do not have to do it on each request.
+local function init_config(config)
+  if not config or not config.rules then return {} end
+
+  local res = {}
+
+  for _, rule in ipairs(config.rules) do
+    local parsed_url = resty_url.parse(rule.url)
+    table.insert(res, { regex = rule.regex, url = parsed_url })
+  end
+
+  return res
+end
+
+--- Initialize an upstream policy.
+-- @tparam[opt] table config Contains the host rewriting rules.
+-- Each rule consists of:
+--
+--   - regex: regular expression to be matched.
+--   - url: new url in case of match.
+function _M.new(config)
+  local self = new(config)
+  self.rules = init_config(config)
+  return self
+end
+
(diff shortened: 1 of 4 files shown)


gateway/src/apicast/policy/upstream/policy.lua
```diff
@@ -0,0 +1,65 @@
...
+local tonumber = tonumber
+local match = ngx.re.match
+local table = table
+
+local _M = require('apicast.policy').new('Upstream policy')
+
+local new = _M.new
+
+local function change_upstream(url)
+  ngx.ctx.upstream = resty_resolver:instance():get_servers(
```
davidor: This code is very similar to what we already have in `Proxy.get_upstream()` https://github.com/3scale/apicast/blob/6049ab33581483b0e69fc1ff6ab2fea38aa4a696/gateway/src/apicast/proxy.lua#L180

If we think that there might be other policies that need the same functionality, we should refactor what we have there to avoid duplicating code in the policies.

t/apicast-policy-upstream.t
```diff
@@ -0,0 +1,125 @@
+use lib 't';
+use Test::APIcast::Blackbox 'no_plan';
+
+run_tests();
+
+__DATA__
+
```
davidor: @mikz  I also wanted to check a case with a single rule that does not match. However, I don't know what's the best way to test that. The request would try to go to an invalid upstream. Not sure what would be the best way to mock it.

gateway/src/apicast/policy/upstream/policy.lua
```diff
@@ -0,0 +1,65 @@
...
+    url.host, { port = tonumber(url.port) })
```
davidor: I need to check if `tonumber()` here is needed or if the `resty_url.parse()` call guarantees that we'll get a number here.

davidor: It does : https://github.com/3scale/lua-resty-url/blob/6fe0896c6f3688d6042a0f28d32ba577cb250515/src/resty/url.lua#L63

Changed.

t/apicast-policy-upstream.t
```diff
@@ -0,0 +1,186 @@
...
+          { "name": "apicast.policy.apicast" }
+        ]
+      }
+    }
+  ]
+}
+--- upstream
+  location / {
+     content_by_lua_block {
+       ngx.say('yay, api backend');
```
mikz: Maybe we could print here the `ngx.var.uri` to verify what was received?

t/apicast-policy-upstream.t
```diff
@@ -0,0 +1,186 @@
...
+          { "pattern": "/", "http_method": "GET", "metric_system_name": "hits", "delta": 2 }
+        ],
+        "policy_chain": [
+          {
+            "name": "apicast.policy.upstream",
+            "configuration":
+              {
+                "rules": [
+                  { "regex": "/i_dont_match", "url": "http://example.com" },
+                  { "regex": "/", "url": "http://test:$TEST_NGINX_SERVER_PORT" },
```
mikz: I think we should test a case when the url has a path to see what is forwarded upstream.

davidor: Agreed. Added in new commit.

gateway/src/apicast/policy/upstream/policy.lua
davidor: I'd rather address this on a separate PR as this is an internal change that will not affect the schema of the policy.

t/apicast-policy-upstream.t
davidor: The path in the location already checks schema, host, and port.
For the path we'll need to use a location different from `/` to be sure. I think that would be enough.
However, we could also use luassert inside the `content_by_lua` block to make these checks more explicit.
What do you think?

gateway/src/apicast/policy/upstream/policy.lua
mikz: Totally 👍

t/apicast-policy-upstream.t
```diff
@@ -0,0 +1,246 @@
...
+              }
+          },
+          { "name": "apicast.policy.apicast" }
+        ]
+      }
+    }
+  ]
+}
+--- upstream
+  location /a_path {
```
mikz: This can match `/a_path/whatever`. Would be good to at least define it as `location = /a_path`. But I'd also go the extra step and print `ngx.var.uri` in the response. So we can for example verify the args.

davidor: I added a luassert test that checks the value of `ngx.var.request`. It turns out that args are not sent when the url of the rule contains a path.

mikz: 👍 I knew this was going to be a bit tricky to send everything correctly.

gateway/src/apicast/policy/upstream/policy.lua
```diff
@@ -0,0 +1,70 @@
...
+local _M = require('apicast.policy').new('Upstream policy')
+
+local new = _M.new
+
+local function change_upstream(url)
+  ngx.ctx.upstream = resty_resolver:instance():get_servers(
+    url.host, { port = url.port })
+
+  if url.path then
+    ngx.var.proxy_pass = format('%s://upstream%s?%s', url.scheme, url.path, ngx.var.args)
```
mikz: This is going to always append `?`, right? Even though there are no args.

davidor: I haved extracted this part of the code to its own method to make it clearer.

t/apicast-policy-upstream.t
davidor: I solved this, but it looks like I have a similar problem with POST bodies.

davidor:
@mikz This is ready. I added asserts in the `location` blocks to make sure that we are receiving the request as we expect it.

PR #564, opened 2018-01-25
[wip] policy manifest and it's JSON schema

mikz:


gateway/src/apicast/policy/caching/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "Caching 3scale backend response",
+  "description": [
+    "Configures a cache for the authentication calls against the 3scale backend.",
+    "The 3scale backend can authorize (status code = 200) and deny (status code = 4xx) calls.",
+    "When it fails, it returns a 5xx code.."
+  ],
+  "version": "1.1",
+  "configuration": {
+    "$ref": "schema.json#"
```
mikz: @davidor I'd like to hear what is your though on this. 

This would allow us to have schema and manifest in different JSON files, but we would need to use some tool (like [json-refs](https://github.com/whitlockjc/json-refs)) to merge the doc. It is not a big deal, just pass  it through, I verified it works and creates correct output.

Or we can have the schema embedded in the manifest and have just one, bigger JSON.

gateway/src/apicast/policy/manifest-schema.json
```diff
@@ -0,0 +1,58 @@
...
+      ],
+      "title": "The Policy Description",
+      "description": "Longer description of what the policy does.",
+      "examples": [
+        "Extract authentication credentials from the HTTP Authorization header and pass them to 3scale backend.",
+        [ "Redirect request to different upstream: ", " - based on path", "- set different Host header"]
+      ]
+    },
+    "version": {
+      "$ref": "#/definitions/version"
```
davidor: Why did you decided to extract this to its own definition and not do the same with `description` or `name` ?

mikz: I wanted to try it out and I was expecting this might appear on multiple places, so defining a primitive for other validations.

mikz:
Closing in favour of #565

PR #565, opened 2018-01-25
Add policy manifest and make existing schemas conform with it

davidor:
Adds the policy manifest from @mikz in #564 and makes all the current schemas conform to it.
This PR also adds descriptions to all the configuration fields.

/cc @alanmoran


gateway/src/apicast/policy/caching/schema.json
```diff
@@ -1,9 +1,28 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "title": "Caching policy configuration",
+  "description":
```
davidor: @mikz  This is the best way I found to write long descriptions. We'll need to specify in the meta-schema that the descriptions might be strings or arrays of strings.

mikz:
@davidor we should define where to have those descriptions. I'm not sure the schema is the right one, because we also need some manifest that defines a version etc. as started in https://github.com/3scale/apicast/pull/564

gateway/src/apicast/policy/caching/schema.json
mikz: Yep, that is done in https://github.com/3scale/apicast/pull/564, but the description is not part of the schema there.

gateway/src/apicast/policy/caching/apicast-policy.json
```diff
@@ -0,0 +1,47 @@
...
+     "down.\n",
+     "- Allow: caches authorized and denied calls. When backend is ",
+     "unavailable, it will cache an authorization. In practice, this means ",
+     "that when backend is down _any_ request will be authorized unless last ",
+     "call to backend for that request returned 'deny' (status code = 4xx). ",
+     "Make sure to understand the implications of that before using this ",
+     "mode. It makes sense only in very specific use cases.\n",
+     "- None: disables caching."],
+  "version": "0.1",
+  "configuration": {
```
mikz: We might need to force `$schema` key in this object, so it can be easier for other libraries to use this and know what schema it is.

ddcesare: No need to force the `$schema` key, the lib figure this out.

davidor: I realized that the `lint-schema` target is not working correctly. It's not checking that `configuration`complies with `http://json-schema.org/draft-07/schema#"`.
Adding `$schema` in the configuration object does not solve the issue.

mikz: I was using https://www.jsonschemavalidator.net for validation and it was ok.
But now it fails on missing `$id` in the `definitions/id` and `definitions/schema`. 
Just adding there `"$id": "#/definitions/schema",` and `"$id":  "#/definitions/version",` makes it validate in that online editor.

What other field was not validated correctly? I tried to change the "type" column from "string" to some invalid value and that is caught.

davidor: @mikz When I run `make lint-schema`, the error you mention is not detected. Not sure why.
I added those ids in your commit.

Another thing I noticed is that, if I remove the "type" attr from `caching_type` in the caching policy config, https://mozilla-services.github.io/react-jsonschema-form/ complains, but `make lint-schema` does not.

PR #566, opened 2018-01-26
Sandbox policy loading

mikz:
Depends on https://github.com/3scale/apicast/pull/578

Policies will be able to load only own code (in their subpath).
Loading the policy twice results in different policy. 
Different policies (and versions) can have different dependencies.

Very light sandbox is not meant for security, but for ease of use and preventing mistakes. Policies should not be able to access the global environment (except `ngx` variable).

Spec: https://gist.github.com/mikz/ceb8f141de2e8f74206761abda231f70

# TODO (in future PRs)

- [ ] change all `policy.lua` to `${policy_name}.lua` and create `init.lua` for each policy (according to the spec) (https://github.com/3scale/apicast/pull/579)

### DIFF
--- a/gateway/src/apicast/policy_chain.lua
+++ b/gateway/src/apicast/policy_chain.lua
@@ -11,12 +11,15 @@ local rawset = rawset
 local type = type
 local require = require
 local insert = table.insert
+local sub = string.sub
+local format = string.format
 local noop = function() end
 
 require('apicast.loader')
 
 local linked_list = require('apicast.linked_list')
 local policy_phases = require('apicast.policy').phases
+local policy_loader = require('apicast.policy_loader')
 
 local _M = {
 
@@ -46,7 +49,7 @@ function _M.build(modules)
 
     for i=1, #list do
         -- TODO: make this error better, possibly not crash and just log and skip the module
-        chain[i] = _M.load(list[i]) or error("module " .. list[i] .. ' could not be loaded')
+        chain[i] = _M.load_policy(list[i]) or error(format('module %q could not be loaded', list[i]))
     end
 
     return _M.new(chain)
@@ -72,10 +75,14 @@ end
 -- @tparam string|table module the module or its name
 -- @tparam ?table ... params needed to initialize the module
 -- @treturn object The module instantiated
-function _M.load(module, ...)
+function _M.load_policy(module, version, ...)
     if type(module) == 'string' then
-        ngx.log(ngx.DEBUG, 'loading policy module: ', module)
-        local mod = require(module)
+        if sub(module, 1, 14) == 'apicast.policy' then
+            module = sub(module, 16)
+            version = 'builtin'
+        end
+
+        local mod = policy_loader(module, version or 'builtin')
 
         if mod then
             return mod.new(...)
--- a/gateway/src/apicast/policy_loader.lua
+++ b/gateway/src/apicast/policy_loader.lua
@@ -1,0 +1,280 @@
+--- Policy loader
+-- This module loads a policy defined by its name and version.
+-- It uses sandboxed require to isolate dependencies and not mutate global state.
+-- That allows for loading several versions of the same policy with different dependencies.
+-- And even loading several independent copies of the same policy with no shared state.
+-- Each object returned by the loader is new table and shares only shared APIcast code.
+
+local format = string.format
+local error = error
+local type = type
+local loadfile = loadfile
+local getenv = os.getenv
+local insert = table.insert
+local setmetatable = setmetatable
+local concat = table.concat
+
+local _G = _G
+local _M = {}
+
+local searchpath = package.searchpath
+local root_loaded = package.loaded
+
+local root_require = require
+
+local preload = package.preload
+
+--- create a require function not using the global namespace
+-- loading code from policy namespace should have no effect on the global namespace
+-- but poliocy can load shared libraries that would be cached globally
+local function gen_require(package)
+
+  local function not_found(modname, err)
+    return error(format("module '%s' not found:%s", modname, err), 0)
+  end
+
+  --- helper function to safely use the native require function
+  local function fallback(modname)
+    local mod
+
+    mod = package.loaded[modname]
+
+    if not mod then
+      ngx.log(ngx.DEBUG, 'native require for: ', modname)
+      mod = root_require(modname)
+    end
+
+    return mod
+  end
+
+  --- helper function to find and return correct loader for a module
+  local function find_loader(modname)
+    local loader, file, err, ret
+
+    -- http://www.lua.org/manual/5.2/manual.html#pdf-package.searchers
+
+    -- When looking for a module, require calls each of these searchers in ascending order,
+    for i=1, #package.searchers do
+      -- with the module name (the argument given to require) as its sole parameter.
+      ret, err = package.searchers[i](modname)
+
+      -- The function can return another function (the module loader)
+      -- plus an extra value that will be passed to that loader,
+      if type(ret) == 'function' then
+        loader = ret
+        file = err
+        break
+      -- or a string explaining why it did not find that module
+      elseif type(ret) == 'string' then
+        err = ret
+      end
+      -- (or nil if it has nothing to say).
+    end
+
+    return loader, file, err
+  end
+
+  --- reimplemented require function
+  -- - return a module if it was already loaded (globally or locally)
+  -- - try to find loader function
+  -- - fallback to global require
+  -- @tparam string modname module name
+  -- @tparam boolean exclusive load only policy code, turns off the fallback loader
+  return function(modname, exclusive)
+    -- http://www.lua.org/manual/5.2/manual.html#pdf-require
+    ngx.log(ngx.DEBUG, 'sandbox require: ', modname)
+
+    -- The function starts by looking into the package.loaded table
+    -- to determine whether modname is already loaded.
+    -- NOTE: this is different from the spec: use the top level package.loaded,
+    --       otherwise it would try to sandbox load already loaded shared code
+    local mod = root_loaded[modname]
+
+    --  If it is, then require returns the value stored at package.loaded[modname].
+    if mod then return mod end
+
+    -- Otherwise, it tries to find a loader for the module.
+    local loader, file, err = find_loader(modname)
+
+    -- Once a loader is found,
+    if loader then
+      ngx.log(ngx.DEBUG, 'sandboxed require for: ', modname, ' file: ', file)
+      -- require calls the loader with two arguments:
+      --   modname and an extra value dependent on how it got the loader.
+      -- (If the loader came from a file, this extra value is the file name.)
+      mod = loader(modname, file)
+    elseif not exclusive then
+      ngx.log(ngx.DEBUG, 'fallback loader for: ', modname, ' error: ', err)
+      mod = fallback(modname)
+    else
+      -- If there is any error loading or running the module,
+      -- or if it cannot find any loader for the module, then require raises an error.
+      return not_found(modname, err)
+    end
+
+    -- If the loader returns any non-nil value,
+    if mod ~= nil then
+      -- require assigns the returned value to package.loaded[modname].
+      package.loaded[modname] = mod
+
+      -- If the loader does not return a non-nil value
+      -- and has not assigned any value to package.loaded[modname],
+    elseif not package.loaded[modname] then
+      -- then require assigns true to this entry.
+      package.loaded[modname] = true
+    end
+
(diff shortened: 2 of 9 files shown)


gateway/src/apicast/policy_loader.lua
```diff
@@ -0,0 +1,198 @@
...
+  --- helper function to safely use the native require function
+  local function fallback(modname, err)
+    local mod
+
+    mod = package.loaded[modname]
+
+    if not mod then
+      local ok
+      ngx.log(ngx.DEBUG, 'native require for: ', modname)
+      ok, mod = prequire(modname)
```
mikz: This might not be needed if we want to raise on the next line anyway.

gateway/src/apicast/policy_loader.lua
```diff
@@ -0,0 +1,198 @@
...
+  end
+
+  --- helper function to find and return correct loader for a module
+  local function find_loader(modname)
+    local loader, file, err, ret
+
+    for i=1, #package.searchers do
+      ret, err = package.searchers[i](modname)
+
+      if type(ret) == 'function' then
```
mikz: Would be good to paste parts of the Lua documentation describing what loaders can return.

gateway/src/apicast/policy_loader.lua
```diff
@@ -0,0 +1,198 @@
...
+
+local searchpath = package.searchpath
+local root_loaded = package.loaded
+
+local root_require = require
+local pcall = pcall
+
+local preload = package.preload
+
+preload['apicast.policy_loader'] = function() return _M end
```
mikz: This might not be needed anymore.

spec/policy_loader_spec.lua
```diff
@@ -0,0 +1,24 @@
+local _M = require 'apicast.policy_loader'
+
+describe('APIcast Policy Loader', function()
+
+  describe('.call', function()
```
davidor: I think we need tests that show that the code works correctly when it loads two different versions of the same policy. I guess that testing this it's a bit difficult, because we'll need, at least, to create some test policies in the `fixtures` folder.

Maybe worth taking a look whether injecting `apicast_dir` would make that test easier or clearer.

gateway/src/apicast/policy_chain.lua
```diff
@@ -72,10 +75,14 @@ end
...
 -- @tparam ?table ... params needed to initialize the module
 -- @treturn object The module instantiated
-function _M.load(module, ...)
+function _M.load_policy(module, version, ...)
     if type(module) == 'string' then
-        ngx.log(ngx.DEBUG, 'loading policy module: ', module)
-        local mod = require(module)
+        if sub(module, 1, 14) == 'apicast.policy' then
+            module = sub(module, 16)
+            version = 'builtin'
```
davidor: This implies that we will not be offering several versions of the same policy in a given version of Apicast. Not saying I don't agree with that, just mentioning it so we're aware of this important decision.

gateway/cpanfile
```diff
@@ -1,2 +1,2 @@
-requires 'Test::APIcast', '0.03';
+requires 'Test::APIcast', '0.05';
```
davidor: I see that you released `0.06` some days ago. Any reason not to use that one?

mikz: No reason. 0.06 was released after I changed this. Will update 👍

gateway/src/apicast/policy_chain.lua
mikz: Yep, no two versions of builtin policies. For now. This is quite easy change to make if we want to. 

IMO it would be quite difficult to maintain > 1 versions of builtin policies, so I think it is not worth the effort. Internals of this method could be changed if we change that in the future.

spec/executor_spec.lua
```diff
@@ -1,42 +1,48 @@
...
         stub(policy, phase)
       end
     end
 
     -- For each one of the nginx phases, verify that when called on the
     -- executor, each one of the policies executes the code for that phase.
-    for _, phase in ipairs(phases) do
-      executor[phase](executor)
-      for _, policy in ipairs(default_executor_chain) do
+    for _, phase in Policy.phases() do
```
davidor: I was going to say that the test would be incomplete after this change, but I see that in the specs of policy we're already testing that `Policy.phases()` returns the correct phases. Nice cleanup :+1:

mikz: Actually the test was incomplete before this change :) Because It didn't had `content`. I think there might be one more place where it has own list.

spec/policy_loader_spec.lua
```diff
@@ -0,0 +1,24 @@
...
+    it('finds apicast builtin policy', function()
+      local root = require('apicast.policy.apicast')
+      local sandbox = _M:call('apicast')
+
+      assert.is_table(sandbox)
+      assert.are_not.same(root, sandbox)
+      assert.equal(root._NAME, sandbox._NAME)
+      assert.equal(root._VERSION, sandbox._VERSION)
```
davidor: These `asserts` are good, but how can we know that this is really the Apicast policy :eyes: ? Should we at least check that it contains some of its more important methods? The phases for example. I feel like we need something more than testing that it's just a table with the correct name and version. Ideas?

gateway/src/apicast/policy_loader.lua
```diff
@@ -0,0 +1,199 @@
+--- Policy loader
+-- This module loads a policy defined by its name and version.
+-- It uses sandboxed require to isolate dependencies and not mutate global state.
```
davidor: Maybe emphasize somehow that this loader allows us to load several versions of the same policy without problems?

spec/policy_loader_spec.lua
mikz: The issue is that if you load the same policy twice by the policy loader it is actually different table with different functions. The file is evaluated and returned twice.  And it is always going to be different than the one returned by require.
Each policy has all the phases, so we can't really compare that.

I tried adding there some random id, but it is different every time you load the policy. I really have no nice way of comparing them.

gateway/src/apicast/policy_loader.lua
mikz: Yep. It even allows loading the same policy several times without global state. Will emphasize that.

gateway/src/apicast/policy_loader.lua
```diff
@@ -0,0 +1,199 @@
...
+      ngx.log(ngx.DEBUG, 'fallback loader for: ', modname, ' error: ', err)
+      mod = fallback(modname)
+    else
+      return not_found(modname, err)
+    end
+
+    if mod ~= nil then
+      package.loaded[modname] = mod
+    else
+      package.loaded[modname] = true
```
davidor: Why this assignment?

gateway/src/apicast/policy_loader.lua
```diff
@@ -0,0 +1,199 @@
...
+    end
+
+    return package.loaded[modname]
+  end
+end
+
+--- this is environment exposed to the policies
+-- that means this is very light sandbox so policies don't mutate global env
+-- and most importantly we replace the require function with our own
+_M.env = {
```
davidor: Just curious, how did you decide what to include and what not to include here ?

gateway/src/apicast/policy_loader.lua
mikz: Will add there the comment from the docs 👍

gateway/src/apicast/policy_loader.lua
mikz: I want to include everything that is in standard Lua env. And it was just trial & error. So I guess something might be missing still.
I could go through every key in the normal env and make sure it is here.

mikz: We probably don't want to expose code loading functions that could escape the sandbox, right?
The sandbox is meant to be just light shell, not a thorough security measure, but still.

Here is global from repl (without package):
```lua
<1>{
  _ = <table 1>,
  _G = <table 1>,
  _VERSION = "Lua 5.1",
  assert = <function 1>,
  bit = {
    arshift = <function 2>,
    band = <function 3>,
    bnot = <function 4>,
    bor = <function 5>,
    bswap = <function 6>,
    bxor = <function 7>,
    lshift = <function 8>,
    rol = <function 9>,
    ror = <function 10>,
    rshift = <function 11>,
    tobit = <function 12>,
    tohex = <function 13>
  },
  collectgarbage = <function 14>,
  coroutine = {
    create = <function 15>,
    isyieldable = <function 16>,
    resume = <function 17>,
    running = <function 18>,
    status = <function 19>,
    wrap = <function 20>,
    yield = <function 21>
  },
  debug = {
    debug = <function 22>,
    getfenv = <function 23>,
    gethook = <function 24>,
    getinfo = <function 25>,
    getlocal = <function 26>,
    getmetatable = <function 27>,
    getregistry = <function 28>,
    getupvalue = <function 29>,
    getuservalue = <function 30>,
    setfenv = <function 31>,
    sethook = <function 32>,
    setlocal = <function 33>,
    setmetatable = <function 34>,
    setupvalue = <function 35>,
    setuservalue = <function 36>,
    traceback = <function 37>,
    upvalueid = <function 38>,
    upvaluejoin = <function 39>
  },
  dofile = <function 40>,
  error = <function 41>,
  gcinfo = <function 42>,
  getfenv = <function 43>,
  getmetatable = <function 44>,
  io = {
    close = <function 45>,
    flush = <function 46>,
    input = <function 47>,
    lines = <function 48>,
    open = <function 49>,
    output = <function 50>,
    popen = <function 51>,
    read = <function 52>,
    stderr = <userdata 1>,
    stdin = <userdata 2>,
    stdout = <userdata 3>,
    tmpfile = <function 53>,
    type = <function 54>,
    write = <function 55>
  },
  ipairs = <function 56>,
  jit = {
    arch = "x64",
    attach = <function 57>,
    flush = <function 58>,
    off = <function 59>,
    on = <function 60>,
    opt = {
      start = <function 61>
    },
    os = "OSX",
    status = <function 62>,
    version = "LuaJIT 2.1.0-beta3",
    version_num = 20100
  },
  load = <function 63>,
  loadfile = <function 64>,
  loadstring = <function 65>,
  math = {
    abs = <function 66>,
    acos = <function 67>,
    asin = <function 68>,
    atan = <function 69>,
    atan2 = <function 70>,
    ceil = <function 71>,
    cos = <function 72>,
    cosh = <function 73>,
    deg = <function 74>,
    exp = <function 75>,
    floor = <function 76>,
    fmod = <function 77>,
    frexp = <function 78>,
    huge = inf,
    ldexp = <function 79>,
    log = <function 80>,
    log10 = <function 81>,
    max = <function 82>,
    min = <function 83>,
    modf = <function 84>,
    pi = 3.1415926535898,
    pow = <function 85>,
    rad = <function 86>,
    random = <function 87>,
    randomseed = <function 88>,
    sin = <function 89>,
    sinh = <function 90>,
    sqrt = <function 91>,
    tan = <function 92>,
    tanh = <function 93>
  },
  module = <function 94>,
  newproxy = <function 95>,
  next = <function 96>,
  os = {
    clock = <function 97>,
    date = <function 98>,
    difftime = <function 99>,
    execute = <function 100>,
    exit = <function 101>,
    getenv = <function 102>,
    remove = <function 103>,
    rename = <function 104>,
    setlocale = <function 105>,
    time = <function 106>,
    tmpname = <function 107>
  },
  pairs = <function 108>,
  pcall = <function 109>,
  print = <function 110>,
  rawequal = <function 111>,
  rawget = <function 112>,
  rawlen = <function 113>,
  rawset = <function 114>,
  require = <function 115>,
  select = <function 116>,
  setfenv = <function 117>,
  setmetatable = <function 118>,
  string = {
    byte = <function 119>,
    char = <function 120>,
    dump = <function 121>,
    find = <function 122>,
    format = <function 123>,
    gmatch = <function 124>,
    gsub = <function 125>,
    len = <function 126>,
    lower = <function 127>,
    match = <function 128>,
    rep = <function 129>,
    reverse = <function 130>,
    sub = <function 131>,
    upper = <function 132>
  },
  table = {
    concat = <function 133>,
    foreach = <function 134>,
    foreachi = <function 135>,
    getn = <function 136>,
    insert = <function 137>,
    maxn = <function 138>,
    move = <function 139>,
    pack = <function 140>,
    remove = <function 141>,
    sort = <function 142>,
    unpack = <function 143>
  },
  tonumber = <function 144>,
  tostring = <function 145>,
  type = <function 146>,
  unpack = <function 143>,
  xpcall = <function 147>
}
```

mikz: Made a new sandbox in https://github.com/3scale/apicast/pull/566/commits/6884ae0a6c78e024fb890e2fc245b2d8da10f2b9

PR #567, opened 2018-01-30
SOAP policy

davidor:
This policy adds support for a very small subset of SOAP.

This policy basically expects a SOAPAction URI in the SOAPAction header or the content-type header.
The SOAPAction header is used in v1.1: https://www.w3.org/TR/2000/NOTE-SOAP-20000508/#_Toc478383528, whereas the Content-Type header is used in v1.2: https://www.w3.org/TR/soap12-part2/#ActionFeature

The SOAPAction URI is matched against the mapping rules defined in the configuration of the policy and the usage resulting from that is authorized and reported against 3scale's backend.

### DIFF
--- a/gateway/src/apicast/policy/soap/init.lua
+++ b/gateway/src/apicast/policy/soap/init.lua
@@ -1,0 +1,1 @@
+return require('soap')
--- a/gateway/src/apicast/policy/soap/soap.lua
+++ b/gateway/src/apicast/policy/soap/soap.lua
@@ -1,0 +1,127 @@
+--- SOAP Policy
+-- This policy adds support for a very small subset of SOAP.
+-- This policy basically expects a SOAPAction URI in the SOAPAction header or
+-- the content-type header.
+-- The SOAPAction header is used in v1.1 of the SOAP standard:
+-- https://www.w3.org/TR/2000/NOTE-SOAP-20000508/#_Toc478383528, whereas the
+-- Content-Type header is used in v1.2 of the SOAP standard:
+-- https://www.w3.org/TR/soap12-part2/#ActionFeature
+-- The SOAPAction URI is matched against the mapping rules defined in the
+-- policy and calculates a usage based on that so it can be authorized and
+-- reported against 3scale's backend.
+
+local sub = string.sub
+local len = string.len
+local ipairs = ipairs
+local insert = table.insert
+local re = require('ngx.re')
+
+local MappingRule = require('apicast.mapping_rule')
+local Usage = require('apicast.usage')
+local mapping_rules_matcher = require('apicast.mapping_rules_matcher')
+
+local policy = require('apicast.policy')
+
+local _M = policy.new('SOAP policy')
+
+local soap_action_header = 'SOAPAction'
+local soap_action_ctype = 'application/soap+xml;'
+
+local new = _M.new
+
+local function starts_with(str, start)
+  return sub(str, 1, len(start)) == start
+end
+
+-- Extracts a SOAP action from the SOAPAction header. Returns nil when not
+-- present.
+local function soap_action_in_header(headers)
+  return headers[soap_action_header]
+end
+
+-- Extracts the SOAP action from a string that contains the parameters of a
+-- Content-Type header. The string has this format:
+-- a_param=x;action=soap_action;another_param=y
+-- This method returns the value of 'action' or nil when it's not present.
+local function soap_action_from_ctype_params(params)
+  local params_split = re.split(params, ";")
+
+  for _, param in ipairs(params_split) do
+    local header_param_split = re.split(param, "=")
+    local name = header_param_split[1]
+    local value = header_param_split[2]
+    if name == "action" then
+      return value
+    end
+  end
+
+  return nil
+end
+
+-- Extracts a SOAP action from the Content-Type header. In SOAP, the
+-- type/subtype is application/soap+xml, and the action is specified as a
+-- param in that header. When there is no SOAP action, this method returns nil.
+local function soap_action_in_ctype(headers)
+  local ctype = headers['Content-Type']
+
+  if ctype and starts_with(ctype, soap_action_ctype) then
+    local header_params = sub(ctype, len(soap_action_ctype) + 1, -1)
+    return soap_action_from_ctype_params(header_params)
+  else
+    return nil
+  end
+end
+
+-- Extracts a SOAP action URI from the SOAP Action and the Content-Type
+-- headers. When both contain a SOAP action, the Content-Type one takes
+-- precedence.
+local function extract_soap_uri()
+  local headers = ngx.req.get_headers() or {}
+  return soap_action_in_ctype(headers) or soap_action_in_header(headers)
+end
+
+local function usage_from_matching_rules(soap_action_uri, rules)
+  return mapping_rules_matcher.get_usage_from_matches(
+    nil, soap_action_uri, {}, rules)
+end
+
+local function mapping_rules_from_config(config)
+  if not (config and config.mapping_rules) then return {} end
+
+  local res = {}
+
+  for _, config_rule in ipairs(config.mapping_rules) do
+    local rule = MappingRule.from_proxy_rule(config_rule)
+    insert(res, rule)
+  end
+
+  return res
+end
+
+--- Initialize a SOAP policy
+-- @tparam[opt] table config Configuration
+function _M.new(config)
+  local self = new(config)
+  self.mapping_rules = mapping_rules_from_config(config)
+  return self
+end
+
+--- Rewrite phase
+-- When a SOAP Action is received via the SOAPAction or the Content-Type
+-- headers, the policy matches it against the mapping rules defined in the
+-- configuration of the policy and calculates the associated usage.
+-- This usage is merged with the one received in the shared context.
+-- @tparam table context Shared context between policies
+function _M:rewrite(context)
+  local soap_action_uri = extract_soap_uri()
+
+  if soap_action_uri then
+    local soap_usage = usage_from_matching_rules(
+      soap_action_uri, self.mapping_rules)
+
+    context.usage = context.usage or Usage.new()
+    context.usage:merge(soap_usage)
+  end
+end
+
+return _M
--- a/gateway/src/apicast/policy/soap/soap.lua
+++ b/gateway/src/apicast/policy/soap/soap.lua
@@ -12,9 +12,9 @@
 
 local sub = string.sub
 local len = string.len
+local lower = string.lower
 local ipairs = ipairs
 local insert = table.insert
-local re = require('ngx.re')
 
 local MappingRule = require('apicast.mapping_rule')
 local Usage = require('apicast.usage')
@@ -39,23 +39,30 @@ local function soap_action_in_header(headers)
   return headers[soap_action_header]
 end
 
+local regex_del_leading_spaces = [[^\s*]]
+local regex_del_spaces_around_semicolon = [[\s*;\s*]]
+
+-- There can be spaces in the Content-Type, values can be wrapped with '"',
+-- etc.
+-- See: https://tools.ietf.org/html/rfc7231#section-3.1.1.1
+local regex_action_from_ctype = [[action=(?:"(.+)"|([^;"]+))\s*(?:;|$)]]
+
 -- Extracts the SOAP action from a string that contains the parameters of a
 -- Content-Type header. The string has this format:
 -- a_param=x;action=soap_action;another_param=y
 -- This method returns the value of 'action' or nil when it's not present.
-local function soap_action_from_ctype_params(params)
-  local params_split = re.split(params, ";")
+local function soap_action_from_ctype_params(ctype_params)
+  local params = ngx.re.sub(
+    ctype_params, regex_del_leading_spaces, '', 'oj')
 
-  for _, param in ipairs(params_split) do
-    local header_param_split = re.split(param, "=")
-    local name = header_param_split[1]
-    local value = header_param_split[2]
-    if name == "action" then
-      return value
-    end
-  end
+  local params_without_blanks = ngx.re.gsub(
+    params, regex_del_spaces_around_semicolon, ';', 'oj')
 
-  return nil
+  local matches = ngx.re.match(
+    lower(params_without_blanks), regex_action_from_ctype, 'oj')
+
+  if not matches then return nil end
+  return matches[1] or matches[2] -- There are 2 paranthesized captures
 end
 
 -- Extracts a SOAP action from the Content-Type header. In SOAP, the
@@ -64,7 +71,9 @@ end
 local function soap_action_in_ctype(headers)
   local ctype = headers['Content-Type']
 
-  if ctype and starts_with(ctype, soap_action_ctype) then
+  -- The Content-Type can be a mix of upper and lower-case chars. Convert it to
+  -- include only lower-case chars to be able to compare it.
+  if ctype and starts_with(lower(ctype), soap_action_ctype) then
     local header_params = sub(ctype, len(soap_action_ctype) + 1, -1)
     return soap_action_from_ctype_params(header_params)
   else
--- a/gateway/src/apicast/policy/soap/soap.lua
+++ b/gateway/src/apicast/policy/soap/soap.lua
@@ -25,7 +25,7 @@ local policy = require('apicast.policy')
 local _M = policy.new('SOAP policy')
 
 local soap_action_header = 'SOAPAction'
-local soap_action_ctype = 'application/soap+xml;'
+local soap_action_ctype = 'application/soap+xml'
 
 local new = _M.new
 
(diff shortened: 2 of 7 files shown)


davidor:
@mikz  I didn't add a JSON schema. I'll wait until https://github.com/3scale/apicast/pull/565 is merged.

gateway/src/apicast/policy/soap/policy.lua
```diff
@@ -0,0 +1,68 @@
...
+-- contains pairs of (metric, usage).
+-- @tparam table context Shared context between policies
+function _M:rewrite(context)
+  local soap_action_uri = extract_soap_uri()
+
+  if soap_action_uri then
+    local soap_usage = context.service:get_usage(
+      ngx.req.get_method(), soap_action_uri)
+
+    context.add_to_usage = soap_usage
```
davidor: @mikz I considered several options here but this seemed the cleanest one. We could pass the URI instead and do the mapping rules matching later.

gateway/src/apicast/proxy.lua
```diff
@@ -231,11 +241,11 @@ function _M:call(service)
 
   return function()
     -- call access phase
-    return self:access(service)
+    return self:access(service, context)
   end
 end
 
-function _M:access(service)
+function _M:access(service, context)
```
davidor: @mikz  I decided to pass the context instead of passing just the usage to be increased because in the future there might be other policies passing other information in the context. This way we won't need to change the signature of the access() method again.

gateway/src/apicast/policy/soap/policy.lua
```diff
@@ -0,0 +1,68 @@
+--- SOAP Policy
+-- This policy adds support for a very small subset of SOAP.
+-- This policy basically expects a SOAPAction URI in the SOAPAction header or
+-- the content-type header.
+-- The SOAPAction header is used in v1.1:
```
andrewdavidmackenzie: "....v1.1 of the SOAP Standard"  
(just to avoid any confusion in readers about versions of apicast or something?)

gateway/src/apicast/policy/soap/policy.lua
```diff
@@ -0,0 +1,68 @@
...
+-- https://www.w3.org/TR/2000/NOTE-SOAP-20000508/#_Toc478383528, whereas the
+-- Content-Type header is used in v1.2:
+-- https://www.w3.org/TR/soap12-part2/#ActionFeature
+-- The SOAPAction URI is matched against the mapping rules defined for the
+-- service and the usage resulting from that is authorized and reported
+-- against 3scale's backend.
```
andrewdavidmackenzie: "...and reported to the API Management platform using the Service Management API" or similar?

gateway/src/apicast/policy/soap/policy.lua
davidor: :+1:

gateway/src/apicast/policy/soap/policy.lua
davidor: I usually refer to backend as '3scale's backend' sounds simpler than that, but would be good to reach an agreement on that. Should we call it by its _new upstream name_?

mikz: I'd call it 3scale's backend.  Definitely try not to write its full name because I could not get it right and have to copy paste it from somehwere.

gateway/src/apicast/proxy.lua
mikz: I'm not sure. https://github.com/3scale/apicast/pull/556/files#diff-31313c92616b54028dac1ba183e6f79aR287 was all the info it needed. It is quite a lot of params, but not that bad imo.

gateway/src/apicast/policy/soap/policy.lua
mikz: I wonder how several policies would interact. Or if this would be twice in the chain.

My idea was to make each policy to increment the values in the context. That way the policy can decide what to do: add, replace, remove, ...

gateway/src/apicast/policy/soap/policy.lua
```diff
@@ -0,0 +1,68 @@
...
+
+local sub = string.sub
+local len = string.len
+
+local policy = require('apicast.policy')
+local _M = policy.new('SOAP policy')
+
+local soap_action_header = 'SOAPAction'
+local soap_action_ctype = 'application/soap+xml?action='
```
mikz: I think this has to be parsed as URI args. Because we can't control the action appears right after as it can have `&` and be the last param.

Also quick google search shows some people use `;` to concat params. We should verify if that can really be the case.

gateway/src/apicast/policy/soap/policy.lua
```diff
@@ -0,0 +1,68 @@
...
+--- Rewrite phase
+-- Adds to the field 'add_to_usage' of the shared context a table that
+-- contains pairs of (metric, usage).
+-- @tparam table context Shared context between policies
+function _M.rewrite(_, context)
+  local soap_action_uri = extract_soap_uri()
+
+  if soap_action_uri then
+    local soap_usage = context.service:get_usage(
+      ngx.req.get_method(), soap_action_uri)
```
mikz: So this is using service mapping rules? I'm not sure users can define full URLs there.
I though we will have some mapping rules in this policy.

davidor: I assumed it was possible to define something like `/a_path#some_action` in the mapping rules. That's why I decided to do it this way. I see now that it's not possible. With that constraint, I think we need to include mapping rules in the configuration of this policy.

gateway/src/apicast/policy/soap/policy.lua
davidor: Agreed @mikz

gateway/src/apicast/policy/soap/policy.lua
davidor: You're right @mikz . I wrongly assumed that `action` would be the only param.

gateway/src/apicast/proxy.lua
davidor: I'm OK with adding 2 or 3 extract params. I'm more concerned about changing the signature of `access()` in the future and breaking backwards compatibility.

mikz: I agree. Maybe we can build on https://github.com/3scale/apicast/pull/556/ and try to figure out out there and then just use it from here?

davidor: Sounds good :+1:

gateway/src/apicast/policy/soap/policy.lua
davidor: Definitely. If in the future we have more policies doing the same thing, we'll need to update `add_to_usage` instead of simply assign it.

mikz: What I mean is that we can just `context.usage.hits = (context.usage.hits or 0) + 1` in the rewrite phase. Or expose some table merger function to merge those. Or have them in an array and merge them in the proxy:access. It just does not feel right to have `add_to_usage` when the `usage` key is there too.

gateway/src/apicast/policy/soap/policy.lua
davidor: @mikz  Let me know what you think about this solution:
- Add a mapping rules array in the policy config. Each element will contain: `pattern`, `http_method`, `metric_name`, and `delta`. The same fields we have for the service mapping rules.
- Add a `add_mapping_rule()` method in `Service`. This would simply accept the fields above, add the ones needed to construct the rule, and then add it to the `self.rules` table. We'll need to make sure that `add_mapping_rule()` is only called once during the lifetime of the policy so we don't add duplicated rules.
- The above will let us call the existing `Service:get_usage()`. This method returns the metrics to auth and report based on the mapping rules defined. We'll call this method only when we receive a SOAP action.

mikz: @davidor interesting idea. I'm not sure if I like the idea of merging then into one list. I think they should be applied separately. IMO there could be some rules that match both and because some are absolute and some relative it could be quite nasty as `/` would increment hits all the time and twice.
Maybe we need to expose`UsageExtractor:get_usage` which is initialized with list of mapping rules. Basically exposing the low level API where you can inject own mapping rules.

davidor: I get your first point, but wouldn't that be coherent with the way mapping rules work currently?

Suppose that the policy adds this mapping rule: `/a_path#a_soap_action => +1 a_soap`. If the user has already defined `/a_path => +1 a_path`, we should increase both `a_soap` and `a_path` by 1. If the user already defined `/ => +1 hits`, and `hits` is a parent metric of the other 2, well, it's going to be increased by 3(!), but that's how it currently works no?

Regarding your second suggestion, could you please expand a bit on it or provide some pseudo-code or examples?

mikz: I think the issue is when you have following rules:

/ => hits
/a => a

And in soap you'll have 
http://example.com/soap#a => a

If you match `http://example.com/soap#a` it will return `hits=1&a=1`.

So it depends if this **replaces** mapping rules or **complements** them.
If it would first match mapping rules and then soap rules and merged the results then for a request:
```http
GET /
Content-Type: application/xml+soap?action=http://example.com/soap#a
```

it would match `hits=2&a=1`. Which is wrong IMO. Because the path mapping rules should not be applied to soap mapping rules (because they are relative and not absolute).
I'm not thinking about parent/child metrics, just about applying mapping rules to the request path/ soap action.
The point is that current path relative mapping rules like `/a` can unintentionally match soap services like `http://example.com/a` and double counting. And also match different services like `http://example.com/a` and `http://foobar.com/a`.

And the signature of UsageExtractor would be quite simple. 
```
local usage_rules = UsageExtractor.new({
  { pattern = 'http://example.com/a', http_method = 'GET', metric_name = 'a', delta = 1 }
})
local usage = usage_rules:get_usage(...)
```

And the `Service:get_usage` would use this too.  But we would just have an API to create own independent list of rules.

davidor: @mikz  Thanks for the example.

It's clear to me now why you think that a url received via SOAP Action should not be matched against the service mapping rules. It should only be matched against the mapping rules specified in the SOAP policy. I agree with that.

Now, what happens when the SOAP policy is enabled but a request does not contain a SOAP Action ? I think we should apply the service mapping rules in that case.

I like the idea of extracting a `UsageExtractor` module from `Service` and `Configuration`. This is something I've been thinking about. Those 2 modules have many responsabilities and need to be splitted. This seems to be a good excuse to do it :) I'll address this task in a separate PR.

gateway/src/apicast/policy/soap/policy.lua
davidor: About using `;`, I think we need to support that: https://developer.mozilla.org/es/docs/Web/HTTP/Headers/Content-Type

mikz: Yep. https://tools.ietf.org/html/rfc7231#section-3.1.1.1 says parameters are separated by `;` and  https://tools.ietf.org/html/rfc3902 says it is a parameter. I wonder where I've seen the example with `?`. Can't find it now.

gateway/src/apicast/policy/soap/init.lua
```diff
@@ -0,0 +1 @@
+return require('soap')
```
mikz: So I started to think about this. And I think we can expose some support tooling in different module, so it can be unit tested.

Lets say the main policy code is in `soap_policy.lua`. Then there can be `soap.lua` that has stuff like "extracting the soap action" and it can be unit tested in busted. 

That would allow us not exposing extra methods on the policy, but still exposing it internally (if the loading works) for tests. And policies should not be able to load other policies (but that is not enforced yet), so we should be fine and the code should be used only from tests.

Just some food for though. I'd like to hear your take.

davidor:
This is ready @mikz . There have been a few changes. Mainly because the code is now using the modules extracted in #571 , #573 , and #580 .

I also addressed your comments.

gateway/src/apicast/policy/soap/soap.lua
```diff
@@ -0,0 +1,127 @@
...
+-- headers. When both contain a SOAP action, the Content-Type one takes
+-- precedence.
+local function extract_soap_uri()
+  local headers = ngx.req.get_headers() or {}
+  return soap_action_in_ctype(headers) or soap_action_in_header(headers)
+end
+
+local function usage_from_matching_rules(soap_action_uri, rules)
+  return mapping_rules_matcher.get_usage_from_matches(
+    nil, soap_action_uri, {}, rules)
```
davidor: @mikz Notice that I'm sending http_method = nil here. Not sure if we should take it into account for SOAP actions.

mikz: Hm. SOAP can be used with [GET](https://www.w3.org/TR/2007/REC-soap12-part0-20070427/#L26854) and [POST](https://www.w3.org/TR/2007/REC-soap12-part0-20070427/#L26866).
Aha! The GET does not have the `Content-Type` header. So technically it should be only POST. I guess.

But it looks like the GET is used as some REST hybrid, so I guess we could leave that to the mapping rules.
I'd possibly even go for hardcording POST there.

spec/policy/soap/policy_spec.lua
```diff
@@ -0,0 +1,101 @@
...
+    describe('when the SOAP action is in the SOAPAction and the Content-Type headers', function()
+      it('calculates the usage and merges it with the one in the context', function()
+        ngx.req.get_headers = function()
+          return {
+            SOAPAction = '/soap_action',
+            ["Content-Type"] = "application/soap+xml;action=/soap_action_ctype"
+          }
+        end
+
+        soap_policy:rewrite(context)
```
mikz: Oh. This is so good 🥇  Great we can unit test this.

gateway/src/apicast/policy/soap/soap.lua
```diff
@@ -0,0 +1,127 @@
...
+  return nil
+end
+
+-- Extracts a SOAP action from the Content-Type header. In SOAP, the
+-- type/subtype is application/soap+xml, and the action is specified as a
+-- param in that header. When there is no SOAP action, this method returns nil.
+local function soap_action_in_ctype(headers)
+  local ctype = headers['Content-Type']
+
+  if ctype and starts_with(ctype, soap_action_ctype) then
```
mikz: I think this won't be that easy 🤔 
The [RFC](https://tools.ietf.org/html/rfc7231#section-3.1.1.1) says those are all equivalent:

>   text/html;charset=utf-8
     text/html;charset=UTF-8
     Text/HTML;Charset="utf-8"
     text/html; charset="utf-8"

And looks like it can have whitespace before `;` too. Look at the definition: ` media-type = type "/" subtype *( OWS ";" OWS parameter )`. And `OWS` is optional white space.

gateway/src/apicast/policy/soap/soap.lua
davidor: I don't really know about this. The Content-Type is included in the request, not in the SOAP action URI.
Depending on what we decided here we'll need to add it to the JSON schema also. Notice that for now, I only included 'pattern', 'metric', and 'delta', and left out other fields that are present in the `proxy rules`.

mikz: I see. We are initializing the mapping rules directly from the config. I was thinking that we could just override the `http_method` attribute with `POST` there. And then pass the real http method here. So we match it only when the request is POST. 
But it is probably not really important.

gateway/src/apicast/policy/soap/init.lua
davidor: That would be a nice improvement.

gateway/src/apicast/policy/soap/soap.lua
davidor: You're right :+1:

gateway/src/apicast/policy/soap/soap.lua
```diff
@@ -0,0 +1,127 @@
...
+-- This method returns the value of 'action' or nil when it's not present.
+local function soap_action_from_ctype_params(params)
+  local params_split = re.split(params, ";")
+
+  for _, param in ipairs(params_split) do
+    local header_param_split = re.split(param, "=")
+    local name = header_param_split[1]
+    local value = header_param_split[2]
+    if name == "action" then
+      return value
```
mikz: The value can be either `"url"` or just `url`. We should strip the quotes.

gateway/src/apicast/policy/soap/soap.lua
davidor: Check last commit.

gateway/src/apicast/policy/soap/soap.lua
davidor: This is also solved in the new commit.

gateway/src/apicast/policy/soap/soap.lua
```diff
@@ -43,14 +44,25 @@ end
 -- Content-Type header. The string has this format:
 -- a_param=x;action=soap_action;another_param=y
 -- This method returns the value of 'action' or nil when it's not present.
+-- Note that params can be a mix of lower and upper case chars, it can have
+-- spaces, and wrap the value with '"':
+-- https://tools.ietf.org/html/rfc7231#section-3.1.1.1
 local function soap_action_from_ctype_params(params)
-  local params_split = re.split(params, ";")
+  local params_without_blanks = ngx.re.gsub(params, '\\s', '')
+  local params_split = re.split(params_without_blanks, ";")
```
mikz: You could also split by `[[\s*;\s*]]` to remove spaces around `;`, right?
Seems a bit evil to strip all spaces from it. What if there is a space in the url? Now when I think about it. 
What happens when there is `;` in soap action? That would be quite broken.

gateway/src/apicast/policy/soap/soap.lua
```diff
@@ -64,7 +76,9 @@ end
 local function soap_action_in_ctype(headers)
   local ctype = headers['Content-Type']
 
-  if ctype and starts_with(ctype, soap_action_ctype) then
+  -- The Content-Type can be a mix of upper and lower-case chars. Convert it to
+  -- include only lower-case chars to be able to compare it.
+  if ctype and starts_with(lower(ctype), soap_action_ctype) then
```
mikz: I think this would not work for `application/xml+soap ; action=""` because of the whitespace.

spec/policy/soap/policy_spec.lua
```diff
@@ -0,0 +1,128 @@
...
+  describe('.rewrite', function()
+    local context -- Context shared between policies
+
+    -- Define a config with 2 rules. One increases hits by 10 and the other by
+    -- 20. Their patterns have values that allow us to easily associate them
+    -- with a SOAP action receive via SOAPAction header or via Content-Type.
+    local policy_config = {
+      mapping_rules = {
+        {
+          pattern = '/soap_action$',
```
mikz: Because the RFC says these should be full URL we should test also with full URLs. To verify all the escaping works.

PR #570, opened 2018-02-01
policy/cors: stop setting headers twice in preflight reqs

davidor:
This PR introduces a minor improvement in the CORS policy.
In the case of a preflight request, we were setting the CORS headers twice. This PR fixes that.

Thanks @mayorova for reporting.


gateway/src/apicast/policy/cors/policy.lua
```diff
@@ -72,9 +72,9 @@ local function is_cors_preflight()
          ngx.var.http_access_control_request_method
 end
 
-function _M:rewrite()
+function _M.rewrite(_)
```
davidor: Changed this so codeclimate does not complain about unused `self`.

PR #571, opened 2018-02-02
Extract MappingRule from Configuration

davidor:
Minor refactoring.

This PR extracts MappingRule from Configuration. There were many local methods in `Configuration` that were related only to mapping rules, so I think this makes sense.
Just to be clear, there are no changes in the implementation nor the test coverage.

I think that some of the code from `Service` should also be in this new module, but let's go step by step.

Let me know what you think @mikz . If you agree I'll add some documentation for the public methods.

### DIFF
--- a/gateway/src/apicast/configuration.lua
+++ b/gateway/src/apicast/configuration.lua
@@ -5,20 +5,18 @@ local _M = {
 local len = string.len
 local pairs = pairs
 local type = type
-local error = error
 local tostring = tostring
 local next = next
 local lower = string.lower
 local insert = table.insert
 local setmetatable = setmetatable
-local re_match = ngx.re.match
 
-local inspect = require 'inspect'
 local re = require 'ngx.re'
 local env = require 'resty.env'
 local resty_url = require 'resty.url'
 local util = require 'apicast.util'
 local policy_chain = require 'apicast.policy_chain'
+local mapping_rule = require 'apicast.mapping_rule'
 
 local mt = { __index = _M, __tostring = function() return 'Configuration' end }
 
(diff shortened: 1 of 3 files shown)


gateway/src/apicast/configuration.lua
```diff
@@ -159,19 +109,7 @@ function _M.parse_service(service)
...
-          pattern = proxy_rule.pattern,
-          regexpified_pattern = regexpify(proxy_rule.pattern),
-          parameters = proxy_rule.parameters,
-          querystring_params = function(args)
-            return check_querystring_params(querystring_parameters, args)
-          end,
-          system_name = proxy_rule.metric_system_name or error('missing metric name of rule ' .. inspect(proxy_rule)),
-          delta = proxy_rule.delta
-        }
+        return mapping_rule.from_proxy_rule(proxy_rule)
```
mikz: This can be further simplified to `map(mapping_rule.from_proxy_rule, proxy.proxy_rules or empty)`

gateway/src/apicast/mapping_rule.lua
```diff
@@ -0,0 +1,94 @@
...
+        match = false
+        break
+      end
+    end
+  end
+
+  return match
+end
+
+function _M.new(http_method, pattern, params, querystring_params, metric, delta)
```
mikz: I'd keep this method private because there is high chance the initializer will change with that many params.

mikz:
@davidor definitely agree! This is a good path to split the configuration.lua. The `.new(...)` looks like has too many params, so maybe I'd not expose it just yet. But looks good anyway!

gateway/src/apicast/mapping_rule.lua
davidor: Good point. Let's wait until we bring more functionality from `Service`.

gateway/src/apicast/configuration.lua
davidor: Fixed :+1:

PR #573, opened 2018-02-05
Move mapping rule checking to MappingRule and add missing tests

davidor:
Refactor.
This PR moves the method that checks a mapping rule againts a request from the `Service` module to the `MappingRule` module. I think it clearly belongs there.
This PR also adds some unit tests for the matching of mapping rules.

### DIFF
--- a/gateway/src/apicast/mapping_rule.lua
+++ b/gateway/src/apicast/mapping_rule.lua
@@ -8,6 +8,7 @@ local setmetatable = setmetatable
 local pairs = pairs
 local error = error
 local type = type
+local format = string.format
 local re_match = ngx.re.match
 local insert = table.insert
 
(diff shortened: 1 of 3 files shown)


gateway/src/apicast/mapping_rule.lua
```diff
@@ -109,4 +114,19 @@ function _M.from_proxy_rule(proxy_rule)
   )
 end
 
+--- Checks if the mapping rule matches a given request method, path, and args
+--
+-- @tparam string method HTTP method (GET, POST, etc.).
+-- @tparam string path Path of an HTTP request.
+-- @tparam table args Table with the args and values of an HTTP request.
+-- @treturn boolean Whether the mapping rule matches the given request.
+function _M:matches(method, path, args)
```
mikz: If we introduce SOAP rules that are full URL the `path` will no longer be just path, but a full URI. But a `pathj` is an `URI`. So maybe lets call it `uri` ?

davidor: Good point.

davidor: Changed :+1:

PR #581, opened 2018-02-07
[cli] allow changing policy load path

mikz:
Allow configuring policy load paths from the CLI.
That makes easier testing and development.

### DIFF
--- a/examples/policies/README.md
+++ b/examples/policies/README.md
@@ -1,0 +1,45 @@
+# Policies examples
+
+The ngx_example policy is setting request header based on the policy configuration.
+
+You can see the APIcast configuration in [examples/policies/example.json](./example.json).
+It defines a policy chain with two policies: `ngx_example` and `upstream`.
+The `upstream` policy just forwards the request to some upstream server based on its' configuration.
+The `ngx_example` policy is configured to set HTTP header `Example` to value `Value`.
+
+You can start example configuration with example policy by running this from the APIcast root directory:
+
+```shell
+bin/apicast -c examples/policies/example.json -b --policy-load-path examples/policies -v
+```
+
+And then make a request by curl:
+
+```shell
+curl 'localhost:8080/test'
+```
+
+And see a response:
+
+```http
+GET /test HTTP/1.1
+X-Real-IP: 127.0.0.1
+Host: echo
+User-Agent: curl/7.57.0
+Accept: */*
+Example: Value
+```
+
+And see similar line in the APIcast log:
+
+```
+2018/02/08 10:01:48 [notice] 72249#10132050: *9 [lua] init.lua:19: op(): setting header: Example to: Value, client: 127.0.0.1, server: _, request: "GET /test HTTP/1.1", host: "localhost:8080"
+```
+
+You can see the policy entry point in [examples/policies/ngx-example/1.0.0/init.lua](./ngx-example/1.0.0/init.lua) and the source code in
+[examples/policies/ngx-example/1.0.0/ngx_example.lua](./ngx-example/1.0.0/ngx_example.lua) to verify it indeed should print such log entry.
+
+For the policy to be complete it needs a manifest with description and the configuration schema.
+The file has to be named [`apicast-policy.json`](./ngx-example/1.0.0/apicast-policy.json) and has
+to be valid according to our [manifest JSON schema](../../gateway/src/apicast/policy/manifest-schema.json).
+
--- a/examples/policies/ngx-example/1.0.0/apicast-policy.json
+++ b/examples/policies/ngx-example/1.0.0/apicast-policy.json
@@ -1,0 +1,31 @@
+{
+  "$schema": "http://apicast.io/policy-v1/schema#manifest#",
+  "name": "Ngx example policy",
+  "description":
+  ["This policy is meant to be just an example.",
+    "It sets request headers based on the configuration.",
+    "And prints a log entry for each header set. " ],
+  "version": "0.1",
+  "configuration": {
+    "type": "object",
+    "properties": {
+      "set_header": {
+        "type": "array",
+        "items": {
+          "type": "object",
+          "properties": {
+            "name": {
+              "type": "string",
+              "description": "HTTP header name"
+            },
+            "value": {
+              "type": "string",
+              "description": "HTTP header value"
+            }
+          },
+          "required": ["name", "value"]
+        }
+      }
+    }
+  }
+}
--- a/examples/policies/ngx-example/1.0.0/init.lua
+++ b/examples/policies/ngx-example/1.0.0/init.lua
@@ -1,0 +1,1 @@
+return require('ngx_example')
--- a/gateway/conf.d/apicast.conf
+++ b/gateway/conf.d/apicast.conf
@@ -82,23 +82,23 @@ location @upstream {
 }
 
 location / {
-  set $cached_key null;
-  set $credentials null;
-  set $usage null;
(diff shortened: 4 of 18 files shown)


gateway/src/apicast/policy_loader.lua
```diff
@@ -211,16 +214,24 @@ local mt = {
   __call = function(loader, ...) return loader.env.require(...) end
 }
 
-function _M.new(name, version, dir)
-  local apicast_dir = dir or getenv('APICAST_DIR') or '.'
+do
```
davidor: Why is this in a `do` block?

mikz: It makes a variable that is not really needed to be exposed for the rest of the module. 

I've seen this in resty-core and other resty libraries. I think it is quite nice pattern to create temporary variables.

davidor: I see :+1:

gateway/conf.d/apicast.conf
```diff
@@ -82,23 +82,23 @@ location @upstream {
 }
 
 location / {
-  set $cached_key null;
-  set $credentials null;
-  set $usage null;
-  set $service_id null;
-  set $proxy_pass null;
-  set $secret_token null;
+  set $cached_key '';
```
davidor: There's something I don't understand about this. I see in the commit comment that you're doing this to avoid sending this in headers for example. Why are they sent when set to `null` and not sent when set to `''` ?

examples/policies/ngx-example/1.0.0/init.lua
```diff
@@ -0,0 +1,35 @@
+
```
davidor: I think we should follow the same pattern we follow for our policies, that is, having a `init.lua` with only a `require` and the code in a different .lua file.

davidor: We should probably add a json schema as well to show a complete example.

examples/policies/README.md
```diff
@@ -0,0 +1,31 @@
+# Policies examples
+
+You can start example configuration with example policy by runing this from the APIcast root directory:
```
davidor: I think that in order to be able to follow the example, we need to specify here what's the policy being loaded in the example configuration file and what it does.

examples/policies/README.md
```diff
@@ -0,0 +1,31 @@
...
+Host: echo
+User-Agent: curl/7.57.0
+Accept: */*
+Example: Value
+```
+
+And see similar line in the APIcast log:
+
+```
+2018/02/08 10:01:48 [warn] 72249#10132050: *9 [lua] init.lua:19: op(): setting header: Example to: Value, client: 127.0.0.1, server: _, request: "GET /test HTTP/1.1", host: "localhost:8080"
```
davidor: This shouldn't be a warning, no? It's the purpose of the policy. Warning sounds like something is not working as it should, I'd use a different log level for the example.

examples/policies/ngx-example/1.0.0/init.lua
```diff
@@ -0,0 +1,35 @@
...
+    end)
+  end
+
+  self.ops = ops
+
+  return self
+end
+
+function _M:rewrite()
+  for _,op in pairs(self.ops) do
```
davidor: It should be `ipairs`, no?

gateway/conf.d/apicast.conf
mikz: Nginx is not going to send values that are empty strings. But it was sending `null`. Which was pretty ugly. Imo better to keep it empty than send `null` as a string.

examples/policies/ngx-example/1.0.0/init.lua
mikz: Yep!

examples/policies/README.md
mikz: Ok. Will tune it to info/notice level.

examples/policies/ngx-example/1.0.0/init.lua
mikz: True!

examples/policies/README.md
mikz: Better now?

examples/policies/README.md
mikz: Done.

examples/policies/ngx-example/1.0.0/init.lua
mikz: Done.

examples/policies/ngx-example/1.0.0/init.lua
mikz: First part done. Will do the schema now.

gateway/conf.d/apicast.conf
davidor: :+1: 
That's a surprising behavior.

examples/policies/README.md
davidor: :+1:

examples/policies/ngx-example/1.0.0/init.lua
mikz: Done in https://github.com/3scale/apicast/pull/581/commits/e6785ae55b818ec83b0a2f273b4a2b5fbe89a88e

examples/policies/ngx-example/1.0.0/apicast-policy.json
```diff
@@ -0,0 +1,16 @@
...
+  "$schema": "http://apicast.io/policy-v1/schema#manifest#",
+  "name": "Ngx example policy",
+  "description":
+  ["This policy is meant to be just an example.",
+    "It sets request headers based on the configuration.",
+    "And prints a log entry for each header set. " ],
+  "version": "0.1",
+  "configuration": {
+    "type": "object",
+    "additionalProperties": {
```
davidor: Maybe this would be easier to understand if we made the fields explicit by specifying a `set_header` object with two properties: `header` and `value`. What do you think?

mikz: True. That is what I wanted to do :)  I might need to change it to an array, but that is probably for the better.

mikz: Changed to an array.

gateway/http.d/init.conf
```diff
@@ -12,6 +12,13 @@ init_by_lua_block {
     local context = env:context()
 
     local resty_env = require('resty.env')
+    -- this file both liquid template and pure nginx file
+    -- {% raw %} and {{ ENV }} is both valid Lua code and Liquid variable interpolation {% endraw %}
+    -- so we can push ENV from outside when this file is included and inject extra environment
```
mikz: @davidor I definitely have to improve this and properly document it, because this is serious mindfuck.

I was thinking about something like:
```lua
-- WARNING, WARNING, WARNING: this is insane hack and should not be touched
-- This file is not templated by Liquid, because it is used by not blackbox integration tests (Search for TEST_NGINX_HTTP_CONFIG).
-- So it can't be templated by liquid, to lift env variables and persist them in the configuration.
-- This is a workaround so we can store environment at the time of building the config into the config itself.
--  {% raw %}  {{ ENV }}  {% endraw %} is both valid Liquid template and Lua code.
-- In Lua it is a table with another empty table inside. In Liquid it prints variable ENV.
-- This ENV variable is defined in the main nginx.conf.liquid and injected when including this partial.
-- The content of the ENV variable is a Lua table, so when rendered it actually can run ipairs on it.
```

PR #585, opened 2018-02-09
[bin] honour APICAST_LOG_LEVEL env

mikz:
* so the first resty instance is started with proper log level


davidor:
Pushed to solve conflicts in the changelog.

PR #588, opened 2018-02-12
[circleci] update s2i-openresty

mikz:
should result in faster builds
https://github.com/3scale/s2i-openresty/releases/tag/1.13.6.1-rover6


mikz:
Looks like this saves about 30s per build. So 30s in s2i-builder, 30s in s2i-runtime and 60s in deploy.

PR #590, opened 2018-02-13
[cli] decouple apicast environment and 3scale configuration channel

mikz:
* `THREESCALE_DEPLOYMENT_ENV` controls the 3scale proxy configuration
channel (staging/production) and can be passed as `-3 --channel` from the CLI
* `APICAST_ENVIRONMENT` controls the APIcast settings (caching, ...) and can be passed by `-e --environment` from the CLI
* `APICAST_LOADED_ENVIRONMENTS` can control always loaded APIcast environments (internal use only)
* update Test-APIcast to include https://github.com/3scale/Test-APIcast/pull/3
* `production` is still the default configuration channel, that loads the `production` environment configuration which configures the configuration loader to "boot". In short this makes APIcast to require `THREESCALE_PORTAL_ENDPOINT` or `--configuration` when started without any other params. That makes `bin/apicast` to not start because it is missing configuration.

Needed for https://github.com/3scale/apicast-cloud-hosted/pull/1

### DIFF
--- a/gateway/config/sandbox.lua
+++ b/gateway/config/sandbox.lua
@@ -1,0 +1,1 @@
+staging.lua
--- a/gateway/config/staging.lua
+++ b/gateway/config/staging.lua
@@ -1,0 +1,6 @@
+return {
+    master_process = 'on',
+    lua_code_cache = 'on',
+    configuration_loader = 'lazy',
+    configuration_cache = 0,
+}
--- a/gateway/http.d/init.conf
+++ b/gateway/http.d/init.conf
@@ -15,7 +15,7 @@ init_by_lua_block {
     -- This ENV variable is defined in the main nginx.conf.liquid and injected when including this partial.
     -- The content of the ENV variable is a Lua table, so when rendered it actually can run ipairs on it.
     for k,v in pairs({{ ENV }}) do
-      if type(k) == 'string' and k ~= 'APICAST_CONFIGURATION_LOADER' then
+      if type(k) == 'string' and not resty_env.value(k) then
         resty_env.set(k,v)
       end
     end
--- a/gateway/src/apicast/cli/environment.lua
+++ b/gateway/src/apicast/cli/environment.lua
@@ -75,17 +75,22 @@ _M.default_config = {
 
 local mt = { __index = _M }
 
+--- Return loaded environments defined as environment variable.
+-- @treturn {string,...}
+function _M.loaded()
+    local value = resty_env.value('APICAST_LOADED_ENVIRONMENTS')
+    return re.split(value or '', [[\|]], 'jo')
+end
+
 --- Load an environment from files in ENV.
 -- @treturn Environment
 function _M.load()
-    local value = resty_env.value('APICAST_LOADED_ENVIRONMENTS')
     local env = _M.new()
+    local environments = _M.loaded()
 
-    if not value then
+    if not environments then
         return env
     end
-
-    local environments = re.split(value, '\\|', 'jo')
 
     for i=1,#environments do
         assert(env:add(environments[i]))
--- a/gateway/src/apicast/cli/environment.lua
+++ b/gateway/src/apicast/cli/environment.lua
@@ -148,7 +148,7 @@ function _M:add(env)
 
     local config = loadfile(path, 't', {
         print = print, inspect = require('inspect'), context = self._context,
-        tonumber = tonumber, tostring = tostring,
+        tonumber = tonumber, tostring = tostring, os = { getenv = resty_env.value },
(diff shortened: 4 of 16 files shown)


gateway/config/sandbox.lua
```diff
@@ -0,0 +1 @@
+staging.lua
```
davidor: Why `sandbox` and why it's the same as staging?

mikz: Because some configurations can still have `THREESCALE_DEPLOYMENT_ENV=sandbox`. It was the original value in AMP 2.0 IIRC. We alias it now to `staging` which makes more sense.

davidor: :+1:

gateway/config/staging.lua
```diff
@@ -0,0 +1,6 @@
+return {
+    master_process = 'on',
+    lua_code_cache = 'on',
+    configuration_loader = 'lazy',
+    configuration_cache = 0,
```
davidor: Why not `os.getenv('APICAST_CONFIGURATION_CACHE') or 0` ?

gateway/src/apicast/cli/environment.lua
```diff
@@ -148,7 +148,7 @@ function _M:add(env)
 
     local config = loadfile(path, 't', {
         print = print, inspect = require('inspect'), context = self._context,
-        tonumber = tonumber, tostring = tostring,
+        tonumber = tonumber, tostring = tostring, os = { getenv = resty_env.value },
```
davidor: I think that overriding `getenv` here can be a bit confusing.
Although I get that you're doing this to be able to use it in the environment files.

gateway/config/staging.lua
mikz: I think our staging should always reload configuration. And because of the order of `options` and `context` in the `start.lua` the CLI options take precedence. So setting `APICAST_CONFIGURATION_CACHE` env would override this anyway.

That makes the production.lua `os.getenv` a bit redundant, but I wanted it there as an example of what can be done. But if it is confusing I could change/remove it.

t/configuration-loading-boot-with-config.t
```diff
@@ -1,10 +1,8 @@
 use lib 't';
 use Test::APIcast::Blackbox 'no_plan';
 
-$ENV{APICAST_CONFIGURATION_LOADER} = 'boot';
-
-env_to_nginx(
-    'APICAST_CONFIGURATION_LOADER',
+env_to_apicast(
+    'APICAST_CONFIGURATION_LOADER' => 'boot'
```
davidor: Was this allowed before or it's supported since Test::APIcast 0.08 ?

gateway/src/apicast/cli/environment.lua
mikz: Yep. I think it is more useful to expose `resty_env.value` than plain `resty_env.get` or `os.getenv`.
And the only difference is that it does not return empty strings. Which is actually useful for default values.

t/configuration-loading-boot-with-config.t
mikz: This wasn't really working before Test::APIcast 0.08. `env_to_nginx` was not working in Test::APIcast::Blackbox at all.

gateway/http.d/init.conf
```diff
@@ -15,7 +15,7 @@ init_by_lua_block {
     -- This ENV variable is defined in the main nginx.conf.liquid and injected when including this partial.
     -- The content of the ENV variable is a Lua table, so when rendered it actually can run ipairs on it.
     for k,v in pairs({{ ENV }}) do
-      if type(k) == 'string' and k ~= 'APICAST_CONFIGURATION_LOADER' then
+      if type(k) == 'string' and not resty_env.value(k) then
```
davidor: I don't get this change. Can you explain it, please?

mikz: It sets only missing/empty variables.

This leaves the flexibility of overriding the ENV when starting nginx and treating the `{{ ENV }}` only as default values.

Because Test::APIcast runs `APICAST_CONFIGURATION_LOADER=test bin/apicast` and expects it to work. But in cases where `APICAST_CONFIGURATION_LOADER=boot` is already set in the ENV this would not apply and `APICAST_CONFIGURATION_LOADER` would have value `boot` during initialization. Basically ignoring what ENV it was started in.

So this change is for two reasons:

* I think it makes sense for ENV of the shell to take priority than what is in the config file
* Test::APIcast needs to start APIcast with APICAST_CONFIGURATION_LOADER=test when testing configuration with `--test` but then later start it with `boot` when the config is started for real.

PR #592, opened 2018-02-14
Endpoint to fetch JSON manifests of policies

davidor:
This PR introduces an endpoint in the management API to get all the schemas of the policies, both the built-in ones and the ones loaded using the `--policy-load-path` option.

/cc @3scale/system @macejmic the Policy API has now final shape.

This is the output with the policies we have:
```bash
curl "http://172.18.0.3:8090/policies/" | jq
{
  "policies": [
    {
      "description": [
        "This policy allows to modify the path of a request. ",
        "The operations supported are sub and gsub based on ngx.re.sub and ",
        "ngx.re.gsub provided by OpenResty. Please check ",
        "https://github.com/openresty/lua-nginx-module for more details on how ",
        "to define regular expressions and learn the options supported."
      ],
      "$schema": "http://apicast.io/policy-v1/schema#manifest#",
      "name": "URL rewriting policy",
      "configuration": {
        "properties": {
          "commands": {
            "items": {
              "properties": {
                "op": {
                  "oneOf": [
                    {
                      "const": "sub",
                      "description": "Substitutes the first match of the regex applied."
                    },
                    {
                      "const": "gsub",
                      "description": "Substitutes all the matches of the regex applied."
                    }
                  ],
                  "description": "Operation to be applied (sub or gsub)",
                  "type": "string"
                },
                "regex": {
                  "type": "string",
                  "description": "Regular expression to be matched"
                },
                "break": {
                  "type": "boolean",
                  "description": "when set to true, if the command rewrote the URL, it will be the last one applied"
                },
                "replace": {
                  "type": "string",
                  "description": "String that will replace what is matched by the regex"
                },
                "options": {
                  "type": "string",
                  "description": "Options that define how the regex matching is performed"
                }
              },
              "required": [
                "op",
                "regex",
                "replace"
              ],
              "type": "object"
            },
            "description": "List of rewriting commands to be applied",
            "type": "array"
          }
        },
        "type": "object"
      },
      "version": "0.1"
    },
    {
      "description": "This policy allows to modify the host of a request based on its path.",
      "$schema": "http://apicast.io/policy-v1/schema#manifest#",
      "name": "Upstream policy",
      "configuration": {
        "properties": {
          "rules": {
            "items": {
              "properties": {
                "url": {
                  "type": "string",
                  "description": "new URL in case of match"
                },
                "regex": {
                  "type": "string",
                  "description": "regular expression to be matched"
                }
              },
              "required": [
                "regex",
                "url"
              ],
              "type": "object"
            },
            "description": "list of rules to be applied",
            "type": "array"
          }
        },
        "type": "object"
      },
      "version": "0.1"
    },
    {
      "description": "This policy enables CORS (Cross Origin Resource Sharing) request handling.",
      "$schema": "http://apicast.io/policy-v1/schema#manifest#",
      "name": "CORS policy",
      "configuration": {
        "properties": {
          "allow_headers": {
            "items": {
              "type": "string"
            },
            "description": "Allowed headers",
            "type": "array"
          },
          "allow_credentials": {
            "type": "boolean",
            "description": "Whether the request can be made using credentials"
          },
          "allow_methods": {
            "items": {
              "enum": [
                "GET",
                "HEAD",
                "POST",
                "PUT",
                "DELETE",
                "PATCH",
                "OPTIONS",
                "TRACE",
                "CONNECT"
              ],
              "type": "string"
            },
            "description": "Allowed methods",
            "type": "array"
          },
          "allow_origin": {
            "type": "string",
            "description": "Origins for which the response can be shared with"
          }
        },
        "type": "object"
      },
      "version": "0.1"
    },
    {
      "description": [
        "This policy prints the request back to the client and optionally sets ",
        "a status code."
      ],
      "$schema": "http://apicast.io/policy-v1/schema#manifest#",
      "name": "Echo policy",
      "configuration": {
        "properties": {
          "status": {
            "type": "integer",
            "description": "HTTP status code to be returned"
          },
          "exit": {
            "oneOf": [
              {
                "const": "request",
                "description": "Interrupts the processing of the request."
              },
              {
                "const": "set",
                "description": "Only skips the rewrite phase."
              }
            ],
            "description": "Exit mode",
            "type": "string"
          }
        },
        "type": "object"
      },
      "version": "0.1"
    },
    {
      "description": [
        "This policy allows to include custom headers that will be sent to the ",
        "upstream as well as modify or delete the ones included in the original ",
        "request. Similarly, this policy also allows to add, modify, and delete ",
        "the headers included in the response."
      ],
      "$schema": "http://apicast.io/policy-v1/schema#manifest#",
      "name": "Headers policy",
      "configuration": {
        "definitions": {
          "commands": {
            "items": {
              "properties": {
                "op": {
                  "oneOf": [
                    {
                      "const": "add",
                      "description": "Adds a value to an existing header."
                    },
                    {
                      "const": "set",
                      "description": "Creates the header when not set, replaces its value when set."
                    },
                    {
                      "const": "push",
                      "description": "Creates the header when not set, adds the value when set."
                    }
                  ],
                  "description": "Operation to be applied",
                  "type": "string"
                },
                "header": {
                  "type": "string",
                  "description": "Header to be modified"
                },
                "value": {
                  "type": "string",
                  "description": "Value that will be added, set or pushed in the header"
                }
              },
              "required": [
                "op",
                "header",
                "value"
              ],
              "type": "object"
            },
            "description": "List of operations to apply to the headers",
            "type": "array"
          }
        },
        "properties": {
          "response": {
            "$ref": "#/definitions/commands"
          },
          "request": {
            "$ref": "#/definitions/commands"
          }
        },
        "type": "object"
      },
      "version": "0.1"
    },
    {
      "description": [
        "Configures a cache for the authentication calls against the 3scale ",
        "backend. This policy support three kinds of caching: \n",
        " - Strict: it only caches authorized calls. Denied and failed calls ",
        "invalidate the cache entry.\n",
        " - Resilient: caches authorized and denied calls. Failed calls do not ",
        "invalidate the cache. This allows us to authorize and deny calls ",
        "according to the result of the last request made even when backend is ",
        "down.\n",
        "- Allow: caches authorized and denied calls. When backend is ",
        "unavailable, it will cache an authorization. In practice, this means ",
        "that when backend is down _any_ request will be authorized unless last ",
        "call to backend for that request returned 'deny' (status code = 4xx). ",
        "Make sure to understand the implications of that before using this ",
        "mode. It makes sense only in very specific use cases.\n",
        "- None: disables caching."
      ],
      "$schema": "http://apicast.io/policy-v1/schema#manifest#",
      "name": "Caching policy",
      "configuration": {
        "properties": {
          "caching_type": {
            "oneOf": [
              {
                "const": "resilient",
                "description": "Authorize according to last request when backend is down."
              },
              {
                "const": "strict",
                "description": "It only caches authorized calls."
              },
              {
                "const": "allow",
                "description": "When backend is down, allow everything unless seen before and denied."
              },
              {
                "const": "none",
                "description": "Disables caching."
              }
            ],
            "description": "Caching mode",
            "type": "string"
          }
        },
        "type": "object"
      },
      "version": "0.1"
    },
    {
      "description": [
        "This policy adds support for a very small subset of SOAP. \n",
        "It expects a SOAP action URI in the SOAPAction header or the Content-Type ",
        "header. The SOAPAction header is used in v1.1 of the SOAP standard: ",
        "https://www.w3.org/TR/2000/NOTE-SOAP-20000508/#_Toc478383528 , whereas ",
        "the Content-Type header is used in v1.2 of the SOAP standard: ",
        "https://www.w3.org/TR/soap12-part2/#ActionFeature \n",
        "The SOAPAction URI is matched against the mapping rules defined in the ",
        "policy and calculates a usage based on that so it can be authorized and ",
        "reported against 3scale's backend."
      ],
      "$schema": "http://apicast.io/policy-v1/schema#manifest#",
      "name": "SOAP policy",
      "configuration": {
        "properties": {
          "mapping_rules": {
            "items": {
              "properties": {
                "delta": {
                  "type": "integer",
                  "description": "Value."
                },
                "metric_system_name": {
                  "type": "string",
                  "description": "Metric."
                },
                "pattern": {
                  "type": "string",
                  "description": "Pattern to match against the request."
                }
              },
              "required": [
                "pattern",
                "metric_system_name",
                "delta"
              ],
              "type": "object"
            },
            "description": "Mapping rules.",
            "type": "array"
          }
        },
        "type": "object"
      },
      "version": "0.1"
    }
  ]
}

```

### DIFF
--- a/t/management-policies-endpoint-custom.t
+++ b/t/management-policies-endpoint-custom.t
@@ -1,0 +1,54 @@
+use lib 't';
+use Test::APIcast::Blackbox 'no_plan';
+
+BEGIN {
+    $ENV{TEST_NGINX_MANAGEMENT_SERVER_NAME} = 'management';
+}
+
+env_to_apicast(
+    'APICAST_CONFIGURATION_LOADER' => 'test',
+    'APICAST_POLICY_LOAD_PATH' => "$ENV{PWD}/t/fixtures/policies_endpoint_test/policies"
+);
+
+# Converts what's in the 'expected_json' block and the body to JSON and
+# compares them. Raises and error when they do not match.
+add_response_body_check(sub {
+    my ($block, $body) = @_;
+
+    use JSON;
+    use Data::Compare;
+
+    my $h1 = JSON->new->utf8->decode($body);
+    my $h2 = JSON->new->utf8->decode($block->expected_json);
+    my $c = Data::Compare->new($h1, $h2);
+    if (!$c->Cmp) {
+        bail_out "JSON returned does not match the expected one.";
+    }
+
+});
+
+run_tests();
+
+__DATA__
+
+=== TEST 1: GET /policies
+Check that the endpoint returns the manifests of all the built-in policies plus
+the one in paths specified via config.
+We have a directory with policies in 't/fixtures'. There are 2 policies there,
+a valid one (example1), and one where the version specified in the manifest and
+the one in the policy directory do not match (example2). We need to check that
+the later is not returned.
+--- request
+GET /policies
+--- more_headers
+Host: management
+--- response_headers
+Content-Type: application/json; charset=utf-8
+--- expected_json eval
+use Cwd;
+my $dir = getcwd();
+my $cmd = "script/get_built_in_policy_manifests.sh $dir/t/fixtures/policies_endpoint_test/policies/example1";
+`$cmd`
+--- error_code: 200
+--- no_error_log
+[error]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - CLI is running with proper log level set by `APICAST_LOG_LEVEL` [PR #585](https://github.com/3scale/apicast/pull/585)
 - 3scale configuration (staging/production) can be passed as `-3` or `--channel` on the CLI [PR #590](https://github.com/3scale/apicast/pull/590)
 - APIcast CLI loads environments defined by `APICAST_ENVIRONMENT` variable [PR #590](https://github.com/3scale/apicast/pull/590)
+- Endpoint in management API to retrieve all the JSON manifests of the policies [PR #592](https://github.com/3scale/apicast/pull/592)
 
 ## Fixed
 
--- a/gateway/cpanfile
+++ b/gateway/cpanfile
@@ -1,2 +1,3 @@
 requires 'Test::APIcast', '0.08';
 requires 'Crypt::JWT';
+requires 'Data::Compare';
--- a/gateway/src/apicast/management.lua
+++ b/gateway/src/apicast/management.lua
@@ -8,6 +8,9 @@ local configuration_loader = require('apicast.configuration_loader')
 local inspect = require('inspect')
 local resolver_cache = require('resty.resolver.cache')
 local env = require('resty.env')
+local policy_manifests_loader = require('apicast.policy_manifests_loader')
+
+local setmetatable = setmetatable
 
 local live = { status = 'live', success = true }
 
@@ -124,6 +127,12 @@ function _M.info()
   })
 end
 
+function _M.get_all_policy_manifests()
+  local manifests = policy_manifests_loader.get_all()
+  setmetatable(manifests, cjson.array_mt)
+  return json_response({ policies = manifests })
+end
+
 local routes = {}
 
 function routes.disabled(r)
@@ -134,6 +143,8 @@ function routes.status(r)
   r:get('/status/info', _M.info)
   r:get('/status/ready', _M.ready)
   r:get('/status/live', _M.live)
+
+  routes.policies(r)
 end
 
 function routes.debug(r)
@@ -147,6 +158,12 @@ function routes.debug(r)
   r:get('/dns/cache', _M.dns_cache)
 
   r:post('/boot', _M.boot)
+
+  routes.policies(r)
+end
+
+function routes.policies(r)
+  r:get('/policies/', _M.get_all_policy_manifests)
 end
 
 function _M.router()
--- a/gateway/src/apicast/policy_manifests_loader.lua
+++ b/gateway/src/apicast/policy_manifests_loader.lua
@@ -1,0 +1,99 @@
+--- Policy manifests loader
+-- Finds the manifests for the builtin policies and the other ones loaded.
+
+local pl_file = require('pl.file')
+local pl_dir = require('pl.dir')
+local pl_path = require('pl.path')
+local cjson = require('cjson')
+local format = string.format
+local ipairs = ipairs
+local insert = table.insert
+local policy_loader = require('apicast.policy_loader')
+
+local _M = {}
+
+local builtin_policy_load_path = policy_loader.builtin_policy_load_path
+local policy_load_paths = policy_loader.policy_load_paths
+
+local policy_manifest_name = 'apicast-policy.json'
+
+local function dir_iter(dir)
+  return ipairs(pl_dir.getdirectories(dir))
+end
+
+-- Returns the manifests for all the built-in policies.
+local function all_builtin_policy_manifests()
+  local manifests = {}
+
+  for _, policy_dir in dir_iter(builtin_policy_load_path()) do
+    local manifest_file = format('%s/%s', policy_dir, policy_manifest_name)
+    local manifest = pl_file.read(manifest_file)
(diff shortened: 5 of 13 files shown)


gateway/src/apicast/policy_schemas_loader.lua
```diff
@@ -0,0 +1,99 @@
...
+end
+
+-- Returns all the non-built-in schemas from the paths specified by the user.
+-- These paths always follow the same pattern. It contains a directory for each
+-- policy, and each of those contain a directory for each version of that
+-- policy. The json schema is in that 'version' directory.
+local function all_loaded_policy_schemas()
+  local schemas = {}
+
+  for _, load_path in ipairs(policy_load_paths) do
```
mikz: `policy_load_paths` changes in https://github.com/3scale/apicast/pull/593 to a function so it is refreshed with every call

gateway/src/apicast/policy_schemas_loader.lua
```diff
@@ -0,0 +1,99 @@
...
+    if pl_path.exists(load_path) then
+      local policy_dirs = pl_dir.getdirectories(load_path)
+      for _, policy_dir in ipairs(policy_dirs) do
+        local version_dirs = pl_dir.getdirectories(policy_dir)
+        for _, version_dir in ipairs(version_dirs) do
```
mikz: I think it could be a bit nicer if we would reduce the number of variables by defining something like `dir_iter(dir)` that would do `return ipairs(pl_dir.getdirectories(dir))`

t/management-policies-endpoint-custom.t
```diff
@@ -0,0 +1,55 @@
...
+GET /policies
+--- more_headers
+Host: management
+--- response_headers
+Content-Type: application/json; charset=utf-8
+--- expected_json eval
+use Cwd;
+my $dir = getcwd();
+my $cmd = "script/get_built_in_policy_schemas.sh $dir/t/fixtures/policies_endpoint_test/policies/example1";
+`$cmd`
```
mikz: 🐗 this is quite something :) Born Perl Monger.

gateway/src/apicast/policy_schemas_loader.lua
mikz: Extracted into https://github.com/3scale/apicast/pull/595

gateway/src/apicast/policy_schemas_loader.lua
davidor: Thanks. I changed it. Looks better :ok_hand:

gateway/src/apicast/policy_schemas_loader.lua
davidor: Changed.

gateway/src/apicast/policy_schemas_loader.lua
```diff
@@ -0,0 +1,99 @@
...
+  return ipairs(pl_dir.getdirectories(dir))
+end
+
+-- Returns the schemas for all the built-in policies.
+local function all_builtin_policy_schemas()
+  local schemas = {}
+
+  for _, policy_dir in dir_iter(builtin_policy_load_path()) do
+    local schema_file = format('%s/%s', policy_dir, policy_schema_name)
+    local schema = pl_file.read(schema_file)
```
mikz: Just a nitpick. It is not a schema, but a manifest, right?

davidor: :+1:

t/management-policies-endpoint-custom.t
```diff
@@ -0,0 +1,55 @@
...
+    my $h1 = JSON->new->utf8->decode($body);
+    my $h2 = JSON->new->utf8->decode($block->expected_json);
+    my $c = Data::Compare->new($h1, $h2);
+    if (!$c->Cmp) {
+        bail_out "JSON returned does not match the expected one.";
+    }
+
+});
+
+repeat_each(1);
```
mikz: Any reason to repeat just once instead of the default 2 ?

davidor: I always forget about this and copy it from the wrong place. I'll just delete this as Test::APIcast uses repeat_each(2) by default when not set.

mikz: Perils of copy&pasting ☠️

mikz:
:shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit:

hallelujah:
Thanks! @davidor ! 🚀

PR #596, opened 2018-02-15
[sandbox] environment policies have the same env as policy loader

mikz:
* Extracts sandbox from policy loader to own module.
* Load environment configurations with the same sandbox env as policies.
* Expose `arg` in sandbox.
* Expose also `cli` as an alias to `arg` when loading environment configurations.
  So policies can detect if they are being executed on the CLI
  and decide to do something different like not loading themselves.
  
  Because some might require shared dict that is not available.

### DIFF
--- a/gateway/src/apicast/cli/environment.lua
+++ b/gateway/src/apicast/cli/environment.lua
@@ -147,7 +147,7 @@ function _M:add(env)
     end
 
     local config = loadfile(path, 't', {
-        print = print, inspect = require('inspect'), context = self._context,
+        print = print, inspect = require('inspect'), context = self._context, arg = arg, cli = arg,
         tonumber = tonumber, tostring = tostring, os = { getenv = resty_env.value },
         pcall = pcall, require = require, assert = assert, error = error,
     })
--- a/gateway/src/apicast/cli/environment.lua
+++ b/gateway/src/apicast/cli/environment.lua
@@ -6,13 +6,12 @@
 local pl_path = require('pl.path')
 local resty_env = require('resty.env')
 local linked_list = require('apicast.linked_list')
+local sandbox = require('resty.sandbox')
 local util = require('apicast.util')
 local setmetatable = setmetatable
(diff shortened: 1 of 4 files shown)


gateway/src/resty/sandbox.lua
```diff
@@ -0,0 +1,255 @@
+--- Sandbox
```
davidor: Nice refactor :+1: 
I think this module does not contain any apicast-specific code so it could potentially be extracted into its own library, and that's why it's in the resty directory. However, there are some comments that mention apicast and policies. I think we should get rid of them. For example, the description of the module mentions APIcast, and the comment for `local env` mentions policies.

mikz: Yep. I did not want to bother much with it now, but rather when it is extracted. Would that be ok?

davidor: Sure.

gateway/src/apicast/cli/environment.lua
```diff
@@ -146,11 +145,12 @@ function _M:add(env)
         return nil, 'no configuration found'
     end
 
-    local config = loadfile(path, 't', {
-        print = print, inspect = require('inspect'), context = self._context, arg = arg, cli = arg,
-        tonumber = tonumber, tostring = tostring, os = { getenv = resty_env.value },
-        pcall = pcall, require = require, assert = assert, error = error,
-    })
+    local box = sandbox.new()
```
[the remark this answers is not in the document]
mikz: Yeah. We don't **need** one. But is is quite convenient way of adding some extra env and not using global variables. Will add it there.

gateway/src/resty/sandbox.lua
mikz: Done.

gateway/src/apicast/cli/environment.lua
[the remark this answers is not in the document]
mikz: Fixed.

PR #598, opened 2018-02-15
[cli] load openresty libraries when rover is active

mikz:
Rover would remove all shared paths, including paths
where openresty distributes own code.

So try go get those paths and set them back after loading rover.

It is not really possible to write a test because APIcast crashes in the CLI (before generating nginx config) and we have no CLI tests.


gateway/bin/apicast
```diff
@@ -109,3 +109,61 @@ for my $inc ($apicast_src, 'src') {
 my @args = ('resty', @resty_args, $bin, @ARGV);
 
 exec '/usr/bin/env', @args;
+
+__DATA__
```
mikz: This is the easiest way to distribute the lua helper.

gateway/bin/apicast
```diff
@@ -109,3 +109,61 @@ for my $inc ($apicast_src, 'src') {
...
+#!/usr/bin/env resty
```
davidor: I think we need an explanation here about why this is needed. Like the one included in the PR description.

mikz: Added more comments. Most of them are in the lua helper code.

PR #601, opened 2018-02-16
Add JSON manifest for APIcast policy

davidor:
Without a manifest, it will not show up in the policies endpoint.

### DIFF
--- a/gateway/src/apicast/policy/apicast/apicast-policy.json
+++ b/gateway/src/apicast/policy/apicast/apicast-policy.json
@@ -1,0 +1,10 @@
+{
+  "$schema": "http://apicast.io/policy-v1/schema#manifest#",
+  "name": "APIcast policy",
+  "description": "Main functionality of APIcast.",
+  "version": "0.1",
+  "configuration": {
+    "type": "object",
+    "properties": { }
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Added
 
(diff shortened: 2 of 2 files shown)


gateway/src/apicast/policy/apicast/apicast-policy.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "http://apicast.io/policy-v1/schema#manifest#",
+  "name": "Apicast policy",
+  "description": "Main functionality of APIcast.",
+  "version": "0.1",
```
davidor: This can be tricky @mikz .

mikz: I think we can have there`builtin`. The loaded already expects that for all builtin policies.

davidor: I'm not sure about that. It's not a problem for now, but what if in the next release we introduce a new config param for this policy? We'd need to somehow distinguish the 2 versions.

mikz: I mean it seriously needs to be builtin because those have different load path. It will not work at all with normal version string unless we move all policies to different folders. 

I think it is fine because they are not released separately. We will have to maintain backwards compatibility. 

Why we would need to distinguish two versions ? There is going to be just one version of APIcast policy. We could add some extra property to the policy API to show version of APIcast itself.

mikz: I mean it seriously needs to be builtin because those have different load path. It will not work at all with normal version string unless we move all policies to different folders. 

I think it is fine because they are not released separately. We will have to maintain backwards compatibility. 

Why we would need to distinguish two versions ? There is going to be just one version of APIcast policy. We could add some extra property to the policy API to show version of APIcast itself.

PR #603, opened 2018-02-19
An APIcast policy that enables SSL/TLS client certificate authentication

nmasse-itix:
Hi team,

This pull request is a preliminary support for SSL/TLS client certificate authentication that fits the new policy system. 

# Customer needs

## Need 1: handle COTS software 

Some customers needs to handle "Commercial-off-the-shelf" (COTS) applications within 3scale. 
They need to handle access control, quotas and track usages of those apps just like any normal app within 3scale. 

Usually, those COTS application can be configured with an alternate URL that can be set to the APIcast gateway but there is usually no way to add something as HTTP headers or in the Query String. 

However, most of these apps can be configured to use a TLS Key Store and authenticate using TLS client certificate authentication. 

## Need 2: have a higher level of security and authenticate the client application using Certificates

Some customers have requirements to use SSL/TLS client certificate authentication to authenticate the client application. This is a strong requirement in the EU (and especially in France where the ANSSI strongly pushes the usage of certificates everywhere). 

# Design choices

In order to be manageable, the SSL/TLS client certificate authentication needs to enforced at the APIcast level, but access control and the whole management needs to remain in the 3scale Admin Portal. 

So, I piggybacked the SSL/TLS client certificate authentication onto the existing API Key support of 3scale. Namely, a successful client certificate authentication is mapped to an API Key that has the following value: `certificate-<SHA1>` where `<SHA1>` is the SHA1 hash of the Common Name(s) of the client certificate concatenated with the service_id. 

# Current state

 - make the policy configurable (which information contributes to the identification data)
 - unit testing
 - discuss and validate the design more thoroughly 

I also had issues during the development with `Rover` so I had to fallback to `luarocks` but the final implementation should use `Rover` instead. 

Many thanks for your help !

Nicolas


nmasse-itix:
cc @mikz, @andrewdavidmackenzie, @lucamaf, @kevprice83

gateway/apicast-0.1-0.rockspec
```diff
@@ -12,6 +12,7 @@ dependencies = {
   'liquid',
   'argparse',
   'penlight',
+  'sha1'
```
mikz: There should be no need. There is [ngx.sha1_bin](https://github.com/openresty/lua-nginx-module/blob/master/README.markdown#ngxsha1_bin) and [`resty.sha1`](https://github.com/openresty/lua-resty-string#synopsis).

gateway/apicast.d/ssl.conf
```diff
@@ -0,0 +1,9 @@
+listen 8443 ssl;
+
+ssl_certificate {{ prefix }}/conf/serving-cert/tls.crt;
```
mikz: Unfortunately those files can't be templated by liquid yet as they are also included in plain nginx in our test suite. We are making a workaround in https://github.com/3scale/apicast/pull/610 so we could apply something similar here.

gateway/src/custom/policy/tls_client_certificate_authentication.lua
```diff
@@ -0,0 +1,83 @@
+local policy = require('apicast.policy')
+local _M = policy.new('Authenticate a client using his API Key or his SSL/TLS client certificate')
+
+local openssl = require('openssl')
```
mikz: Depending on C library would be troublesome so the best approach is to use FFI like: https://github.com/SkyLothar/lua-resty-jwt/blob/master/lib/resty/evp.lua

mikz:
@nmasse-itix I think this would make sense to me with some small details.

I belive this policy should do one thing and one thing only:

extract the `commonName` and use it as `user-key` or `app-id` or whatever is configured in the service.

It should not check and enforce the auth scheme is set to "headers" or verify the service is loaded. There are other policies that do that.
This policy should complement the default implementation and not try to override it. 
The same way OIDC extracts app_id/key from the JWT this could extract it from the cert.
And I'd imagine there could be some option to pass a secret key and do HMAC to create app_key from the commonName (or any other configured field).

If we could provide `ssl_verify_client` set to `optional` or `optional_no_ca` in the default configuration then all the work of verifying the cert could be delegated to the policy.

And can't be the common name from the client cert be extracted from nginx variable like [$ssl_client_s_dn](http://nginx.org/en/docs/http/ngx_http_ssl_module.html#var_ssl_client_s_dn)?

nmasse-itix:
Hi @mikz ! Thanks for the feedback !

I initially used the commonName and the service id to identify the app to have different user-key/app-id if the app needs to access multiple services. I thought we might have issues since the user-key/app-id is used as an identifier to find the app. I double checked the API doc for AuthRep and it seems the service id is also passed along with the user-key / app-id. Do you confirm that different apps that uses different services can have the same user-key / app-id ? 

I agree on the idea that the policy should be minimal and do only one thing. 

Regarding the customer need [1], I wanted to let the customer have a mix of usual user-keys and certificates. But for need [2] it would definitely makes more sense to only set the user-key/app-id with the certificate chosen identification data. 

Also for need [1], since I allowed a mix of usual user-keys and certificates, I had to make sure that someone would not pass an HTTP header with a chosen user-key identifying another application (since certificate data are considered public, one could generate the user-key / app-id). That's why I had to check for that case. 

So, to support use case [1] and [2], we could split this policy in two:
 A. one that checks if the client authenticated with a certificate and if yes, use the configured identification data to generate and set the credentials
 B. one that does surface checks on the provided credentials (ensure that user-key / app-id matches a certain pattern)

For use case [1], we would need first policy B followed by policy A. 
For use case [2], we would need first a policy that enforces the uses of a TLS client certificate (`require_tls_certificate.lua` in #607) followed by policy A

Finally, I used the openssl library to properly decode the subject DN of the certificate since parsing DN as a string is very complicated to get right and error prone. There are a lot of corner cases / quoting / encoding issues, etc. 

And I wanted to allow the customer to compute the identification data by himself (so that the custom user-key / app-id could be computed externally and set automatically through the 3scale APIs) so that means having a reliable mechanism for extracting data from the certificate. 

So if I try to sum up the remaining work:
- split the policy in two as described earlier
- make them configurable
- handle other credential types (app-id / user-key)
- handle other credential locations (query-string / headers)
- use the suggested libraries (resty.sha1 / ffi)
- remove the service id from the identification data

Last question: do we set the credentials by manipulating the request (adding headers / query_string) as I did or do you planned to have more formal representation of the credentials (set a context variable ?) that could be read/written by policies ?

Thanks again for your feedback.

mikz:
@nmasse-itix `user_id`/`app_id` do not have to be unique across services in AMP deployment. This is not true on SaaS where they have to be unique. But at some point we could lift this constraint everywhere.

Definitely get that for [1] you need a mix of certs and tokens. And for [2] you don't want to provide fallback but rather cert be the only auth option. This could just a configuration option.

The detail how to exactly provide credentials fallback / override is a task for further exploration (policy context, setting header/param, ...). But more on that later.

I understand the concern of verifying the commonName against the user_key/app_id. 
Basically none of the public information in the certificate can be used as user_key/app_key.
I guess what we want to do is the app_id+app_key auth:
If the client certificate is valid take the commonName/serialNumber and use it as app_id and authenticate with 3scale backend **without app_key**.
If that would not be possible then we could try to use the OAuth2 API that allows to lookup app_key for any app_id. Then APIcast would cache the app_key and use for authrep.
Otherwise require app_id+app_key in the request.

I think this could be just simple configuration option that would terminate the request if the client certificate is not valid (for case [2]). We are still experimenting with what is better to split and what is better to keep together, but I guess the rule could be: "can it be used standalone". As for the "B" policy I think it should not be needed at all. 

I have no preference whether to use OpenSSL as FFI or write pure Lua [RFC 2253](https://tools.ietf.org/html/rfc2253) parser. If going for the pure Lua there is https://github.com/pygy/LuLPeg that might be help and some examples how to use lpeg: https://github.com/daurnimator/lpeg_patterns. Do you have any example of a cert with complicated metadata? To me it seems similar to other formats and might not be that hard to parse in Lua.

And how to define the credentials. I'm not sure. I guess over time we will definitely want to use the context. But maybe for now it might be easier to change the request. Right now it actually might be pretty hard even to just replace extracting credentials.  Easiest option might be to override `service.extract_credentials`.  Suggestion welcome. This will require some prototyping various versions and then deciding what is the best approach.
For prototyping I'd go with changing https://github.com/3scale/apicast/blob/1bfd8b791ca2bcf85ddaf481e8a9f42b6285445a/gateway/src/apicast/proxy.lua#L268
into:
```lua
local credentials, err
if context.credentials then
  credentials = context.credentials
else
  credentials, err = service:extract_credentials()
end
```

Thank you so much! This is great 👍

mikz:
Here is FFI way to get commonName from PEM cert:

```lua
local ssl = require('ngx.ssl')
local pem = [[-----BEGIN CERTIFICATE-----
MIIBRzCB7gIJAPHi8uNGM8wDMAoGCCqGSM49BAMCMCwxFjAUBgNVBAoMDVRlc3Q6
OkFQSWNhc3QxEjAQBgNVBAMMCWxvY2FsaG9zdDAeFw0xODA2MDUwOTQ0MjRaFw0y
ODA2MDIwOTQ0MjRaMCwxFjAUBgNVBAoMDVRlc3Q6OkFQSWNhc3QxEjAQBgNVBAMM
CWxvY2FsaG9zdDBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABI3IZUvpJsaQbiLy
/yfthJDd/+BIaKzAbgMAimth4ePOi3a/YICwsHyq6sBxbgvMeTwxNJIHpe3td4tB
VZ5Wr10wCgYIKoZIzj0EAwIDSAAwRQIhAPRkfbxowt0H7p5xZYpwoMKanUXz9eKQ
0sGkOw+TqqGXAiAMKJRqtjnCF2LIjGygHG6BlgjM4NgIMDHteZPEr4qEmw==
-----END CERTIFICATE-----]]

local chain = ssl.parse_pem_cert(pem)

local ffi = require "ffi"

ffi.load("ssl", true)
ffi.load("crypto", true)

ffi.cdef([[
typedef struct stack_st OPENSSL_STACK;

int OPENSSL_sk_num(const OPENSSL_STACK *);
void *OPENSSL_sk_value(const OPENSSL_STACK *, int);

typedef void X509_NAME;
typedef void X509_NAME_ENTRY;
typedef void X509;
typedef void ASN1_STRING;

X509_NAME *X509_get_subject_name(const X509 *x);
int X509_NAME_get_index_by_NID(X509_NAME *name, int nid, int lastpos);
 
int X509_NAME_entry_count(const X509_NAME *name);
X509_NAME_ENTRY *X509_NAME_get_entry(const X509_NAME *name, int loc);
ASN1_STRING * X509_NAME_ENTRY_get_data(const X509_NAME_ENTRY *ne);
unsigned char * ASN1_STRING_data(ASN1_STRING *x);

]])

local C = ffi.C
local n = C.OPENSSL_sk_num(chain)

print('certifictes in chain: ', n)

local NID_commonName = 13

for i=0, n-1 do
	local crt = C.OPENSSL_sk_value(chain, i)
	local nm = C.X509_get_subject_name(crt)

	local lastpos = -1;
	local entry, data, str

 	while true do
		lastpos = C.X509_NAME_get_index_by_NID(nm, NID_commonName, lastpos);
		if lastpos == -1 then break end
		entry = C.X509_NAME_get_entry(nm, lastpos);
		data = C.X509_NAME_ENTRY_get_data(entry);
		str = C.ASN1_STRING_data(data);
		print('str: ', ffi.string(str))
	end
end
```

BTW https://zakird.com/2013/10/13/certificate-parsing-with-openssl is very nice article and has plenty of examples of common operations.

mikz:
@nmasse-itix would be good to make a summary into a todo list of what we want to achieve here and tick off items that are done.

- [ ] Listen on HTTPS (https://github.com/3scale/apicast/pull/622)
- [ ] Configuration options for client certificates 
- [ ] Define more steps

mikz:
A policy that enables client certificate validation (#966) has been merged.

It is now possible to configure a whitelist of CAs and individual certificates.

It does not validate the CommonName or any other properties of the certificate other than it matches the whitelist of was issued by the CA in the whitelist.

PR #604, opened 2018-02-19
OpenID Connect policies

nmasse-itix:
Hi team,

Here are two policies for the new APIcast policy system that helps to handle some of the OpenID Connect use cases. 

# Needs

Some customer would like to use the information provided in the OpenID Connect token to base their decisions or to enrich the HTTP request. 

Some use cases could be : 
 - let the request if only if the OIDC token has a proper scope (or realm_access if using RH-SSO)
 - add HTTP headers to the request with values from the OIDC token so that the API Backend can know who is the user and from which application the request comes from. 

# Implementation

There are two policies, one for decoding the OIDC token, one for exploding its values as HTTP headers. 

The one that decodes the OIDC token and store the payload as a context value is the most questionable since it would have been better done directly by APIcast during the token validation. 

I also added a sample apicast configuration files with the required OIDC configuration fields since it took me quite some time to figure out the proper format. 

# Remaining work

 - unit testing
 - discuss if the token decoding is left in separate policy or performed directly by APIcast during the token validation

Many thanks for your help !

Nicolas

### DIFF
--- a/examples/configuration/oidc.json
+++ b/examples/configuration/oidc.json
@@ -1,0 +1,53 @@
+{
+  "services": [
+    {
+      "id": 123,
+      "backend_version": "oauth",
+      "backend_authentication_type": "service_token",
+      "backend_authentication_value": "456",
+      "proxy": {
+        "api_backend": "http://echo-api.3scale.net",
+        "hosts": [
+          "localhost",
+          "127.0.0.1"
+        ],
+        "backend": {
+          "endpoint": "https://su1.3scale.net"
+        },
+        "oidc_issuer_endpoint": "https://sso.app.openshift.test/auth/realms/3scale",
+        "authentication_method": "oidc",
+        "proxy_rules": [
+          {
+            "http_method": "GET",
+            "pattern": "/",
+            "metric_system_name": "hits",
+            "delta": 1
+          },
+          {
+            "http_method": "POST",
+            "pattern": "/",
+            "metric_system_name": "hits",
+            "delta": 1
+          },
+          {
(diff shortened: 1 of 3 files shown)


davidor:
@nmasse-itix : I think that the latest PRs that we merged can help you implement the same functionality, at least partially, without implementing new policies.

In #716 we added support for liquid templates in the headers policy, and in #718 , we exposed the decoded jwt in the policies context. This means that in the headers policy config you can have values like `{{ jwt.aud }}`.

nmasse-itix:
Excellent ! 👍

mikz:
@nmasse-itix we can close this as it can be implemented as https://github.com/3scale/apicast/pull/716 ?

nmasse-itix:
Hi @mikz ! 

That's good news ! There was two features in this PR: 
 1. let the request if only if the OIDC token has a proper scope (or realm_access if using RH-SSO)
 2. add HTTP headers to the request with values from the OIDC token so that the API Backend can know who is the user and from which application the request comes from.

I understood that 2. is implemented, that good. 👍 How would you like to implement the remaining part (1.) : "let the request if only if the OIDC token has a proper scope (or realm_access if using RH-SSO)" ?

Do you want me to open a new pull request only for this point ?

mikz:
@nmasse-itix I think the first part is in some other PR. 
This one really just looks like passing JWT as headers.

nmasse-itix:
Yes, indeed. You are right (and I'm tired 😁). 

You can safely close this PR. 👍

PR #606, opened 2018-02-19
SSL/TLS Client certificate authentication with OpenID Connect

nmasse-itix:
Hi team,

Here is a pull request for some policies / customization of APIcast that implements SSL/TLS client certificate authentication in combination with the OpenID Connect authentication. When possible, the customization uses the new policy system. 

# Needs

In the banking industry, a new regulation emerged and is being implemented : PSD2. This regulation mandates that banks are forced to open up and must expose APIs to allow payments from third parties and to allow third parties to fetch information of the user accounts. 

To use the bank's APIs, the current regulation as implemented here, requires an OpenID Connect token from the end-user as well as a SSL/TLS client certificate authentication from the third party. 

Since the OpenID Connect token also identifies the third party, there must be a binding between the SSL/TLS client certificate authentication and the OpenID Connect token. 

# Design

I described it in details in the [CUSTOM.md file](https://github.com/nmasse-itix/3scale-apicast/blob/tls-jwt-client-authentication/CUSTOM.md). 

# Implementation

There are two endpoints exposed:
- one to register a client certificate with the OIDC client_id
- one to validate during the token request that the SSL/TLS client certificate is the same that signed the JWT client assertion. 

With perspective, I think the first one would be better implemented as a Camel route in Fuse and exposed in 3scale as a regular API. 

# Current state

 - unit testing
 - discuss and challenge the design
 - implement caching (currently there is none so the performances are bad)

Many thanks for your help !

Nicolas


mikz:
HTTPS support in APIcast is something we've been thinking about a for a while. Lets take that out of the scope and just assume it is there.

Looks like we agree that this should not be scope of the gateway.
We are trying to not expose any internal endpoints on the gateway. First because that could be introducing some internal state to the gateway which should be stateless. And second because it introduces another port that needs to be exposed or could clash with the API endpoints.
I seriously think the gateway should not have write access to the IDP. That is pretty similar to https://github.com/3scale/apicast/pull/435. 

In this case I think this should be the IDP responsibility no? Using signed JWT like described in http://www.keycloak.org/docs/3.2/securing_apps/topics/oidc/java/client-authentication.html.
I'd expect that exposes enough information about the certificate in the JWT so it could be matched when it is sent to APIcast.
It should be just matter of exposing the certificate attributes in the JWT. Possibly we could try to verify that certificate against the IDP. But bonding the certificate with the JWT should not be concern of the gateway.

tmogi001:
We are also considering to implement PSD2 security, and interested in the discussion. In our understanding, PSD2 does not have specific technical requirement, but one of the most promising security requirement related to PSD2 is OpenID Foundation’s FAPI. http://openid.net/wg/fapi/
What kind of specification do you mean by “PSD2”?  FAPI? Or others?

FYI: Also we are implementing other FAPI requirements in Keycloak,
https://issues.jboss.org/browse/KEYCLOAK-6767
If someone in 3scale is interested, help(and discussion) is very appreciated.
Also we implemented one of FAPI requirements (token binding based on MTLS), and sent PR to Keycloak.
https://github.com/keycloak/keycloak/pull/5083/
Above may be related this PR.

eloycoto:
This is already done by TLS policy. closing this one.

PR #607, opened 2018-02-19
Complimentary Policies that are needed for PSD2 APIs

nmasse-itix:
Hi team,

Here is a pull request for some policies / customization of APIcast that implements SSL/TLS client certificate authentication in combination with the OpenID Connect authentication. When possible, the customization uses the new policy system.

# Needs

In the banking industry, a new regulation emerged and is being implemented : PSD2. This regulation mandates that banks are forced to open up and must expose APIs to allow payments from third parties and to allow third parties to fetch information of the user accounts.

To use the bank's APIs, the current regulation as implemented here, requires an OpenID Connect token from the end-user as well as a SSL/TLS client certificate authentication from the third party.

Since the OpenID Connect token also identifies the third party, there must be a binding between the SSL/TLS client certificate authentication and the OpenID Connect token.

Also the current implementation of this regulation requires to use OAuth scopes to restrict accesses to the Bank's APIs. 

# Implementation

First, of all, I'm very sorry for the poor presentation of this pull request. It is based on a previous branch for which I also submitted a pull request. So you might have a feeling of déjà-vu. 

The important bits are:
- `gateway/src/custom/policy/require_tls_certificate.lua`: enforces the use of a SSL/TLS client certificate to authenticate
- `gateway/src/custom/policy/bind_tls_certificate_with_oidc_token.lua`: this policy makes sure the certificate used during the SSL/TLS client certificate authentication is the one registered in RH-SSO for the used client_id. 
- `gateway/src/custom/policy/check_oidc_roles.lua`: this policy makes sure that the user has the required roles in his OIDC token

# Current state
 - unit testing
 - discuss and challenge the design
 - implement caching (currently there is none so the performances are bad)

Many thanks for your help !

Nicolas

### DIFF
--- a/examples/configuration/oidc.json
+++ b/examples/configuration/oidc.json
@@ -1,0 +1,53 @@
+{
+  "services": [
+    {
+      "id": 123,
+      "backend_version": "oauth",
+      "backend_authentication_type": "service_token",
+      "backend_authentication_value": "456",
+      "proxy": {
+        "api_backend": "http://echo-api.3scale.net",
+        "hosts": [
+          "localhost",
+          "127.0.0.1"
+        ],
+        "backend": {
+          "endpoint": "https://su1.3scale.net"
+        },
+        "oidc_issuer_endpoint": "https://sso.app.openshift.test/auth/realms/3scale",
+        "authentication_method": "oidc",
+        "proxy_rules": [
+          {
+            "http_method": "GET",
+            "pattern": "/",
+            "metric_system_name": "hits",
+            "delta": 1
+          },
+          {
+            "http_method": "POST",
+            "pattern": "/",
+            "metric_system_name": "hits",
+            "delta": 1
+          },
+          {
+            "http_method": "PUT",
+            "pattern": "/",
+            "metric_system_name": "hits",
+            "delta": 1
+          }
+        ]
+      }
+    }
+  ],
+  "oidc": [
+    {
+      "issuer": "https://sso.app.openshift.test/auth/realms/3scale",
+      "config": {
+        "public_key": "YOUR-REALM-PUBLIC-KEY",
+        "openid": {
+          "id_token_signing_alg_values_supported": [ "RS256" ]
+        }
+      }
+    }
+  ]
+}
--- a/gateway/.s2i/bin/assemble
+++ b/gateway/.s2i/bin/assemble
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+echo "---> Calling the openresty-s2i assemble script..."
+/usr/libexec/s2i/assemble
+
+echo "---> Installing dependencies (custom fix)..."
+rover install --roverfile /tmp/src/Roverfile --path .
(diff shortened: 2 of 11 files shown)


mikz:
Lets tackle this issue separately from https://github.com/3scale/apicast/pull/606.

What I differences I see here are:

* Passing JWT fields to the upstream service by HTTP Headers
* Verify roles in JWT match metadata in policy configuration  

Regarding passing the JWT fields as headers. I doubt its usefulness apart from debugging/development. JWT can be decoded by the upstream API. Could you share why that is needed ? Or was it for debugging / development purposes? In that case it might be better to just use `ngx.log`.

And regarding roles/scopes there is a ongoing discussion how that should work in [THREESCALE-311]()https://issues.jboss.org/browse/THREESCALE-311). 
There is few interesting links around that topic: http://www.keycloak.org/docs/3.3/authorization_services/topics/overview/architecture.html
https://developer.okta.com/authentication-guide/implementing-authentication/set-up-authz-server.html#create-rules-for-each-access-policy

I'm not really sure what will come out of it, but it is considered for the next release. Would that cover your use case?
Could you share some example JWT that has those roles defined?

nmasse-itix:
Hi @mikz ! Sorry for the delay. 

Passing JWT fields to the upstream service by HTTP Headers is a common feature requested by most customers. 

But you are right when you say that the token can be decoded on the API backend. Namely, you can pickup a JWT library (or implement a rough decoding by yourself) and get those values. 

But from the customer point of view, it is has the following *perceived* drawbacks:
 - the API Backend needs to be modified which implies costs and risks
 - you need to find an off-the-shelf JWT library (or implement a rough decoding) for your programming language. This is easy for most recent programming languages but can be tedious when the API Backend is a mainframe (AS/400, system Z, etc.), an old unsupported Unix (Solaris), or when your API Backend is developed as shell scripts (CGI scripts) or SQL Stored Procedures. 
 - every running code is a bit more load on the API Backend which is fine when the server is a common x86 server but when the API Backend is a big mainframe priced by MIPS (the more you use the CPU, the more you pay), it ends up with higher licensing costs. 

Whereas picking values from HTTP Headers is already implemented on most applications found on customer site. In fact, the Load Balancers on customer site already use HTTP Headers to add information to the request, so customers are used to this working. 

I will try to formalize the Access Control part and comment on the Jira Ticket. 

Thanks for your help !

mikz:
@nmasse-itix makes sense. Thanks for taking the time.

We plan to add templating capabilities to the headers policy, so it could be used to extract fields from JWT.

For example it should be possible to set header "JWT-aud" with template string `{{ jwt.aud }}` and that would evaluate within context of the current request.
Surely we could have own policy with some preset jwt settings, but it seems not worth doing when it could be achieved by different policy in the future.

y-tabata:
@mikz We also want to parse jwt and pass scopes and user info to the backend by HTTP headers. Could you please tell us the plan to add templating capacilities?

mikz:
@y-tabata the plan is to have templating in the next release (~3 months), but it might be one of the first things we do.

y-tabata:
@mikz Sounds good!

eloycoto:
Already covered.

PR #610, opened 2018-02-19
[THREESCALE-328] Allow to set a client SSL cert for upstream

davidor:
This PR introduces 2 new ENVs (`APICAST_PROXY_HTTPS_CERTIFICATE_KEY ` and `APICAST_PROXY_HTTPS_CERTIFICATE`), they are used to set a client SSL certificate using the `proxy_ssl_certificate` and the `proxy_ssl_certificate_key` NGINX directives.

As a limitation, this certificate needs to be valid for all the services configured.

### DIFF
--- a/examples/ssl-verification/badssl.json
+++ b/examples/ssl-verification/badssl.json
@@ -35,6 +35,28 @@
           }
         ]
       }
+    },
+    {
+      "id": "client-cert",
+      "backend_version": 1,
+      "proxy": {
+        "api_backend": "https://client.badssl.com",
+        "backend": {
+          "endpoint": "https://echo-api.3scale.net"
+        },
+        "hosts": [
+          "client.badssl.com",
+          "localhost"
+        ],
+        "proxy_rules": [
+          {
+            "http_method": "GET",
+            "pattern": "/",
+            "metric_system_name": "hits",
+            "delta": 1
+          }
+        ]
+      }
     }
   ]
 }
--- a/gateway/conf.d/apicast.conf
+++ b/gateway/conf.d/apicast.conf
@@ -75,6 +75,34 @@ location @upstream {
   proxy_set_header X-3scale-debug "";
   proxy_set_header Connection "";
 
+  # This is a bit tricky. It uses liquid to set a SSL client certificate. In
+  # NGINX, all this is not executed as it is commented with '#'. However, in
+  # Liquid, all this will be evaluated. As a result, the following directives
+  # are set optionally: proxy_ssl_certificate, proxy_ssl_certificate_key,
+  # proxy_ssl_session_reuse, and proxy_ssl_password_file.
+
+  # {% if proxy_ssl_certificate != empty and proxy_ssl_certificate_key != empty %}
+  #   {% capture proxy_ssl %}
+  #{#}   proxy_ssl_certificate {{ proxy_ssl_certificate }};
+  #{#}   proxy_ssl_certificate_key {{ proxy_ssl_certificate_key }};
+  #   {% endcapture %}
+  #   {{ proxy_ssl | replace: "#{#}", "" }}
+  #
+  #   {% if proxy_ssl_password_file != empty %}
+  #     {% capture proxy_ssl %}
+  #{#}   proxy_ssl_password_file {{ proxy_ssl_password_file }}
+  #     {% endcapture %}
+  #   {{ proxy_ssl | replace: "#{#}", "" }}
+  #   {% endif %}
+  #
+  #   {% if proxy_ssl_session_reuse != empty %}
+  #     {% capture proxy_ssl %}
+  #{#}   proxy_ssl_session_reuse {{ proxy_ssl_session_reuse }}
+  #     {% endcapture %}
+  #   {{ proxy_ssl | replace: "#{#}", "" }}
+  #   {% endif %}
+  # {% endif %}
+
   # these are duplicated so when request is redirected here those phases are executed
   post_action @out_of_band_authrep_action;
   body_filter_by_lua_block { require('apicast.executor'):body_filter() }
--- a/gateway/src/apicast/cli/environment.lua
+++ b/gateway/src/apicast/cli/environment.lua
@@ -75,6 +75,13 @@ local function cpus()
     return tonumber(nproc)
 end
 
+local env_value_mt = {
+    __tostring = function(t) return resty_env.value(t.name) end
+}
+
+local function env_value_ref(name)
+    return setmetatable({ name = name }, env_value_mt)
+end
 
 local _M = {}
 ---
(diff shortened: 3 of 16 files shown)


gateway/conf.d/apicast.conf
```diff
@@ -75,6 +75,29 @@ location @upstream {
...
+  # NGINX, all this is not executed as it is commented with '#'. However, in
+  # Liquid, all this will be evaluated. As a result, the following directives
+  # are set optionally: proxy_ssl_certificate, proxy_ssl_certificate_key,
+  # and proxy_ssl_session_reuse. The last one is 'on' by default, and I found
+  # that the certificate defined in this location was used in other ones.
+  # Setting it to 'off' fixes the issue.
+
+  # {% if proxy_ssl_certificate and proxy_ssl_certificate_key %}
+  #   {% capture proxy_ssl_session_reuse %}
+  #   proxy_ssl_session_reuse off;
```
mikz: Maybe we should expose this as a config option with a default value so it can be customized.

davidor: Maybe, but I couldn't make this work properly with `proxy_ssl_session_reuse on;`

gateway/conf.d/apicast.conf
```diff
@@ -75,6 +75,29 @@ location @upstream {
...
+  #   {% endcapture %}
+  #   {% capture ssl_certificate %}
+  #   proxy_ssl_certificate {{ proxy_ssl_certificate }};
+  #   {% endcapture %}
+  #   {% capture ssl_certificate_key %}
+  #   proxy_ssl_certificate_key {{ proxy_ssl_certificate_key }};
```
mikz: Lets not forget [proxy_ssl_password_file](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_ssl_password_file).

gateway/src/apicast/cli/environment.lua
```diff
@@ -56,13 +56,17 @@ _M.default_environment = 'production'
...
+-- @tfield ?string proxy_ssl_certificate path to SSL certificate
+-- @tfield ?string proxy_ssl_certificate_key path to SSL certificate key
 -- @tfield ?policy_chain policy_chain @{policy_chain} instance
 -- @tfield ?{string,...} nameservers list of nameservers
 -- @tfield ?string package.path path to load Lua files
 -- @tfield ?string package.cpath path to load libraries
 -- @table environment.default_config default configuration
 _M.default_config = {
     ca_bundle = resty_env.value('SSL_CERT_FILE'),
+    proxy_ssl_certificate = resty_env.value('SSL_PROXY_CERTIFICATE'),
```
mikz: We probably should covert them to absolute paths because it fails with relative ones as it changes nginx workdir/prefix with:
> nginx: [emerg] BIO_new_file("/tmp/public.crt") failed (SSL: error:02001002:system library:fopen:No such file or directory:fopen('/tmp/public.crt','r') error:2006D080:BIO routines:BIO_new_file:no such file)

But we also should support `engine:name:id ` format which would load the key from OpenSSL.

gateway/conf.d/apicast.conf
```diff
@@ -75,6 +75,29 @@ location @upstream {
...
+  #   {% endcapture %}
+  #   {{ proxy_ssl_session_reuse | replace: "#", "" }}
+  #   {{ ssl_certificate | replace: "#", "" }}
+  #   {{ ssl_certificate_key | replace: "#", "" }}
```
mikz: I find this working quite well:

```liquid
  # {% if proxy_ssl_certificate and proxy_ssl_certificate_key %}
  #   {% capture proxy_ssl %}
  #{#}   proxy_ssl_session_reuse off;
  #{#}   proxy_ssl_certificate {{ proxy_ssl_certificate }};
  #{#}   proxy_ssl_certificate_key {{ proxy_ssl_certificate_key }};
  #   {% endcapture %}
  #   {{ proxy_ssl | replace: "#{#}", "" }}
  # {% endif %}
```

This makes it obvious that it should strip only the first part of the comment and allows commenting lines within comments.

gateway/src/apicast/cli/environment.lua
```diff
@@ -56,13 +56,17 @@ _M.default_environment = 'production'
...
+    proxy_ssl_certificate_key = resty_env.value('SSL_PROXY_CERTIFICATE_KEY'),
```
mikz: We probably could name these like `APICAST_PROXY_*` (`APICAST_PROXY_CERTIFICATE` and `APICAST_PROXY_CERTIFICATE_KEY`). 

`APICAST_PROXY` is not the best name. Maybe `APICAST_UPSTREAM_HTTPS` ?

gateway/conf.d/apicast.conf
davidor: Yes. Otherwise, we'll need to type the password when Apicast boots.

gateway/conf.d/apicast.conf
davidor: Good idea :+1:

gateway/conf.d/apicast.conf
```diff
@@ -75,6 +75,23 @@ location @upstream {
...
+  # NGINX, all this is not executed as it is commented with '#'. However, in
+  # Liquid, all this will be evaluated. As a result, the following directives
+  # are set optionally: proxy_ssl_certificate, proxy_ssl_certificate_key,
+  # and proxy_ssl_session_reuse. The last one is 'on' by default, and I found
+  # that the certificate defined in this location was used in other ones.
+  # Setting it to 'off' fixes the issue.
+
+  # {% if proxy_ssl_certificate and proxy_ssl_certificate_key %}
+  #   {% capture proxy_ssl %}
+  #{#}   # proxy_ssl_session_reuse off;
```
davidor: We'll need to set this one according to `APICAST_PROXY_SESSION_REUSE`.

t/mutual-ssl.t
```diff
@@ -0,0 +1,111 @@
+use lib 't';
+use Test::APIcast::Blackbox 'no_plan';
+
+env_to_apicast(
+    'APICAST_PROXY_CERTIFICATE' => '../html/server.crt',
```
davidor: This should be `client.crt` and `client.key`.

t/mutual-ssl.t
```diff
@@ -0,0 +1,111 @@
...
+        ngx.exit(200)
+    }
+  }
+--- upstream env
+  listen $TEST_NGINX_RANDOM_PORT ssl;
+
+  ssl_certificate $TEST_NGINX_SERVROOT/html/server.crt;
+  ssl_certificate_key $TEST_NGINX_SERVROOT/html/server.key;
+
+  ssl_client_certificate $TEST_NGINX_SERVROOT/html/server.crt;
```
davidor: client.crt

examples/ssl-verification/badssl.json
```diff
@@ -35,6 +35,28 @@
...
         ]
       }
+    },
+    {
+      "id": "client-cert",
+      "backend_version": 1,
+      "proxy": {
+        "api_backend": "https://client.badssl.com",
+        "backend": {
+          "endpoint": "https://echo-api.3scale.net"
```
davidor: ?

davidor: Did you use the echo to test if the cert is sent to the backend too when the reuse ssl session option is enabled?

mikz: Yep. But I could not figure it out from the debug logs. Haven't done wireshark yet.

PR #612, opened 2018-02-20
[THREESCALE-630] passthrough invalid HTTP headers

mikz:
APIcast should not drop any headers and just pass them to upstream. Fixes https://issues.jboss.org/browse/THREESCALE-630.

### DIFF
--- a/gateway/conf/nginx.conf.liquid
+++ b/gateway/conf/nginx.conf.liquid
@@ -41,6 +41,8 @@ http {
   lua_package_path "{{ lua_path | default: package.path }}";
   lua_package_cpath "{{ lua_cpath | default: package.cpath }}";
 
+  ignore_invalid_headers off;
+
   {% if nameservers %}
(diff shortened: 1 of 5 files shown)


gateway/conf/nginx.conf.liquid
```diff
@@ -41,6 +41,9 @@ http {
   lua_package_path "{{ lua_path | default: package.path }}";
   lua_package_cpath "{{ lua_cpath | default: package.cpath }}";
 
+  underscores_in_headers on;
```
davidor: Is this needed if `ignore_invalid_headers` is set to `off`?
From the docs:

```
 When the use of underscores is disabled, request header fields whose names contain underscores are marked as invalid and become subject to the ignore_invalid_headers directive.
```

mikz: Good point. Lets try.

mikz: Indeed. Fixed.

PR #615, opened 2018-02-21
Add unit tests for the headers policy

davidor:
This policy only had integration tests.

### DIFF
--- a/spec/policy/headers/headers_spec.lua
+++ b/spec/policy/headers/headers_spec.lua
@@ -1,0 +1,204 @@
+local HeadersPolicy = require('apicast.policy.headers')
+
+describe('Headers policy', function()
+  -- Apply the operations to the same header in all the tests for simplicity
+  local header = 'test_header'
+
+  describe('.rewrite', function()
+    local request_headers = 'request' -- rewrite() only modifies request headers
+
+    local ngx_req_headers = {}
+
+    before_each(function()
+      -- ngx.req.get_headers and ngx.set_headers are not available in these tests
+      -- so we need to mock them.
+      -- Note: ngx.req_headers()['some_header'] can return a single value or a
+      -- table.
+
+      ngx_req_headers = {}
+      stub(ngx.req, 'get_headers', function() return ngx_req_headers end)
+      stub(ngx.req, 'set_header', function(name, value) ngx_req_headers[name] = value end)
+    end)
+
+    describe('the push operation', function()
+      local push = 'push'
+
+      it('creates the header with the given value when it is not set', function()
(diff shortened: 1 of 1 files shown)


spec/policy/headers/headers_spec.lua
```diff
@@ -0,0 +1,204 @@
...
+
+  describe('.rewrite', function()
+    local header_type = 'request' -- rewrite() only modifies request headers
+
+    -- ngx.req.get_headers and ngx.set_headers are not available in these tests
+    -- so we need to mock them.
+    -- Note: ngx.req_headers()['some_header'] can return a single value or a
+    -- table.
+    local ngx_req_headers = {}
+    ngx.req = {
```
mikz: Hm. This is going to override it locally. Would it be possible to have this done by `stub` in some before_each hook? I think this will leak through tests.

davidor: You are right. What's defined here also applies to the `describe(.header_filter')` block below. Which is fine in this case, but I can see how this can create problems in other cases.

davidor: Moved the stubs to a `before_each` clause so what's defined there does not leak to other tests.

spec/policy/headers/headers_spec.lua
```diff
@@ -0,0 +1,204 @@
...
+        local headers_policy = HeadersPolicy.new(config)
+
+        headers_policy:header_filter()
+
+        assert.is_nil(ngx.header[header])
+      end)
+
+      it('adds the value to the list of values for that header when its set', function()
+        ngx.header[header] = { '1', '2' }
+        local config = { [header_type] = { { op = op, header = header, value = '3' } } }
```
mikz: Might be nicer to rename `op` to `add` so it is obvious that it is `add` operation when reading the test.
And `header_type` might deserve the same treatment. 
Possibly even `test_header`.

davidor: I thought about that, but I decided to define each var as close as possible to its `describe`. Like this:
```lua
describe('the set operation', function()
  local op = 'set'
```

This way I can be more certain that I won't call the wrong operation. Having said that, I don't really have a strong opinion on this one. So if you find the other solution clearer I can change this.

mikz: Yeah. I think that is fine - using variables instead of strings is definitely better. But just renaming variable  `op` to `add` would improve readability of the test. It is hard to look up for context.

davidor: I misundertood your first comment. I thought you were proposing hardcoding the string there instead of defining the var.

The solution you propose is OK. I changed this.

PR #616, opened 2018-02-21
[t] use configuration fixtures

mikz:
Simple way of having fixtures for common blackbox configurations.

depends on https://github.com/3scale/Test-APIcast/pull/5

### DIFF
--- a/t/sanity.t
+++ b/t/sanity.t
@@ -1,6 +1,7 @@
 use lib 't';
 use Test::APIcast::Blackbox 'no_plan';
 
+filters { configuration => 'fixture=echo.json' };
 
 run_tests();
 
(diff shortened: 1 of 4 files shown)


t/sanity.t
```diff
@@ -1,6 +1,7 @@
 use lib 't';
 use Test::APIcast::Blackbox 'no_plan';
 
+filters { configuration => 'fixture=echo.json' };
```
davidor: I understand that the `fixture=something` part is implemented in Test::APIcast 0.10, but where is the `filters { something => something_else } ` defined? Is it just valid for blocks?

mikz: @davidor that comes from http://search.cpan.org/~ingy/Test-Base-0.88/lib/Test/Base.pod#FILTERS 
That is what powers the Test::Nginx declarative testing blocks.

davidor: :+1:

PR #618, opened 2018-02-21
Split resolv.conf in lines

maneta:
So the code does not brake like with comented #nameserver lines


gateway/src/resty/resolver.lua
```diff
@@ -131,12 +131,15 @@ function _M.parse_nameservers(path)
...
     insert(nameservers, resolver)
   end
-
-  for server in gmatch(resolv_conf, 'nameserver%s+([^%s]+)') do
-    -- TODO: implement port matching based on https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=549190
-    if server ~= resolver then
-      insert(nameservers, nameserver.new(server))
-    end
+  
+  for line in gmatch(resolv_conf, '(.*)\n]') do
```
mikz: I think this could be better as:
```lua
for _,line in ipairs(re.split(resolv_conf, [[\n+$]])) do
```

mikz: And actually I think it would be nice to extract parsing of the file into own function so it can be unit tested. I can do that if you want. 

Also looks like there is similar bug on line 116 in this file. `search` would also be extracted from a comment line.

maneta: @mikz go ahead this was just me playing around with lua regex. But lua was winning me :)

mikz: If you want I'm happy to provide guidance and let you do it ;)

maneta: yay! Thanks @mikz I will do it then

t/resolver.t
```diff
@@ -34,6 +34,15 @@ GET /t
 nameservers: 3 127.0.1.15353 1.2.3.453 4.5.6.753
 --- user_files
 >>> resolv.conf
+# nameserver updated  in comentary
```
davidor: This is good. It's testing more things than before, so I think it's worth expanding the description of the test (line 16).

spec/resty/resolver_spec.lua
```diff
@@ -163,6 +163,14 @@ describe('resty.resolver', function()
     before_each(function()
       tmpname = io.tmpfile()
 
+      tmpname:write('# nameserver updated  in comentary\n')
```
davidor: How about calling `write` just once? I think it looks better:

```lua
      tmpname:write(
        [[# nameserver updated  in comentary\n
          #nameserver updated  in comentary\n
          #comentary nameserver\n
          ..............]])
```

(Not sure if the `\n` needs to be there)

davidor:
Travis tests are failing because https://olivinelabs.com/busted/ is down, so don't worry about those.

gateway/src/resty/resolver.lua
```diff
@@ -131,10 +124,23 @@ function _M.parse_nameservers(path)
...
+        ngx.log(ngx.DEBUG, 'search domain: ', domain)
+        insert(search, domain)
+      end
+    end
 
-  for server in gmatch(resolv_conf, 'nameserver%s+([^%s]+)') do
+    local server = match(line, '^nameserver%s+([^%s]+)')
     -- TODO: implement port matching based on https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=549190
-    if server ~= resolver then
+    if server and server ~= resolver then
```
davidor: I think we don't have tests for this case.

t/resolver.t
maneta: done in d49b712

spec/resty/resolver_spec.lua
maneta: done in d49b712

gateway/src/resty/resolver.lua
maneta: Actual this test case fails.

We have agreed (me and @davidor)  that we should address those problems in another PR.

The issue which treat this case is #630

davidor:
@maneta the changelog entry is not under the correct version. It should be under `unreleased`.
Regarding the subsection, I think this should be under `fixed` instead of `changed` because it fixes a bug.

maneta:
Got it @davidor now it should be good to go :)

mikz:
@maneta there are conflicts with master. You'll need to rebase/merge master branch.

maneta:
@mikz Rebased

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Fixed
 
 - Error loading policy chain configuration JSON with null value [PR #626](https://github.com/3scale/apicast/pull/626)
+- Splitted `resolv.conf` in lines,to avoid commented lines  [PR #618](https://github.com/3scale/apicast/pull/618)
```
mikz: Just a nitpic. Changelog should focus on "what" and not "how".
Here would be better fit: Skip lines starting with a comment when parsing `resolv.conf`

mikz:
Thank you @maneta ! Much appreciated.

CHANGELOG.md
maneta: I bear that in mind. Thanks

PR #619, opened 2018-02-22
Add token introspection feature as a policy.

tmogi001:
Hi team,

I created a new policy to enable [RFC7662 OAuth2.0 Token Introspection](https://tools.ietf.org/html/rfc7662).
This PR is a rework of the previous request (https://github.com/3scale/apicast/pull/435) as a policy.

And Also added sample configuration in the `examples/policies` directory.

### DIFF
--- a/gateway/src/apicast/policy/token_introspection/apicast-policy.json
+++ b/gateway/src/apicast/policy/token_introspection/apicast-policy.json
@@ -12,11 +12,11 @@
         "description": "Introspection Endpoint URL",
         "type": "string"
       },
-      "clientId": {
-        "description": "Clilent ID for the Token Introspection Endpoint",
+      "client_id": {
+        "description": "Client ID for the Token Introspection Endpoint",
         "type": "string"
       },
-      "clientSecret": {
+      "client_secret": {
         "description": "Client Secret for the Token Introspection Endpoint",
         "type": "string"
       }
--- a/gateway/src/apicast/policy/token_introspection/apicast-policy.json
+++ b/gateway/src/apicast/policy/token_introspection/apicast-policy.json
@@ -1,0 +1,25 @@
+{
+  "$schema": "http://apicast.io/poolicy-v1/schema#manifest#",
+  "name": "oauth2 token introspection policy",
+  "description": 
+  ["This policy executes OAuth 2.0 Token Introspection",
+   "(https://tools.ietf.org/html/rfc7662) for every API calls."],
+  "version": "builtin",
+  "configuration": {
+    "type": "object",
+    "properties": {
+      "introspectionUrl": {
+        "description": "Introspection Endpoint URL",
+        "type": "string"
+      },
+      "clientId": {
+        "description": "Clilent ID for the Token Introspection Endpoint",
+        "type": "string"
+      },
+      "clientSecret": {
+        "description": "Client Secret for the Token Introspection Endpoint",
+        "type": "string"
+      }
+    }
+  }
+}
--- a/gateway/src/apicast/policy/token_introspection/apicast-policy.json
+++ b/gateway/src/apicast/policy/token_introspection/apicast-policy.json
@@ -8,7 +8,7 @@
   "configuration": {
     "type": "object",
     "properties": {
-      "introspectionUrl": {
+      "introspection_url": {
         "description": "Introspection Endpoint URL",
         "type": "string"
       },
--- a/gateway/src/apicast/policy/token_introspection/token_introspection.lua
+++ b/gateway/src/apicast/policy/token_introspection/token_introspection.lua
@@ -1,7 +1,7 @@
 local policy = require('apicast.policy')
 local _M = policy.new('Token Introspection Policy')
 
-local cjson = require('cjson')
+local cjson = require('cjson.safe')
 local http_authorization = require 'resty.http_authorization'
 local http_ng = require 'resty.http_ng'
 local user_agent = require 'apicast.user_agent'
@@ -35,13 +35,17 @@ local function introspect_token(self, token)
   --- Parameters for the token introspection endpoint.
   -- https://tools.ietf.org/html/rfc7662#section-2.1
   local res, err = self.http_client.post(self.introspection_url , { token = token, token_type_hint = 'access_token'})
-  if res and err then
+  if err then
     ngx.log(ngx.WARN, 'token introspection error: ', err, ' url: ', self.introspection_url)
     return { active = false }
   end
 
   if res.status == 200 then
-    local token_info = cjson.decode(res.body)
+    local token_info, decode_err = cjson.decode(res.body)
+    if decode_err then
+      ngx.log(ngx.ERR, 'failed to parse token introspection response:', decode_err)
+      return { active = false }
+    end
     return token_info
   else
     ngx.log(ngx.WARN, 'failed to execute token introspection. status: ', res.status)
--- a/gateway/src/apicast/policy/token_introspection/token_introspection.lua
+++ b/gateway/src/apicast/policy/token_introspection/token_introspection.lua
@@ -12,6 +12,10 @@ local new = _M.new
 function _M.new(config)
   local self = new()
   self.config = config or {}
+  --- authorization for the token introspection endpoint.
+  -- https://tools.ietf.org/html/rfc7662#section-2.2
+  self.credential = 'Basic ' .. ngx.encode_base64(table.concat({ self.config.client_id or '', self.config.client_secret or '' }, ':'))
+  self.introspection_url = config.introspection_url
   self.http_client = http_ng.new{
     backend = config.client,
     options = {
@@ -22,42 +26,38 @@ function _M.new(config)
   return self
 end
 
+--- OAuth 2.0 Token Introspection defined in RFC7662.
+-- https://tools.ietf.org/html/rfc7662
 local function introspect_token(self, token)
-  local config = self.config
-  if not config then
-    return false
-  end
-
-  local introspection_url = config.introspection_url
-  local client = config.client_id
-  local secret = config.client_secret
-  local credential = 'Basic ' .. ngx.encode_base64(table.concat({ client or '', secret or '' }, ':'))
   local opts = {
     headers = {
-      ['Authorization'] = credential
+      ['Authorization'] = self.credential
     }
   }
-
-  local res, err = self.http_client.post(introspection_url , { token = token, token_type_hint = 'access_token'}, opts)
+  --- Parameters for the token introspection endpoint.
+  -- https://tools.ietf.org/html/rfc7662#section-2.1
+  local res, err = self.http_client.post(self.introspection_url , { token = token, token_type_hint = 'access_token'}, opts)
   if res and err then
-    ngx.log(ngx.WARN, 'token introspection error: ', err, ' url: ', introspection_url)
-    return false
+    ngx.log(ngx.WARN, 'token introspection error: ', err, ' url: ', self.introspection_url)
+    return { active = false }
   end
 
   if res.status == 200 then
     local token_info = cjson.decode(res.body)
-    return token_info.active
+    return token_info
   else
     ngx.log(ngx.WARN, 'failed to execute token introspection. status: ', res.status)
-    return false
+    return { active = false }
   end
 end
 
 function _M:access(context)
-  if self.config.introspection_url then
+  if self.introspection_url then
     local authorization = http_authorization.new(ngx.var.http_authorization)
     local access_token = authorization.token
-    if not introspect_token(self, access_token) then
+    --- Introspection Response must have an "active" boolean value.
+    -- https://tools.ietf.org/html/rfc7662#section-2.2
+    if not introspect_token(self, access_token).active then
       ngx.status = context.service.auth_failed_status
       ngx.say(context.service.error_auth_failed)
       return ngx.exit(ngx.status)
--- a/gateway/src/apicast/policy/token_introspection/token_introspection.lua
+++ b/gateway/src/apicast/policy/token_introspection/token_introspection.lua
@@ -42,11 +42,12 @@ local function introspect_token(self, token)
 
   if res.status == 200 then
     local token_info, decode_err = cjson.decode(res.body)
-    if decode_err then
+    if type(token_info) == 'table' then
+      return token_info
+    else 
       ngx.log(ngx.ERR, 'failed to parse token introspection response:', decode_err)
       return { active = false }
     end
-    return token_info
   else
     ngx.log(ngx.WARN, 'failed to execute token introspection. status: ', res.status)
     return { active = false }
@@ -59,7 +60,7 @@ function _M:access(context)
     local access_token = authorization.token
     --- Introspection Response must have an "active" boolean value.
     -- https://tools.ietf.org/html/rfc7662#section-2.2
-    if not introspect_token(self, access_token).active then
+    if not introspect_token(self, access_token).active == true then
       ngx.status = context.service.auth_failed_status
       ngx.say(context.service.error_auth_failed)
       return ngx.exit(ngx.status)
--- a/gateway/src/apicast/policy/token_introspection/token_introspection.lua
+++ b/gateway/src/apicast/policy/token_introspection/token_introspection.lua
@@ -14,12 +14,15 @@ function _M.new(config)
   self.config = config or {}
   --- authorization for the token introspection endpoint.
   -- https://tools.ietf.org/html/rfc7662#section-2.2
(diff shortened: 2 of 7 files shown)


gateway/src/apicast/policy/token_introspection/token_introspection.lua
```diff
@@ -0,0 +1,68 @@
...
+local function introspect_token(self, token)
+  local config = self.config
+  if not config then
+    return false
+  end
+
+  local introspection_url = config.introspection_url
+  local client = config.client_id
+  local secret = config.client_secret
+  local credential = 'Basic ' .. ngx.encode_base64(table.concat({ client or '', secret or '' }, ':'))
```
mikz: This header can be moved to the _M.new, right?

gateway/src/apicast/policy/token_introspection/token_introspection.lua
```diff
@@ -0,0 +1,68 @@
...
+    ngx.log(ngx.WARN, 'failed to execute token introspection. status: ', res.status)
+    return false
+  end
+end
+
+function _M:access(context)
+  if self.config.introspection_url then
+    local authorization = http_authorization.new(ngx.var.http_authorization)
+    local access_token = authorization.token
+    if not introspect_token(self, access_token) then
```
mikz: I think it would be a bit nicer for `introspect_token` to return a table so this would look like:

```lua
if introspect_token(self, access_token).active then
  ...
else
  ... 
end
```

Or rename `introspect_token` to `active_token` because it actually returns true/false if the token is active or not.

gateway/src/apicast/policy/token_introspection/token_introspection.lua
```diff
@@ -0,0 +1,68 @@
...
+
+  local res, err = self.http_client.post(introspection_url , { token = token, token_type_hint = 'access_token'}, opts)
+  if res and err then
+    ngx.log(ngx.WARN, 'token introspection error: ', err, ' url: ', introspection_url)
+    return false
+  end
+
+  if res.status == 200 then
+    local token_info = cjson.decode(res.body)
+    return token_info.active
```
mikz: Is this `active` field defined in some RFC? Could we link to the relevant RFC section in a comment?

gateway/src/apicast/policy/token_introspection/token_introspection.lua
```diff
@@ -0,0 +1,68 @@
...
+  local client = config.client_id
+  local secret = config.client_secret
+  local credential = 'Basic ' .. ngx.encode_base64(table.concat({ client or '', secret or '' }, ':'))
+  local opts = {
+    headers = {
+      ['Authorization'] = credential
+    }
+  }
+
+  local res, err = self.http_client.post(introspection_url , { token = token, token_type_hint = 'access_token'}, opts)
```
mikz: Would be good to link to the RFC section describing that introspection endpoint receives POST.

spec/policy/token_introspection/token_introspection_spec.lua
```diff
@@ -0,0 +1,94 @@
...
+        respond_with{
+          status = 200,
+          body = cjson.encode({
+              active = true
+          })
+        }
+      local token_policy = require('apicast.policy.token_introspection').new(policy_config)
+      token_policy:access(context)
+
+      test_backend.verify_no_outstanding_expectations()
```
mikz: This line would be better in `after_each` block so it is not repeated in every test.

spec/policy/token_introspection/token_introspection_spec.lua
```diff
@@ -0,0 +1,94 @@
...
+    it('execute token introspection success', function()
+      local introspection_url = "http://example/token/introspection"
+      local policy_config = {
+        client = test_backend,
+        introspection_url = introspection_url,
+        client_id = "client",
+        client_secret = "secret"
+      }
+
+      test_backend.expect{ url = introspection_url }.
```
mikz: It probably should expect right http method  (POST)  too ? And some other parameters like token_type_hint ? And possibly the headers?

gateway/src/apicast/policy/token_introspection/token_introspection.lua
mikz: @davidor I think we should provide a way to do this in http_ng constructor. Because it is done one several places IIRC.

gateway/src/apicast/policy/token_introspection/token_introspection.lua
```diff
@@ -0,0 +1,68 @@
...
+local new = _M.new
+
+function _M.new(config)
+  local self = new()
+  self.config = config or {}
+  self.http_client = http_ng.new{
+    backend = config.client,
+    options = {
+      header = { ['User-Agent'] = user_agent() },
+      ssl = { verify = resty_env.enabled('OPENSSL_VERIFY') }
```
mikz: @davidor we probably should provide a default value for these options/headers in the http_ng constructor to simplify its initialization.

mikz:
@tmogi001 this is very good! 👍 🥇 My comments will be just nitpicks. This should be just minimal effort to be mergeable.
One of the things missing would be an integration test similar to other policy tests: https://github.com/3scale/apicast/blob/master/t/apicast-policy-cors.t
I think we could help with those a bit.

gateway/src/apicast/policy/token_introspection/token_introspection.lua
reedstrm: https://tools.ietf.org/html/rfc7662#section-2.1 - only REQUIRED param is the token

reedstrm:
@mikz  - As a random drive-by developer, with a group working on making our PR review policies more explicit (and automating as much as possible :-)  - I landed here following examples of use of the CHANGELOG github app) I just wanted to say that I liked your review style here - not only do you review the code-as-presented, you call out issues w/ the shared testing infrastructure, to make future coding easier.

gateway/src/apicast/policy/token_introspection/token_introspection.lua
reedstrm: Came by to look at your PR review automation, stayed to help out w/ a little PR doc pointing. :-) Seems the  appropriate RFC doc entry point is: https://tools.ietf.org/html/rfc7662#section-2.2 Where `active` is the only REQUIRED member of the returned JSON introspection result.

mikz: @reedstrm very nice. Thanks! This looks like the right doc.

@tmogi001 for cases where is RFC available I try to implement it as:
https://github.com/3scale/apicast/blob/5168008db2abb2e217d23f925599fd33da5248cb/gateway/src/resty/http_ng/cache_store.lua#L85-L110

Because the RFC is immutable the comments won't get out of sync and act as relevant context with links to the other parts of the RFC. In this case just the relevant parts to the code in question to document the requests and responses for example.

BTW it is pity[ OpenID Connect Discovery](http://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata) does not define token introspection endpoint. Some IDPs provide it (like Keycloak as `token_introspection_endpoint`), but I don't see it being part of the spec.

mikz:
@reedstrm thank you very much! Means a lot.

Regarding the Changelog app (if you mean https://github.com/mikz/probot-changelog).
I'd say we are mostly happy. But because the changelog is changed in almost every PR it is hard when there is a lot of concurrent changes because they conflict. That means every PR then has to be resolved after there is another one merged. That can be quite frustrating. Maybe not as much when your testsuite runs in less than few minutes, but if you have to wait several minutes it is easy to forget.

I'm quite obsessed with automation and other approaches I was considering were:
* https://github.com/conventional-changelog/conventional-changelog
* generating changelog from PR description + labels

But a changelog is very simple approach and can scale up to some level. Even some huge projects like Ruby on Rails still use one changelog file per component and it works.

tmogi001:
@mikz , @reedstrm Thank you for reviewing. I pushed fixed code and also added integration test case.

gateway/src/apicast/policy/token_introspection/token_introspection.lua
```diff
@@ -0,0 +1,68 @@
...
+  local self = new()
+  self.config = config or {}
+  --- authorization for the token introspection endpoint.
+  -- https://tools.ietf.org/html/rfc7662#section-2.2
+  self.credential = 'Basic ' .. ngx.encode_base64(table.concat({ self.config.client_id or '', self.config.client_secret or '' }, ':'))
+  self.introspection_url = config.introspection_url
+  self.http_client = http_ng.new{
+    backend = config.client,
+    options = {
+      header = { ['User-Agent'] = user_agent() },
```
mikz: @tmogi001 any reason why not to add the 'Authorization' header here too?

gateway/src/apicast/policy/token_introspection/token_introspection.lua
```diff
@@ -0,0 +1,68 @@
...
+local function introspect_token(self, token)
+  local opts = {
+    headers = {
+      ['Authorization'] = self.credential
+    }
+  }
+  --- Parameters for the token introspection endpoint.
+  -- https://tools.ietf.org/html/rfc7662#section-2.1
+  local res, err = self.http_client.post(self.introspection_url , { token = token, token_type_hint = 'access_token'}, opts)
+  if res and err then
```
mikz: Is the `if res` really necessary? Shouldn't this be just `if err` ?

gateway/src/apicast/policy/token_introspection/token_introspection.lua
```diff
@@ -0,0 +1,68 @@
...
+    ngx.log(ngx.WARN, 'token introspection error: ', err, ' url: ', self.introspection_url)
+    return { active = false }
+  end
+
+  if res.status == 200 then
+    local token_info = cjson.decode(res.body)
+    return token_info
```
mikz: This can return `nil` when `cjson.decode` returns `nil, err`. That can happen for example when the response body is not valid json (html usually).

t/apicast-policy-token-introspection.t
```diff
@@ -0,0 +1,156 @@
...
+--- request
+GET /echo
+--- more_headers
+Authorization: Bearer testaccesstoken
+--- error_code: 200
+--- no_error_log
+[error]
+
+=== TEST 2: Token already revoked.
+Token introspection policy return "403 Unauthorized" if access token is already revoked on IdP.
```
mikz: Again worth linking to the RFC: https://tools.ietf.org/html/rfc7662#section-2.3
It should be 401 Unauthorized.

gateway/src/apicast/policy/token_introspection/apicast-policy.json
```diff
@@ -0,0 +1,25 @@
...
+  "version": "builtin",
+  "configuration": {
+    "type": "object",
+    "properties": {
+      "introspectionUrl": {
+        "description": "Introspection Endpoint URL",
+        "type": "string"
+      },
+      "clientId": {
+        "description": "Clilent ID for the Token Introspection Endpoint",
```
mikz: A typo here. Should be **Client ID**.

gateway/src/apicast/policy/token_introspection/apicast-policy.json
```diff
@@ -0,0 +1,25 @@
...
+   "(https://tools.ietf.org/html/rfc7662) for every API calls."],
+  "version": "builtin",
+  "configuration": {
+    "type": "object",
+    "properties": {
+      "introspectionUrl": {
+        "description": "Introspection Endpoint URL",
+        "type": "string"
+      },
+      "clientId": {
```
mikz: Looks like this name is not the same as in the tests/code. Everywhere it uses `client_id`.

gateway/src/apicast/policy/token_introspection/apicast-policy.json
```diff
@@ -0,0 +1,25 @@
...
+        "description": "Clilent ID for the Token Introspection Endpoint",
+        "type": "string"
+      },
+      "clientSecret": {
```
mikz: The same issue as https://github.com/3scale/apicast/pull/619/files#r170914144.

gateway/src/apicast/policy/token_introspection/token_introspection.lua
mikz: Just when fixing https://github.com/3scale/apicast/pull/626 I realised there is plenty of error cases.

If the API would return `null` then this would be `ngx.null` and return the same error as in https://github.com/3scale/apicast/pull/626.  

This function should always return a table even when the json response can be a string, number, null, ...

And it should normalize `active` because `if ngx.null` is truthy so `{ "active": null }` would evaluate to true.

tmogi001:
I fixed my source code and added test cases (unit, integration) of invalid response from IdP.
And also added a change log entry.

gateway/src/apicast/policy/token_introspection/apicast-policy.json
```diff
@@ -0,0 +1,25 @@
...
+  "$schema": "http://apicast.io/poolicy-v1/schema#manifest#",
+  "name": "oauth2 token introspection policy",
+  "description": 
+  ["This policy executes OAuth 2.0 Token Introspection",
+   "(https://tools.ietf.org/html/rfc7662) for every API calls."],
+  "version": "builtin",
+  "configuration": {
+    "type": "object",
+    "properties": {
+      "introspectionUrl": {
```
mikz: This does not match the use in the policy. Here it is `introspectionUrl` but there it is `introspection_url`. 
We plan to have JSON schema validation during Policy initialization to verify those cases in CI, but right now it is just manual process, sorry.

gateway/src/apicast/policy/token_introspection/token_introspection.lua
```diff
@@ -0,0 +1,70 @@
...
+  end
+end
+
+function _M:access(context)
+  if self.introspection_url then
+    local authorization = http_authorization.new(ngx.var.http_authorization)
+    local access_token = authorization.token
+    --- Introspection Response must have an "active" boolean value.
+    -- https://tools.ietf.org/html/rfc7662#section-2.2
+    if not introspect_token(self, access_token).active then
```
mikz: This will incorrectly validate case when the response is `{ "active": null }`. It should check for `active == true`.

gateway/src/apicast/policy/token_introspection/token_introspection.lua
```diff
@@ -0,0 +1,70 @@
...
+    return { active = false }
+  end
+
+  if res.status == 200 then
+    local token_info, decode_err = cjson.decode(res.body)
+    if decode_err then
+      ngx.log(ngx.ERR, 'failed to parse token introspection response:', decode_err)
+      return { active = false }
+    end
+    return token_info
```
mikz: This will return whatever the JSON was. So if the JSON is `null` this would crash a bit later. 

I think it should verify `type(token_info) == 'table'` and return the error otherwise instead of checking just the `decode_err`.

tmogi001: In my opinion, checking the `decode_err` is enough. Because it must be an error when `token_info` is null.
So, `cjson.decode()` always return `error` and `decode_err` must not be null  when `token_info` JSON is `null`. 

I checked `cjson.decode("")`,`cjson.decode("{}")`, `cjson.decode(nil)` and `cjson.decode(ngx.null)` returned `decode_err`.

mikz: If the response body is `null` then the JSON will be correctly decoded to `ngx.null` like:

```lua
cjson.decode('null')
```

That will later crash on trying to access `.active` (attempt to index a userdata value):

```lua
cjson.decode('null').active
```

The same can happen when a number is returned like:

```lua
cjson.decode('42').active
```

What I'm saying is `introspect_token` always has to return a table, because later it tries to get the key `.access` from it. And the only way to verify it always returns table is to check the type.
Because the response body is basically user input it can be any valid JSON, including just string "null" or numbers.

tmogi001: Thank you for information. I missed 'null' and numbers.

Do you think it is good as following? This worked fine with test code.
```lua
if decode_err or not (type(token_info) == 'table') then
  ngx.log(ngx.ERR, 'failed to parse token introspection response:', decode_err)
  return { active = false }
end
```

mikz: I think it could be easier to read as:

```lua
if type(token_info) == 'table' then
  return token_info
else
  ngx.log(ngx.ERR, 'failed to parse token introspection response:', decode_err or 'not a table')
  return { active = false }
end
```

tmogi001: Thank you! I think this is nice.

mikz:
Thank you @tmogi001 for this excellent PR! 🎉 

You did a really good job.

PR #620, opened 2018-02-22
Add more tests for the CORS policy

davidor:
- Adds some missing checks in the integration tests for the CORS policy.
- Fixes a minor thing in the CORS policy code.
- Adds unit tests for the CORS policy.

### DIFF
--- a/spec/policy/cors/cors_spec.lua
+++ b/spec/policy/cors/cors_spec.lua
@@ -1,0 +1,105 @@
+local CORSPolicy = require('apicast.policy.cors')
+
+describe('CORS policy', function()
+  describe('.rewrite', function()
+    local ngx_exit_spy
+
+    describe('when the request is a CORS preflight', function()
+      setup(function()
+        ngx_exit_spy = spy.on(ngx, 'exit')
+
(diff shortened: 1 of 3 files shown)


spec/policy/cors/cors_spec.lua
```diff
@@ -0,0 +1,105 @@
+local CORSPolicy = require('apicast.policy.cors')
+
+describe('CORS policy', function()
+  describe('.rewrite', function()
+    local ngx_exit_spy
+
+    describe('when the request is a CORS preflight', function()
+      setup(function()
+        ngx_exit_spy = spy.on(ngx, 'exit')
```
mikz: This should also work as `stub(ngx,'exit')`. Learned that from https://github.com/3scale/apicast/pull/619/commits/1214674a92956ebde121bd48344eea599a2b3410#diff-5798bd05ef49d5c1a084d8f5464520c1R65

davidor: :+1: 
I don't see much difference in this particular case.

mikz: Just that it does not need to be stored in a variable.

Issue #609, opened 2018-02-19
Define the nginx TLS configuration using environment variables


nmasse-itix:
# Need

Some use cases require that nginx has a valid SSL/TLS configuration (such as client certificate authentication for instance). 

Currently, the customer has to write his nginx configuration templates that will be loaded by apicast in order to generate the final nginx configuration file. 

However, nginx configuration files defined in OpenShift Config Maps is error prone and not easily maintainable. 

It would be better if APIcast could generate the nginx configuration by itself using environment variables. 

# Proposition 

We could use the following environment variables : 
- APICAST_TLS_LISTEN={enable/disable} (optional, default = disable)
- APICAST_TLS_PORT=8443 (optional, default = 8443 if tls is enabled)
- APICAST_TLS_SERVING_CERTIFICATE=/path/to/tls.crt
- APICAST_TLS_SERVING_KEY=/path/to/tls.key
- APICAST_TLS_CRL=/path/to/file.crl (optional, none defined by default)
- APICAST_TLS_TRUSTED_CERTIFICATE=/path/to/ca-bundle.crt (optional, default to system bundle)
- APICAST_TLS_VERIFY_CLIENT={on/off/optional/optional_no_ca}
- APICAST_TLS_VERIFY_DEPTH= (optional, none defined by default)

There are other SSL/TLS configuration directives but they are less likely to change so they might be defined as nginx templates in ConfigMap or as custom image of APIcast.

mikz:
@nmasse-itix we are implementing client certificate support for upstream connections in https://github.com/3scale/apicast/pull/610
Would be good if you could share your inputs regarding naming.

mikz:
@nmasse-itix reviewing your proposal here are my suggestions: 

I'd try to get rid of `APICAST_TLS_LISTEN` and make it implicit by setting both the certificate and the key.

I'd change the `APICAST_TLS_` namespace to `APICAST_HTTPS_` to make it explicit that it is for HTTPS termination. There will be different settings for upstream connections: #610

Also I'm not sure if this is the best approach because we would basically expose every directive from http://nginx.org/en/docs/http/ngx_http_ssl_module.html as an option. It is a lot of variables. 

We are kind of postponing this decision until we have a solution we are going to be happy about. 
Maybe having the ability to control the nginx configuration from within a global policy (https://github.com/3scale/apicast/issues/477) would make this possible to be a global policy. Then we could give it some time to mature outside of the core before merging this upstream.

[...]

PR #622, opened 2018-02-23
Support HTTPS and introduce ssl_certificate policy phase

mikz:
This PR has two parts:

# Listen on HTTPS

Controlled by `APICAST_HTTPS_PORT` ENV variable.
Path to the certificate and key is defined by `APICAST_HTTPS_CERTIFICATE` and `APICAST_HTTPS_CERTIFICATE_KEY` ENV variables.

# `ssl_certificate` phase

New policy phase can select a certificate. It can load it from network or a file. See [lua-resty-autossl](https://github.com/GUI/lua-resty-auto-ssl) how to get letsencrypt certificates automatically.

Fixes https://github.com/3scale/apicast/issues/609

Depends on https://github.com/3scale/Test-APIcast/pull/9
Depends on https://github.com/3scale/Test-APIcast/pull/10


gateway/src/apicast/executor.lua
```diff
@@ -52,7 +52,7 @@ end
 -- @tparam string phase Nginx phase
 -- @treturn linked_list The context. Note: The list returned is 'read-write'.
 function _M:context(phase)
-    if phase == 'init' then
+    if phase == 'init' or phase == 'ssl_certificate' then
```
davidor: why?

gateway/conf/nginx.conf.liquid
```diff
@@ -122,7 +122,18 @@ http {
 
     access_log {{ access_log_file | default: "/dev/stdout" }} time;
 
-    listen {{ port.apicast | default: 8080 }};
+    {%- assign http_port = port.apicast | default: 8080 %}
+    {%- assign https_port = env.APICAST_HTTPS_PORT %}
+
+    {% if http_port != https_port -%}
```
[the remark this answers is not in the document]
davidor: I'd add a new test case in `t/listen-https.t` just to be sure.

gateway/src/apicast/executor.lua
mikz: Good question. That is probably a leftover from a debugging session. Will take a look.

gateway/conf/nginx.conf.liquid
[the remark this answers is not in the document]
mikz: So like the variable not set at all? That works - normal blackbox tests run with https_port not set.

The issue was that `{% if env.somevar %}` would be true even if it is empty.
This works just fine because `http_port` always has value.

gateway/conf/nginx.conf.liquid
```diff
@@ -77,6 +77,9 @@ http {
     {% include tracer_conf %}
   {% endif %}
 
+  ssl_session_fetch_by_lua_block { require('apicast.executor'):ssl_session_fetch() }
```
davidor: I don't see `ssl_session_fetch()` and `ssl_session_store()` defined anywhere. What I am missing?

gateway/conf/nginx.conf.liquid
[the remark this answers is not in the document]
davidor: Right :+1:

gateway/src/apicast/policy.lua
```diff
@@ -14,6 +14,8 @@ local PHASES = {
     'content', 'balancer',
     'header_filter', 'body_filter',
     'post_action',  'log', 'metrics',
+
+    'ssl_certificate',
```
davidor: Can you add this new phase to the list that we have in `policies.md` ?

gateway/src/apicast/policy/find_service/find_service.lua
```diff
@@ -70,8 +70,11 @@ function _M.new(...)
...
 end
 
-function _M:rewrite(context)
-  context.service = self.find_service(context.configuration, context.host)
+local function find_service(policy, context)
+  context.service = policy.find_service(context.configuration, context.host)
 end
 
+_M.rewrite = find_service
+_M.ssl_certificate = find_service
```
davidor: why?

gateway/conf/nginx.conf.liquid
```diff
@@ -131,8 +134,10 @@ http {
 
     {% if https_port -%}
     listen {{ https_port }} ssl;
-    ssl_certificate {{ env.APICAST_HTTPS_CERTIFICATE }};
-    ssl_certificate_key {{ env.APICAST_HTTPS_CERTIFICATE_KEY }};
+    ssl_certificate {{ env.APICAST_HTTPS_CERTIFICATE | default: "conf/server.crt" | filesystem | first }};
```
davidor: Why is `first` needed?

gateway/src/apicast/policy/load_configuration/load_configuration.lua
```diff
@@ -30,4 +31,8 @@ function _M:rewrite(context)
     context.configuration = configuration_loader.rewrite(self.configuration, context.host)
 end
 
+function _M.ssl_certificate(_, context)
+    context.host = ssl.server_name()
```
davidor: `ssl.server_name()` can return an error: https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/ssl.md#server_name
Shouldn't we handle it?

gateway/conf/nginx.conf.liquid
mikz: Actually I did not want to commit those lines just say. Will remove them. Want to keep it for the future.

gateway/src/apicast/policy.lua
mikz: Yep.

gateway/src/apicast/policy/find_service/find_service.lua
mikz: ssl_certificate executes first and needs to have access to the service so it can execute the chain

I'm going to verify the context is shared or not after removing https://github.com/3scale/apicast/pull/622#discussion_r193339476. But then it still needs to find the service in both (but can skip if the service is already found).

gateway/conf/nginx.conf.liquid
mikz: because `| filesystem` produces a table and it can't be serialized.

gateway/src/apicast/policy/load_configuration/load_configuration.lua
mikz: If it returns error then `context.host` is going to be nil and no service policy chain will be found.
In that case we serve whatever certificate was set to APICAST_HTTPS_CERTIFICATE or the provided default. 

Maybe we should just log the error with some low log level like debug.

gateway/conf/nginx.conf.liquid
```diff
@@ -122,7 +122,32 @@ http {
...
+
+    {% if http_port != https_port -%}
+      listen {{ http_port }};
+    {% endif %}
+
+    {% if https_port -%}
+    listen {{ https_port }} ssl;
+
+    {%- assign https_certificate = env.APICAST_HTTPS_CERTIFICATE -%}
+    ssl_certificate {% if https_certificate or https_certificate != empty -%}
```
davidor: I'm not sure I understand this. Shouldn't it be an `and` instead of an `or` ?
Does an empty string pass the first part of the `or` in Liquid?

mikz: I was trying to work around an issue with the liquid library. `{% "somestring" == empty %}` crashes. Will remove the `or ` part as that would still not execute when using magic tables like in other parts (where we have to check for == empty).

gateway/conf/nginx.conf.liquid
```diff
@@ -122,7 +122,32 @@ http {
...
+    {% if http_port != https_port -%}
+      listen {{ http_port }};
+    {% endif %}
+
+    {% if https_port -%}
+    listen {{ https_port }} ssl;
+
+    {%- assign https_certificate = env.APICAST_HTTPS_CERTIFICATE -%}
+    ssl_certificate {% if https_certificate -%}
+      {{  https_certificate }}
```
davidor: This one does not need ` | filesystem | first ` ?

mikz: This is path that was given by user in ENV, so we should not need to do any transformations.

mikz: Also, it does not really work. The `| filesystem` filter does not really work with absolute paths (yet).

davidor:
:+1:

PR #624, opened 2018-02-23
Add more tests for the upstream policy

davidor:


spec/policy/upstream/upstream_spec.lua
```diff
@@ -0,0 +1,128 @@
...
+    local test_upstream_matched_proxy_pass = 'http://upstream/a_path:8080'
+    local test_upstream_not_matched = 'http://localhost/a_path:80'
+
+    local context = {} -- Context shared between policies
+
+    before_each(function()
+      context = {}
+
+      -- Set headers_sent to false, otherwise, the upstream is not changed
+      ngx.headers_sent = false
```
mikz: Can we use `stub` instead ? Or would it have to be function?

davidor: This one is a bit tricky.

The value set here leaks to the `describe('.balancer')` below.

I thought about resetting `ngx.headers_sent` to its original value just like we do with other `ngx.x` in `ngx_helper.lua`. However, that does not work either, because when we capture the original ngx values, `ngx.headers_sent` raises the `API disabled` error. Stubs and spies raise the same error in the `before_each` method.

We could set `ngx_headers_sent` to nil in the `after_each` method and be done with it. If some test in the same file ever needs to read this and find a true, we'll rely on Openresty setting it in the test execution, or someone writing `ngx.headers_sent = true` in a `before` method for that test.

mikz: We would have to do the same as we do for `ngx.status`:
https://github.com/3scale/apicast/blob/985a24a76c16ffca4f245fb487414f3c94781362/spec/ngx_helper.lua#L37
because it is implemented the same way: 
https://github.com/openresty/lua-resty-core/blob/a39b8d68b3b1c93f22259f95925c9dc854d69c1c/lib/resty/core/misc.lua#L133-L152

davidor: Implemented in the new commit.

PR #625, opened 2018-02-26
Reset ngx state on every test

davidor:
This PR modifies the `after_each` in the busted helpers to make sure that we start with a clean ngx state on every test.

### DIFF
--- a/spec/ngx_helper.lua
+++ b/spec/ngx_helper.lua
@@ -1,10 +1,11 @@
 local busted = require('busted')
 local misc = require('resty.core.misc')
+local deepcopy = require('pl.tablex').deepcopy
 
-local ngx_var = ngx.var
-local ngx_ctx = ngx.ctx
-local ngx_shared = ngx.shared
-local ngx_header = ngx.header
+local ngx_var_original = deepcopy(ngx.var)
+local ngx_ctx_original = deepcopy(ngx.ctx)
+local ngx_shared_original = deepcopy(ngx.shared)
+local ngx_header_original = deepcopy(ngx.header)
 
 local register_getter = misc.register_ngx_magic_key_getter
 local register_setter = misc.register_ngx_magic_key_setter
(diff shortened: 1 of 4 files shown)


davidor:
Build failing in Travis because https://olivinelabs.com/busted/ is down.

spec/ngx_helper.lua
```diff
@@ -1,10 +1,11 @@
 local busted = require('busted')
 local misc = require('resty.core.misc')
+local deepcopy = require('pl.tablex').deepcopy
 
-local ngx_var = ngx.var
-local ngx_ctx = ngx.ctx
-local ngx_shared = ngx.shared
-local ngx_header = ngx.header
+local ngx_var_original = deepcopy(ngx.var)
```
mikz: Why do we need to copy `ngx.var` ? Could we deepcopy the whole `ngx` ?

mikz: I think the deepcopy would be useful if we would be to copy the whole `ngx` var. 
But for `var`, `ctx` and `header` I think they could be just reference to the original no? They would crash on use.

`ngx.shared` probably needs to be copied as that could be mutated and leak through tests.

davidor: Deep-copying and then restoring the whole `ngx` does not work. I need to investigate why.

I think that deep-copying everything is safer. Take a look at the second commit, setting `ngx.header`, for example, fails in busted tests with the `"API disabled
in the current context"` error. To make it work, I can just overwrite it with `{}` because the method that raises that error is  `.__newindex`. But after that particular test finishes, I'd like to restore the original `ngx.header` to make sure that other tests do not rely on the custom one.

mikz: Some of the `ngx` API might be implemented on C level and could lead to segfaults. I tried to call `getmetatable(ngx).__index()` and it segfaulted. It would be nice, but not sure how feasible it is to deep copy it.

Yep I think that is exactly what it was doing before. After every test setting back the original `ngx.header`.

I guess it is safer to do a deep copy because some test could run `rawset(ngx.header, 'foo', 'val')` skipping the `__newindex` and mutating the global table.

davidor: I didn't find a good way to clone the whole `ngx` table. I'd leave it as it is, at least for now.

PR #628, opened 2018-02-27
Add more tests for the URL rewriting policy

davidor:


davidor:
Travis tests are failing because http://olivinelabs.com/busted/ is down.

PR #629, opened 2018-02-27
[prometheus] exporter prototype

mikz:
* introduce metrics phase to export metrics
* executor prints collects the response from prometheus
* metrics can be also published from any other phase or timer

## Example
See for policy changes and metrics examples:
https://github.com/3scale/apicast-cloud-hosted/pull/5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Error loading policy chain configuration JSON with null value [PR #626](https://github.com/3scale/apicast/pull/626)
 - Splitted `resolv.conf` in lines,to avoid commented lines  [PR #618](https://github.com/3scale/apicast/pull/618)
 
+## Added
+
+- New `metrics` phase that runs when prometheus is collecting metrics [PR #629](https://github.com/3scale/apicast/pull/629)
+
 ## [3.2.0-beta1] - 2018-02-20
 
 ## Added
--- a/gateway/conf/nginx.conf.liquid
+++ b/gateway/conf/nginx.conf.liquid
@@ -106,6 +106,24 @@ http {
     {% include "conf.d/apicast.conf" %}
   }
 
+  {% if port.metrics %}
+    lua_shared_dict prometheus_metrics 16M;
+    server {
+      access_log off;
+      listen {{ port.metrics }};
+      server_name metrics prometheus _;
+
+      location /metrics {
+        content_by_lua_block { require('apicast.executor'):metrics() }
+      }
+
+      location /nginx_status {
+        internal;
+        stub_status;
+      }
+    }
+  {% endif %}
+
   {% for file in "sites.d/*.conf" | filesystem %}
     {% include file %}
   {% endfor %}
--- a/gateway/config/development.lua
+++ b/gateway/config/development.lua
@@ -28,5 +28,6 @@ return {
     lua_code_cache = 'on',
     configuration_loader = 'lazy',
     configuration_cache = 0,
-    configuration = data_url('application/json', configuration)
+    configuration = data_url('application/json', configuration),
+    port = { metrics = 9421 }, -- see https://github.com/prometheus/prometheus/wiki/Default-port-allocations
 }
--- a/gateway/src/apicast/executor.lua
+++ b/gateway/src/apicast/executor.lua
@@ -9,6 +9,7 @@ require('apicast.loader') -- to load code from deprecated paths
 local PolicyChain = require('apicast.policy_chain')
 local policy = require('apicast.policy')
 local linked_list = require('apicast.linked_list')
+local prometheus = require('apicast.prometheus')
 
 local setmetatable = setmetatable
 
@@ -56,6 +57,13 @@ function _M:context(phase)
     end
 
     return shared_build_context(self)
+end
+
+local metrics = _M.metrics
+--- Render metrics from all policies.
+function _M:metrics(...)
+    metrics(self, ...)
+    return prometheus:collect()
 end
 
 return _M.new(PolicyChain.default())
--- a/gateway/src/apicast/policy.lua
+++ b/gateway/src/apicast/policy.lua
@@ -13,7 +13,7 @@ local PHASES = {
     'rewrite', 'access',
     'content', 'balancer',
     'header_filter', 'body_filter',
-    'post_action',  'log'
+    'post_action',  'log', 'metrics',
 }
 
 local setmetatable = setmetatable
--- a/gateway/src/apicast/prometheus.lua
+++ b/gateway/src/apicast/prometheus.lua
@@ -1,0 +1,23 @@
+local prometheus = require('nginx.prometheus')
+local assert = assert
+local dict = 'prometheus_metrics'
+
+if ngx.shared[dict] then
+  local init = prometheus.init(dict)
+
+  local metrics = { }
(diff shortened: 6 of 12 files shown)


gateway/config/development.lua
```diff
@@ -28,5 +28,6 @@ return {
     lua_code_cache = 'on',
     configuration_loader = 'lazy',
     configuration_cache = 0,
-    configuration = data_url('application/json', configuration)
+    configuration = data_url('application/json', configuration),
+    port = { metrics = 9100 },
```
maneta: @mikz maybe reserve a port for the APIcast endpoint so we can publish it on the [Prometheus Default Port Allocation list](https://github.com/prometheus/prometheus/wiki/Default-port-allocations)?

mikz: I think we should agree how we want to proceed first.
We are not going to reserve a port for every app we have, right? We need some generic port for openresty exporter?

maneta: Got It, I thought that at least for APICast we were willing to create a specific "exporter" but yes makes more sense to create a generic port for openresty exporter.

jmprusi: To me it doesn't really make sense to reserve a port for each APP, but this is not a /generic/ openresty prometheus exporter, so, should we define a port for the "APIcast framework" and reuse that one in all our applications?

mikz: That makes sense if we want to converge all our proxies running on the "APIcast framework".

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Error loading policy chain configuration JSON with null value [PR #626](https://github.com/3scale/apicast/pull/626)
 - Splitted `resolv.conf` in lines,to avoid commented lines  [PR #618](https://github.com/3scale/apicast/pull/618)
 
+## Addwed
```
davidor: Typo

gateway/config/development.lua
mikz: Reserved 9421.

jmprusi: We should converge them all, now with policies it should be easier ;)

gateway/conf/nginx.conf.liquid
```diff
@@ -106,6 +106,24 @@ http {
     {% include "conf.d/apicast.conf" %}
   }
 
+  {% if port.metrics %}
+    lua_shared_dict prometheus_metrics 16M;
```
davidor: Should this be configurable?

mikz: Probably, but I'd wait with some real operations use to see if it is really needed.

gateway/conf/nginx.conf.liquid
```diff
@@ -106,6 +106,24 @@ http {
...
+    server {
+      access_log off;
+      listen {{ port.metrics }};
+      server_name metrics prometheus _;
+
+      location /metrics {
+        content_by_lua_block { require('apicast.executor'):metrics() }
+      }
+
+      location /nginx_status {
```
davidor: Why is this needed?

gateway/conf/nginx.conf.liquid
mikz: Hah. This just gave me an idea for another exporter: shmem dict stats about free space.

gateway/config/development.lua
```diff
@@ -28,5 +28,6 @@ return {
     lua_code_cache = 'on',
     configuration_loader = 'lazy',
     configuration_cache = 0,
-    configuration = data_url('application/json', configuration)
+    configuration = data_url('application/json', configuration),
+    port = { metrics = 9421 }, -- see https://github.com/prometheus/prometheus/wiki/Default-port-allocations
```
davidor: :+1:

gateway/conf/nginx.conf.liquid
mikz: It is used in https://github.com/3scale/apicast-cloud-hosted/pull/5/files#diff-047b1780b0ffeb4eba7b3d05beb76d5eR77 to extract the connection stats. 
There is no easy way to add that location from customization, so I added it here as this will become part of upstream at some point.

gateway/src/apicast/policy.lua
```diff
@@ -13,7 +13,7 @@ local PHASES = {
     'rewrite', 'access',
     'content', 'balancer',
     'header_filter', 'body_filter',
-    'post_action',  'log'
+    'post_action',  'log', 'metrics',
```
davidor: We should add new phases to the policies doc: https://github.com/3scale/apicast/blob/master/doc/policies.md#policies

gateway/conf/nginx.conf.liquid
mikz: Done in https://github.com/3scale/apicast-cloud-hosted/pull/5/commits/09f65c93775c4bf6186aac8c92cdd06344e4f63d

gateway/src/apicast/executor.lua
```diff
@@ -58,4 +59,11 @@ function _M:context(phase)
     return shared_build_context(self)
 end
 
+local metrics = _M.metrics
+--- Render metrics from all policies.
+function _M:metrics(...)
```
davidor: Maybe we should add a test for this. Define 2 policies with only a metrics phase that incr some counter and make sure that this returns the sum, or something similar.

gateway/src/apicast/policy.lua
mikz: Done in https://github.com/3scale/apicast/pull/629/commits/c44b062f1047b718ac3a65773363708b1b1a704b

gateway/src/apicast/executor.lua
mikz: Yeah but this does not really "return" anything. It renders the output.
I'll definitely do that when moving some policies with metrics to upstream from https://github.com/3scale/apicast-cloud-hosted/pull/5.

gateway/src/apicast/prometheus.lua
```diff
@@ -0,0 +1,23 @@
+local prometheus = require('nginx.prometheus')
+local assert = assert
+local dict = 'prometheus_metrics'
+
+if ngx.shared[dict] then
+  local init = prometheus.init(dict)
+
+  local metrics = { }
+  local __call = function(_, type, name, ...)
+    local metric_name = assert(name, 'missing metric name')
```
davidor: Why do we check `name` and `type`? Can't we delegate this to the Prometheus lib?

spec/prometheus_spec.lua
```diff
@@ -0,0 +1,41 @@
...
+      }
+    end)
+
+    local prometheus
+
+    before_each(function()
+      prometheus = assert(require('apicast.prometheus'))
+    end)
+
+    it('can be called', function()
```
davidor: Should we change this to test that it delegates `counter`, `gauge`, etc. to the Prometheus lib?

gateway/src/apicast/prometheus.lua
mikz: Because we actually use it. The type gets a function that is later called and the name is used as a key to store the metric. I had few tricky errors so decided to go for explicit ones.

spec/prometheus_spec.lua
mikz: Yep, makes sense.

mikz: Done.

PR #632, opened 2018-02-28
[make] run .t files in examples folder

mikz:
run `.t` files from `t` and `examples` folders

useful for example policies like https://github.com/3scale/apicast/pull/561


davidor:
:+1: 

I think we should do the same for busted tests.

mikz:
@davidor I agree, but lets do that when we actually have some so we can try. `bin/busted` should be ready for it.

PR #646, opened 2018-03-05
Validation for policy configurations

davidor:
This PR introduces a way to validate a policy's configuration against the JSON schema defined for its configuration.

This is what I used to detect the problems in our tests fixed in previous PRs: #641 , #644 , #645 

We need to define and agree on a few things before this becomes mergeable. That's why I tagged it as WIP:
* I used `rapidjson` to perform the validation, and it needs cmake. This means that it should be installed on the base image.
* `rapidjson` uses the draft04 version of the JSON schema. I needed to replace the `const`s in our schemas so it works correctly as `const` was introduced in newer versions.
* I needed to instantiate policies using the same name and version used by the policy loaded to find the manifests easily.
* I think this feature should be opt-in via ENV and always enabled in the tests. (I didn't implement this part yet).

### DIFF
--- a/t/apicast-policy-invalid-config.t
+++ b/t/apicast-policy-invalid-config.t
@@ -1,0 +1,44 @@
+use lib 't';
+use Test::APIcast::Blackbox 'no_plan';
+
+use Cwd qw(abs_path);
+
+BEGIN {
+    $ENV{TEST_NGINX_APICAST_POLICY_LOAD_PATH} = 't/fixtures/policies';
+}
+
+env_to_apicast(
+    'APICAST_POLICY_LOAD_PATH' => abs_path($ENV{TEST_NGINX_APICAST_POLICY_LOAD_PATH}),
+);
+
+run_tests();
+
+__DATA__
+
+=== TEST 1: policy with invalid configuration
+In this test, we use an example policy that requires a 'message' param in the
+configuration. We are going to initialize the policy with an empty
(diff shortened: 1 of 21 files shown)


davidor:
There have been a few changes:
* Using our [fork of ljsonschema](https://github.com/3scale/ljsonschema) instead of `rapidjson`, because it's a pure lua dependencies and does not require to install additional software in the base image.
* Found an alternative way to validate schemas in the `policy_loader` that does not require changing the existing names and versions used when instantiating policies.
* Added a new ENV to make the feature opt-in. We only use it in tests. Do you think we should document it @mikz ? Or just leave it undocumented for now?

t/apicast-policy-invalid-config.t
```diff
@@ -0,0 +1,56 @@
...
+    }
+  }
+--- upstream
+  location / {
+     echo 'yay, api backend: $http_host';
+  }
+--- request
+GET /?user_key=value
+--- error_code: 200
+--- must_die
```
mikz: The must die signals nginx should quit, right? does it need the --- backend and upstream then? And the error_code ?

davidor: You're right. These are leftovers because at first, I was logging an error instead of exiting.

davidor: Fixed.

PR #648, opened 2018-03-08
Add rate limiting to the entire service policy.

y-tabata:
Hi team,

I created a new policy that adds rate limiting to the entire Service.
The RFE about this has been opened [1].

[1] https://access.redhat.com/solutions/3308541

### DIFF
--- a/examples/policies/rate_limiting_to_service_configuration.lua
+++ b/examples/policies/rate_limiting_to_service_configuration.lua
@@ -1,13 +1,31 @@
 local policy_chain = require('apicast.policy_chain').default()
 
 local rate_limit_policy = require('apicast.policy.rate_limiting_to_service').new({
-  limit = 10,
-  period = 10,
-  service_name = "rate_limit_service"
+  limitters = {
+  {
+    limitter = "resty.limit.conn",
+    key = "limit1",
+    values = {20, 10, 0.5}
+  },
+  {
+    limitter = "resty.limit.req",
+    key = "limit2",
+    values = {18, 9}
+  },
+  {
+    limitter = "resty.limit.count",
+    key = "limit3",
+    values = {10, 10}
+  }},
+  redis_info = {
+    host = '127.0.0.1',
+    port = 6379,
+    db = 1
+  }
 })
 
 policy_chain:insert(rate_limit_policy, 1)
 
 return {
-        policy_chain = policy_chain
+  policy_chain = policy_chain
 }
--- a/examples/policies/rate_limiting_to_service_configuration.lua
+++ b/examples/policies/rate_limiting_to_service_configuration.lua
@@ -17,11 +17,7 @@ local rate_limit_policy = require('apicast.policy.rate_limiting_to_service').new
     key = "limit3",
     values = {10, 10}
   }},
-  redis_info = {
-    host = '127.0.0.1',
-    port = 6379,
-    db = 1
-  }
+  redis_url = "redis://localhost:6379/1"
 })
 
 policy_chain:insert(rate_limit_policy, 1)
--- a/examples/policies/rate_limiting_to_service_configuration.lua
+++ b/examples/policies/rate_limiting_to_service_configuration.lua
@@ -1,19 +1,19 @@
 local policy_chain = require('apicast.policy_chain').default()
 
 local rate_limit_policy = require('apicast.policy.rate_limiting_to_service').new({
-  limitters = {
+  limiters = {
   {
-    limitter = "resty.limit.conn",
+    limiter = "resty.limit.conn",
     key = "limit1",
     values = {20, 10, 0.5}
   },
   {
-    limitter = "resty.limit.req",
+    limiter = "resty.limit.req",
     key = "limit2",
     values = {18, 9}
   },
   {
-    limitter = "resty.limit.count",
+    limiter = "resty.limit.count",
     key = "limit3",
     values = {10, 10}
   }},
--- a/examples/policies/rate_limiting_to_service_configuration.lua
+++ b/examples/policies/rate_limiting_to_service_configuration.lua
@@ -3,19 +3,23 @@ local policy_chain = require('apicast.policy_chain').default()
 local rate_limit_policy = require('apicast.policy.rate_limiting_to_service').new({
   limiters = {
   {
-    limiter = "resty.limit.conn",
+    name = "connections",
     key = "limit1",
-    values = {20, 10, 0.5}
+    conn = 20,
+    burst = 10,
+    delay = 0.5
   },
   {
-    limiter = "resty.limit.req",
+    name = "leaky_bucket",
     key = "limit2",
-    values = {18, 9}
+    rate = 18,
+    burst = 9
   },
   {
-    limiter = "resty.limit.count",
+    name = "fixed_window",
     key = "limit3",
-    values = {10, 10}
+    count = 10,
+    window = 10
   }},
   redis_url = "redis://localhost:6379/1"
 })
--- a/examples/policies/rate_limiting_to_service_configuration.lua
+++ b/examples/policies/rate_limiting_to_service_configuration.lua
@@ -1,0 +1,13 @@
+local policy_chain = require('apicast.policy_chain').default()
+
+local rate_limit_policy = require('apicast.policy.rate_limiting_to_service').new({
+  limit = 10,
+  period = 10,
+  service_name = "rate_limit_service"
+})
+
+policy_chain:insert(rate_limit_policy, 1)
+
+return {
+        policy_chain = policy_chain
+}
--- a/gateway/src/apicast/policy/rate_limit/apicast-policy.json
+++ b/gateway/src/apicast/policy/rate_limit/apicast-policy.json
@@ -32,7 +32,7 @@
                 "description": "the number of excessive concurrent requests (or connections) allowed to be delayed"
               },
               "delay": {
-                "type": "integer",
+                "type": "number",
                 "description": "the default processing latency of a typical connection (or request)"
               }
             }
--- a/gateway/src/apicast/policy/rate_limit/apicast-policy.json
+++ b/gateway/src/apicast/policy/rate_limit/apicast-policy.json
@@ -1,7 +1,7 @@
 {
   "$schema": "http://apicast.io/policy-v1/schema#manifest#",
-  "name": "rate limiting to service policy",
-  "description": ["This policy adds rate limiting to the entire Service.",
+  "name": "rate limit policy",
+  "description": ["This policy adds rate limit.",
     ""
   ],
   "version": "builtin",
--- a/gateway/src/apicast/policy/rate_limit/apicast-policy.json
+++ b/gateway/src/apicast/policy/rate_limit/apicast-policy.json
@@ -79,6 +79,10 @@
       "redis_url": {
         "description": "URL of Redis",
         "type": "string"
+      },
+      "status_code_rejected": {
+        "type": "integer",
+        "description": "the status code when requests over the limit, default 429"
       }
     }
   }
--- a/gateway/src/apicast/policy/rate_limit/apicast-policy.json
+++ b/gateway/src/apicast/policy/rate_limit/apicast-policy.json
@@ -1,9 +1,7 @@
 {
   "$schema": "http://apicast.io/policy-v1/schema#manifest#",
   "name": "rate limit policy",
-  "description": ["This policy adds rate limit.",
-    ""
-  ],
+  "description": ["This policy adds rate limit."],
   "version": "builtin",
   "configuration": {
     "type": "object",
--- a/gateway/src/apicast/policy/rate_limit/apicast-policy.json
+++ b/gateway/src/apicast/policy/rate_limit/apicast-policy.json
@@ -11,9 +11,9 @@
         "description": "List of limiters to be applied",
         "type": "array",
         "items": {
-          "type": "object",
-          "properties": {
-            "oneOf": [{
+          "oneOf": [{
+            "type": "object",
+            "properties": {
               "name": {
                 "type": "string",
                 "enum": ["connections"],
@@ -35,7 +35,10 @@
                 "type": "integer",
                 "description": "the default processing latency of a typical connection (or request)"
               }
-            }, {
+            }
+          }, {
+            "type": "object",
+            "properties": {
               "name": {
                 "type": "string",
                 "enum": ["leaky_bucket"],
@@ -53,7 +56,10 @@
                 "type": "integer",
                 "description": "the number of excessive requests per second allowed to be delayed"
               }
-            }, {
+            }
+          }, {
+            "type": "object",
+            "properties": {
               "name": {
                 "type": "string",
                 "enum": ["fixed_window"],
@@ -71,8 +77,8 @@
                 "type": "integer",
                 "description": "the time window in seconds before the request count is reset"
               }
-            }]
-          }
+            }
+          }]
         }
       },
       "redis_url": {
--- a/gateway/src/apicast/policy/rate_limit/apicast-policy.json
+++ b/gateway/src/apicast/policy/rate_limit/apicast-policy.json
@@ -83,6 +83,10 @@
       "status_code_rejected": {
         "type": "integer",
         "description": "the status code when requests over the limit, default 429"
+      },
+      "logging_only": {
+        "type": "boolean",
+        "description": "if true, the request goes through when there is some issue with rate limiting, default false"
       }
     }
   }
--- a/gateway/src/apicast/policy/rate_limit/apicast-policy.json
+++ b/gateway/src/apicast/policy/rate_limit/apicast-policy.json
@@ -1,6 +1,7 @@
 {
   "$schema": "http://apicast.io/policy-v1/schema#manifest#",
   "name": "rate limit policy",
+  "summary": "Adds rate limit.",
   "description": ["This policy adds rate limit."],
   "version": "builtin",
   "configuration": {
--- a/gateway/src/apicast/policy/rate_limit/rate_limit.lua
+++ b/gateway/src/apicast/policy/rate_limit/rate_limit.lua
@@ -43,9 +43,6 @@ local function init_limiter(config)
       return nil, e
     end
   )
-  if not lim then
-    return nil, limerr
-  end
 
   return lim, nil
 end
--- a/gateway/src/apicast/policy/rate_limit/rate_limit.lua
+++ b/gateway/src/apicast/policy/rate_limit/rate_limit.lua
@@ -1,5 +1,5 @@
 local policy = require('apicast.policy')
-local _M = policy.new('Rate Limiting to Service Policy')
+local _M = policy.new('Rate Limit Policy')
 
 local resty_limit_conn = require('resty.limit.conn')
 local resty_limit_req = require('resty.limit.req')
--- a/gateway/src/apicast/policy/rate_limit/rate_limit.lua
+++ b/gateway/src/apicast/policy/rate_limit/rate_limit.lua
@@ -103,28 +103,15 @@ function _M:access()
   local limiters = {}
   local keys = {}
 
-  if not self.redis_url then
-    -- Only one (the first) limiter is enable.
-    -- Key will be shdict_key ('limiter').
-    local lim, initerr = init_limiter(self.limiters[1])
+  for _, limiter in ipairs(self.limiters) do
+    local lim, initerr = init_limiter(limiter)
     if not lim then
-      ngx.log(ngx.ERR, "unknown limiter: ", initerr)
+      ngx.log(ngx.ERR, "unknown limiter: ", limiter.name, ", err: ", initerr)
       error(self.logging_only, 500)
       return
     end
 
-    limiters[1] = lim
-    keys[1] = shdict_key
-
-  else
-    for _, limiter in ipairs(self.limiters) do
-      local lim, initerr = init_limiter(limiter)
-      if not lim then
-        ngx.log(ngx.ERR, "unknown limiter: ", limiter.name, ", err: ", initerr)
-        error(self.logging_only, 500)
-        return
-      end
-
+    if self.redis_url then
       local rediserr
       lim.dict, rediserr = redis_shdict(self.redis_url)
       if not lim.dict then
@@ -132,12 +119,13 @@ function _M:access()
         error(self.logging_only, 500)
         return
       end
-
-      limiters[#limiters + 1] = lim
-      keys[#keys + 1] = limiter.key
-
     end
+
+    limiters[#limiters + 1] = lim
+    keys[#keys + 1] = limiter.key
+
   end
+
 
   local states = {}
   local connections_committed = {}
--- a/gateway/src/apicast/policy/rate_limit/rate_limit.lua
+++ b/gateway/src/apicast/policy/rate_limit/rate_limit.lua
@@ -110,7 +110,7 @@ function _M:access()
     if not lim then
       ngx.log(ngx.ERR, "unknown limiter: ", initerr)
       error(self.logging_only, 500)
-      goto done
+      return
     end
 
     limiters[1] = lim
@@ -122,7 +122,7 @@ function _M:access()
       if not lim then
         ngx.log(ngx.ERR, "unknown limiter: ", limiter.name, ", err: ", initerr)
         error(self.logging_only, 500)
-        goto done
+        return
       end
 
       local rediserr
@@ -130,7 +130,7 @@ function _M:access()
       if not lim.dict then
         ngx.log(ngx.ERR, "failed to connect Redis: ", rediserr)
         error(self.logging_only, 500)
-        goto done
+        return
       end
 
       limiters[#limiters + 1] = lim
@@ -148,11 +148,11 @@ function _M:access()
     if comerr == "rejected" then
       ngx.log(ngx.WARN, "Requests over the limit.")
       error(self.logging_only, self.status_code_rejected)
-      goto done
+      return
     end
     ngx.log(ngx.ERR, "failed to limit traffic: ", comerr)
     error(self.logging_only, 500)
-    goto done
+    return
   end
 
   for i, lim in ipairs(limiters) do
@@ -173,7 +173,6 @@ function _M:access()
     ngx.sleep(delay)
   end
 
-  ::done::
 end
 
 local function checkin(_, ctx, time, semaphore, redis_url, logging_only)
@@ -188,14 +187,14 @@ local function checkin(_, ctx, time, semaphore, redis_url, logging_only)
       if not lim.dict then
         ngx.log(ngx.ERR, "failed to connect Redis: ", rediserr)
         error(logging_only, 500)
-        goto done
+        return
       end
     end
     local conn, err = lim:leaving(keys[i], latency)
     if not conn then
       ngx.log(ngx.ERR, "failed to record the connection leaving request: ", err)
       error(logging_only, 500)
-      goto done
+      return
     end
   end
 
@@ -203,7 +202,6 @@ local function checkin(_, ctx, time, semaphore, redis_url, logging_only)
     semaphore:post(1)
   end
 
-  ::done::
 end
 
 function _M:log()
--- a/gateway/src/apicast/policy/rate_limit/rate_limit.lua
+++ b/gateway/src/apicast/policy/rate_limit/rate_limit.lua
@@ -168,9 +168,9 @@ end
 local function checkin(_, ctx, time, semaphore)
   local limiters = ctx.limiters
   local keys = ctx.keys
+  local latency = tonumber(time)
 
   for i, lim in ipairs(limiters) do
-    local latency = tonumber(time)
     local conn, err = lim:leaving(keys[i], latency)
     if not conn then
       ngx.log(ngx.ERR, "failed to record the connection leaving request: ", err)
--- a/gateway/src/apicast/policy/rate_limit/rate_limit.lua
+++ b/gateway/src/apicast/policy/rate_limit/rate_limit.lua
@@ -90,6 +90,7 @@ function _M.new(config)
   self.config = config or {}
   self.limiters = config.limiters
   self.redis_url = config.redis_url
+  self.status_code_rejected = config.status_code_rejected or 429
 
   return self
 end
@@ -139,7 +140,7 @@ function _M:access()
   if not delay then
     if comerr == "rejected" then
       ngx.log(ngx.WARN, "Requests over the limit.")
-      return ngx.exit(429)
+      return ngx.exit(self.status_code_rejected)
     end
     ngx.log(ngx.ERR, "failed to limit traffic: ", comerr)
     return ngx.exit(500)
--- a/gateway/src/apicast/policy/rate_limit/rate_limit.lua
+++ b/gateway/src/apicast/policy/rate_limit/rate_limit.lua
@@ -138,7 +138,7 @@ function _M:access()
   local delay, comerr = limit_traffic.combine(limiters, keys, states)
   if not delay then
     if comerr == "rejected" then
-      ngx.log(ngx.ERR, "Requests over the limit.")
+      ngx.log(ngx.WARN, "Requests over the limit.")
       return ngx.exit(429)
     end
     ngx.log(ngx.ERR, "failed to limit traffic: ", comerr)
--- a/gateway/src/apicast/policy/rate_limit/rate_limit.lua
+++ b/gateway/src/apicast/policy/rate_limit/rate_limit.lua
@@ -44,7 +44,7 @@ local function init_limiter(config)
     end
   )
 
-  return lim, nil
+  return lim, limerr
 end
 
 local function redis_shdict(url)
--- a/gateway/src/apicast/policy/rate_limit/rate_limit.lua
+++ b/gateway/src/apicast/policy/rate_limit/rate_limit.lua
@@ -146,7 +146,7 @@ function _M:access()
   end
 
   for i, lim in ipairs(limiters) do
-    if lim.iscommitted and lim:is_committed() then
+    if lim.is_committed and lim:is_committed() then
       connections_committed[#connections_committed + 1] = lim
       keys_committed[#keys_committed + 1] = keys[i]
     end
@@ -165,12 +165,20 @@ function _M:access()
 
 end
 
-local function checkin(_, ctx, time, semaphore)
+local function checkin(_, ctx, time, semaphore, redis_url)
   local limiters = ctx.limiters
   local keys = ctx.keys
   local latency = tonumber(time)
 
   for i, lim in ipairs(limiters) do
+    if redis_url then
+      local rediserr
+      lim.dict, rediserr = redis_shdict(redis_url)
+      if not lim.dict then
+        ngx.log(ngx.ERR, "failed to connect Redis: ", rediserr)
+        return ngx.exit(500)
+      end
+    end
     local conn, err = lim:leaving(keys[i], latency)
     if not conn then
       ngx.log(ngx.ERR, "failed to record the connection leaving request: ", err)
@@ -188,7 +196,7 @@ function _M:log()
   local limiters = ctx.limiters
   if limiters and next(limiters) ~= nil then
     local semaphore = ngx_semaphore.new()
-    ngx.timer.at(0, checkin, ngx.ctx, ngx.var.request_time, semaphore)
+    ngx.timer.at(0, checkin, ngx.ctx, ngx.var.request_time, semaphore, self.redis_url)
     semaphore:wait(10)
   end
 end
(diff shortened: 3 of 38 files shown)


kevprice83:
@y-tabata thanks for the contribution. However, it would make more sense to use the [lua-resty-limit-traffic module](https://github.com/openresty/lua-resty-limit-traffic) to implement the rate limiting here. This will allow you to create true periods using the [count module](https://github.com/openresty/lua-resty-limit-traffic/blob/master/lib/resty/limit/count.md). The limits can be defined using the [traffic module](https://github.com/openresty/lua-resty-limit-traffic/blob/master/lib/resty/limit/traffic.md).

You can combine the **traffic module** with the [request module](https://github.com/openresty/lua-resty-limit-traffic/blob/master/lib/resty/limit/traffic.md) to utilise the _"leaky bucket"_ method to rate limit requests.

The idea overall should be to wrap the redis instance so that it is a shared cache and opt in, the local `dict` can be used if not redis. You can see a proof of concept [here](https://github.com/3scale/conn-limiter/commit/2210dccf9370bfe2a6bbdf3f8030ec21ed951c74). The shared redis memory would be written into the API shown [here](https://github.com/openresty/lua-nginx-module#ngxshareddict). The default should be the `shdict` and Redis used if configured.

The final policy should be easily extensible for future contributions.

cc @mikz if you want to add anything to my comments here.

ytabata:
@kevprice83 Thanks for your comment. I'm improving my code now and I have a concern.
As you said, the default should be the shdict and Redis used if configured.
Considering using the shdict, I have to define the size of the shdict by using lua_shared_dict.
The existing lua_shared_dict is defined in nginx.conf.liquid and shdict.conf, but if I define lua_shared_dict in these files, the number of limiters is limited.
Considering rate limiting to the service, I don't want to limit the number of limiters.
Do you have any idea?

y-tabata:
@kevprice83 @mikz I combined the lua-resty-limit-traffic module to implement the rate limiting.

gateway/src/apicast/policy/rate_limiting_to_service/rate_limiting_to_service.lua
```diff
@@ -0,0 +1,201 @@
...
+    lim.dict, rediserr = redis_shdict(self.redis_info.host, self.redis_info.port, self.redis_info.db)
+    if not lim.dict then
+      ngx.log(ngx.ERR, "failed to connect Redis: ", rediserr)
+      return ngx.exit(500)
+    end
+
+    limitters[#limitters + 1] = lim
+    keys[#keys + 1] = limitter.key
+
+    if limitter.limitter == "resty.limit.conn" then
```
mikz: I think this could be abstracted by checking the the limiter has `is_commited` key.

gateway/src/apicast/policy/rate_limiting_to_service/rate_limiting_to_service.lua
```diff
@@ -0,0 +1,201 @@
...
+
+end
+
+local function checkin(_, ctx, time, semaphore, redis_info)
+  local limitters = ctx.limitters
+  local keys = ctx.keys
+
+  for i, lim in ipairs(limitters) do
+    local rediserr
+    lim.dict, rediserr = redis_shdict(redis_info.host, redis_info.port, redis_info.db)
```
mikz: This shouldn't be needed as the limiters are already initialized and have correct redis object in the `.dict` property, right?

gateway/src/apicast/policy/rate_limiting_to_service/rate_limiting_to_service.lua
```diff
@@ -0,0 +1,201 @@
...
+
+function _M:access()
+  local limitters = {}
+  local keys = {}
+  local states = {}
+
+  local limitters_limit_conn = {}
+  local keys_limit_conn = {}
+
+  local limitters_limit_conn_committed = {}
```
mikz: We should try to minimize number of tables allocated per request. This would cause GC pressure. 
My preferred way would be not doing this at all per request and just initializing the limiters once when initializing the policy.

gateway/src/apicast/policy/rate_limiting_to_service/apicast-policy.json
```diff
@@ -0,0 +1,53 @@
...
+              "description": "Values to be applied",
+              "type": "array",
+              "items": {
+                "type": "integer"
+              }
+            }
+          }
+        }
+      },
+      "redis_info": {
```
mikz: I think it might be better to use just one property: redis_url.
That can embed authentication, database, port, ...
Wdyt @davidor ?

gateway/src/apicast/policy/rate_limiting_to_service/rate_limiting_to_service.lua
```diff
@@ -0,0 +1,201 @@
...
+  local keys_limit_conn_committed = {}
+
+  if not self.redis_info then
+    ngx.log(ngx.ERR, "No Redis information.")
+    return ngx.exit(500)
```
mikz: I think we want this policy to work without redis. It should just work with shared memory. Redis should be opt-in.

gateway/src/apicast/policy/rate_limiting_to_service/apicast-policy.json
```diff
@@ -0,0 +1,53 @@
...
+            },
+            "key": {
+              "description": "Key of Redis or Name of lua_shared_dict shm zone",
+              "type": "string"
+            },
+            "values": {
+              "description": "Values to be applied",
+              "type": "array",
+              "items": {
+                "type": "integer"
```
mikz: This is pretty non descriptive and won't have really useful UI.

IMO we could use `anyOf` and write a description for each limiter and describing each field (`burst`, `rate`, `delay`, ...).

gateway/src/apicast/policy/rate_limiting_to_service/rate_limiting_to_service.lua
```diff
@@ -0,0 +1,201 @@
...
+  end
+
+  for _, limitter in ipairs(self.limitters) do
+    local limit
+    local class_not_found = false
+    try(
+      function()
+        limit = require (limitter.limitter)
```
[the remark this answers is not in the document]
mikz: And possibly even writing functions to take the configuration and passing it properly (https://github.com/3scale/apicast/pull/648#discussion_r177011986)  to the initializers like:

```lua
local resty_limit_count = require('resty.limit.count')
local resty_limit_req = require('resty.limit.req')
local traffic_limiters = {
  fixed_window = function(config)
    return  resty_limit_count.new(shdict, config.rate, config.burst)
  end,
  connections = function(config)
    return resty_limit_conn.new(shdict, config.conn, config.burst, config.delay)
  end,
}
```

gateway/src/apicast/policy/rate_limiting_to_service/rate_limiting_to_service.lua
```diff
@@ -0,0 +1,201 @@
...
+    )
+    if class_not_found then
+      return ngx.exit(500)
+    end
+
+    local lim, limerr
+    local failed_to_instantiate = false
+    try(
+      function()
+        lim, limerr = limit.new(shdict_key, unpack(limitter.values))
```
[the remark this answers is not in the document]
y-tabata: If we use shared memory, and I move this to new, "shared dict not found" error occurs. 
Is my understanding correct?

y-tabata:
@mikz Thank you for reviewing. I pushed fixed code.

Regarding the part of initializing the limiter, considering using shared memory, if I move the initializing part to new, "shared dict not found" error occurs.
Is my understanding correct?

gateway/src/apicast/policy/rate_limiting_to_service/rate_limiting_to_service.lua
```diff
@@ -0,0 +1,196 @@
...
+    if comerr == "rejected" then
+      ngx.log(ngx.ERR, "Requests over the limit.")
+      return ngx.exit(429)
+    end
+    ngx.log(ngx.ERR, "failed to limit traffic: ", comerr)
+    return ngx.exit(500)
+  end
+
+  for i, lim in ipairs(limiters) do
+    if lim.iscommitted and lim:is_committed() then
```
mikz: This looks wrong.

gateway/src/apicast/policy/rate_limiting_to_service/rate_limiting_to_service.lua
```diff
@@ -0,0 +1,196 @@
...
+
+  local states = {}
+  local connections_committed = {}
+  local keys_committed = {}
+
+  local delay, comerr = limit_traffic.combine(limiters, keys, states)
+  if not delay then
+    if comerr == "rejected" then
+      ngx.log(ngx.ERR, "Requests over the limit.")
+      return ngx.exit(429)
```
mikz: This status will have to be configurable.

gateway/src/apicast/policy/rate_limiting_to_service/rate_limiting_to_service.lua
```diff
@@ -0,0 +1,196 @@
...
+    end
+    ngx.log(ngx.ERR, "failed to limit traffic: ", comerr)
+    return ngx.exit(500)
```
mikz: I think this might be another configuration option. I'm not sure that everyone wants to respond with an error when there is some issue with rate limiting. Some people might want to let the request go through.

gateway/src/apicast/policy/rate_limiting_to_service/rate_limiting_to_service.lua
```diff
@@ -0,0 +1,196 @@
...
+  end
+
+end
+
+local function checkin(_, ctx, time, semaphore)
+  local limiters = ctx.limiters
+  local keys = ctx.keys
+
+  for i, lim in ipairs(limiters) do
+    local latency = tonumber(time)
```
mikz: This can be just converted once before the loop.

gateway/src/apicast/policy/rate_limiting_to_service/rate_limiting_to_service.lua
```diff
@@ -0,0 +1,196 @@
...
+  self.redis_url = config.redis_url
+
+  return self
+end
+
+function _M:access()
+  local limiters = {}
+  local keys = {}
+
+  if not self.redis_url then
```
mikz: I don't get this. You could use several limiters with one shared dictionary, right?

gateway/src/apicast/policy/rate_limiting_to_service/rate_limiting_to_service.lua
```diff
@@ -0,0 +1,196 @@
...
+  end
+
+  local states = {}
+  local connections_committed = {}
+  local keys_committed = {}
+
+  local delay, comerr = limit_traffic.combine(limiters, keys, states)
+  if not delay then
+    if comerr == "rejected" then
+      ngx.log(ngx.ERR, "Requests over the limit.")
```
mikz: I don't think this is necessarily an "error".

examples/policies/rate_limiting_to_service_configuration.lua
```diff
@@ -0,0 +1,31 @@
+local policy_chain = require('apicast.policy_chain').default()
+
+local rate_limit_policy = require('apicast.policy.rate_limiting_to_service').new({
```
mikz: I think this policy can have way simpler name than `rate_limiting_to_service`. Just `rate_limit` should be enough.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -0,0 +1,222 @@
...
+
+local function checkin(_, ctx, time, semaphore, redis_url, logging_only)
+  local limiters = ctx.limiters
+  local keys = ctx.keys
+  local latency = tonumber(time)
+
+  for i, lim in ipairs(limiters) do
+    if redis_url then
+      local rediserr
+      lim.dict, rediserr = redis_shdict(redis_url)
```
y-tabata: This is needed because the connection has been already closed.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -0,0 +1,222 @@
...
+      error(self.logging_only, 500)
+      goto done
+    end
+
+    limiters[1] = lim
+    keys[1] = shdict_key
+
+  else
+    for _, limiter in ipairs(self.limiters) do
+      local lim, initerr = init_limiter(limiter)
```
y-tabata: > I meant to move the limiter initialization to _M.new so they are created just once per policy.

@mikz If I move the limiter initialization to _M.new, "shared dict not found" error occurs. Maybe you know this behavior, so you put the limiter initialization to _M.access [here](https://github.com/3scale/conn-limiter/blob/master/src/conn-limiter.lua#L44), right?

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -0,0 +1,222 @@
...
+  return self
+end
+
+function _M:access()
+  local limiters = {}
+  local keys = {}
+
+  if not self.redis_url then
+    -- Only one (the first) limiter is enable.
+    -- Key will be shdict_key ('limiter').
```
y-tabata: > I don't get this. You could use several limiters with one shared dictionary, right?

@mikz No. Although it is a tentative implementation, we can use only one limiter with one shared dictionary. When user defines 3 limiters ("limiter1","limiter2","limiter3"), only "limiter1" is enabled.

gateway/src/apicast/policy/rate_limit/apicast-policy.json
```diff
@@ -0,0 +1,93 @@
+{
+  "$schema": "http://apicast.io/policy-v1/schema#manifest#",
+  "name": "rate limit policy",
+  "description": ["This policy adds rate limit.",
+    ""
```
davidor: Why is the `""` needed ?

t/apicast-policy-rate-limit.t
```diff
@@ -0,0 +1,995 @@
...
+                      burst = 0,
+                      delay = 2
+                    }
+                  },
+                  redis_url = "redis://localhost:6379/1",
+                  logging_only = true
+                }
+              },
+              {
+                name = "apicast.policy.rate_limit",
```
davidor: This policy is duplicated.

t/apicast-policy-rate-limit.t
```diff
@@ -0,0 +1,995 @@
...
+                      conn = 1,
+                      burst = 0,
+                      delay = 2
+                    }
+                  },
+                  redis_url = "redis://localhost:6379/1"
+                }
+              },
+              {
+                name = "apicast.policy.rate_limit",
```
davidor: This policy is duplicated.

davidor: Please review this test file. I think that policies are duplicated in several tests.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -0,0 +1,222 @@
...
+  local lim, limerr
+  try(
+    function()
+      lim, limerr = traffic_limiters[config.name](config)
+    end,
+    function(e)
+      return nil, e
+    end
+  )
+  if not lim then
```
davidor: Can't we return `lim, limerr` and delete this ìf` ?

gateway/src/apicast/policy/rate_limit/rate_limit.lua
davidor: This initializes a Redis client instance on every request. It might be better to have a connection pool and fetch and return connections as needed, but maybe we can leave this for a future PR.

spec/policy/rate_limit/rate_limit_spec.lua
```diff
@@ -0,0 +1,175 @@
...
+      local config = {
+        limiters = {
+          {name = "connections", key = 'test1', conn = 1, burst = 1, delay = 2}
+        },
+        redis_url = 'redis://localhost:6379/1'
+      }
+      local rate_limit_policy = RateLimitPolicy.new(config)
+      rate_limit_policy:access()
+      rate_limit_policy:access()
+      assert.spy(ngx_sleep_spy).was_called_more_than(0.001)
```
davidor: This is not correct. `was_called_more_than` refers to the number of times the spy was called. However, what you need is to check that it was called with a value that is > 0.001.

Same in the test below this one.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -0,0 +1,222 @@
...
+  local keys = {}
+
+  if not self.redis_url then
+    -- Only one (the first) limiter is enable.
+    -- Key will be shdict_key ('limiter').
+    local lim, initerr = init_limiter(self.limiters[1])
+    if not lim then
+      ngx.log(ngx.ERR, "unknown limiter: ", initerr)
+      error(self.logging_only, 500)
+      goto done
```
davidor: You could call `return` and avoid the `goto`, but this is a matter of preference.

davidor:
Thanks for your contribution @y-tabata 

This looks great. Good job :+1: . However, in order to merge your PR we'll need to make sure that both all the unit and integration tests pass. There are other problems that @mikz and I commented, but those are minor and could be solved in a future PR.

Some busted tests are failing:

```
Rate limit policy .access rejected (req)
spec/policy/rate_limit/rate_limit_spec.lua:123: Function was not called with the arguments

Rate limit policy .access delay (conn)
spec/policy/rate_limit/rate_limit_spec.lua:147: Expected to be called more than 0.001 time(s), but was called 0 time(s)

Rate limit policy .access delay (req)
spec/policy/rate_limit/rate_limit_spec.lua:159: Expected to be called more than 0.001 time(s), but was called 0 time(s)

Rate limit policy .access rejected (count)
spec/policy/rate_limit/rate_limit_spec.lua:135: Function was not called with the arguments

Rate limit policy .access rejected (conn)
spec/policy/rate_limit/rate_limit_spec.lua:111: Function was not called with the arguments
```

Two of those (`Rate limit policy .access delay (conn)` and `Rate limit policy .access delay (req)`) are failing because of the problem with `was_called_more_than` that I pointed out in another comment.

All the others fail because you specified `redis_url = 'redis://localhost:6379/1'` in the config, but in the environment where the tests are run there isn't a redis configured, so we'll need to mock it. When I started a local Redis all the tests passed :+1:

spec/policy/rate_limit/rate_limit_spec.lua
```diff
@@ -0,0 +1,175 @@
...
+
+  describe('.access', function()
+    it('success with multiple limiters', function()
+      local config = {
+        limiters = {
+          {name = "connections", key = 'test1', conn = 20, burst = 10, delay = 0.5},
+          {name = "leaky_bucket", key = 'test2', rate = 18, burst = 9},
+          {name = "fixed_window", key = 'test3', count = 10, window = 10}
+        },
+        redis_url = 'redis://localhost:6379/1'
```
davidor: Some of the tests use `redis://localhost`. This will not work in some environments like the dockerized development one. 

No need to fix this now. This is just a note so we remember to fix it in a different PR. It would be good to use an env with the Redis host like we do in the integration tests with `TEST_NGINX_REDIS_HOST`.

t/apicast-policy-rate-limit.t
```diff
@@ -0,0 +1,995 @@
...
+      redis:connect('127.0.0.1', 6379)
+      redis:select(1)
+      redis:del("test7")
+    }
+  }
+
+--- pipelined_requests eval
+["GET /flush_redis","GET /"]
+--- error_code eval
+[200, 429]
```
davidor: I'm not sure about this one. `conn` is set to 1 in the config. This means that it will only allow 1 concurrent connection.
So if I understood this correctly, the first `GET /` should be a 200 for sure. Even if we added more, they could be 200s as well as long as there isn't more than 1 concurrent connection.

t/apicast-policy-rate-limit.t
```diff
@@ -0,0 +1,995 @@
...
+      redis:connect('127.0.0.1', 6379)
+      redis:select(1)
+      redis:del("test8")
+    }
+  }
+
+--- pipelined_requests eval
+["GET /flush_redis","GET /"]
+--- error_code eval
+[200, 503]
```
davidor: I don't think this is correct. `rate` is set to 1 in the config, so the first `GET /` should return 200 for sure. The second one would return 503 only if it was done in the same second, which is very likely in this test but might be a bit difficult to ensure that it's done that way.

t/apicast-policy-rate-limit.t
y-tabata: @davidor I define this policy twice here, because we can open 2 connections by "GET /".

t/apicast-policy-rate-limit.t
y-tabata: @davidor I define this policy twice here, because we can open 2 connections by "GET /".

t/apicast-policy-rate-limit.t
y-tabata: @davidor I define the policy twice here, so we can open 2 connections by "GET /", and the first "GET /" is a 429.

t/apicast-policy-rate-limit.t
y-tabata: @davidor I define the policy twice here, so we can open 2 connections by "GET /", and the first "GET /" is a 503.

t/apicast-policy-rate-limit.t
```diff
@@ -0,0 +1,995 @@
...
+      redis:connect('127.0.0.1', 6379)
+      redis:select(1)
+      redis:del("test9")
+    }
+  }
+
+--- pipelined_requests eval
+["GET /flush_redis","GET /"]
+--- error_code eval
+[200, 429]
```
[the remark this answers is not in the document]
y-tabata: @davidor I define the policy twice here, so we can count 2 by "GET /", and the first "GET /" returns 429.

t/apicast-policy-rate-limit.t
davidor: I see now why you duplicated the policy in the config in several of these tests.

I find this way of testing the policy a bit counter-intuitive. To use this policy, users should include it just once and add there all the limiters they need. The problem I see with these tests is that the policy appears duplicated for testing convenience, but the configuration does not reflect a realistic scenario, because users will include the policy just once.

I think we should try to include the policy just once and change the tests so they work with that configuration. For example, in this particular test, we can make 3 requests. The first one to `/flush_redis`  (to make sure we start from a clean state), the second one to `/`(which should return 200), and the third one to `/` (which should return 503 because we defined a limit of 1 in the configuration).

What do you think?

gateway/src/apicast/policy/rate_limit/rate_limit.lua
davidor: Sorry, I don't understand this either.
Why is it a problem to allow several limiters?

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -0,0 +1,217 @@
...
+      keys[#keys + 1] = limiter.key
+
+    end
+  end
+
+  local states = {}
+  local connections_committed = {}
+  local keys_committed = {}
+
+  local delay, comerr = limit_traffic.combine(limiters, keys, states)
```
davidor: When using Redis and several limiters, the call to `combine()` will run several Redis commands.

For efficiency, we might want to find a way to send all those commands in a single network trip using pipelines: https://github.com/openresty/lua-resty-redis#init_pipeline

This is an optimization we can leave for a future PR.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
y-tabata: @davidor If we allow to use several limiters with shdict, we have to define the size of the shdict by using lua_shared_dict. If a user use 3 limiters, we have to define 3 lua_shared_dict, and if a user use 10 limiters, we have to define 10 lua_shared_dict. I don't want to limit the number of limiters, but there is no good way to define lua_shared_dict automatically as many as a user defines. So I defined only one lua_shared_dict tentatively.

t/apicast-policy-rate-limit.t
y-tabata: @davidor Your method works well for "leaky_bucket" and "fixed_window", but doesn't work well for "connections", I think. Regarding "connections", the second request returns 200 and then the number of connections is reset to zero, so the third request also returns 200.
However, I strongly agree with that the configuration does not reflect a realistic scenario in these tests, so I will change the tests of "leaky_bucket" and "fixed_window".

gateway/src/apicast/policy/rate_limit/rate_limit.lua
y-tabata: @davidor I'd like to leave this for a future PR because it's a little bit difficult.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
y-tabata: @davidor I'd like to leave this for a future PR because it's a little bit difficult.

t/apicast-policy-rate-limit.t
mikz: This policy has to be reusable by multiple services at the same time. It will be loaded and active more than once in the same time.

We can't be defining shdicts per policy, so they all have to reuse one shdict. If some parts do not support that we can't support them. I think they should be supported because we control the rate limiting keys and can for example prefix them with the rate limiter name, so they are unique per rate limiter.

This is duplicated discussion of https://github.com/3scale/apicast/pull/648#discussion_r178231827.

y-tabata: @mikz Let me confirm about your comment. Do you mean I should change the integration test as following?

before:
policy_chain = {{name = "apicast.policy.rate_limit", ...},{name = "apicast.policy.rate_limit", ...}}

after:
policy_chain = {{name = "apicast.policy.rate_limit", 
configuration = {limiters = {name = "connections", ...},{name = "connections", ...}}}}

mikz: I'm not really saying that.

I'm saying there can be several instances of connection/leaky bucket/fixed window rate limiting. APIcast can run several services and in case of our Cloud Hosted deployment it runs several thousands of services and configurations - each of them could use one or more rate limiting policies - and still have to be using just one or two shared dictionaries.

No matter how many rate limiting policies you have they have to be able to run from ideally one shared dictionary. 

Because each policy can have unique rate limiting "key" ( https://github.com/openresty/lua-resty-limit-traffic/blob/master/lib/resty/limit/req.md#incoming , passed to .combine as table of keys https://github.com/openresty/lua-resty-limit-traffic/blob/master/lib/resty/limit/traffic.md#combine) they could operate on same shared dictionary, but just using different keys.

Now - defining those keys should be left to the user so they can implement for example cross service rate limiting. But if the same key being used between fixed window / leaky bucket / connection limiter causes trouble we can prefix those keys with limiter name. So user can still do cross service rate limiting but only with the same limiter (user setting key "foo" would use it as "fixed_window-foo" or "leaky_bucket-foo" or "connections-foo"). My understanding is those keys can't be shared between limiters anyway.

y-tabata: @mikz Now I defined only one shdict called 'limiter' in nginx.conf.liquid. But I have to let the user to use multiple limiters with one shdict, right? The way I came up with is using FFI cdata as a value (for example [here](https://github.com/openresty/lua-resty-limit-traffic/blob/master/lib/resty/limit/req.lua#L111)). FFI cdata has the actual key and the value of limiter. Is my image matching with yours?

y-tabata: @mikz I misunderstood how shdict works. Could you please review the latest commit?

davidor:
Great progress @y-tabata :+1: 

I think that the only remaining issue is that some integration tests are failing. Take a look at the `s2i-builder` task, tests 6,7,8, and 9 are failing. 

The reason is that in `location /flush_redis`, `TEST_NGINX_REDIS_HOST` is not defined and because of that `redis:connect()` fails.

davidor:
@y-tabata , the tests that I pointed out in my previous comment are failing for two reasons:

1) The `TEST_NGINX_REDIS_HOST` and `TEST_NGINX_REDIS_PORT` envs are nil in `location /flush_redis`. You need to fetch them like you did in the policy config. So for example, the `flush_redis` location for test 6 should look like this:

```
  location /flush_redis {
    content_by_lua_block {
      local redis_host = "$TEST_NGINX_REDIS_HOST" or '127.0.0.1'
      local redis_port = "$TEST_NGINX_REDIS_PORT" or 6379
      local redis = require('resty.redis'):new()
      redis:connect(redis_host, redis_port)
      redis:select(1)
      redis:del("test6_1", "test6_2", "test6_3")
    }
  }
```

2) `TEST_NGINX_REDIS_HOST` is set to "redis" in the test environment. We need to set a resolver to get the IP, otherwise the call to `redis:connect` will fail. We can do this the same way we do in other tests.
You just need to define 
```perl
$ENV{TEST_NGINX_RESOLVER} ||= `grep nameserver /etc/resolv.conf | awk '{print \$2}' | head -1 | tr '\n' ' '`;`
```
at the beginning of the file, just below the other 2 envs you already have.
And finally, we need to tell nginx to use the resolver with the `resolver`directive. So the complete config for the test 6 should be:
```
--- config
  resolver $TEST_NGINX_RESOLVER;

  location /flush_redis {
    content_by_lua_block {
      local redis_host = "$TEST_NGINX_REDIS_HOST" or '127.0.0.1'
      local redis_port = "$TEST_NGINX_REDIS_PORT" or 6379
      local redis = require('resty.redis'):new()
      redis:connect(redis_host, redis_port)
      redis:select(1)
      redis:del("test6_1", "test6_2", "test6_3")
    }
  }
```

I think this should make all the tests pass. Let me know if you need any help.

y-tabata:
@davidor Thank you for your advice! I will modify the tests.

y-tabata:
The reason why JSON schema validation failed is because "connections" limiter has 4 items. Other limiters have 3 items. I confirmed that the validation succeeded when I deleted "delay" item in apicast-policy json. Do you have any idea?

y-tabata:
Alternative I considered is the followings:
~~~
"limiters": {
  "description": "List of limiters to be applied",
  "type": "array",
  "items": {
    "type": "object",
    "properties": {
      "name": {},
      "key": {},
      "conn": {},
      "burst": {},
      "delay": {},
      "rate": {},
      "count": {},
      "window": {}
    }
  }
}
~~~

~~~
"connections": {
  "type": "array",
  "items": {}
},
"leaky_bucket": {
  "type": "array",
  "items": {}
},
"fixed_window": {
  "type": "array",
  "items": {}
}
~~~

davidor:
@y-tabata in tests 5 and 6 you specified `delay = 0.5`. The problem is that the schema says that delay should be an `integer`. Changing that to `number` fixes the issue:

```json
 "delay": {
   "type": "number",
    "description": "the default processing latency of a typical connection (or request)"
}
```

y-tabata:
> @y-tabata in tests 5 and 6 you specified delay = 0.5. The problem is that the schema says that delay should be an integer. Changing that to number fixes the issue:
> 
>  "delay": {
>    "type": "number",
>     "description": "the default processing latency of a typical connection (or request)"
> }

@davidor Thanks! I changed the schema.

gateway/src/apicast/policy/rate_limit/apicast-policy.json
```diff
@@ -0,0 +1,98 @@
...
+        "description": "List of limiters to be applied",
+        "type": "array",
+        "items": {
+          "oneOf": [{
+            "type": "object",
+            "properties": {
+              "name": {
+                "type": "string",
+                "enum": ["connections"],
+                "description": "limiting request concurrency (or concurrent connections)"
```
mikz: Would be good to start with capital letter. This applies to other comments too.

gateway/src/apicast/policy/rate_limit/apicast-policy.json
```diff
@@ -0,0 +1,98 @@
...
+              },
+              "key": {
+                "description": "Key of limiter",
```
mikz: This is not very useful description.

gateway/src/apicast/policy/rate_limit/apicast-policy.json
```diff
@@ -0,0 +1,98 @@
...
+            }
+          }, {
+            "type": "object",
+            "properties": {
+              "name": {
+                "type": "string",
+                "enum": ["fixed_window"],
+                "description": "limiting request counts"
+              },
+              "key": {
```
mikz: When we allow this to be a "template" in the future we will somehow have to distinguish between plain strings and templated strings.

We probably should change this to be more extensible in the future before the final release.

Also right now those keys have global scopes. That does not really work in our Cloud Hosted environment. We probably should offer a way to have "scope" of the key that could be either "global" or "service" and the key would be automatically namespaced.

gateway/src/apicast/policy/rate_limit/apicast-policy.json
```diff
@@ -0,0 +1,98 @@
...
+      },
+      "redis_url": {
+        "description": "URL of Redis",
+        "type": "string"
+      },
+      "status_code_rejected": {
+        "type": "integer",
+        "description": "the status code when requests over the limit, default 429"
+      },
+      "logging_only": {
```
mikz: I think this should not be boolean but rather enum like `error_handling` with options `log` and `exit`.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -0,0 +1,205 @@
...
+local function try(f, catch_f)
+  local status, exception = pcall(f)
+  if not status then
+    catch_f(exception)
+  end
+end
+
+local function init_limiter(config)
+  local lim, limerr
+  try(
```
mikz: This is not really idiomatic Lua and we really should not be expecting exceptions.

All resty code will return an error as a second return value, so we can just use that.
Otherwise our code should not crash because that would trash performance of the whole gateway.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -0,0 +1,205 @@
...
+
+local function redis_shdict(url)
+  local options = { url = url }
+  local redis, err = ts.connect_redis(options)
+  if not redis then
+    return nil, err
+  end
+
+  return {
+    incr = function(_, key, value, init)
```
mikz: These functions can be defined just once in the top closure and then get the `redis` reference from `self`.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -0,0 +1,205 @@
...
+
+  local states = {}
+  local connections_committed = {}
+  local keys_committed = {}
+
+  local delay, comerr = limit_traffic.combine(limiters, keys, states)
+  if not delay then
+    if comerr == "rejected" then
+      ngx.log(ngx.WARN, "Requests over the limit.")
+      error(self.logging_only, self.status_code_rejected)
```
mikz: Not sure if we want to let the requests through when "logging only". That would effectively disable the policy.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -0,0 +1,205 @@
...
+      return
+    end
+    ngx.log(ngx.ERR, "failed to limit traffic: ", comerr)
+    error(self.logging_only, 500)
+    return
+  end
+
+  for i, lim in ipairs(limiters) do
+    if lim.is_committed and lim:is_committed() then
+      connections_committed[#connections_committed + 1] = lim
```
mikz: This is getting number of elements in the table in each iteration, better to use `table.insert`.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -0,0 +1,205 @@
...
+    end
+  end
+
+  if next(connections_committed) ~= nil then
+    local ctx = ngx.ctx
+    ctx.limiters = connections_committed
+    ctx.keys = keys_committed
+  end
+
+  if delay >= 0.001 then
```
mikz: if `delay > 0 ` would be more precise.

spec/policy/rate_limit/rate_limit_spec.lua
```diff
@@ -0,0 +1,177 @@
...
+        limiters = {
+          {name = "connections", key = 'test1', conn = 20, burst = 10, delay = 0.5},
+          {name = "leaky_bucket", key = 'test2', rate = 18, burst = 9},
+          {name = "fixed_window", key = 'test3', count = 10, window = 10}
+        }
+      }
+      local rate_limit_policy = RateLimitPolicy.new(config)
+      rate_limit_policy:access()
+    end)
+    it('invalid redis url', function()
```
mikz: Would be nice to have some whitespace between the test blocks.

y-tabata:
@mikz @davidor Thank you for reviewing! So after merging, I will work on the comments.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
y-tabata: @mikz Do you mean `disable` option should be added to `error_handling`?

PR #649, opened 2018-03-08
Dockerized dev env: ensure correct dependencies are used

davidor:
The problem was that when mounting the volume, ./lua_modules was used instead of the one generated inside the container.

To solve the issue, lua_modules is moved to a tmp directory when building the container and we make sure to use it when running the tests in `script/test`.

### DIFF
--- a/Dockerfile-development
+++ b/Dockerfile-development
@@ -25,6 +25,8 @@ COPY Makefile ${USER_HOME}
 WORKDIR ${USER_HOME}
 
 RUN make dependencies
+RUN mv lua_modules /tmp/lua_modules
+
 RUN cpanm --installdeps ./gateway
 
 COPY . ${USER_HOME}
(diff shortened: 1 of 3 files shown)


script/test
```diff
@@ -5,12 +5,16 @@ PROJECT_PATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && cd .. && pwd )"
 ROVER="/usr/local/openresty/luajit/bin/rover"
 
 function run_busted_tests {
+    sudo cp -R /tmp/lua_modules .
```
maneta: I think that this would be faster using `rsync` so it would merge the differences something like `rsync -az  /tmp/lua_modules .`

davidor: Not sure it's worth in this case, the time spent copying the files is _nothing_ compared to the time spent running the integration tests.

maneta: ok fair enough

Dockerfile-development
```diff
@@ -25,6 +25,8 @@ COPY Makefile ${USER_HOME}
 WORKDIR ${USER_HOME}
 
 RUN make dependencies
+RUN mv lua_modules /tmp/lua_modules
+
 RUN cpanm --installdeps ./gateway
 
 COPY . ${USER_HOME}
```
mikz: We could add `lua_modules` to `.dockerignore` so it is not copied?

davidor: Yes, it's because of the volume.

PR #651, opened 2018-03-12
Allow setting cache loader = boot on staging and never reload the config [THREESCALE-756] 

davidor:
APIcast crashes when it starts with:

```
THREESCALE_DEPLOYMENT_ENV: staging
APICAST_CONFIGURATION_LOADER: boot
APICAST_CONFIGURATION_CACHE: ''
```

Error msg:

```
[lua] configuration_loader.lua:139: init(): cache is off, cannot store configuration, exiting
```

This is because by default, the staging deployment env sets `configuration_cache` to 0, which is incompatible with the 'boot' mode. This PR solves the issue by avoiding setting that default.

Ref: https://issues.jboss.org/browse/THREESCALE-756

### DIFF
--- a/gateway/src/apicast/cli/command/start.lua
+++ b/gateway/src/apicast/cli/command/start.lua
@@ -110,15 +110,10 @@ local function openresty_binary(candidates)
 end
 
 local function build_env(options, config, context)
-    local configuration_cache = options.cache or context.configuration_cache
-    if configuration_cache then -- Avoid calling tostring() on nil, would return "nil".
-        configuration_cache = tostring(configuration_cache)
-    end
-
     return {
         APICAST_CONFIGURATION = options.configuration or context.configuration,
-        APICAST_CONFIGURATION_LOADER = tostring(options.configuration_loader or context.configuration_loader or 'lazy'),
-        APICAST_CONFIGURATION_CACHE = configuration_cache,
+        APICAST_CONFIGURATION_LOADER = options.configuration_loader or context.configuration_loader or 'lazy',
+        APICAST_CONFIGURATION_CACHE = options.cache or context.configuration_cache,
         THREESCALE_DEPLOYMENT_ENV = context.configuration_channel or options.channel or config.name,
         APICAST_POLICY_LOAD_PATH = concat(options.policy_load_path or context.policy_load_path, ':'),
     }
--- a/gateway/src/apicast/cli/command/start.lua
+++ b/gateway/src/apicast/cli/command/start.lua
@@ -110,10 +110,15 @@ local function openresty_binary(candidates)
 end
 
 local function build_env(options, config, context)
+    local configuration_cache = options.cache or context.configuration_cache
+    if configuration_cache then -- Avoid calling tostring() on nil, would return "nil".
+        configuration_cache = tostring(configuration_cache)
+    end
+
     return {
         APICAST_CONFIGURATION = options.configuration or context.configuration,
         APICAST_CONFIGURATION_LOADER = tostring(options.configuration_loader or context.configuration_loader or 'lazy'),
-        APICAST_CONFIGURATION_CACHE = tostring(options.cache or context.configuration_cache or 0),
+        APICAST_CONFIGURATION_CACHE = configuration_cache,
         THREESCALE_DEPLOYMENT_ENV = context.configuration_channel or options.channel or config.name,
         APICAST_POLICY_LOAD_PATH = concat(options.policy_load_path or context.policy_load_path, ':'),
     }
--- a/gateway/src/apicast/configuration_loader.lua
+++ b/gateway/src/apicast/configuration_loader.lua
@@ -60,7 +60,7 @@ end
 _M.mock = mock_loader.save
 
 local function ttl()
-  return tonumber(env.get('APICAST_CONFIGURATION_CACHE'), 10)
+  return tonumber(env.value('APICAST_CONFIGURATION_CACHE') or 0, 10)
 end
 
 function _M.global(contents)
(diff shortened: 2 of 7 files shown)


gateway/src/apicast/cli/command/start.lua
```diff
@@ -109,11 +109,28 @@ local function openresty_binary(candidates)
...
 end
 
+-- Assign different defaults to configuration_cache depending on
+-- configuration_loader. When it's lazy, we want 0 so it reloads on every
+-- request. However 0 it's incompatible with 'boot' mode, so we just keep
+-- it as nil.
+local function default_config_cache(options, context, configuration_loader)
+    local configuration_cache = options.cache or context.configuration_cache
+
+    if not configuration_cache and configuration_loader == 'lazy' then
```
mikz: Can't we delegate this to the environment configurations ?

davidor: Yes, that's better :+1:

gateway/src/apicast/cli/command/start.lua
```diff
@@ -110,10 +110,15 @@ local function openresty_binary(candidates)
 end
 
 local function build_env(options, config, context)
+    local configuration_cache = options.cache or context.configuration_cache
+    if configuration_cache then -- Avoid calling tostring() on nil, would return "nil".
+        configuration_cache = tostring(configuration_cache)
```
mikz: Maybe it is not necessary to call `tostring` here because it is done in the `update_env` function?

mikz:
@davidor Can we try to test this? Would blackbox test with env do the trick?

davidor:
@mikz : I tried and it didn't work. I've been investigating, and I found that `THREESCALE_DEPLOYMENT_ENV` is not respected in the integration tests. `config/production.lua` is used instead of `config/staging.lua` even with `TREESCALE_DEPLOYMENT_ENV=staging`.

It's read here when it hasn't been set yet:
https://github.com/3scale/apicast/blob/69ff05c03e934daf1fa3a9b2cfb4231773b5a22a/gateway/src/apicast/cli/command/start.lua#L184

I'll try to fix it.

davidor:
@mikz It turns out that it was easier than I thought, using `$ENV` instead of `env_to_apicast` was enough. Check the new regression test. I verified that it fails in master.

gateway/src/apicast/configuration_loader.lua
```diff
@@ -60,7 +60,7 @@ end
 _M.mock = mock_loader.save
 
 local function ttl()
-  return tonumber(env.get('APICAST_CONFIGURATION_CACHE'), 10)
+  return tonumber(env.get('APICAST_CONFIGURATION_CACHE') or 0, 10)
```
mikz: `env.get` can return empty string, so it should do the `or` after `tonumber` or use `env.value` ?

t/configuration-loading-boot-staging.t
```diff
@@ -0,0 +1,29 @@
+use lib 't';
+use Test::APIcast::Blackbox 'no_plan';
+
+$ENV{THREESCALE_DEPLOYMENT_ENV} = 'staging';
+$ENV{APICAST_CONFIGURATION_LOADER} = 'boot';
+$ENV{APICAST_CONFIGURATION_CACHE} = '';
+
+filters { configuration => 'fixture=echo.json' };
```
mikz: ❤️

gateway/src/apicast/cli/command/start.lua
davidor: I think it's needed. The result of this method is not only used in `update_env`, it's also used in the call to `exec` at the end of `mt:__call`.

davidor: https://github.com/3scale/apicast/blob/30984b947c3abaef6335d5dbb591e0136c07f247/gateway/src/apicast/cli/command/start.lua#L174

mikz: But `nil` keys are being removed from the table, so it won't even get there.

davidor: `update_env` does not mutate the table that it receives in the params. so if we do not apply `tostring` here, it will never be applied before it reaches that `exec`.

gateway/src/apicast/configuration_loader.lua
davidor: Right. Changed :+1:

gateway/src/apicast/cli/command/start.lua
mikz: exec does not call `tostring` on the value: https://github.com/3scale/lua-resty-execvp/blob/bfa07e3f82fdd8bfb41be5c152438f03ccd60b9f/src/resty/execvp.lua#L48
and `update_env` does: https://github.com/3scale/apicast/blob/30984b947c3abaef6335d5dbb591e0136c07f247/gateway/src/apicast/cli/command/start.lua#L47 

that means the values in this table should be either a `string` or `nil`. 
lua tables don't have `nil` values, so assigning nil will make the key disappear so it is never iterated by `pairs`

davidor: Not sure I follow. I probably didn't explain this well. Let me try:
1) `build_env` constructs a table `t` and returns it. We're discussing whether we should call `tostring` in some values of `t` here or not. I think we should.
2) The table `t` returned in the previous step is sent to `update_env`. But that function _does not_ mutate `t`.
3) `t` is sent as a param of the call to `exec`. As 2) didn't modify it, it will be sent as we left it in 1). That's why we need to call `tostring()` in `build_env`. Otherwise, `exec` might receive a table that contains things that are not strings.

mikz: Looks like I was really confused 😄 


We are saying the same. This table now needs to contain `nil` or `string` and that is what you are ensuring. 

I think it would be good if we could clean it up by changing the `exec` to serialize the env: 
https://github.com/3scale/lua-resty-execvp/blob/bfa07e3f82fdd8bfb41be5c152438f03ccd60b9f/src/resty/execvp.lua#L48

Then we don't have to really think about the serialization anymore.

davidor: Right. I think that fixes the root problem: https://github.com/3scale/lua-resty-execvp/pull/1

PR #655, opened 2018-03-13
Require summary in policy manifests and add it to the ones that do not have one

davidor:
This PR makes summary mandatory in the policy manifests.
The reason is that users of the UI might be confused if they see a blank space in the list of policies instead of a summary.


mikz:
Maybe we should require the summary then?

davidor:
Yes, I think we should make `summary` (shorter) required and `description` (longer) optional.

vramosp:
UI limits summary at 75 characters I believe.

davidor:
@vramosp We are enforcing that in our `manifest-schema.json` too: https://github.com/3scale/apicast/blob/69ff05c03e934daf1fa3a9b2cfb4231773b5a22a/gateway/src/apicast/policy/manifest-schema.json#L42

ddcesare:
As we talked yesterday with @davidor and @thomasmaas, we decided `summary` should be mandatory, and `description`.

PR #657, opened 2018-03-16
Add option to never reload configuration when configuration loader is lazy

davidor:
It's not possible to use `APICAST_CONFIGURATION_LOADER=lazy` and never refresh the configuration. This PR solves the issue by allowing setting negatives values in `APICAST_CONFIGURATION_CACHE`.

### DIFF
--- a/spec/configuration_store_spec.lua
+++ b/spec/configuration_store_spec.lua
@@ -151,20 +151,24 @@ describe('Configuration Store', function()
 
     it('returns stale records by default', function()
       local store = configuration.new()
-      local service =  { id = '21', hosts = { 'example.com', 'localhost' } }
+      local service =  { id = '21', hosts = { 'example.com' } }
 
-      store:add(service, -1)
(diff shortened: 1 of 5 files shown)


spec/configuration_store_spec.lua
```diff
@@ -151,20 +151,24 @@ describe('Configuration Store', function()
...
       assert.same({ service }, store:find_by_host('example.com'))
     end)
 
     it('does not return stale records when disabled', function()
       local store = configuration.new()
-      local service =  { id = '21', hosts = { 'example.com', 'localhost' } }
+      local service =  { id = '21', hosts = { 'example.com' } }
 
-      store:add(service, -1)
+      -- Can't add stale info with config.add. Need to do it through the
```
mikz: We could add one with expiry 0 ?

davidor: I just tried it and it does not work. I guess it needs a bit of time to delete it even with ttl=0.

PR #658, opened 2018-03-19
OIDC: verify typ is Bearer when included

davidor:
This is Keycloak specific. It verifies 'typ' included in the token payload only when it's present.

Closes https://github.com/3scale/apicast/issues/643

### DIFF
--- a/spec/oauth/oidc_spec.lua
+++ b/spec/oauth/oidc_spec.lua
@@ -135,6 +135,43 @@ describe('OIDC', function()
       assert.match('invalid alg', err, nil, true)
       assert.falsy(credentials, err)
     end)
+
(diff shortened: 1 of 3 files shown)


spec/oauth/oidc_spec.lua
```diff
@@ -135,6 +135,24 @@ describe('OIDC', function()
...
+          aud = 'notused',
+          azp = 'ce3b2e5e',
+          nbf = 0,
+          exp = ngx.now() + 10,
+          typ = 'Not-Bearer'
+        },
+      })
+
+      local credentials, _, err = oidc:transform_credentials({ access_token = access_token })
+      assert.falsy(credentials, err)
```
mikz: should we match the `err` ?

PR #659, opened 2018-03-19
release: v3.2.0-beta2

davidor:


davidor:
The build is failing because we're referencing a version in the readme that does not exist yet, so it's safe to merge.

PR #661, opened 2018-03-20
release: v3.2.0-beta3

davidor:


davidor:
Travis build is failing because the changelog references tags that will be created after this is merged (`3.2.0-beta3`).

PR #663, opened 2018-04-04
policy manifests: replace descriptions with titles in OneOfs

davidor:
This is a change proposed by @ddcesare 

The reason is that the library used in the UI to render policy configs does not show descriptions, but it shows titles.

From an APIcast point of view I don't see a difference, so I think this should be OK.


davidor:
Added a changelog entry.

PR #664, opened 2018-04-09
script/test: set redis host when running busted tests

davidor:
Busted tests are failing in the dockerized development environment.

The reason is that the unit tests of the rate limit policy use `TEST_NGINX_REDIS_HOST`, which was only set in `script/test` when running the `Test::Nginx` tests.

This PR simply sets that env when running busted tests.

### DIFF
--- a/docker-compose-devel.yml
+++ b/docker-compose-devel.yml
@@ -7,5 +7,11 @@ services:
     entrypoint: "bash"
     volumes:
       - .:/home/centos/
+    environment:
+      TEST_NGINX_REDIS_HOST: redis
+      TEST_NGINX_BINARY: openresty
+      PROJECT_PATH: /home/centos
+      TEST_NGINX_APICAST_PATH: /home/centos/gateway
+      ROVER: /usr/local/openresty/luajit/bin/rover
   redis:
     image: redis
--- a/script/test
+++ b/script/test
@@ -1,13 +1,13 @@
 #!/bin/bash
 
 COMMAND=$1
-PROJECT_PATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && cd .. && pwd )"
-ROVER="/usr/local/openresty/luajit/bin/rover"
+PROJECT_PATH=${PROJECT_PATH-$( cd "$( dirname "${BASH_SOURCE[0]}" )" && cd .. && pwd )}
+ROVER=${ROVER-/usr/local/openresty/luajit/bin/rover}
 
 function run_busted_tests {
     sudo cp -R /tmp/lua_modules .
-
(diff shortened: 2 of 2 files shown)


script/test
```diff
@@ -7,7 +7,8 @@ ROVER="/usr/local/openresty/luajit/bin/rover"
 function run_busted_tests {
     sudo cp -R /tmp/lua_modules .
 
-    cd "${PROJECT_PATH}" || exit; ${ROVER} exec ./bin/busted
+    cd "${PROJECT_PATH}" || exit; \
+        TEST_NGINX_REDIS_HOST=redis ${ROVER} exec ./bin/busted
```
mikz: Shouldn't this env var be pushed from outside? Docker compose/makefile ?

davidor: Yes, what I meant is that if we make the change you propose, everyone should be able to use this script because `redis` will not be hardcoded and will be set depending on the user environment.

mikz: I see. Sorry.

davidor: I've moved the envs to the docker-compose file.

docker-compose-devel.yml
```diff
@@ -7,5 +7,11 @@ services:
     entrypoint: "bash"
     volumes:
       - .:/home/centos/
+    environment:
```
mikz: I think this might be better as:
```yaml
environment:
  TEST_NGINX_REDIS_HOST: redis
```

Using hash instead of array.

script/test
davidor: Yes. I removed the rest and forgot about this one. Fixed.

docker-compose-devel.yml
davidor: Both options should be equivalent. I've changed it because the hash version looks a bit nicer.

script/test
```diff
@@ -1,24 +1,22 @@
...
 function run_nginx_unit_tests {
     cd "${PROJECT_PATH}" || exit
 
     sudo cp -R /tmp/lua_modules .
 
-    sudo TEST_NGINX_APICAST_PATH="${PROJECT_PATH}/gateway" \
-         TEST_NGINX_BINARY=openresty \
-         TEST_NGINX_REDIS_HOST=redis \
-         ${ROVER} exec prove
+    sudo -E "${ROVER}" exec prove
```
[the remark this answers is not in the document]
davidor: I'm not sure. Then I'd need to run `sudo -E script/test` every time instead of just `script/test`.
Another solution would be to modify the image in a way that this does not require `sudo`.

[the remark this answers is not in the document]
mikz: `make development` can run the docker with `-u root` and then there is no need for sudo.
If we aim for this to be executed by users then it surely can't use sudo.

PR #665, opened 2018-04-10
Improve rate limit policy

y-tabata:
This PR improves the rate limit policy according to the comments of [PR #648](https://github.com/3scale/apicast/pull/648).

### DIFF
--- a/gateway/src/apicast/policy/rate_limit/apicast-policy.json
+++ b/gateway/src/apicast/policy/rate_limit/apicast-policy.json
@@ -17,23 +17,23 @@
               "name": {
                 "type": "string",
                 "enum": ["connections"],
-                "description": "limiting request concurrency (or concurrent connections)"
+                "description": "Limiting request concurrency (or concurrent connections)"
               },
               "key": {
-                "description": "Key of limiter",
+                "description": "The user key corresponding to the limiter object. This must be unique in the scope.",
                 "type": "string"
               },
               "conn": {
                 "type": "integer",
-                "description": "the maximum number of concurrent requests allowed"
+                "description": "The maximum number of concurrent requests allowed"
               },
               "burst": {
                 "type": "integer",
-                "description": "the number of excessive concurrent requests (or connections) allowed to be delayed"
+                "description": "The number of excessive concurrent requests (or connections) allowed to be delayed"
               },
               "delay": {
                 "type": "number",
-                "description": "the default processing latency of a typical connection (or request)"
+                "description": "The default processing latency of a typical connection (or request)"
               }
             }
           }, {
@@ -42,19 +42,19 @@
               "name": {
                 "type": "string",
                 "enum": ["leaky_bucket"],
-                "description": "limiting request rate"
+                "description": "Limiting request rate"
               },
               "key": {
-                "description": "Key of limiter",
+                "description": "The user key corresponding to the limiter object. This must be unique in the scope.",
                 "type": "string"
               },
               "rate": {
                 "type": "integer",
-                "description": "the specified request rate (number per second) threshold"
+                "description": "The specified request rate (number per second) threshold"
               },
               "burst": {
                 "type": "integer",
-                "description": "the number of excessive requests per second allowed to be delayed"
+                "description": "The number of excessive requests per second allowed to be delayed"
               }
             }
           }, {
@@ -63,19 +63,19 @@
               "name": {
                 "type": "string",
                 "enum": ["fixed_window"],
-                "description": "limiting request counts"
+                "description": "Limiting request counts"
               },
               "key": {
-                "description": "Key of limiter",
+                "description": "The user key corresponding to the limiter object. This must be unique in the scope.",
                 "type": "string"
               },
               "count": {
                 "type": "integer",
-                "description": "the specified number of requests threshold"
+                "description": "The specified number of requests threshold"
               },
               "window": {
                 "type": "integer",
-                "description": "the time window in seconds before the request count is reset"
+                "description": "The time window in seconds before the request count is reset"
               }
             }
           }]
@@ -87,11 +87,11 @@
       },
       "status_code_rejected": {
         "type": "integer",
-        "description": "the status code when requests over the limit, default 429"
+        "description": "The status code when requests over the limit, default 429"
       },
       "logging_only": {
         "type": "boolean",
-        "description": "if true, the request goes through when there is some issue with rate limiting, default false"
+        "description": "If true, the request goes through when there is some issue with rate limiting, default false"
       }
     }
   }
--- a/gateway/src/apicast/policy/rate_limit/apicast-policy.json
+++ b/gateway/src/apicast/policy/rate_limit/apicast-policy.json
@@ -11,7 +11,7 @@
         "description": "List of limiters to be applied",
         "type": "array",
         "items": {
-          "oneOf": [{
+          "anyOf": [{
             "type": "object",
             "properties": {
               "name": {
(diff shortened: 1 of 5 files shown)


gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -104,7 +83,7 @@ function _M:access()
   local keys = {}
 
   for _, limiter in ipairs(self.limiters) do
-    local lim, initerr = init_limiter(limiter)
+    local lim, initerr = traffic_limiters[limiter.name](limiter)
```
davidor: I'm not sure about this change.

`resty.limit` calls `assert` when instantiating some limiters. See this for example:
https://github.com/openresty/lua-resty-limit-traffic/blob/master/lib/resty/limit/req.lua#L56

I'd expect that the library returned that kind of errors in the second var of the response, but that's not the case. For that reason, I think that we should catch the error like we did before.

What do you think @mikz  ?

gateway/src/apicast/policy/rate_limit/apicast-policy.json
```diff
@@ -20,7 +20,7 @@
                 "description": "Limiting request concurrency (or concurrent connections)"
               },
               "key": {
-                "description": "Key of limiter",
+                "description": "The user key corresponding to the limiter object. This must be unique in the scope.",
```
davidor: This is just a detail, but I think we should change the `user key` part.  The reason is that one of the credentials that 3scale backend accepts is also called `user key`, and it's unrelated to this. In order to avoid confusion, I'd just call it `key`.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -147,16 +149,20 @@ local function checkin(_, ctx, time, semaphore, redis_url, logging_only)
...
+    red, rederr = redis_shdict(redis_url)
+    if not red then
+      ngx.log(ngx.ERR, "failed to connect Redis: ", rederr)
+      error(logging_only, 500)
+      return
     end
+  end
+
+  for i, lim in ipairs(limiters) do
+    lim.dict = red and red or lim.dict
```
davidor: This should be `red or lim.dict`.

gateway/src/apicast/policy/rate_limit/apicast-policy.json
```diff
@@ -80,16 +80,40 @@
...
+            "type": "object",
+            "properties": {
+              "type": {
+                "type": "string",
+                "enum": ["limits_exceeded"],
+                "description": "The error setting when requests over the limit"
+              },
+              "status_code": {
+                "type": "integer",
+                "description": "The status code when requests over the limit, default 429"
```
davidor: You can use the `default` field.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -61,19 +72,35 @@ local function redis_shdict(url)
...
+  if error_settings[type]["error_handling"] == "exit" then
+    return ngx.exit(error_settings[type]["status_code"])
+  end
+end
+
+local function init_error_settings(config_error_settings)
+  local error_settings = default_error_settings
+  if config_error_settings then
+    for _, error_setting in pairs(config_error_settings) do
+      if error_setting.type then
```
davidor: Do we need this check? I think the type should be required according to the schema.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -61,19 +72,35 @@ local function redis_shdict(url)
...
+        if error_setting.status_code then
```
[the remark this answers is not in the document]
davidor: You're right :+1: 
I was confused, sorry.

davidor:
Tests are failing because there are errors in the schema. The error is in the `s2i-builder` check:
```
gateway/src/apicast/policy/rate_limit/apicast-policy.json invalid
[ { keyword: 'type',
    dataPath: '.configuration.properties[\'limiters\'].items.oneOf[0].properties[\'$ref\']',
    schemaPath: '#/type',
    params: { type: 'object,boolean' },
    message: 'should be object,boolean' },
  { keyword: 'type',
    dataPath: '.configuration.properties[\'limiters\'].items',
    schemaPath: '#/type',
    params: { type: 'array' },
    message: 'should be array' },
  { keyword: 'anyOf',
    dataPath: '.configuration.properties[\'limiters\'].items',
    schemaPath: '#/properties/items/anyOf',
    params: {},
    message: 'should match some schema in anyOf' } ]
```

y-tabata:
@davidor The `deploy` error has nothing to do with this PR, isn't it?

davidor:
@y-tabata This looks good :+1: 

Could you squash some of your commits before merging, please? We try to keep a clean history, although we are not strict about it. In your PR, there are many commits that fix issues introduced in previous commits. For example,  the `remove redundant` commit fixes a problem introduced in `change to define redis just once` commit. So would be nice to squash those 2 before merging.

Just try to squash those that seem obvious, like the ones I mentioned, to clean the PR a bit. Do not spend too much time on this. Please let me know if you need any help with this.

Also, @mikz do you want to do a quick review before merging?

ytabata:
@davidor I squashed some of my commits. Is there any unnecessary commits?

davidor:
Thanks @y-tabata :+1:

gateway/src/apicast/policy/rate_limit/apicast-policy.json
```diff
@@ -11,29 +11,54 @@
...
+                    "default": "global",
+                    "oneOf": [{
+                      "enum": ["global"],
+                      "description": "Global scope, affecting to all services"
+                    }, {
+                      "enum": ["service"],
+                      "description": "Service scope, affecting to one service"
+                    }]
+                  },
+                  "service_name": {
```
mikz: I thought this won't be supplied by users but rather just automatically using the service_id available in the context. So all service scope limiters are automatically scoped and can't be shared between services. This could still be shared by multiple services when they fill the same identifier.

gateway/src/apicast/policy/rate_limit/apicast-policy.json
```diff
@@ -11,29 +11,54 @@
...
+                "type": "object",
+                "properties": {
+                  "name": {
+                    "type": "string",
+                    "description": "The name of the key, must be unique in the scope"
+                  },
+                  "scope": {
+                    "type": "string",
+                    "description": "Scope of the key",
+                    "default": "global",
```
mikz: I think the default should be service.

gateway/src/apicast/policy/rate_limit/apicast-policy.json
```diff
@@ -11,29 +11,54 @@
...
-                "description": "Key of limiter",
-                "type": "string"
+                "description": "The key corresponding to the limiter object",
+                "type": "object",
+                "properties": {
+                  "name": {
+                    "type": "string",
+                    "description": "The name of the key, must be unique in the scope"
+                  },
+                  "scope": {
```
mikz: I think we could extract this into shared definition in this schema an reuse it so it is not duplicated 3 times. 

See an example: 
https://github.com/3scale/apicast/blob/4ab2ac80d1bd6f952de4a0068932b5a992619e0b/gateway/src/apicast/policy/manifest-schema.json#L11-L21

mikz: Possibly even the whole `key` property could be shared between all limiters.

y-tabata:
@mikz @davidor Thank you for your review! I continue to improve the policy.

gateway/src/apicast/policy/rate_limit/apicast-policy.json
mikz: @y-tabata it should be:

```json
{ "properties": {
"key": {
      "$ref": "#/definitions/key"
    }
}
```

You have `"$ref"` as a property of `"properties"`, not `"key" property`.

y-tabata: @mikz Thank you!

PR #673, opened 2018-04-19
Fix bug in export() applied to policies in the local chain

### LINKED ISSUE #672: export() does not work correctly in policies included in the local chain
Policies can define an `export` function to expose the data they need to share with other policies. 

`export` works fine for policies defined in the global chain, but it does not for policies include in a local (service level) chain. Whatever is returned in `export` does not end in `context`.

When this bug is solved, we could also solve #671 by simply using `export()` instead of `rewrite()` in the caching policy to expose the shared cache handler.

davidor:
Fixes #672

### DIFF
--- a/gateway/src/apicast/policy/local_chain/local_chain.lua
+++ b/gateway/src/apicast/policy/local_chain/local_chain.lua
@@ -28,7 +28,11 @@ local function build_chain(context)
   local proxy = Proxy.new(context.configuration)
 
   context.proxy = context.proxy or proxy
-  context.policy_chain = find_policy_chain(context)
+
+  local policy_chain = find_policy_chain(context)
+  context.policy_chain = policy_chain
+
+  return policy_chain
 end
 
 -- forward all policy methods to the policy chain
@@ -45,9 +49,9 @@ end
 local rewrite = _M.rewrite
 
 function _M:rewrite(context, ...)
-  build_chain(context)
+  local policy_chain = build_chain(context)
 
-  context = LinkedList.readwrite(context, find_policy_chain(context):export())
+  context = LinkedList.readwrite(context, policy_chain:export())
 
   rewrite(self, context, ...)
 end
--- a/gateway/src/apicast/policy/local_chain/local_chain.lua
+++ b/gateway/src/apicast/policy/local_chain/local_chain.lua
@@ -2,6 +2,7 @@ local resty_env = require('resty.env')
 
 local policy = require('apicast.policy')
 local Proxy = require('apicast.proxy')
+local LinkedList = require('apicast.linked_list')
 local _M = policy.new('Local Policy Chain')
 
 local function build_default_chain()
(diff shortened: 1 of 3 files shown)


gateway/src/apicast/policy/local_chain/local_chain.lua
```diff
@@ -45,6 +46,9 @@ local rewrite = _M.rewrite
 
 function _M:rewrite(context, ...)
   build_chain(context)
```
mikz: Wouldn't it be nicer if `build_chain` would return the policy_chain object? Then we would not have to call `find_policy_chain` twice.

gateway/src/apicast/policy/local_chain/local_chain.lua
```diff
@@ -45,6 +46,9 @@ local rewrite = _M.rewrite
...
+
+  context = LinkedList.readwrite(context, find_policy_chain(context):export())
```
mikz: Shouldn't this  call `self:export()` instead of `find_local_policy_chain` ?

gateway/src/apicast/policy/local_chain/local_chain.lua
davidor: You're right. It looks more readable. Changed in new commit.

gateway/src/apicast/policy/local_chain/local_chain.lua
davidor: I don't see it. `export` is not defined in the local chain class.

mikz: I see. Would it make sense to add it to:
https://github.com/3scale/apicast/blob/85a14f2b5cecb199b856264df1b2e2de6b4cd9da/gateway/src/apicast/policy.lua#L11-L17

Then we should be able to just call it and it would create proper delegator.

davidor: I'm not sure. The case of `export` is a bit different from the rest of phases.

In the rest of phases, we just call that phase for each of the policies. `export` needs to build a list instead. Like in https://github.com/3scale/apicast/blob/869aa91a3e435cdb1f30cdc5fac698d3adba83f5/gateway/src/apicast/policy_chain.lua#L114

mikz: Only the PolicyChain merges those exports. The rest of policies just export something. I think that is exactly the same as the other phases. And if every policy would have an export function we could get rid of the `or`: 
https://github.com/3scale/apicast/blob/869aa91a3e435cdb1f30cdc5fac698d3adba83f5/gateway/src/apicast/policy_chain.lua#L120
Making it easier for LuaJIT.

No need to change this. I just thought it would be even nicer cleanup to treat all phases the same.

davidor: I'd leave this for a future PR.

davidor:
Rebased to solve the changelog conflicts.

PR #675, opened 2018-04-19
[THREESCALE-799] policy: document in schemas when the order in the chain matters

davidor:
Some policies only work properly when placed before or after the Apicast policy. Others have a different behavior depending on their position with respect to Apicast in the chain.

For the policies affected, this PR adds a clarification in the description of their schemas.

Ref: https://issues.jboss.org/browse/THREESCALE-799

### DIFF
--- a/gateway/src/apicast/policy/upstream/apicast-policy.json
+++ b/gateway/src/apicast/policy/upstream/apicast-policy.json
@@ -5,7 +5,9 @@
   "description":
     ["This policy allows to modify the host of a request based on its path. ",
      "It accepts regular expressions that, when matched against the path, ",
-     "replace it with a given string."],
+     "replace it with a given string. \n",
+     "When combined with the APIcast policy, the upstream policy should be ",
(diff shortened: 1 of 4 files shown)


gateway/src/apicast/policy/upstream/apicast-policy.json
```diff
@@ -5,7 +5,9 @@
   "description":
     ["This policy allows to modify the host of a request based on its path. ",
      "It accepts regular expressions that, when matched against the path, ",
-     "replace it with a given string."],
+     "replace it with a given string. \n",
+     "When combined with the Apicast policy, the upstream policy should be ",
```
mikz: It should be spelled as APIcast policy:
https://github.com/3scale/apicast/blob/99d17ba34fa3ad7b461049b45e25a1088560d809/gateway/src/apicast/policy/apicast/apicast-policy.json#L3

davidor: Right. Fixed.

PR #681, opened 2018-04-26
Fix title of the token introspection and the rate limit policies in their manifests

davidor:
A minor change proposed by @thomasmaas

This makes the title of the policy look nicer in the UI and more consistent with the rest of policies.


davidor:
I also changed the title of the rate limit policy.

PR #688, opened 2018-05-01
[policy] run __gc metamethod when policies are garbage collected

mikz:
This would allow doing cleanup when policies are reloaded by changed configuration (like shutting down timers).

### DIFF
--- a/gateway/src/apicast/configuration_loader.lua
+++ b/gateway/src/apicast/configuration_loader.lua
@@ -84,7 +84,9 @@ function _M.configure(configuration, contents)
   end
 
   if config then
-    return configuration:store(config, ttl())
+    configuration:store(config, ttl())
+    collectgarbage()
+    return config
   end
 end
 
--- a/gateway/src/apicast/gc.lua
+++ b/gateway/src/apicast/gc.lua
@@ -2,6 +2,8 @@
 -- works in "userdata". This module introduces a workaround to make it work
 -- with tables.
 
+local tab_clone = require "table.clone"
+
 local rawgetmetatable = debug.getmetatable
 local getmetatable = getmetatable
 local setmetatable = setmetatable
@@ -9,7 +11,7 @@ local newproxy = newproxy
 local ipairs = ipairs
 local pairs = pairs
 local pcall = pcall
-local table = table
+local remove = table.remove
 local unpack = unpack
 local error = error
 local tostring = tostring
@@ -17,7 +19,9 @@ local tostring = tostring
 local _M = {}
 
 local function original_table(proxy)
-  return rawgetmetatable(proxy).__table
+  local mt = rawgetmetatable(proxy)
+
+  return mt and mt.__table
 end
 
 local function __gc(proxy)
@@ -37,7 +41,7 @@ local function __call(proxy, ...)
   -- Try to run __call() and if it's not possible, try to run it in a way that
   -- it returns a meaningful error.
   local ret = { pcall(t, ...) }
-  local ok = table.remove(ret, 1)
+  local ok = remove(ret, 1)
 
   if ok then
     return unpack(ret)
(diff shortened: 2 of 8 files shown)


gateway/src/apicast/policy.lua
```diff
@@ -30,6 +30,29 @@ local function __eq(policy, other)
...
+    local policy = getmetatable(proxy).__index -- proxy always points to the policy
+    local mt = getmetatable(policy) or empty-- in case someone changed the policy metatable
+
+    local gc = mt.__gc or noop
+
+    return gc(policy)
+end
+
+local function metatable(policy)
+    local proxy = newproxy(true)
```
davidor: To be honest, I learnt about `newproxy` while reviewing this PR.

If I understood it correctly we need to do this to include `__gc` in the table. It looks like this is not needed in Lua 5.2+ and will not be needed in future versions of LuaJIT: https://github.com/LuaJIT/LuaJIT/issues/47

I think this merits a comment.

mikz: It definitely needs plenty of comments. 

I seriously doubt it will land in LuaJIT because it depends on https://github.com/LuaJIT/LuaJIT/issues/38. Even though LuaJIT/LuaJIT#47 is closed it is not implemented.

Another approach is to use void pointer from ffi and use ffi.gc like:

```lua
local ffi = require('ffi')
local p = ffi.new('void *')
 ffi.gc(p, function() print('__gc') end)
collectgarbage()
p = nil
collectgarbage()
```

gateway/src/apicast/configuration_loader.lua
```diff
@@ -219,6 +219,8 @@ function lazy.rewrite(configuration, host)
     ngx.log(ngx.WARN, 'failed to acquire lock to lazy load: ', host, ' error: ', err)
   end
 
+  collectgarbage()
```
davidor: In some configurations when reloading, this line is not executed. I think we need to call `collectgarbage` also in `refresh_configuration`, or somewhere else that runs when re-loading in 'boot' mode.

gateway/src/apicast/policy.lua
```diff
@@ -45,14 +47,18 @@ function _M.new(name, version)
     local mt = { __index = policy, __tostring = __tostring, policy = policy }
 
     function policy.new()
-        return setmetatable({}, mt)
+        local p = setmetatable({}, mt)
+
+        p.gc = GC.delegate(p)
```
davidor: Why assign to `.gc` ?

mikz: They have both pointer to the other, so their life is tied together.
When the `gc` object is GC'd it will call the `__gc` method on the policy metatable.

Using the `userdata` as proxy had significant performance issue in my early benchmarks, but can't reproduce it now, so they might have been invalid.

There are several other issues with the delegation on policies: calling `next` or `type`. Those are used by `inspect` and by some our code as well. We would have to patch `next` and `type` methods to rely on some other metamethods. 

I agree this is less than ideal and will explore the approach of allocating metatable per policy instance, so it can be GCd with the policy.

gateway/src/apicast/module.lua
```diff
@@ -39,7 +39,7 @@ local name = env.value('APICAST_MODULE') or 'apicast.policy.apicast'
 local ok, mod = prequire(name)
 
 if ok and mod then
-  if type(mod) == 'table' then
+  if type(mod) == 'table' or true then
```
davidor: This is always true.

mikz: I know. Needed to do this temporarily because `type(mod)` returns `userdata`.

gateway/src/apicast/configuration_loader.lua
davidor: Pushed a commit fixing this.

gateway/src/apicast/policy.lua
```diff
@@ -18,7 +18,8 @@ local PHASES = {
     'ssl_certificate',
 }
 
-local setmetatable = setmetatable
+local GC = require('apicast.gc')
+local setmetatable = GC.setmetatable
```
davidor: Maybe we should rename this var to avoid confusion.
When I read this diff for the first time I assumed that we were using the normal `setmetatable` function.

gateway/src/apicast/gc.lua
```diff
@@ -91,4 +93,28 @@ function _M.set_metatable_gc(t, metatable)
   return proxy
 end
 
+local delegate_mt = {
+  __gc = function(self)
+    local mt = getmetatable(self.table)
+
+    if mt and mt.__gc then return mt.__gc(table) end
```
davidor: This should be `self.table`.

gateway/src/apicast/policy.lua
mikz: Yep.

mikz:
@davidor sorry for rebasing your commits, but I think it is better when the relevant changes (code, tests) are in one commit and I needed to make a change anyway to cover the `self.table` issue.

davidor:
@mikz  No worries, I agree with that approach.

PR #693, opened 2018-05-08
.s2i/bin/assemble: do not add expiry time to master images

davidor:
`master` images expire after 7 days in quay. This PR solves the issue.

### DIFF
--- a/gateway/.s2i/bin/assemble
+++ b/gateway/.s2i/bin/assemble
@@ -2,7 +2,7 @@
 
 mkdir -p /tmp/.s2i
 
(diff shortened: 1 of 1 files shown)


mikz:
@davidor what about just rebuilding master every day? 

https://circleci.com/docs/2.0/configuration-reference/#schedule
https://circleci.com/docs/2.0/workflows/#scheduling-a-workflow

davidor:
We could add that too. I think we could do both. Sometimes builds fail because of things like not being able to download a dependency. I think that removing the expire option can be helpful in those cases.

Issue #695, opened 2018-05-10
Validate that policy config schemas are compatible with react-jsonschema-form


davidor:
The 3scale UI uses `react-jsonschema-form` to build forms from the policy config schemas.

We need to add some kind of validation to make sure that apicast policy config schemas can be correctly rendered by that library. We have some limitations. For example, using an array of objects with `oneOf` does not work correctly: https://github.com/mozilla-services/react-jsonschema-form#q-does-rjsf-support-oneof-anyof-multiple-types-in-an-array-etc

In particular, the JSON schema of the rate limit policy is affected by this problem. The `limiters` and the `error_settings` properties in the schema are arrays that are not compatible with the `react-jsonschema-form` library.

There's a web tool to verify that a JSON schema can be rendered by the library: https://mozilla-services.github.io/react-jsonschema-form/

\cc @y-tabata

davidor:
I started a WIP branch to modify the rate limit schema and make it compatible with the `react-jsonschema-form` library:  https://github.com/3scale/apicast/tree/adapt-rate-limit-policy-schema-to-react-jsonschema

I still need to modify the code and the tests to adapt them to the new schema.

davidor:
The schema of the rate-limit policy was modified in https://github.com/3scale/apicast/pull/696

The issue of introducing an automated way of validating the schemas to make sure they're compatible with `react-jsonschema-form` remains open.

PR #696, opened 2018-05-10
Adapt rate limit config schema to make it compatible with react-jsonschema-form

davidor:
The rate limit policy has a valid configuration JSON schema. However, it cannot be rendered by `react-jsonschema-form` the library used in the 3scale UI to build HTML forms from the config schemas. 

This PR solves the issue by adapting the schema. The library cannot render arrays of objects with different types, so I've split them into several arrays that have items of only one type. In particular, the properties causing issues in the UI library were `limiters` and `error_settings`.

This is a breaking change, but not between final versions. If you were using this policy in a previous release candidate or beta, your configuration will become invalid.

Ref: https://github.com/3scale/apicast/issues/695


y-tabata:
@davidor I reviewed this PR and I created [PR #697](https://github.com/3scale/apicast/pull/697).
Others have no problem.

PR #697, opened 2018-05-11
examples/rate_limit: adapt to new config schema

y-tabata:
The example also needs to adopt to new config schema.


mikz:
Fixed by https://github.com/3scale/apicast/pull/698

PR #699, opened 2018-05-11
3.2.0 rc2

mikz:
Cherry picked from master:

*  #684
* #696
* #690
* #691


davidor:
Reminder: we'll also need to add the new version to the changelog and `version.lua`.

davidor:
We should add #698 too

Issue #667, opened 2018-04-16
Rate limit policy does not take into account changes in the config when using Redis


davidor:
The rate limit policy has some limitations when configured to use Redis.

For example:
- Create a `fixed window` limit for a key, with a given `count` and `window`.
- Restart the gateway with the same config, just changing the `window`.
- The policy ignores the change and will not adapt the TTL of the limiting keys to the new `window`.

or

- Start Apicast with a limit in the config.
- Restart the gateway without that limit.
- Add the same limit again and restart the gateway.
- Apicast will apply limits using the keys created in the first step. That means that for example, the count and the ttl are not reset.

In order to solve this problems, we need to store more information in Redis and perform some checks every time Apicast boots with the rate limit policy config. For example, the policy needs to check the existing limits and see if it needs to increase the TTL of the limiting keys, decrease it, remove them, etc.

y-tabata:
If it's acceptable to reset rate limits when the gateway restarts, we can add timestamp or a random string to the key in the initialization.

davidor:
@y-tabata I think it will depend a lot on the use case.

For limits with small periods (seconds, minutes) it might be acceptable. However, 3scale supports big periods for limits (like a month, or a year). In that case, resetting the counters and the limiters is not acceptable.

y-tabata:
Regarding `connections` and `leaky bucket`, it's no problem to reset rate limits because when the gateway restarts, there is no connection.

So we need to focus on `fixed window` only.
There are 5 cases on each parameter as illustrated in the following example.

window:

| old window | current | new window | expected window | reset? |
|---:|---:|---:|---:|:---:|
| 10 | 5 | 3 | 3 | yes |
| 10 | 5 | 5 | 5 | yes |
| 10 | 5 | 7 | 7 | yes |
| 10 | 5 | 10 | 5 | no |
| 10 | 5 | 15 | 15 | yes |

count:

| old count | current | new count | expected count | reset? |
|---:|---:|---:|---:|:---:|
| 10 | 5 | 3 | 3 | yes |
| 10 | 5 | 5 | 5 | yes |
| 10 | 5 | 7 | 7 | yes |
| 10 | 5 | 10 | 5 | no |
| 10 | 5 | 15 | 15 | yes |

Regarding `window`, if the new `window` is shorter than the current `window`, we have to reset the rate limit.
[...]

PR #703, opened 2018-05-14
Rate Limit policy to take into account changes in the config

y-tabata:
This PR addresses to Issue #667.

### DIFF
--- a/gateway/src/apicast/policy/rate_limit/rate_limit.lua
+++ b/gateway/src/apicast/policy/rate_limit/rate_limit.lua
@@ -80,12 +80,13 @@ local function init_error_settings(limits_exceeded_error, configuration_error)
   return error_settings
 end
 
-local function build_limiters_and_keys(type, limiters, redis, error_settings, context)
+local function build_limiters_and_keys(type, limiters, redis, error_settings, context, timestamp)
   local res_limiters = {}
   local res_keys = {}
 
   for _, limiter in ipairs(limiters) do
     local lim, initerr = traffic_limiters[type](limiter)
+    local window = limiter.window or 0
     if not lim then
       ngx.log(ngx.ERR, "unknown limiter: ", type, ", err: ", initerr)
       error(error_settings, "configuration_issue")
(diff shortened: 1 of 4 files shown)


mikz:
@y-tabata to help with the code review, could we leave out https://github.com/3scale/apicast/pull/703/commits/b04fbf2eaefa0d189de3978450209d14304b4386 to its own PR?

y-tabata:
@mikz Sure, I will divide this PR.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -160,6 +161,7 @@ function _M.new(config)
   self.redis_url = config.redis_url
   self.error_settings = init_error_settings(
     config.limits_exceeded_error, config.configuration_error)
+  self.timestamp = ngx.time()
```
mikz: There is no need to store time in `self.timestamp`. Local variable will be ok too.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -133,9 +134,9 @@ local function build_limiters_and_keys(type, limiters, redis, error_settings, co
     local key = limiter.template_string:render(
       ngx_variable.available_context(context))
     if limiter.key.scope == "global" then
-      key = format("%s_%s", type, key)
+      key = format("%d_%s_%s", timestamp + window, type, key)
```
mikz: You can't use just timestamp + window. In case of multiple gateways they all need to generate the same key. That can be achieved by generating a time bucket.

The formula is: `time - (time % window)`. That will for the whole time of `window` return the same value and then flip to the next one when the `window` is over.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
y-tabata: Fixed.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
y-tabata: Fixed.

PR #704, opened 2018-05-14
Improve rate limit policy

y-tabata:
This PR improves the rate limit policy according to the comments of PR #665.
From #703.

### DIFF
--- a/gateway/src/apicast/policy/rate_limit/apicast-policy.json
+++ b/gateway/src/apicast/policy/rate_limit/apicast-policy.json
@@ -23,7 +23,7 @@
                 "scope": {
                   "type": "string",
                   "description": "Scope of the key",
-                  "default": "global",
+                  "default": "service",
                   "oneOf": [{
                     "enum": ["global"],
                     "description": "Global scope, affecting to all services"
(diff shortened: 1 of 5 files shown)


gateway/src/apicast/policy/rate_limit/apicast-policy.json
```diff
@@ -23,18 +23,14 @@
                 "scope": {
                   "type": "string",
                   "description": "Scope of the key",
-                  "default": "global",
+                  "default": "service",
```
davidor: I think this change of behavior should be documented in the changelog.

davidor: This policy was included in a 'release candidate' version that can become final. If it does, this change will be included in the next release and be considered a breaking change. I think we should be clear about that in the changelog.

davidor:
Apart from what I commented above, the PR looks good to me :+1:

gateway/src/apicast/policy/rate_limit/apicast-policy.json
mikz: @davidor I'm not sure if it is a breaking change. Just changing the default value will affect only new instances of this policy.

davidor: You're right @mikz :+1: 
This is not a breaking change, sorry for the confusion.

PR #705, opened 2018-05-14
CircleCI: Don't try to push docker images in PRs from forks

davidor:
The `deploy` task in CircleCI fails in PRs opened from forks. The reason is that this task tries to push docker images to our registry and the docker credentials are not available in PRs from forks.
This PR solves the issue by skipping the `push` instruction when the credentials are not set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -221,7 +221,10 @@ jobs:
       - checkout
       - run: travis/install-s2i.sh /usr/local/bin # needs to be after checkout because it uses script from the repo
       - deploy:
+          # In PRs coming from forks, the docker credentials are not set.
+          # In that case, we just skip the step of pushing to the registry.
           command: |
+            ([ -z "${DOCKER_USERNAME}" ] || [ -z "${DOCKER_PASSWORD}" ]) && return 0;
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -104,8 +107,14 @@ jobs:
     machine: true
     steps:
       - checkout
-      - run: IMAGE_TAG=${CIRCLE_TAG:-${CIRCLE_BRANCH}} make benchmark
-      - run: IMAGE_TAG=master make benchmark
+      # If the docker creds are not set, the image will not be available, and
+      # we should skip the benchmark step.
+      - run:
+          command: |
+            if ([ -n "${DOCKER_USERNAME}" ] || [ -n "${DOCKER_PASSWORD}" ]); then
```
mikz: Those values are populated by setting `context: org-global` in the workflow definition. Benchmark never had access to these.

.circleci/config.yml
```diff
@@ -145,3 +148,6 @@ workflows:
       - benchmark:
           requires:
             - deploy
+          filters:
+            branches:
+              only: /^[^:]+$/
```
mikz: Might be good to have a comment there, it is not entirely obvious. I assume it looks for `:` in the branch name, right?

davidor:
:+1:

PR #707, opened 2018-05-14
lua port of benchmark-ips

mikz:
```
Warming up --------------------------------------
                 ffi
    708409 i/100ms

             syscall
    593019 i/100ms

Calculating -------------------------------------
 11385724.4 (±1.9%) i/s -   57381129 in   5.041731s
 10427682.8 (±1.0%) i/s -   52185672 in   5.005021s
```

It ports awesome https://github.com/evanphx/benchmark-ips to Lua.
The idea is that we have clear benchmarks for some constructs that can show performance difference.
I could see some LuaJIT features like turning off JIT for some functions to see the impact of JIT on performance. 

You can see benchmark results in https://github.com/3scale/apicast/pull/714#discussion_r188866775

### DIFF
--- a/benchmark/ips.lua
+++ b/benchmark/ips.lua
@@ -1,0 +1,388 @@
+local function to_i(str) return tonumber(str:gsub('[^%d]', ''), 10) end
+
+local NANOSECONDS_PER_SECOND = to_i('1 000 000 000')
+local NANOSECONDS_PER_100MS = to_i('100 000 000')
+local Timing = { }
+
+
+local ffi = require 'ffi'
+local tab_new = require "table.new"
+
+ffi.cdef [[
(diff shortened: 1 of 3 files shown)


benchmark/ips.lua
```diff
@@ -0,0 +1,388 @@
...
+function Timing.clean_env()
+  collectgarbage('collect')
+end
+
+-- Return the number of microseconds between the 2 moments
+function Timing.time_ns(before, after)
+  return after - before
+end
+
+-- Add one second to the time represenetation
```
davidor: Adds `s` seconds.

mikz:
@davidor yep, but when we get some milage on this.

PR #710, opened 2018-05-15
3scale batching policy: fix batching period assignment from config

davidor:
This PR fixes a bug in the 3scale batching policy. It didn't assign the batching period according to the given configuration.
While at it, this PR also adds some debugging logs and tests that helped to detect this issue.


t/apicast-policy-3scale-batcher.t
```diff
@@ -255,6 +255,7 @@ push $res, "Host: one";
 $res
 --- no_error_log
 [error]
+3scale batcher report timer got
```
mikz: 🕺

PR #716, opened 2018-05-17
Headers policy: allow templating using liquid

### LINKED ISSUE #708: Headers policy: allow templating using liquid
# Use Cases

Add a HTTP Header to the request that contains value parsed from the JWT used for authentication (https://github.com/3scale/apicast/pull/607#issuecomment-370784965).

Using HMAC to hash and sign some value and pass it upstream in the HTTP Request Header.

# Details

## We should offer both "plain" and "liquid" replacement
It is possible some clients will have "{{ }}" inside their replacement and that would cause an issue.
But maybe that change is negligible and we should always template the string by "liquid".
And last concern is performance - templating it by liquid right now could have performance impact before we make it all JITable.

## This can't be a breaking change to the policy configuration.

https://github.com/3scale/apicast/blob/5ec4848a8527143d3764d43bb0c447afdac5b52a/gateway/src/apicast/policy/headers/apicast-policy.json#L42-L45

We could let "value" to be a string or an object `{"type":"plain|liquid", "value": "..."}`, but that would not work with the UI library (https://github.com/3scale/apicast/issues/6950.

Another way would be to introduce another key "value_type" that is optional enum defaulting to "plain".

## Other policies will try to do the same. 

Try to keep similar interface and see how other policies need to use liquid templating. One other candidate is the rate limiting policy.

davidor:
Closes #708

### DIFF
--- a/gateway/src/apicast/template_string.lua
+++ b/gateway/src/apicast/template_string.lua
@@ -1,0 +1,116 @@
+local Liquid = require 'liquid'
+local LiquidTemplate = Liquid.Template
+local LiquidInterpreterContext = Liquid.InterpreterContext
+local LiquidFilterSet = Liquid.FilterSet
+local LiquidResourceLimit = Liquid.ResourceLimit
+local ngx = ngx
+
+local setmetatable = setmetatable
+local pairs = pairs
+local format = string.format
+
+local LiquidTemplateString = {}
+local liquid_template_string_mt = { __index = LiquidTemplateString }
+
+-- Expose only ngx.* functions that we think could be useful and that do not
+-- have any side-effects.
+
+-- Table of ngx.* functions that we think can be useful for templates and do
+-- not have any side-effects.
+-- @field escape_uri https://github.com/openresty/lua-nginx-module#ngxescape_uri
+-- @field unescape_uri https://github.com/openresty/lua-nginx-module#ngxunescape_uri
+-- @field encode_base64 https://github.com/openresty/lua-nginx-module#ngxencode_base64
+-- @field decode_base64 https://github.com/openresty/lua-nginx-module#ngxdecode_base64
+-- @field crc32_short https://github.com/openresty/lua-nginx-module#ngxcrc32_short
+-- @field crc32_long https://github.com/openresty/lua-nginx-module#ngxcrc32_long
+-- @field hmac_sha1 https://github.com/openresty/lua-nginx-module#ngxhmac_sha1
+-- @field md5 https://github.com/openresty/lua-nginx-module#ngxmd5
+-- @field md5_bin https://github.com/openresty/lua-nginx-module#ngxmd5_bin
+-- @field sha1_bin https://github.com/openresty/lua-nginx-module#ngxsha1_bin
+-- @field quote_sql_str https://github.com/openresty/lua-nginx-module#ngxquote_sql_str
+-- @field today https://github.com/openresty/lua-nginx-module#ngxtoday
+-- @field time https://github.com/openresty/lua-nginx-module#ngxtime
+-- @field now https://github.com/openresty/lua-nginx-module#ngxnow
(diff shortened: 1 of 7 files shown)


gateway/src/apicast/template_string.lua
```diff
@@ -0,0 +1,87 @@
...
+local setmetatable = setmetatable
+local pairs = pairs
+
+local LiquidTemplateString = {}
+local liquid_template_string_mt = { __index = LiquidTemplateString }
+
+-- Expose only ngx.* functions that we think could be useful and that do not
+-- have any side-effects.
+local liquid_filters = {
+  ngx_excape_uri = ngx.escape_uri,
```
mikz: I'm torn on this. 

On one hand customer should not care if it is `ngx` or what. All functions to `escape_uri` should behave the same.
On the other hand it is the nginx function and in theory we might provide some other in the future.

Also, would be good to add ldoc with links to those functions so we can see it in the documentation.

gateway/src/apicast/template_string.lua
```diff
@@ -0,0 +1,87 @@
...
+local _M = {}
+
+--- Initialize a template
+-- Initialize a liquid or a plain text template according to the given type.
+-- @tparam string value String to construct the template from
+-- @tparam string type Render the template as this type.
+--   Can be 'liquid' or 'plain'
+-- @treturn a template string
+function _M.new(value, type)
+  return template[type].new(value)
```
mikz: This will crash when it gets unknown type. It should handle missing types and return `nil, err` IMO.

mikz:
@davidor might be good as a verification to try to replicate https://github.com/nmasse-itix/3scale-apicast/blob/34a01da49ddb90dcbb85027b0d54be503d8ed236/gateway/src/custom/policy/explode_oidc_token_as_http_headers.lua with this.

gateway/src/apicast/template_string.lua
davidor: @mikz I had the same doubt while writing this code.

I'm thinking about users configuring this from 3scale's UI. I think that for them it'd be more natural to write something like 'md5' rather than 'ngx_md5'. In theory, they don't even need to know that apicast is based on nginx.

gateway/src/apicast/template_string.lua
davidor: Changed :+1:

gateway/src/apicast/template_string.lua
davidor: Changed.

mikz: Exactly 👍

davidor:
@mikz I added an integration test with jwt tokens as you suggested. Check the last 2 commits.
@nmasse-itix : you might be interested in the example. It's based in one of the custom policies of your apicast fork.

t/fixtures/policies/decode_oidc_token/1.0.0/decode_oidc_token.lua
```diff
@@ -0,0 +1,23 @@
+local http_authorization = require 'resty.http_authorization'
```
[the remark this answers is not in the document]
davidor: @mikz I'll explain this better.

In order to test that the templating introduced in this PR allows setting headers with some jwt information, I needed to create an example policy specifically for testing that. The reason is that none of the current policies expose a decoded jwt in the policies context.

After adding that example policy, I realized that we should probably expose the decoded jwt in the policies context so other policies can use it, for example, the rate-limit one. That's why I created a new branch that does that.

My suggestion is to delete the commits of this branch that implement the example policy and the test (last 2 commits), merge, and then, create a new PR from the new branch I created.

What do you think?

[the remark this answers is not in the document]
davidor: Done :+1:
I'll continue working on the other PR tomorrow.

Issue #713, opened 2018-05-15
Rate limit policy: template the rate limiting key by liquid


mikz:
# Use Cases

Rate limit by a Client IP address by using `{{ ngx.var.remote_addr }}` for the rate limiting key. This value would get evaluated and replaced with real value in the rewrite phase.

Rate limit by a value of parsed JWT by using `{{ jwt.sub }}` (to rate limit End Users for example).

# Details

Use liquid templating as in #708 and #711. 

# Concerns

Liquid library we use is not really written with LuaJIT in mind, so the code is not really JITable. We might need to optimize parts of it to get better performance.

PR #718, opened 2018-05-18
Expose jwt in the policies context

davidor:
This PR exposes jwt in the policies context in OAuth flows.

This is useful because it allows other policies to access those tokens. For example, the rate-limit policy ( #713 ) could benefit from this.

This PR also adds an integration test that uses OAuth and the headers policy and shows how it can set some headers with information extracted from the decoded jwt.

### DIFF
--- a/t/apicast-policy-headers.t
+++ b/t/apicast-policy-headers.t
@@ -1,6 +1,20 @@
 use lib 't';
 use Test::APIcast::Blackbox 'no_plan';
 
+use Cwd qw(abs_path);
+
+our $rsa = `cat t/fixtures/rsa.pem`;
+
+# Make fixtures policies available. There is a test that needs the "decode
+# oidc token" example policy.
+BEGIN {
+    $ENV{TEST_NGINX_APICAST_POLICY_LOAD_PATH} = 't/fixtures/policies';
+}
+
+env_to_apicast(
+    'APICAST_POLICY_LOAD_PATH' => abs_path($ENV{TEST_NGINX_APICAST_POLICY_LOAD_PATH}),
+);
+
 run_tests();
 
 __DATA__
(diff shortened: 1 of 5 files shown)


t/apicast-policy-headers.t
```diff
@@ -1,6 +1,20 @@
...
 use Test::APIcast::Blackbox 'no_plan';
 
+use Cwd qw(abs_path);
+
+our $rsa = `cat t/fixtures/rsa.pem`;
+
+# Make fixtures policies available. There is a test that needs the "decode
+# oidc token" example policy.
+BEGIN {
+    $ENV{TEST_NGINX_APICAST_POLICY_LOAD_PATH} = 't/fixtures/policies';
```
mikz: No longer needed?

davidor: Right. Fixed :+1:

t/apicast-policy-headers.t
```diff
@@ -518,3 +522,86 @@ yay, api backend
...
+      ngx.exit(200)
+    }
+  }
+--- configuration
+{
+  "oidc": [
+    {
+      "issuer": "https://example.com/auth/realms/apicast",
+      "config": {
+        "public_key": "MFwwDQYJKoZIhvcNAQEBBQADSwAwSAJBALClz96cDQ965ENYMfZzG+Acu25lpx2KNpAALBQ+catCA59us7+uLY5rjQR6SOgZpCz5PJiKNAdRPDJMXSmXqM0CAwEAAQ==",
```
davidor: @mikz Ideas to avoid hardcoding it here?

mikz: You can do ` | eval` filter and use Perl to read it and interpolate it.

mikz: this could possibly work:

```perl
 <<"JSON";
{ 
 "oidc": ...
   "public_key": "$::rsa"
}
JSON

davidor: We can't use `eval` in the configuration block. Maybe I'm missing something.

PR #719, opened 2018-05-22
Rate Limit policy: allow templating using liquid

### LINKED ISSUE #713: Rate limit policy: template the rate limiting key by liquid

y-tabata:
Closes #713.

### DIFF
--- a/gateway/src/apicast/policy/ngx_variable.lua
+++ b/gateway/src/apicast/policy/ngx_variable.lua
@@ -1,0 +1,15 @@
+local _M = {}
+
+local function context_values()
+  return {
+    uri = ngx.var.uri,
+    host = ngx.var.host,
+    remote_addr = ngx.var.remote_addr
+  }
+end
+
+function _M.available_context(policies_context)
+  return setmetatable(context_values(), { __index = policies_context })
+end
+
+return _M
--- a/gateway/src/apicast/policy/rate_limit/rate_limit.lua
+++ b/gateway/src/apicast/policy/rate_limit/rate_limit.lua
@@ -8,6 +8,7 @@ local resty_limit_count = require('resty.limit.count')
 local ngx_semaphore = require "ngx.semaphore"
 local limit_traffic = require "resty.limit.traffic"
 local ts = require ('apicast.threescale_utils')
+local ngx_variable = require ('apicast.policy.ngx_variable')
 local tonumber = tonumber
 local next = next
 local shdict_key = 'limiter'
@@ -17,6 +18,9 @@ local ipairs = ipairs
 local unpack = table.unpack
 local format = string.format
 local concat = table.concat
+
+local TemplateString = require 'apicast.template_string'
+local default_name_type = 'plain'
 
 local new = _M.new
 
@@ -110,7 +114,7 @@ local function init_error_settings(limits_exceeded_error, configuration_error)
   return error_settings
 end
 
-local function build_limiters_and_keys(type, limiters, redis, error_settings, service_id)
+local function build_limiters_and_keys(type, limiters, redis, error_settings, context)
   local res_limiters = {}
   local res_keys = {}
 
@@ -126,17 +130,25 @@ local function build_limiters_and_keys(type, limiters, redis, error_settings, se
 
     insert(res_limiters, lim)
 
-    local key
+    local key = limiter.template_string:render(
+      ngx_variable.available_context(context))
     if limiter.key.scope == "global" then
-      key = format("%s_%s", type, limiter.key.name)
+      key = format("%s_%s", type, key)
     else
-      key = format("%s_%s_%s", service_id, type, limiter.key.name)
+      key = format("%s_%s_%s", context.service.id, type, key)
     end
 
     insert(res_keys, key)
   end
 
   return res_limiters, res_keys
+end
+
+local function build_templates(limiters)
+  for _, limiter in ipairs(limiters) do
+    limiter.template_string = TemplateString.new(
+      limiter.key.name, limiter.key.name_type or default_name_type)
+  end
 end
 
 function _M.new(config)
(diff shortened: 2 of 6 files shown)


t/apicast-policy-rate-limit.t
```diff
@@ -984,3 +990,106 @@ Return 200 code.
 ["GET /","GET /"]
 --- error_code eval
 [200, 200]
+
+=== TEST 17: Liquid templating (jwt.aud).
+Return 429 code.
```
davidor: I think this should be more descriptive.

davidor:
@y-tabata Thanks for your contribution! :+1: 

Can you add a changelog entry, please? Something like `Liquid templating support in the rate-limit policy` under the `Added` section would be enough.

Also, could you add some unit tests for this? I see that you added some integration tests which is good :+1: , but we also need the unit ones. You can use the ones we added for the headers policy as an example: https://github.com/3scale/apicast/pull/716/

t/apicast-policy-rate-limit.t
```diff
@@ -1,9 +1,15 @@
 use lib 't';
 use Test::APIcast 'no_plan';
+use Cwd qw(abs_path);
 
+$ENV{TEST_NGINX_LUA_PATH} = "$Test::APIcast::spec/?.lua;$ENV{TEST_NGINX_LUA_PATH}";
```
davidor: Why is this needed?

y-tabata: In order to resolve `'fixtures.rsa'`.
I referenced [apicast-oidc.t](https://github.com/3scale/apicast/blob/master/t/apicast-oidc.t).

davidor: I see. I think we can find a better solution. I think these tests should make use of the `fixtures` directory in `t` instead of the one in `spec`. However, as you said, we're doing the same in `apicast-oidc.t`. We should fix this, but I think we can do it after this PR.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -151,7 +154,22 @@ end
...
+        local newstr, n, gsuberr = gsub(key_name, m[0], replace)
+        if gsuberr or n <= 0 then
+          break
+        end
+        key_name = newstr
+        m, _ = match(key_name, "{{ *(ngx\\.[^{}]+)}}")
+      end
+      limiter.key_name = key_name
+    else
+      limiter.key_name = limiter.template_string:render(context)
```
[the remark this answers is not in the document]
y-tabata: If we don't use the regex, we have to have the list like `uri = ngx.var.uri` or `host = ngx.var.host` and so on.
The best place is `liquid_filters` of `template_string.lua`, isn't it?
Do you have any idea to avoid the `"API disabled in the current context"` error?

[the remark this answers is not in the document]
davidor: I don't think `liquid_filters` in `template_string.lua` is the best place for this. Those are filters, functions like `md5`  that allow us to have things like  `{{ 'a_value' | md5 }}` in the policy config.

But `uri` and `host` are values, not filters. I think `uri` and `host` need to be included in the context passed to `render`. For that, I think a possible solution is to expose a table in `liquid_templates` with the values (not filters) that we want to expose. Something like:
```lua
_M.values = {
  uri = ngx.var.uri,
  host = ngx.var.host,
  -- possibly many others
}
```
The policies would be responsible for merging the policies context with this table and pass it to the context.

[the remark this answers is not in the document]
y-tabata: Changed.

gateway/src/apicast/policy/rate_limit/rate_limit.lua
```diff
@@ -139,6 +143,19 @@ local function build_limiters_and_keys(type, limiters, redis, error_settings, se
...
+local function build_templates(limiters)
+  for _, limiter in ipairs(limiters) do
+    limiter.template_string = TemplateString.new(
+      limiter.key.name, limiter.key.name_type or default_name_type)
+  end
+end
+
+local function interpolate_key_name(context, limiters)
+  for _, limiter in ipairs(limiters) do
+    limiter.key_name = limiter.template_string:render(context)
```
mikz: This mutates the `limiter` object which is shared across all requests. 

You can just pass `context` to `build_limiters_and_keys` and use it there without storing it in the shared limiter object.

y-tabata:
Do you have any idea to avoid `prove` errors?
These tests succeeded in my local environment, and I remember 3 days ago the tests succeeded here, too.

gateway/src/apicast/policy/ngx_variable.lua
```diff
@@ -0,0 +1,15 @@
...
+local function context_values()
+ return {
+  uri = ngx.var.uri,
+  host = ngx.var.host,
+  remote_addr = ngx.var.remote_addr
+ }
+end
+
+function _M.available_context(policies_context)
+  return setmetatable(policies_context, { __index = context_values() })
```
mikz: This still mutates the context. It should be the other way around. Setting the metatable of `context_values`.

mikz:
@y-tabata the tests fail because of https://github.com/3scale/apicast/pull/719#discussion_r192318502

Issue #711, opened 2018-05-15
Query string policy: Manipulate request URI query string 


mikz:
# Use Cases

- [ ] Map URL `/{orderId}/{accountId}/{incFields}` into `/sales-order/svc/v2/testDoc/{orderId}?account={accountId}&incFields={incFields}`. This should work when request URL contains `account` or `incFields` parameters in the query string.
- [x] Match URL containing `foo=bar` in the query string and replace it by `foo=daz`. (#724)

# Details

This policy should parse URL query string and operate on individual arguments instead of using regular expressions. It should offer similar interface to the Headers policy: add,set,push - possibly even more as we expand use cases.

# Considerations

Query params should be templated in similar way of the Headers proposal: #708

davidor:
The part of modifying query parameters in the URL rewriting policy with push/add/set/delete commands has been done in #724 

The other kind of transformation mentioned in the issue: `Map URL /{orderId}/{accountId}/{incFields} into /sales-order/svc/v2/testDoc/{orderId}?account={accountId}&incFields={incFields}. ` has not been implemented yet.

davidor:
Regarding the first use case mentioned in the issue:

I think we can also include it in the URL rewriting policy. In this policy, we already have 2 types of commands: one that replaces part of the URL by applying regex matching, and other that allows us to add/modify/delete query arguments.

I think we could add a third one for named arguments. This new command would just need 2 properties: `match` and `replace`. In your example, `match = /{orderId}/{accountId}/{incFields}` and `replace = /sales-order/svc/v2/testDoc/{orderId}?account={accountId}&incFields={incFields}`.

Regarding the implementation, I think it shouldn't be too hard. We can probably reuse some of the code that we already have for the 2 commands that we support.

davidor:
There's an aspect that might complicate things a bit.

Right now, we have 2 types of commands in the URL rewriting policy: one replaces part of the URL and the other adds/modifies/deletes query arguments. They're independent, so it does not matter which type we run first (the order of the rules of the same type matters, though). 

The other kind of transformation, which we could call `named_args` (`/{orderId}/{accountId}/{incFields} into /sales-order/svc/v2/testDoc/{orderId}?account={accountId}&incFields={incFields}`, for example) is a mix of both. It can rewrite parts of the URL, but it can also modify the query args.

[...]

PR #724, opened 2018-05-25
URL rewriting policy: Allow to modify query parameters

davidor:
This PR implements part of #711 
It adds the ability to modify query parameters in the URL rewriting policy.

### DIFF
--- a/gateway/src/apicast/policy/url_rewriting/apicast-policy.json
+++ b/gateway/src/apicast/policy/url_rewriting/apicast-policy.json
@@ -77,6 +77,10 @@
                 {
                   "enum": ["push"],
                   "title": "Create the arg when not set, add the value when set"
+                },
+                {
+                  "enum": ["delete"],
+                  "title": "Delete an arg"
                 }
               ]
             },
--- a/gateway/src/apicast/policy/url_rewriting/apicast-policy.json
+++ b/gateway/src/apicast/policy/url_rewriting/apicast-policy.json
@@ -55,6 +55,42 @@
           },
           "required": ["op", "regex", "replace"]
         }
+      },
+      "query_args_commands": {
+        "description": "List of commands to apply to the query string args",
+        "type": "array",
+        "items": {
+          "type": "object",
+          "properties": {
+            "op": {
+              "description": "Operation to apply to the query argument",
+              "type": "string",
+              "oneOf": [
+                {
+                  "enum": ["add"],
+                  "title": "Add a value to an existing argument"
+                },
+                {
+                  "enum": ["set"],
+                  "title": "Create the arg when not set, replace its value when set"
+                },
+                {
+                  "enum": ["push"],
+                  "title": "Create the arg when not set, add the value when set"
+                }
+              ]
+            },
+            "arg": {
+              "description": "Query argument",
+              "type": "string"
+            },
+            "value": {
+              "description": "Value",
+              "type": "string"
+            }
+          }
+        },
+        "required": ["op", "arg", "value"]
       }
     }
   }
--- a/gateway/src/apicast/policy/url_rewriting/apicast-policy.json
+++ b/gateway/src/apicast/policy/url_rewriting/apicast-policy.json
@@ -87,6 +87,21 @@
             "value": {
               "description": "Value",
               "type": "string"
+            },
+            "value_type": {
+              "description": "How to evaluate 'value'",
+              "type": "string",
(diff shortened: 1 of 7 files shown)


gateway/src/apicast/policy/url_rewriting/apicast-policy.json
```diff
@@ -55,6 +55,61 @@
...
+        "items": {
+          "type": "object",
+          "properties": {
+            "op": {
+              "description": "Operation to apply to the query argument",
+              "type": "string",
+              "oneOf": [
+                {
+                  "enum": ["add"],
+                  "title": "Set an argument if it does not exist"
```
mikz: This is different from what we have in the headers policy. There it is "append".

Query string arguments are quite similar and we have to have some "append" operation.

spec/policy/url_rewriting/url_rewriting_spec.lua
```diff
@@ -78,5 +81,135 @@ describe('URL rewriting policy', function()
...
+        }
+      }
+      local url_rewriting = URLRewriting.new(config_to_add_args)
+
+      url_rewriting:rewrite()
+
+      assert.stub(ngx.req.set_uri_args).was_called_with({ an_arg = 'a_val' })
+    end)
+
+    it('can replace arguments from the query', function()
```
mikz: These tests duplicate tests of QueryParams.

I think we are over testing this a bit. We have 3 tests for the same code.
Not sure what is the best path forward though.

gateway/src/apicast/policy/url_rewriting/apicast-policy.json
davidor: You're right. I forgot about values with several arguments.

I've adapted the code to take this case into account. The big changes are in `query_params.lua`. The other changes that I needed to make were basically adapting the rest of modules to use the correct names of the operations.

spec/policy/url_rewriting/url_rewriting_spec.lua
davidor: I see your point. In particular, some of these tests check that `ngx.req.set_uri_args` was called with the proper arguments, but that shouldn't be a concern of `url_rewriting`. It should be of `query_params`, and it's being already tested on its specs.

However, it's a bit tricky to test this properly as `QueryParams.new()` is called from `rewrite()` and it does not make sense to inject an instance of QueryParams there.

davidor: Check the latest commit. It mocks QueryParams to avoid testing its internals.

spec/policy/url_rewriting/url_rewriting_spec.lua
```diff
@@ -83,54 +94,45 @@ describe('URL rewriting policy', function()
...
-          { op = 'push', arg = 'an_arg', value = '2' }
+          { op = 'push', arg = 'an_arg', value = '1' }
         }
       }
       local url_rewriting = URLRewriting.new(config_to_add_args)
 
       url_rewriting:rewrite()
 
-      assert.stub(ngx.req.set_uri_args).was_called_with({ an_arg = { '1', '2' } })
+      assert.spy(spy_mocked_query_params_push).was_called_with(
```
mikz: Looks like all these tests are sharing one spy. Is that safe?

davidor: Changed. They need to be instantiated in the `before_each` to reset their calls.

gateway/src/apicast/policy/url_rewriting/query_params.lua
```diff
@@ -0,0 +1,87 @@
...
+
+  update_uri_args(self.args)
+end
+
+--- Set a value for an argument
+-- 1) When the arg is not set, creates it with the given value.
+-- 2) When the arg is set, replaces its value with the given one.
+-- 3) Deletes an arg when the value is "".
+function _M:set(arg, value)
+  if value == '' then
```
mikz: Hmmm. This might not be right. HTTP headers are not sent at all by nginx when the they are empty, but Query string args can be binary like `?foo&bar&search`. We might need to have "delete" operation after all.

davidor: Right. I forgot about this corner case.

davidor: Fixed. Check the latest commit.

PR #725, opened 2018-05-28
[ci] report unit test code coverage

mikz:
Unfortunately it makes busted build about 7x slower. (from 7s to 48s).

There will be more reports when master gets build with coverage, but for now there is a proof it actually measures stuff: https://codecov.io/gh/3scale/apicast/tree/36c56a6868a69020145862f7aa10f14d6ff62c6f/gateway/src/apicast

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,6 +61,10 @@ jobs:
             - lua_modules
       - run: mkdir -p tmp/junit
       - run: $(make rover) exec make busted
+      - run:
+          name: Report Code Coverage
+          command: bash <(curl -s https://codecov.io/bash)
+          when: always
       - restore_cache:
           keys:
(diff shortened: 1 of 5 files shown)


davidor:
Looks good but I'm concerned about that 7x.

How about disabling the test coverage analysis by default and allow enabling it with an env? We could set that env in our CI systems and also manually when we're interested in the coverage analysis.

mikz:
@davidor I think for the coverage to be useful it should be reported from every PR.

Yep, enabling it by ENV makes total sense (so we run it only on CI).
The "build" job of the CI takes way less than the "s2i-builder" and "s2i-runtime", so it does not slow down the build overall.

davidor:
I agree that it should be run on every PR. My problem is with running it every time we run `make busted` locally.

mikz:
@davidor updated to use `CI` env variable. We are doing the same for JUnit xml reports.

PR #727, opened 2018-05-29
[codecov] configure codecov

mikz:
Disable comments, so we have "quiet" period when we get the 
code coverage numbers.

Set default branch to master. Codecov thinks it is v2.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,3 @@
+codecov:
+  branch: master
+comment: off


.codecov.yml
```diff
@@ -0,0 +1,2 @@
+codecov:
+  branch: stable
```
davidor: `master` ?

mikz: 😂

PR #728, opened 2018-05-29
Add 3scale referrer policy

### LINKED ISSUE #414: Add Referrer filter option
3scale SAAS and on-prem product supports Referrer filter. https://support.3scale.net/docs/api-authentication/authentication-patterns.  We had the option to enable referrer filter on old nginx files https://support.3scale.net/codehub/referer-filter-nginx.

With the recent development of apicast, there is a need to have this option available as an env variable.

Todo.
 - [ ]  While starting the gateway use env variable `APICAST_REFERRER_FILTER` value `true|false`
 - [ ]   When done override 

https://github.com/3scale/apicast/blob/57790a667b6b6a035ce1ef6c17d3a6637095fd94/apicast/conf.d/apicast.conf#L19 

with 

```
set $path /transactions/authrep.xml?$backend_authentication_type=$backend_authentication_value&service_id=$service_id&$usage&$credentials&referrer=$http_referer$log;
```

  - [ ] Test cases

davidor:
Closes #414

### DIFF
--- a/gateway/src/apicast/policy/3scale_referrer/3scale_referrer.lua
+++ b/gateway/src/apicast/policy/3scale_referrer/3scale_referrer.lua
@@ -1,0 +1,16 @@
+local policy = require('apicast.policy')
+local _M = policy.new('3scale Referrer policy')
+
+function _M.rewrite(_, context)
+  local referrer = ngx.var.http_referer
+
+  if referrer then
+    if context.proxy then
+      context.proxy.extra_params_backend_authrep.referrer = referrer
+    else
+      ngx.log(ngx.ERR, 'Did not find a proxy in the policies context.')
+    end
+  end
+end
+
+return _M
--- a/gateway/src/apicast/proxy.lua
+++ b/gateway/src/apicast/proxy.lua
@@ -64,6 +64,11 @@ function _M.new(configuration)
     configuration = assert(configuration, 'missing proxy configuration'),
     cache = cache,
     cache_handler = cache_handler,
+
+    -- Params to send in 3scale backend calls that are not the typical ones
+    -- (credentials, usage, etc.).
+    -- This allows us, for example, to send a referrer.
+    extra_params_backend_authrep = {}
   }, mt)
 end
 
@@ -136,7 +141,7 @@ function _M:authorize(service, usage, credentials, ttl)
     ngx.var.cached_key = nil
 
     local backend = assert(backend_client:new(service, http_ng_ngx), 'missing backend')
-    local res = backend:authrep(formatted_usage, credentials)
+    local res = backend:authrep(formatted_usage, credentials, self.extra_params_backend_authrep)
 
     local authorized, rejection_reason = self:handle_backend_response(cached_key, res, ttl)
     if not authorized then
@@ -315,7 +320,12 @@ end
 local function capture_post_action(self, cached_key, service)
   local backend = assert(backend_client:new(service, http_ng_ngx), 'missing backend')
   local formatted_usage = format_usage(self.usage)
-  local res = backend:authrep(formatted_usage, self.credentials, response_codes_data())
+  local res = backend:authrep(
+    formatted_usage,
+    self.credentials,
+    response_codes_data(),
+    self.extra_params_backend_authrep
+  )
 
   self:handle_backend_response(cached_key, res)
 end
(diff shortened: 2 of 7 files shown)


gateway/src/apicast/policy/3scale_referrer/3scale_referrer.lua
```diff
@@ -0,0 +1,20 @@
...
+local _M = policy.new('3scale Referrer policy')
+
+local new = _M.new
+
+function _M.new(config)
+  local self = new(config)
+  return self
+end
+
+function _M.rewrite(_, context)
```
davidor: It'd be nice to be able to use `export` here but can't because we need to add to the existing table, not overwrite it.

gateway/src/apicast/policy/3scale_referrer/3scale_referrer.lua
```diff
@@ -0,0 +1,20 @@
+local policy = require('apicast.policy')
+local _M = policy.new('3scale Referrer policy')
```
davidor: I added the 3scale prefix here because this is 3scale specific. Like the `3scale_batcher` policy.

gateway/src/apicast/proxy.lua
```diff
@@ -64,6 +64,11 @@ function _M.new(configuration)
     configuration = assert(configuration, 'missing proxy configuration'),
     cache = cache,
     cache_handler = cache_handler,
+
+    -- Params to send in 3scale backend calls that are not the typical ones
+    -- (credentials, usage, etc.).
+    -- This allows us, for example, to send a referrer.
+    extra_params_backend_authrep = {}
```
davidor: I opted for this because it's way easier than propagating these extra params to all the methods that need them.
I think this makes the code more readable.

gateway/src/apicast/policy/3scale_referrer/3scale_referrer.lua
```diff
@@ -0,0 +1,20 @@
...
+function _M.new(config)
+  local self = new(config)
+  return self
+end
+
+function _M.rewrite(_, context)
+  local referrer = ngx.var.http_referer
+
+  if referrer then
+    context.extra_params_backend_authrep = context.extra_params_backend_authrep or {}
```
mikz: We need to find a better way to handle those. They are basically global variables and it starts becoming really ugly.

gateway/src/apicast/policy/3scale_referrer/3scale_referrer.lua
```diff
@@ -0,0 +1,20 @@
+local policy = require('apicast.policy')
+local _M = policy.new('3scale Referrer policy')
+
+local new = _M.new
+
+function _M.new(config)
```
mikz: We don't even need to override new, right?

gateway/src/apicast/policy/3scale_referrer/3scale_referrer.lua
mikz: Can't this policy manipulate the proxy object directly?

gateway/src/apicast/policy/3scale_referrer/3scale_referrer.lua
davidor: You're right. This is the first policy we have with an empty configuration, so we don't need to override `new`. Didn't realize that :+1:

gateway/src/apicast/proxy.lua
mikz: I guess it is fine. Proxy object mutates internal state anyway, so it is not really ready to be used by multiple clients.

gateway/src/apicast/policy/3scale_referrer/3scale_referrer.lua
davidor: Yes, but the proxy object should also be in `context`. Do you think that's a better solution?

mikz: I think it already is.

davidor: Yes. The proxy is stored in `context` in `local_chain.lua` (edited my comment at the same time you were writing).

Now that I think about it, I feel like manipulating here the proxy object is a slightly better solution because it avoids the extra step of doing it in the APIcast policy. Instead of doing that, every policy would be responsible for adding the extra backend params it wants. Is that the benefit you were thinking about?

mikz: Yes, that would reduce the number of objects we store in the context root. And then changing the params basically becomes proxy public API.

gateway/src/apicast/policy/3scale_referrer/3scale_referrer.lua
davidor: Fixed.

gateway/src/apicast/policy/3scale_referrer/3scale_referrer.lua
davidor: Fixed.

davidor:
@mikz I addressed your comments. Ready for review.

PR #729, opened 2018-05-29
[t] generate luacov reports from t files

mikz:
Load luacov in APIcast when `CI` environment variable is set. 

This allows collecting code coverage metrics from blackbox integration tests as well as live APIcast instances (if luacov is available).

It also splits code coverage into `busted` and `prove` so we know which covers what.

### DIFF
--- a/gateway/http.d/init.conf
+++ b/gateway/http.d/init.conf
@@ -3,6 +3,41 @@ init_by_lua_block {
     -- require('jit.p').start('vl')
     -- require('jit.dump').start('bsx', 'jit.log')
 
+    if os.getenv('CI') == 'true' then
+      pcall(require, 'luacov.runner')
+    end
+
+    local luacov = package.loaded['luacov.runner']
+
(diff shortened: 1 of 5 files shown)


gateway/http.d/init.conf
```diff
@@ -3,6 +3,41 @@ init_by_lua_block {
...
+
+        local config = { }
+        for _, option in ipairs({"statsfile", "reportfile"}) do
+          -- properly expand current working dir, workaround for https://github.com/openresty/resty-cli/issues/35
+          config[option] = pwd .. luacov.defaults[option]
+          luacov.defaults[option] = pwd
+        end
+
+        luacov.defaults.savestepsize = 3
+        jit.off()
```
davidor: why?

mikz: It was mentioned in several places like the mailing list and github. Looks like coverage can't really be tracked on the JITed code.

PR #740, opened 2018-05-31
CI improvements

mikz:
* s2i build caching
* more parallelization
* fixed benchmark target
* show busted tests in log too
* uses carton (again) to cache perl dependencies

screen shot 2018-05-31 at 16 15 15

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,12 +34,15 @@ CIRCLE_NODE_INDEX ?= 0
 CIRCLE_STAGE ?= build
 COMPOSE_PROJECT_NAME ?= apicast_$(CIRCLE_STAGE)_$(CIRCLE_NODE_INDEX)
 
-ROVER ?= $(shell which rover 2> /dev/null)
+which = $(shell command -v $(1) 2> /dev/null)
+
+ROVER ?= $(call which, rover)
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -64,18 +67,33 @@ busted: dependencies $(ROVER) ## Test Lua.
...
+	$(CARTON) install --deployment --cached
+	$(CARTON) bundle 2> /dev/null
 
 find-file = $(shell find $(2) -type f -name $(1))
 
 prove: HARNESS ?= TAP::Harness
 prove: PROVE_FILES ?= $(filter-out $(call find-file, "*.t", examples/scaffold),$(call find-file, *.t, t examples))
 prove: export TEST_NGINX_RANDOMIZE=1
-prove: $(ROVER) nginx cpan ## Test nginx
+prove: $(ROVER) nginx ## Test nginx
```
davidor: Why carton instead of cpan?

mikz: Because it can produce .tar.gz of installed dependencies that can be used as a cache. 
Also it does a bit better locking as it has cpanfile.snapshot.

We removed it because it was pain to install it cross system, but now it should be ok with the latest trick to also install it into `./local`.

PR #741, opened 2018-06-01
[THREESCALE-586] Default credentials policy

### LINKED ISSUE #734: Default credentials policy: provide default credentials to unauthenticated requests [THREESCALE-586]
Closes [THREESCALE-586](https://issues.jboss.org/browse/THREESCALE-586)

# Use Cases

Exposing legacy public service without authentication. By providing default credentials with some limits (or not) provider can expose some API without requiring authentication.

# Details

A policy should be able to fill in missing credentials for unauthenticated requests.
The policy configuration should have have option to select user_key / app_id+app_key and fill a value.
Then any request with missing credentials will get these by default.

davidor:
Closes #734

### DIFF
--- a/gateway/src/apicast/policy/default_credentials/apicast-policy.json
+++ b/gateway/src/apicast/policy/default_credentials/apicast-policy.json
@@ -1,0 +1,69 @@
+{
+  "$schema": "http://apicast.io/policy-v1/schema#manifest#",
+  "name": "Default credentials",
+  "summary": "Provides default credentials for unauthenticated requests",
+  "description":
+    ["This policy allows to expose a service without authentication. ",
+     "It can be useful, for example, for legacy apps that cannot be adapted to ",
+     "send the auth params. ",
+     "When the credentials are not provided in the request, this policy ",
+     "provides the default ones configured. ",
+     "An app_id + app_key or a user_key should be configured."],
+  "version": "builtin",
+  "configuration": {
+    "type":"object",
+    "properties":{
+      "auth_type":{
+        "type":"string",
+        "enum":[
+          "user_key",
+          "app_id_and_app_key"
+        ],
+        "default":"user_key"
+      }
+    },
+    "required":[
+      "auth_type"
+    ],
+    "dependencies":{
+      "auth_type":{
+        "oneOf":[
+          {
+            "properties":{
+              "auth_type":{
+                "enum":[
+                  "user_key"
+                ]
+              },
+              "user_key":{
+                "type":"string"
+              }
+            },
+            "required":[
+              "user_key"
+            ]
+          },
+          {
+            "properties":{
+              "auth_type":{
+                "enum":[
+                  "app_id_and_app_key"
+                ]
+              },
+              "app_id":{
+                "type":"string"
+              },
+              "app_key":{
+                "type":"string"
+              }
+            },
+            "required":[
+              "app_id",
+              "app_key"
+            ]
+          }
+        ]
+      }
+    }
+  }
+}
--- a/gateway/src/apicast/policy/default_credentials/default_credentials.lua
+++ b/gateway/src/apicast/policy/default_credentials/default_credentials.lua
@@ -1,0 +1,102 @@
+--- Default credentials policy
+
+local tostring = tostring
+
+local policy = require('apicast.policy')
+local _M = policy.new('Default credentials policy')
+
+local new = _M.new
+
+function _M.new(config)
+  local self = new(config)
+
+  if config then
+    self.default_credentials = {
+      user_key = config.user_key,
+      app_id = config.app_id,
+      app_key = config.app_key
+    }
+  else
+    self.default_credentials = {}
+  end
+
+  return self
+end
+
+local function creds_missing(service)
+  local service_creds = service:extract_credentials()
+  if not service_creds then return true end
+
+  local backend_version = tostring(service.backend_version)
+
+  if backend_version == '1' then
+    return service_creds.user_key == nil
+  elseif backend_version == '2' then
+    return service_creds.app_id == nil and service_creds.app_key == nil
+  end
+end
+
+local function provide_creds_for_version_1(service, default_creds)
+  if default_creds.user_key then
+    -- follows same format as Service.extract_credentials()
+    service.extracted_credentials = {
+      default_creds.user_key,
+      user_key = default_creds.user_key
+    }
+
+    ngx.log(ngx.DEBUG, 'Provided default creds for request')
+  else
+    ngx.log(ngx.WARN, 'No default user key configured')
+  end
+end
+
+local function provide_creds_for_version_2(service, default_creds)
+  if default_creds.app_id and default_creds.app_key then
+    -- follows same format as Service.extract_credentials()
+    service.extracted_credentials = {
+      default_creds.app_id,
+      default_creds.app_key,
+      app_id = default_creds.app_id,
+      app_key = default_creds.app_key
+    }
+
+    ngx.log(ngx.DEBUG, 'Provided default creds for request')
+  else
+    ngx.log(ngx.WARN, 'No default app_id + app_key configured')
+  end
+end
+
+local creds_provider = {
+  ["1"] = provide_creds_for_version_1,
+  ["2"] = provide_creds_for_version_2
+}
+
+local function backend_version_is_supported(backend_version)
+  return creds_provider[backend_version] ~= nil
+end
+
+local function provide_creds(service, default_creds)
+  local backend_version = tostring(service.backend_version)
+  creds_provider[backend_version](service, default_creds)
+end
+
+function _M:rewrite(context)
+  local service = context.service
+
+  if not service then
+    ngx.log(ngx.ERR, 'No service in the context')
+    return
+  end
+
+  local backend_version = tostring(service.backend_version)
+  if not backend_version_is_supported(backend_version) then
+    ngx.log(ngx.ERR, 'Incompatible backend version: ', backend_version)
+    return
+  end
+
+  if creds_missing(service) then
+    provide_creds(service, self.default_credentials)
+  end
+end
+
+return _M
--- a/gateway/src/apicast/proxy.lua
+++ b/gateway/src/apicast/proxy.lua
@@ -225,7 +225,13 @@ function _M:rewrite(service, context)
 
   ngx.var.secret_token = service.secret_token
 
-  local credentials, err = service:extract_credentials()
+  -- Another policy might have already extracted the creds.
+  local credentials = service.extracted_credentials
+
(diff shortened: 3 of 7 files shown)


gateway/src/apicast/policy/default_credentials/default_credentials.lua
```diff
@@ -0,0 +1,98 @@
...
+      app_key = config.app_key
+    }
+  else
+    self.default_credentials = {}
+  end
+
+  return self
+end
+
+local function creds_missing(service)
```
davidor: This should be moved to the Service module

davidor: I think that in order to do this, we'd need to extract a `Credentials` class. Credentials are used in many parts of the code (service, backend_client, proxy, etc.), so this would not be an easy refactor. That's why I'm leaning towards leaving it for a future PR.

gateway/src/apicast/policy/default_credentials/apicast-policy.json
```diff
@@ -0,0 +1,21 @@
...
+     "It can be useful, for example, for legacy apps that cannot be adapted to ",
+     "send the auth params. ",
+     "When the credentials are not provided in the request, this policy ",
+     "provides the default ones configured. ",
+     "An app_id + app_key or a user_key should be configured."],
+  "version": "builtin",
+  "configuration": {
+    "type": "object",
+    "properties": {
+      "user_key": { "type": "string" },
```
mikz: I was hoping this actually could be "oneOf" and user would chose user_key/app_id. But that probably can't happen because of the limitation of the react library?

gateway/src/apicast/policy/default_credentials/default_credentials.lua
mikz: I think we should try to avoid overloading the Service module. Just because it is coupled to 3scale data model and in the future we should work without it.

So I'd try to rather try to make very small independent components and Credentials class could work pretty well I think. But it should not really depend on "service".
I agree it is much bigger effort, because it should be designed in a way it will support various authentication methods like TLS, JWT, ...

gateway/src/apicast/policy/default_credentials/apicast-policy.json
davidor: I didn't find a way to do that.
Something like this does not work:
```json
{  
   "type":"object",
   "properties":{  
      "credentials":{  
         "type":"object",
         "oneOf":[  
            {  
               "type":"object",
               "properties":{  
                  "user_key":{  
                     "type":"string"
                  }
               }
            },
            {  
               "type":"object",
               "properties":{  
                  "app_key":{  
                     "type":"string"
                  },
                  "app_id":{  
                     "type":"string"
                  }
               }
            }
         ]
      }
   }
}
```

gateway/src/apicast/proxy.lua
```diff
@@ -225,7 +225,13 @@ function _M:rewrite(service, context)
 
   ngx.var.secret_token = service.secret_token
 
-  local credentials, err = service:extract_credentials()
+  -- Another policy might have already extracted the creds.
+  local credentials = service.extracted_credentials
```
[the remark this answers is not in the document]
davidor: Because `.credentials` already exists in service and it has a different kind of information. It contains the location for credentials (query, headers, etc.) and also the name used to look for the different types of keys:

https://github.com/3scale/apicast/blob/master/gateway/src/apicast/configuration.lua#L106

gateway/src/apicast/policy/default_credentials/apicast-policy.json
mikz: Thats exactly what I had in mind. 

What about:

```json
{
  "type": "object",
  "properties": {
    "credentials": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "name": {
            "type": "string",
            "enum": ["app_id", "app_key", "user_key"]
          },
          "value": {
            "type": "string"
          }
        }
      }
    }
  }
}
```

davidor: That can be loaded in the react lib, but it has other problems.

With that solution, an array with app_id, app_key and user_key is valid. Also, an array with user_key and app_key for example, or an array with only app_key.

Also, with that solution including, for example, 2 different user_keys is valid.

mikz: If we keep the backend version detection, then it works exactly the same as current solution.
It has exactly the same problems as it is now: you can fill any combination of them.

We can make one field required if other field is filled in: https://github.com/mozilla-services/react-jsonschema-form#property-dependencies (but that would not work with this array).

I think with https://github.com/mozilla-services/react-jsonschema-form#dynamic we could do a selector to chose a backend version and then let them fill app_id+app_key / user_key.

I'm not advocating for actually doing it, but I'd like to evaluate our options. 
Also I think we don't really have to strongly validate everything. If they want to fill in just app_key, let them. 


BTW what if we want to expand it to support JWT/OAuth/TLS in the future? Would one structure be better over another? Without `oneOf` not really, right?

gateway/src/apicast/proxy.lua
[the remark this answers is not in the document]
mikz: Oh boy. This would definitely would benefit from some cleanup.

gateway/src/apicast/policy/default_credentials/default_credentials.lua
```diff
@@ -0,0 +1,98 @@
...
+end
+
+local creds_provider = {
+  ["1"] = provide_creds_for_version_1,
+  ["2"] = provide_creds_for_version_2
+}
+
+local function provide_creds(service, default_creds)
+  local backend_version = tostring(service.backend_version)
+  creds_provider[backend_version](service, default_creds)
```
mikz: This will crash on invalid version.

gateway/src/apicast/policy/default_credentials/default_credentials.lua
```diff
@@ -0,0 +1,98 @@
...
+function _M:rewrite(context)
+  local service = context.service
+
+  if not service then
+    ngx.log(ngx.ERR, 'No service in the context')
+    return
+  end
+
+  local backend_version = tostring(service.backend_version)
+  if backend_version ~= '1' and backend_version ~= '2' then
```
mikz: This duplicates the `creds_provider`. Imo it should check if there is creds provider and then run it else return an error instead of duplicating the checks.

spec/policy/default_credentials/default_credentials_spec.lua
```diff
@@ -0,0 +1,255 @@
...
+local DefaultCredentialsPolicy = require('apicast.policy.default_credentials')
+
+describe('Default credentials policy', function()
+  local policy_with_default_user_key
+  local policy_with_default_app_id_and_key
+  local policy_without_default_creds
+  local policy_with_default_app_id
+  local policy_with_default_app_key
+
+  local service_with_user_key_in_req
```
mikz: Wouldn't it be better for this to be a function? Then you could drop the before_each and just use local variables.

gateway/src/apicast/policy/default_credentials/apicast-policy.json
davidor: `dependencies` would fail to validate things like incluing an app_id with a user_key.
The dynamic schema would work. You can try it in https://mozilla-services.github.io/react-jsonschema-form/ to see how it looks:
```json
{
    "type":"object",
    "properties":{
      "auth_type":{
        "type":"string",
        "enum":[
          "user_key",
          "app_id_and_app_key"
        ],
        "default":"user_key"
      }
    },
    "required":[
      "auth_type"
    ],
    "dependencies":{
      "auth_type":{
        "oneOf":[
          {
            "properties":{
              "auth_type":{
                "enum":[
                  "user_key"
                ]
              },
              "user_key":{
                "type":"string"
              }
            },
            "required":[
              "user_key"
            ]
          },
          {
            "properties":{
              "auth_type":{
                "enum":[
                  "app_id_and_app_key"
                ]
              },
              "app_id":{
                "type":"string"
              },
              "app_key":{
                "type":"string"
              }
            },
            "required":[
              "app_id",
              "app_key"
            ]
          }
        ]
      }
    }
  }
```

With that schema, the config would be something like:
```json
{
    "auth_type": "user_key",
    "user_key": "some_user_key"
}
```

or

```json
{
  "auth_type": "app_id_and_app_key",
  "app_id": "some_app_id",
  "app_key": "some_app_key"
}
```

It looks good to me. What do you think?

gateway/src/apicast/policy/default_credentials/default_credentials.lua
davidor: But the method that gets called before `provide_creds` (`missing_creds`), also needs to check if the `backend_version` is supported by the policy.

Doing the check where it's done now allows us to assume in the rest of the code (`missing_creds` and `provide_creds`) that the backend version is supported. If we don't do that, we'll need to add checks and `ngx.log(ngx.ERR, ...`) in those 2 methods.

gateway/src/apicast/policy/default_credentials/apicast-policy.json
mikz: That looks good to me!

gateway/src/apicast/policy/default_credentials/default_credentials.lua
mikz: But it still can be changed to `if not creds_provider[backend_version]`, right?

My point is that if you remove/add element to `creds_provider` this line would also have to change.

gateway/src/apicast/proxy.lua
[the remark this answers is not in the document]
mikz: What about storing this in`context.credentials` ? Or better we do that when we have proper `Credentials` class/object ?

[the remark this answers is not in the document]
davidor: I would wait until we extract the `Credentials` class.

gateway/src/apicast/policy/default_credentials/apicast-policy.json
davidor: Changed :+1:

gateway/src/apicast/policy/default_credentials/default_credentials.lua
davidor: OK. I rewrote this in a way that adding/removing supported backend_versions would imply changes only in one part of the code.

spec/policy/default_credentials/default_credentials_spec.lua
davidor: Right. Also, the `before_each` seems to imply that everything there is needed for every test, which is not true.

Changed :+1:

PR #751, opened 2018-06-07
[dev] development docker persistent environment using volumes

mikz:
Use named volumes for lua_modules and cpan dependencies.

There is no need to build them into the image becuase they'd be
remounted anyway.

We probably could drop the dockerfile entirely if we would want to.


davidor:
With this solution, the user is still responsible for keeping the dependencies up-to-date by running `make cpan` and `make dependencies` (maybe we should create a new target that installs both types of dependencies). It's the same now, but builds will be faster with this method.

eguzki:
In makefile, we could update development task to
```
development: build-development ## Run bash inside the development image
    docker run -v $(PWD):/home/centos --rm -it $(DEVEL_IMAGE) make dependencies
    $(DOCKER_COMPOSE) -f $(DEVEL_DOCKER_COMPOSE_FILE) run --rm development
```

That way, dependencies will always be up to date when running development container.
In addition, make dependencies should include cpan task as dependency

```
dependencies: $(ROVER) $(S2I_CONTEXT)/Roverfile.lock cpan   
  $(ROVER) install --roverfile=$(S2I_CONTEXT)/Roverfile

```

davidor:
Also, related to this, I think we should remove `script/test`. It's no longer needed with the latest changes we made. We can just run `make busted` and `make prove` inside the container. It'd be nice to have a target that runs both of them.

davidor:
:+1:

Makefile
```diff
@@ -174,11 +174,17 @@ test-runtime-image: clean-containers ## Smoke test the runtime image. Pass any d
...
+	mkdir -p $@
 
-development: build-development ## Run bash inside the development image
-	$(DOCKER_COMPOSE) -f $(DEVEL_DOCKER_COMPOSE_FILE) run --rm development
+ifeq ($(origin USER),environment)
+development: USER := $(shell id -u $(USER))
+endif
+development: .docker/lua_modules .docker/local .docker/cpanm .docker/vendor/cache
+development: ## Run bash inside the development image
+	- docker-compose -f $(DEVEL_DOCKER_COMPOSE_FILE) up --detach
```
eguzki: should be $(DOCKER_COMPOSE) ?

eguzki:
Few comments

- Did not find way to stop containers using makefile. ```make clean-containers``` not working.
- [minor issue] Coverage error ignored
```
Could not load stats file /home/centos/luacov.stats.out.
Run your Lua program with -lluacov and then rerun luacov.
make: [busted] Error 1 (ignored)
```
- [minor issue] circle ci error (does not affect at all)
```
/bin/bash: circleci: command not found (not affected)
```

Makefile
mikz: Yep.

mikz:
@eguzki those issues should be fixed.

PR #754, opened 2018-06-07
[dev] development docker container improvements

eguzki:
* use non provileged user using gosu

* container user same UID as host UID avoiding volume permission issues

* ```dependencies``` makefile task depends on ```carton```  and ```cpan``` tasks

### DIFF
--- a/Dockerfile-development
+++ b/Dockerfile-development
@@ -1,37 +1,37 @@
 ARG BUILDER_IMAGE
 FROM ${BUILDER_IMAGE}
 LABEL maintainer="dortiz@redhat.com"
+ENV GOSU_VERSION 1.10
 
 USER root
-
-# 'default' user is defined in the base image
-ARG USER_NAME=default
-ARG USER_HOME=/home/centos
 
 RUN yum -y install sudo \
  && yum -y autoremove \
  && yum -y clean all
 
-RUN mkdir -p ${USER_HOME}/gateway \
- && chown -R ${USER_NAME}: ${USER_HOME} \
- && echo "${USER_NAME} ALL=(ALL:ALL) NOPASSWD: ALL" > /etc/sudoers.d/${USER_NAME} \
- && chmod 0440 /etc/sudoers.d/${USER_NAME}
+# Install gosu
+RUN set -ex; \
+  \
+  yum -y install epel-release; \
+  yum -y install wget dpkg; \
+  \
+  dpkgArch="$(dpkg --print-architecture | awk -F- '{ print $NF }')"; \
+  wget -O /usr/bin/gosu "https://github.com/tianon/gosu/releases/download/$GOSU_VERSION/gosu-$dpkgArch"; \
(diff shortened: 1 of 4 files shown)


eguzki:
this is directly related to #751

Finally we should come up with a solution which is one, another or a mix.

Dockerfile-development
```diff
@@ -1,37 +1,37 @@
 ARG BUILDER_IMAGE
 FROM ${BUILDER_IMAGE}
 LABEL maintainer="dortiz@redhat.com"
+ENV GOSU_VERSION 1.10
```
mikz: Why? I think this whole Dockerfile can be removed after the cleanup in https://github.com/3scale/apicast/pull/751

Makefile
```diff
@@ -177,7 +179,9 @@ test-runtime-image: clean-containers ## Smoke test the runtime image. Pass any d
 build-development:
 	docker build -f $(DEVEL_DOCKERFILE) -t $(DEVEL_IMAGE) --build-arg BUILDER_IMAGE=$(BUILDER_IMAGE) .
 
+development: export LOCAL_USER_ID = $(shell id -u $(USER))
 development: build-development ## Run bash inside the development image
+	docker run -v $(PROJECT_PATH):/home/centos --rm -it -e LOCAL_USER_ID=$(LOCAL_USER_ID) $(DEVEL_IMAGE) make dependencies
```
mikz: IMO we can call it `USER` and just pass it to docker as `-u`. It works both for user ids and names (so people can pass USER=root).

Makefile
```diff
@@ -186,7 +190,7 @@ rover: $(ROVER)
 $(S2I_CONTEXT)/Roverfile.lock : $(S2I_CONTEXT)/Roverfile
 	$(ROVER) lock --roverfile=$(S2I_CONTEXT)/Roverfile
 
-dependencies: $(ROVER) $(S2I_CONTEXT)/Roverfile.lock
+dependencies: $(ROVER) $(S2I_CONTEXT)/Roverfile.lock cpan carton
```
mikz: Lua dependencies don't depend on cpan or carton.

But there could be new task that depends on this one + carton to install everything.

Makefile
eguzki: using *-u* could be a solution, but less flexible. GID does not exist in container and UID does not exist in */etc/passwd* which may be handy for some commands.

Makefile
eguzki: agree

Makefile
mikz: Passing any (for example the same as on local machine) UID to -u will just work and also usernames that exist in the container /etc/passwd (like root). I don't see any issue with it. And calling it `USER` makes it obvious for people using it.

eguzki: For example, in that base image, there exist a user called *default* UID=1001

I can either run container with -u default

```
docker exec -ti -u default apicast_build_0_development_run_1 /bin/bash
bash-4.2$ whoami
default
bash-4.2$ id -u $USER
1001
```

or run using my local machine UID

```
$ docker exec -ti -u `id -u $USER` apicast_build_0_development_run_1 /bin/bash
bash-4.2$ whoami
whoami: cannot find name for user ID 1000
```
I cannot have both things I guess

PR #755, opened 2018-06-08
Token Introspection Policy to get client information from "oidc_issuer_endpoint".

y-tabata:
Since it is troublesome to define the `client_id` and `client_secret` twice in the `AUTHENTICATION SETTINGS` and in the `POLICIES` of the Admin Portal, this PR adds "use_oidc_issuer_endpoint" property to get client information from `oidc_issuer_endpoint`.

### DIFF
--- a/gateway/src/apicast/policy/token_introspection/apicast-policy.json
+++ b/gateway/src/apicast/policy/token_introspection/apicast-policy.json
@@ -2,24 +2,17 @@
   "$schema": "http://apicast.io/poolicy-v1/schema#manifest#",
   "name": "OAuth 2.0 Token Introspection",
   "summary": "Configures OAuth 2.0 Token Introspection.",
-  "description": 
-  ["This policy executes OAuth 2.0 Token Introspection ",
-   "(https://tools.ietf.org/html/rfc7662) for every API call."],
+  "description": ["This policy executes OAuth 2.0 Token Introspection ",
+    "(https://tools.ietf.org/html/rfc7662) for every API call."
+  ],
   "version": "builtin",
   "configuration": {
     "type": "object",
     "properties": {
-      "introspection_url": {
-        "description": "Introspection Endpoint URL",
-        "type": "string"
-      },
-      "client_id": {
-        "description": "Client ID for the Token Introspection Endpoint",
-        "type": "string"
-      },
-      "client_secret": {
-        "description": "Client Secret for the Token Introspection Endpoint",
-        "type": "string"
+      "auth_type": {
+        "type": "string",
+        "enum": ["use_3scale_oidc_issuer_endpoint", "client_id+client_secret"],
+        "default": "client_id+client_secret"
       },
       "max_ttl_tokens": {
         "description": "Max TTL for cached tokens",
@@ -32,6 +25,43 @@
         "type": "integer",
         "minimum": 0,
         "maximum": 10000
+      }
+    },
+    "required": [
+      "auth_type"
+    ],
+    "dependencies": {
+      "auth_type": {
+        "oneOf": [{
+          "properties": {
+            "auth_type": {
+              "describe": "Use the Client credentials and the Token Introspection Endpoint from the OpenID Connect Issuer setting.",
+              "enum": ["use_3scale_oidc_issuer_endpoint"]
+            }
+          }
+        }, {
+          "properties": {
+            "auth_type": {
+              "describe": "Specify the Token Introspection Endpoint, Client ID, and Client Secret.",
+              "enum": ["client_id+client_secret"]
+            },
+            "client_id": {
+              "description": "Client ID for the Token Introspection Endpoint",
+              "type": "string"
+            },
+            "client_secret": {
+              "description": "Client Secret for the Token Introspection Endpoint",
+              "type": "string"
+            },
+            "introspection_url": {
+              "description": "Introspection Endpoint URL",
+              "type": "string"
+            }
+          },
+          "required": [
+            "client_id", "client_secret", "introspection_url"
+          ]
+        }]
       }
     }
   }
--- a/gateway/src/apicast/policy/token_introspection/token_introspection.lua
+++ b/gateway/src/apicast/policy/token_introspection/token_introspection.lua
@@ -6,6 +6,7 @@ local http_authorization = require 'resty.http_authorization'
 local http_ng = require 'resty.http_ng'
 local user_agent = require 'apicast.user_agent'
 local resty_env = require('resty.env')
+local resty_url = require('resty.url')
 
 local tokens_cache = require('tokens_cache')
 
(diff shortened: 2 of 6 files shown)


gateway/src/apicast/policy/token_introspection/apicast-policy.json
```diff
@@ -21,6 +21,11 @@
         "description": "Client Secret for the Token Introspection Endpoint",
         "type": "string"
       },
+      "use_oidc_issuer_endpoint": {
```
mikz: When this is `true` we don't need ClientID + Secret anymore, right?

I wonder if it makes sense to use dynamic schema like in https://github.com/3scale/apicast/pull/741.
Using either https://github.com/mozilla-services/react-jsonschema-form#property-dependencies or https://github.com/mozilla-services/react-jsonschema-form#schema-dependencies could allow us to have two login methods: clientid+secret or automatic using oidc_issuer_endpoint.

@davidor wdyt?

gateway/src/apicast/policy/token_introspection/token_introspection.lua
```diff
@@ -47,6 +56,13 @@ function _M.new(config)
   return self
 end
 
+function _M:rewrite(context)
+  if self.use_oidc_issuer_endpoint then
+    local components = resty_url.parse(context.service.oidc.issuer_endpoint)
+    set_client(self, components.user, components.password)
```
mikz: It might be easier to pass the headers into the individual http call instead of initializing new client every time. Then we won't mutate `self` and initialize new http client on every request.

You can pass headers when making the request: https://github.com/3scale/apicast/blob/154dc5a4087c97ef9c46209ada6fc8d7d944fea6/spec/resty/http_ng_spec.lua#L78

gateway/src/apicast/policy/token_introspection/apicast-policy.json
y-tabata: > When this is true we don't need ClientID + Secret anymore, right?

Yes. We don't need to define `client_id` and `client_secret` in the schema.

mikz: What I mean is having either client_id+client_secret or use_oidc_issuer_endpoint.


Try following schema:
```json
{
  "type": "object",
  "properties": {
    "auth_type": {
      "type": "string",
      "enum": [
        "oidc_issuer_endpoint",
        "client_id+client_secret"
      ],
      "default": "client_id+client_secret"
    }
  },
  "required": [
    "auth_type"
  ],
  "dependencies": {
    "auth_type": {
      "oneOf": [
        {},
        {
          "properties": {
            "auth_type": {
              "enum": [
                "client_id+client_secret"
              ]
            },
            "client_id": {
              "type": "string"
            },
            "client_secret": {
              "type": "string"
            }
          },
          "required": [
            "client_id",
            "client_secret"
          ]
        }
      ]
    }
  }
}
```

In https://mozilla-services.github.io/react-jsonschema-form/

That would allow user to choose which auth method to use and then fill the credentials.
It should also be extensible to add more properties in the future.

y-tabata: Changed.

gateway/src/apicast/policy/token_introspection/token_introspection.lua
y-tabata: Changed.

gateway/src/apicast/policy/token_introspection/token_introspection.lua
```diff
@@ -55,7 +66,8 @@ local function introspect_token(self, token)
 
   --- Parameters for the token introspection endpoint.
   -- https://tools.ietf.org/html/rfc7662#section-2.1
-  local res, err = self.http_client.post(self.introspection_url , { token = token, token_type_hint = 'access_token'})
+  local res, err = self.http_client.post{self.introspection_url , { token = token, token_type_hint = 'access_token'},
+    headers = {['User-Agent'] = user_agent(), ['Authorization'] = self.credential}}
```
mikz: User-Agent does not change, so it can be passed when creating the HTTP Client.

gateway/src/apicast/policy/token_introspection/token_introspection.lua
```diff
@@ -16,20 +17,23 @@ local new = _M.new
...
 local noop_cache = { get = noop, set = noop }
 
+local function create_credential(client_id, client_secret)
+  return 'Basic ' .. ngx.encode_base64(table.concat({ client_id, client_secret }, ':'))
+end
+
 function _M.new(config)
   local self = new(config)
   self.config = config or {}
+  self.auth_type = config.auth_type or {}
```
mikz: I think the default value for `auth_type` is "client_id+client_secret", right?

y-tabata: Fixed.

gateway/src/apicast/policy/token_introspection/token_introspection.lua
y-tabata: Fixed.

gateway/src/apicast/policy/token_introspection/apicast-policy.json
```diff
@@ -13,13 +13,10 @@
         "description": "Introspection Endpoint URL",
```
mikz: When using the OIDC endpoint, this setting is also unnecessary, right?

mikz: But this endpoint is discoverable through OIDC.

For example: https://keycloak.3sca.net/auth/realms/oidc-demo/.well-known/openid-configuration.

It should be possible to get it from `context.proxy.oauth.config`. You can use `require('resty.repl').start()` to start a REPL and see around the objects available.

gateway/src/apicast/policy/token_introspection/token_introspection.lua
```diff
@@ -77,6 +85,12 @@ local function introspect_token(self, token)
 end
 
 function _M:access(context)
+  if self.auth_type == "oidc_issuer_endpoint" then
+    local components = resty_url.parse(context.service.oidc.issuer_endpoint)
+    self.credential = create_credential(components.user, components.password)
+    self.introspection_url = context.proxy.oauth.config.openid.token_introspection_endpoint
```
mikz: We probably should test how this policy behaves on non OIDC service. I think this would crash.

gateway/src/apicast/policy/token_introspection/apicast-policy.json
```diff
@@ -33,6 +26,43 @@
...
+    },
+    "required": [
+      "auth_type"
+    ],
+    "dependencies": {
+      "auth_type": {
+        "oneOf": [{
+          "properties": {
+            "auth_type": {
+              "describe": "Get the client and the introspection end point from the oidc issuer endpoint",
```
mikz: I think this could be improved a bit.

> Use the Client credentials and the OIDC Issuer Endpoint from the OpenID Connect Issuer setting.

Still not perfect, but we can improve that later.

gateway/src/apicast/policy/token_introspection/apicast-policy.json
```diff
@@ -33,6 +26,43 @@
...
+              "enum": ["oidc_issuer_endpoint"]
+            }
+          }
+        }, {
+          "properties": {
+            "auth_type": {
+              "describe": "Specify the client ID and the client secret",
```
mikz: This probably should say something like:

> Specify the Token Introspection Endpoint, Client ID, and Client Secret.

gateway/src/apicast/policy/token_introspection/token_introspection.lua
```diff
@@ -77,6 +85,17 @@ local function introspect_token(self, token)
 end
 
 function _M:access(context)
+  if self.auth_type == "oidc_issuer_endpoint" then
+    if not context.proxy.oauth then
+      ngx.status = context.service.auth_failed_status
+      ngx.say(context.service.error_auth_failed)
+      return ngx.exit(ngx.status)
+    end
```
mikz: Would be nice to have one whitespace after `end` to visually split those blocks.

gateway/src/apicast/policy/token_introspection/apicast-policy.json
```diff
@@ -33,6 +26,43 @@
...
+    "required": [
+      "auth_type"
+    ],
+    "dependencies": {
+      "auth_type": {
+        "oneOf": [{
+          "properties": {
+            "auth_type": {
+              "describe": "Get the client and the introspection end point from the oidc issuer endpoint",
+              "enum": ["oidc_issuer_endpoint"]
```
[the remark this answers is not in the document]
y-tabata: Changed.

gateway/src/apicast/policy/token_introspection/apicast-policy.json
y-tabata: Changed.

gateway/src/apicast/policy/token_introspection/apicast-policy.json
y-tabata: Changed.

gateway/src/apicast/policy/token_introspection/token_introspection.lua
y-tabata: Changed.

PR #757, opened 2018-06-08
Make 3scale batching policy compatible with the caching policy

### LINKED ISSUE #738: Make 3scale batching policy compatible with the caching policy
The 3scale batching policy is not using the handlers defined in the caching policy. That means that when both policies appear in the chain, the caching one has no effect.

davidor:
Closes #738

### DIFF
--- a/gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
+++ b/gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
@@ -1,6 +1,7 @@
 local backend_client = require('apicast.backend_client')
 local AuthsCache = require('auths_cache')
 local ReportsBatcher = require('reports_batcher')
+local keys_helper = require('keys_helper')
 local policy = require('apicast.policy')
 local errors = require('apicast.errors')
 local reporter = require('reporter')
@@ -37,6 +38,12 @@ function _M.new(config)
     ngx.log(ngx.ERR, "Create semaphore failed: ", err)
   end
   self.semaphore_report_timer = semaphore_report_timer
+
+  -- Cache for authorizations to be used in the event of a 3scale backend
+  -- downtime.
+  -- This cache allows us to use this policy in combination with the caching
+  -- one.
+  self.backend_downtime_cache = ngx.shared.api_keys
 
   return self
 end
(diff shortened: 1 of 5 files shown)


gateway/src/apicast/policy/3scale_batcher/backend_downtime_cache.lua
```diff
@@ -0,0 +1,31 @@
...
+end
+
+function _M:get(transaction)
+  local key = keys_helper.key_for_cached_auth(transaction)
+  return self.shared_dict:get(key)
+end
+
+-- Note: "update_func" should be one of the handlers exposed by the caching
+-- policy.
+function _M:update(transaction, backend_status, update_func)
```
davidor: I'll move `update_func` to the initializer so it becomes an attr of the class.

gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
```diff
@@ -136,6 +164,9 @@ function _M:access(context)
 
   ensure_report_timer_on(self, service_id, backend)
 
+  -- The caching policy sets a cache handler in the context
+  self.cache_handler = context.cache_handler
```
davidor: Ideally, the cache handler should be set in the initializer and passed to `BackendDowntimeCache.new()`. Unfortunately, that's not possible because the exported cache_handler is not available there.

gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
```diff
@@ -111,16 +118,37 @@ local function error(service, rejection_reason)
...
   local rejection_reason = rejection_reason_from_headers(headers)
   self.auths_cache:set(transaction, status, rejection_reason)
   return error(service, rejection_reason)
 end
 
+local function handle_backend_error(self, service, transaction)
+  local cached = self.cache_handler and
+      self.backend_downtime_cache:get(transaction)
+
+  if not cached or cached ~= 200 then
```
mikz: Might be better to do first the positive check ? `if cached == 200 then ... else ... end`

gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
mikz: Is it necessary to mutate `self` ? I know right now each service will get own copy the policy, but that might not be necessary in the future.
I'd like to keep policies "multi tenant" in a way they could operate with the same configuration on multiple services.

gateway/src/apicast/policy/3scale_batcher/backend_downtime_cache.lua
mikz: What about not passing `update_func` at all?

It is passed here only to be called. This function just gets keys for a transaction. 

Actually this class feels a bit weird as it has several collaborators, but they are grouped only by one property: they need keys for the transaction. One reads from shdict and other is a function to be executed. Seems to me they don't really belong to be together.

davidor: You're right. I was having second thoughts about this class when I opened this PR, and that's why I added the comment. The objective was to isolate the management of the backend downtime cache, but we need to use a handler stored in the context and that makes it a bit weird.

In the end, I left the cache handling responsibility in the main class of the policy. I think that's better for now.

gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
davidor: Right :+1:
Fixed.

gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
davidor: Changed. Looks better :+1: 

I was thinking in the case of `cached` being a table, that's why I added that nil check, but it's an integer, so there's no need to do that.

gateway/src/apicast/policy/3scale_batcher/keys_helper.lua
```diff
@@ -5,7 +5,7 @@ local concat = table.concat
 local sort = table.sort
 local unpack = table.unpack
 local ngx_re = ngx.re
-local table = table
+local new_tab = require('resty.core.base').new_tab
```
davidor: @mikz I remember that we discussed this in a previous PR.
I needed to require this in order to make the test pass. I guess it has to do with the OpenResty upgrade that we merged today.

mikz: 👍

gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
```diff
@@ -110,17 +117,42 @@ local function error(service, rejection_reason)
   end
 end
 
-local function handle_backend_ok(self, transaction)
+local function update_downtime_cache(cache, transaction, backend_status, cache_handler)
+  local key = keys_helper.key_for_cached_auth(transaction)
+  cache_handler(cache, key, backend_status)
```
mikz: Isn't this handler being called twice ? Once by APIcast policy and once by this one?

davidor: No, because when this policy is enabled, APIcast does not authorize nor report to the 3scale backend. And the handler is only called when there's a backend response to be handled.

PR #759, opened 2018-06-12
[rate-limit] extract redis shdict

mikz:
* extract redis shdict into own module
* start spec for the redis shdict
* implement init_ttl parameter of incr

/cc @y-tabata my plan is to extract this module and have standalone unit tests for it

### DIFF
--- a/spec/policy/rate_limit/redis_shdict_spec.lua
+++ b/spec/policy/rate_limit/redis_shdict_spec.lua
@@ -1,0 +1,119 @@
+local env = require('resty.env')
+local ts = require ('apicast.threescale_utils')
+
+local redis_host = env.get('TEST_NGINX_REDIS_HOST') or 'localhost'
+local redis_port = env.get('TEST_NGINX_REDIS_PORT') or 6379
+
+local redis_shdict = require('apicast.policy.rate_limit.redis_shdict')
+
+describe('Redis Shared Dictionary', function()
+    local redis
+    local shdict
+
+    before_each(function()
+        local options = { host = redis_host, port = redis_port, db = 1 }
+        redis = assert(ts.connect_redis(options))
+        shdict = assert(redis_shdict.new(options))
+
+        assert(redis:flushdb())
+    end)
+
+    describe('flush_all', function()
+        it('removes all records', function()
+            assert(redis:set('foo', 'bar'))
+            assert.equal('bar', redis:get('foo'))
+
+            shdict:flush_all()
+
+            assert.equal(ngx.null, redis:get('foo'))
+        end)
+    end)
+
+    describe('incr', function()
+        pending('without default', function()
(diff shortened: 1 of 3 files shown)


spec/policy/rate_limit/redis_shdict_spec.lua
```diff
@@ -0,0 +1,55 @@
...
+            local ret = shdict:incr('somekey', 2, 3)
+
+            assert.equal(5, ret)
+            assert.equal('5', redis:get('somekey'))
+        end)
+
+        it('has init_ttl value', function()
+            shdict:incr('somekey', 0, 0, 1)
+
+            assert.equal(1, redis:ttl('somekey'))
```
davidor: This assertion might fail sometimes. Not sure in practice how often it will, though.

spec/policy/rate_limit/redis_shdict_spec.lua
```diff
@@ -0,0 +1,55 @@
...
+
+            assert.equal(ngx.null, redis:get('foo'))
+        end)
+    end)
+
+    describe('incr', function()
+        pending('without default', function()
+            local ret, err = shdict:incr('somekey', 2)
+
+            -- TODO: nginx shdict:incr returns: nil, 'not found' when the key does not exist
```
davidor: This is what the 2 assertions below are checking, right?

davidor: Sorry, didn't notice this one was marked as 'pending'

spec/policy/rate_limit/redis_shdict_spec.lua
```diff
@@ -0,0 +1,55 @@
...
+            assert.is_nil(ret)
+            assert.equal('not found', err)
+        end)
+
+        it('has init value', function()
```
davidor: For completeness, I'd also test the case that calls `incr()` on a key that already exists.

spec/policy/rate_limit/redis_shdict_spec.lua
mikz: Yep.

spec/policy/rate_limit/redis_shdict_spec.lua
mikz: Yeah, I was wondering about it. Tried it plenty of times and never got an issue. Maybe with redis running over network it would be a problem.

spec/policy/rate_limit/redis_shdict_spec.lua
mikz: Oh yeah. Those tests are quite incomplete. I wanted to add the `init_ttl` in a TDD way.

mikz: Fixed.

PR #764, opened 2018-06-15
[wip] load .init* on all policies

mikz:


gateway/src/apicast/executor.lua
```diff
@@ -59,6 +63,28 @@ function _M:context(phase)
     return shared_build_context(self)
 end
 
+local init = _M.init
+function _M:init()
+    init(self)
+
+    for _, policy_mod in ipairs(policy_modules) do
+       if policy_mod.init then
+           policy_mod.init()
```
davidor: I think we'll need to check which policies are in the global policy chain. Otherwise, `init()` might be run twice for those policies. Same for `init_worker()`

spec/ngx_helper.lua
```diff
@@ -28,11 +30,24 @@ local set_status = getlocal(register_setter, 'ngx_magic_key_setters').status
...
+  end
+end
+
 local function reset_ngx_state()
   ngx.var = deepcopy(ngx_var_original)
   ngx.ctx = deepcopy(ngx_ctx_original)
-  ngx.shared = deepcopy(ngx_shared_original)
   ngx.header = deepcopy(ngx_header_original)
+
+  reset_ngx_shared(ngx_shared_original)
```
davidor: Why doesn't `deepcopy()` work in this case?

mikz: `deepcopy` works fine, but `resty.limit.req` takes a reference to `ngx.shared` when it is loaded:
https://github.com/openresty/lua-resty-limit-traffic/blob/ef080731b80fe00a385a4aed8a7f7f575d01c152/lib/resty/limit/req.lua#L11
So replacing the whole table won't have an effect.

davidor: :+1:
I think this deserves a comment.

spec/policy/rate_limit/rate_limit_spec.lua
```diff
@@ -27,6 +27,27 @@ local function is_gt(_, arguments)
...
+  local actual = tonumber(arguments[3])
+
+  local result = not ok and expected == actual
+  -- switch arguments for proper output message
+  util.tinsert(arguments, 1, util.tremove(arguments, 3))
+  state.failure_message = arguments[3]
+  return result
+end
+
+assert:register("assertion", "returns_error", returns_error, "assertion.error.positive", "assertion.error.negative")
```
davidor: `assertion.error.positive` and `assertion.error.negative` are not set.

mikz: But it worked. I just copied it from luassert internals, so I assume they are defined there and just reused.

I plan to extract this to a helper, so it can be reused in other files too.

mikz:
Closed by https://github.com/3scale/apicast/pull/770

PR #766, opened 2018-06-15
[spec] do not replace ngx.shared entirely

mikz:
If some code like `resty.limit.req` is loaded globaly for the whole test suite,
it will take reference to `ngx.shared` and replacing that table would have no effect.


davidor:
Looks good :+1: 
Please fix the codeclimate issue before merging.

PR #770, opened 2018-06-18
init and init_worker phases called on all policies

mikz:
During master process boot all available policies are loaded 
and get their `init` and `init_worker` phases executed.

Increases initial memory use by about 800 kB (which is later shared between workers).

### DIFF
--- a/gateway/src/apicast/executor.lua
+++ b/gateway/src/apicast/executor.lua
@@ -10,12 +10,16 @@ local PolicyChain = require('apicast.policy_chain')
 local policy = require('apicast.policy')
 local linked_list = require('apicast.linked_list')
 local prometheus = require('apicast.prometheus')
+local policy_loader = require('apicast.policy_loader')
 
 local setmetatable = setmetatable
+local ipairs = ipairs
 
 local _M = { }
 
 local mt = { __index = _M }
+
+local policy_modules = policy_loader:get_all()
 
 -- forward all policy methods to the policy chain
 for _,phase in policy.phases() do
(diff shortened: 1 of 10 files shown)


gateway/http.d/init.conf
```diff
@@ -79,7 +79,7 @@ init_by_lua_block {
         package.loaded['apicast.executor'] = module
     end
 
-    module:init()
+    module.init()
```
davidor: Same here

gateway/src/apicast/executor.lua
```diff
@@ -59,6 +61,45 @@ function _M:context(phase)
...
+
+    local init_worker = _M.init_worker
+    function _M:init_worker()
+        local executed = {}
+
+        for _,policy in init_worker(self) do
+            executed[policy.init_worker] = true
+        end
+
+        for _, policy in ipairs(policies or policy_loader:get_all()) do
```
davidor: This caching complicates a lot testing this method. Depending on the execution the value will be cached or not and when it is, mocking `policy_loader:get_all()` will have no effect.
The tests I added in my latest commit are affected by this.

davidor: Are you ok with not caching this @mikz ?

mikz: I see. It would make more sense to me to provide reset for the cache than not having it at all. 

I think it is important to be correct and that init and init_worker should be executed on exactly the same lists. Loading the list for the second time could not ensure that.

davidor: :+1:

PR #773, opened 2018-06-19
Keycloak Role check policy

### LINKED ISSUE #761: OAuth/OIDC authorization
# Work in progress

This is very incomplete proposal and we should gather some use cases how APIcast users want to use OAuth scopes/roles for authorization. Any use cases are much appreciated.

Would be great if we could gather:

- [ ] examples of JWT with some roles/scopes
- [ ] examples of HTTP requests passing through the gateway
- [ ] examples of how these JWT should be authorized
- [ ] examples of API calls made to Keycloak

We need to determine if we can go with simpler scope based authentication or need the full scope of [Authorization Services](https://www.keycloak.org/docs/3.4/authorization_services/index.html).

## Scope based authorization

We could use roles defined by the JWT and require them for some part of the API (for example by using #744)

### Example of Keycloak JWT using Client Credentials and Service Account

```json
{
  "jti": "be99b4c0-e4cd-40d9-858c-33f92930038f",
  "exp": 1528807306,
  "nbf": 0,
  "iat": 1528807006,
  "iss": "http://localhost:8000/auth/realms/APIcast",
  "aud": "someone",
  "sub": "f5d21857-d948-43e4-90a6-5ae537dafc53",
  "typ": "Bearer",
  "azp": "someone",
  "auth_time": 0,
  "session_state": "13ae673a-03fa-4ded-80fb-e78407cc8611",
  "acr": "1",
  "allowed-origins": [],
  "realm_access": {
    "roles": [
      "uma_authorization"
    ]
  },
  "resource_access": {
    "someone": {
      "roles": [
        "uma_protection"
      ]
    },
    "account": {
      "roles": [
        "manage-account",
        "manage-account-links",
        "view-profile"
      ]
    }
  },
  "clientId": "someone",
  "clientHost": "172.17.0.1",
  "preferred_username": "service-account-someone",
  "clientAddress": "172.17.0.1",
  "email": "service-account-someone@placeholder.org"
}
```

### Example of Keycloak JWT using Authorization Code and user

```json
{
  "jti": "55b0d145-635c-4ed4-b86c-1e298f51cd49",
  "exp": 1528808365,
  "nbf": 0,
  "iat": 1528808065,
  "iss": "http://localhost:8000/auth/realms/APIcast",
  "aud": "someone",
  "sub": "4971c0f6-c030-4572-9fa8-ef4b48ab36e0",
  "typ": "Bearer",
  "azp": "someone",
  "auth_time": 1528807928,
  "session_state": "422b4f08-c238-4153-86e7-87571c00e00c",
  "acr": "0",
  "allowed-origins": [],
  "realm_access": {
    "roles": [
      "internal",
      "uma_authorization"
    ]
  },
  "resource_access": {
    "account": {
      "roles": [
        "manage-account",
        "manage-account-links",
        "view-profile"
      ]
    }
  },
  "name": "Michal Cichra",
  "preferred_username": "michal",
  "given_name": "Michal",
  "family_name": "Cichra",
  "email": "mcichra@redhat.com"
}
```

We could for example verify the list of roles required by the policy against  `resource_access.account.roles` or `realm_access.roles`.

## Authorization services

Not sure how portable Keycloak's [Authorization Services](https://www.keycloak.org/docs/3.4/authorization_services/index.html) are.

y-tabata:
Closes #761.


gateway/src/apicast/policy/scope_check/scope_check.lua
```diff
@@ -0,0 +1,88 @@
...
+local ipairs = ipairs
+local format = string.format
+local gsub = ngx.re.gsub
+local match = ngx.re.match
+
+local TemplateString = require 'apicast.template_string'
+
+local new = _M.new
+
+local function build_templates(scopes)
```
mikz: Why using Liquid?

It is not exposed to users and it would be much simpler to just use a table to index roles and then look them up.

gateway/src/apicast/policy/scope_check/scope_check.lua
```diff
@@ -0,0 +1,88 @@
...
+    local uri_regex, _, _ = gsub(scope.resource, "\\*", ".*")
+    uri_regex = format("^%s$", uri_regex)
+
+    local m_uri, _ = match(uri, uri_regex)
+    if m_uri then
+      -- role check
+      local matched = true
+      for _, role in ipairs(scope.roles) do
+        local roles_in_token = role.template_string:render(context)
+        local m_role, _ = match(roles_in_token, role.role)
```
mikz: Users can write regex to match roles?

gateway/src/apicast/policy/scope_check/scope_check.lua
```diff
@@ -0,0 +1,88 @@
...
+  build_templates(self.scopes)
+
+  return self
+end
+
+local function scope_check(self, context)
+  local uri = self.scopes.template_string:render(ngx_variable.available_context(context))
+
+  for _, scope in ipairs(self.scopes) do
+    local uri_regex, _, _ = gsub(scope.resource, "\\*", ".*")
```
mikz: Would it make sense to reuse our wildcard matching we have for mapping rules?

gateway/src/apicast/policy/scope_check/scope_check.lua
```diff
@@ -0,0 +1,88 @@
...
+      local matched = true
+      for _, role in ipairs(scope.roles) do
+        local roles_in_token = role.template_string:render(context)
+        local m_role, _ = match(roles_in_token, role.role)
+        if not m_role then
+          matched = false
+          break
+        end
+      end
+      if matched then return { match = true } end
```
mikz: Every time this method executes a new table will be allocated.

Any reason to not use simple true/false ?

gateway/src/apicast/template_string.lua
```diff
@@ -63,8 +63,7 @@ for name, func in pairs(liquid_filters) do
   liquid_filter_set:add_filter(name, func)
 end
 
--- Set resource limits to avoid loops
-local liquid_resource_limit = LiquidResourceLimit:new(nil, nil, 0)
+local liquid_resource_limit
```
mikz: In case Liquid is actually needed, there is no need for this variable to be defined on the top level.

gateway/src/apicast/policy/scope_check/apicast-policy.json
```diff
@@ -0,0 +1,50 @@
...
+          "properties": {
+            "roles": {
+              "type": "array",
+              "description": "Roles necessary to access the resource",
+              "items": {
+                "type": "object",
+                "properties": {
+                  "role": {
+                    "type": "string",
+                    "description": "Name of the role"
```
mikz: I'm not really sure who's is this role. User's ?

Then we should name it clearly. Right now "role" vs "client" does not really say much.

y-tabata:
@mikz Thank you for your reviewing.
Since I think it is better to have something concrete to discuss the design, I created this one.

> For example I'm not sure about the regex matching of roles. The code supports it, but there are no tests for it so I'm not sure it is desired or not.

The regex is only for matching the wildcard. It checks `/aaa` matches `/*`.

> Also, in OIDC/OAuth2 the terminology is pretty clear and we should try to stick to it. Some things can be related to the Client or the User, so we should be careful and name them properly.

I think so too.

> Also, is there any RFC or examples this is following? How can we be sure the JWT will contain those fields even when using different IDP than RHSSO?

There is no RFC for roles or scopes. The implement of the realm role and the client role is the RHSSO original. We may need to make this to be able to add other IDPs easily.

gateway/src/apicast/policy/scope_check/scope_check.lua
y-tabata: @mikz Yes, you're right. Fixed.

gateway/src/apicast/policy/scope_check/scope_check.lua
y-tabata: @mikz No. Changed not to use regex.

gateway/src/apicast/policy/scope_check/scope_check.lua
y-tabata: Changed.

gateway/src/apicast/policy/scope_check/scope_check.lua
y-tabata: @mikz Now there is no reason. Changed.

gateway/src/apicast/template_string.lua
y-tabata: Removed changes.

gateway/src/apicast/policy/scope_check/apicast-policy.json
y-tabata: Clarified the realm roles and the client roles.

gateway/src/apicast/policy/scope_check/scope_check.lua
```diff
@@ -0,0 +1,105 @@
...
+function _M:access(context)
+  if scope_check(self.scopes, context.jwt) then
+    if self.type == "blacklist" then
+      ngx.status = context.service.auth_failed_status
+      ngx.say(context.service.error_auth_failed)
+      return ngx.exit(ngx.status)
+    end
+  else
+    if self.type == "whitelist" then
+      ngx.status = context.service.auth_failed_status
```
mikz: This is the same as in the block before, right? Would be nice if it would not have to be duplicated.

gateway/src/apicast/policy/scope_check/scope_check.lua
```diff
@@ -0,0 +1,105 @@
...
+          if not check_roles_in_token(role.name, jwt.resource_access[role.client].roles) then
+            matched_all_role = false
+            break
+          end
+        end
+      end
+
+      if scope.realm_roles then
+        for _, role in ipairs(scope.realm_roles) do
+          if not jwt.realm_access then
```
mikz: This is another field that needs some pointer to it's meaning.  Basically all the `jwt.` fields need to be described somehow, so developers can go and read about what they mean and what is the format they can expect.

gateway/src/apicast/policy/scope_check/scope_check.lua
```diff
@@ -0,0 +1,105 @@
...
+  self.scopes = config.scopes or {}
+
+  init_mapping_rules(self.scopes)
+
+  return self
+end
+
+local function check_roles_in_token(role, roles_in_token)
+  local matched_each_role = false
+  for _, role_in_token in ipairs(roles_in_token) do
```
mikz: This whole method can be simplified as:

```lua
for _, role_in_token in ipairs(roles_in_token) do
  if role == role_in_token then return true end
end
```

gateway/src/apicast/policy/scope_check/scope_check.lua
```diff
@@ -0,0 +1,105 @@
...
+  end
+
+  return matched_each_role
+end
+
+local function scope_check(scopes, jwt)
+  local uri = ngx.var.uri
+
+  for _, scope in ipairs(scopes) do
+    -- resource check
```
mikz: I think it would be great to extract this part into a function. This one is too large and does two things.

In the end if it is split into more functions it could be simplified to:

```lua
for _, scope in ipairs(scopes) do
  if match_resource_roles(scope, jwt) and match_client_roles(scope, jwt) then
   return true
  end
end
```

gateway/src/apicast/policy/scope_check/scope_check.lua
```diff
@@ -0,0 +1,105 @@
...
+
+local new = _M.new
+
+local function init_mapping_rules(scopes)
+  for _, scope in ipairs(scopes) do
+    scope.mapping_rule = MappingRule.from_proxy_rule({
+      http_method = 'GET',
+      pattern = scope.resource,
+      querystring_parameters = {},
+      metric_system_name = 'hits',
```
mikz: I know why this is here: https://github.com/3scale/apicast/blob/db9f97e430d1ed01abf78b7aeddfbc493a4a2fc8/gateway/src/apicast/mapping_rule.lua#L89

This is task for us: improve MappingRule be allowed initialize without metric system name, so it can be used for such things. 

`delta` should not be needed too (because we just do matching, not incrementing metrics).

gateway/src/apicast/policy/scope_check/scope_check.lua
```diff
@@ -0,0 +1,105 @@
...
+
+local ipairs = ipairs
+local MappingRule = require('apicast.mapping_rule')
+
+local new = _M.new
+
+local function init_mapping_rules(scopes)
+  for _, scope in ipairs(scopes) do
+    scope.mapping_rule = MappingRule.from_proxy_rule({
+      http_method = 'GET',
```
mikz: I think it would be better if this would be some non http method keyword like 'ANY'. Also it could be in a variable so it is reused when calling `scope.mapping_rule:matches`.

gateway/src/apicast/policy/scope_check/scope_check.lua
```diff
@@ -0,0 +1,105 @@
+local policy = require('apicast.policy')
+local _M = policy.new('Scope Check Policy')
```
mikz: As @nmasse-itix pointed out, would be good to name this policy as Keycloak/RH-SSO specific if it only supports Keycloak/RH-SSO.

spec/policy/scope_check/scope_check_spec.lua
```diff
@@ -0,0 +1,534 @@
...
+
+          assert(scope_check_policy:access(context))
+        end)
+
+        it('wildcard', function()
+          local scope_check_policy = ScopeCheckPolicy.new({
+            scopes = {
+              {
+                client_roles = { { name = "aaa", client = "ccc" } },
+                resource = "/bbb/{wildcard}"
```
mikz: Actually how the MappingRule patters are defined, this should pass even without the wildcard.

`/bbb/` will match `/bbb/*`. To terminate at the end the pattern would have to include `$` like `/bbb/$` would not match `/bbb/a`, but only `/bbb/`.

Is that desired?

spec/policy/scope_check/scope_check_spec.lua
```diff
@@ -0,0 +1,534 @@
...
+                realm_roles = { { name = "ddd" } },
+                resource = "/eeeyyy{wildcard}"
+              },
+              {
+                realm_roles = { { name = "rrr" } },
+                resource = "/{wildcard}ggg"
+              },
+              {
+                client_roles = { { name = "fff", client = "hhh" } },
+                resource = "/z{wildcard}ggg"
```
mikz: Those tests are not very readable to me. I'm not sure what is it, but it is hard to understand what is going on.

Maybe it is just those random characters in roles and urls, or actually that they are quite big and have a lot of configuration. It is hard to know *why* the test should be passing/failing.

nmasse-itix:
Yes, for sure ! I can provide the user stories, test the policy once developed and help for the doc.

gateway/src/apicast/policy/scope_check/scope_check.lua
```diff
@@ -0,0 +1,105 @@
...
+      end
+
+      if scope.realm_roles then
+        for _, role in ipairs(scope.realm_roles) do
+          if not jwt.realm_access then
+            matched_all_role = false
+            break
+          end
+
+          if not check_roles_in_token(role.name, jwt.realm_access.roles) then
```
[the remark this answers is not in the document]
y-tabata: Added comments.

gateway/src/apicast/policy/scope_check/scope_check.lua
```diff
@@ -0,0 +1,105 @@
...
+
+  for _, scope in ipairs(scopes) do
+    -- resource check
+    if scope.mapping_rule:matches('GET', uri) then
+      -- role check
+      local matched_all_role = true
+
+      if scope.client_roles then
+        for _, role in ipairs(scope.client_roles) do
+          if not jwt.resource_access or not jwt.resource_access[role.client] then
```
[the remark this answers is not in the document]
y-tabata: Added comments.

gateway/src/apicast/policy/scope_check/scope_check.lua
y-tabata: Added comments.

gateway/src/apicast/policy/scope_check/scope_check.lua
y-tabata: Changed.

gateway/src/apicast/policy/scope_check/scope_check.lua
y-tabata: Changed.

gateway/src/apicast/policy/scope_check/scope_check.lua
y-tabata: Changed.

gateway/src/apicast/policy/scope_check/scope_check.lua
y-tabata: Removed `delta`.

gateway/src/apicast/policy/scope_check/scope_check.lua
y-tabata: Changed.

gateway/src/apicast/policy/scope_check/scope_check.lua
y-tabata: Changed.

spec/policy/scope_check/scope_check_spec.lua
y-tabata: Changed.

spec/policy/scope_check/scope_check_spec.lua
y-tabata: Changed.

gateway/src/apicast/policy/keycloak_scope_check/keycloak_scope_check.lua
```diff
@@ -0,0 +1,157 @@
+--- Keycloak Scope Check Policy
+-- This policy exposes the realm roles and the client roles in the JWT.
+--
+--- The realm roles are specified when you want to add scope check to the every client's resources or 3scale itself.
+-- https://www.keycloak.org/docs/latest/server_admin/index.html#realm-roles
```
mikz: Would be good to pin this to a particular version, so the link does not break over time.

gateway/src/apicast/policy/keycloak_scope_check/keycloak_scope_check.lua
```diff
@@ -0,0 +1,157 @@
...
+
+local function match_client_roles(scope, jwt)
+  if not scope.client_roles then return true end
+
+  for _, role in ipairs(scope.client_roles) do
+    if not jwt.resource_access or not jwt.resource_access[role.client] then
+      return false
+    end
+
+    if not check_roles_in_token(role.name, jwt.resource_access[role.client].roles) then
```
mikz: Would be good to store `jwt.resource_access[role.client]` in a variable, so you don't have to access it twice.


```lua
local client_roles = jwt.resource_access and jwt.resource_access[role.client]
```

gateway/src/apicast/policy/keycloak_scope_check/apicast-policy.json
```diff
@@ -0,0 +1,63 @@
+{
+  "$schema": "http://apicast.io/policy-v1/schema#manifest#",
+  "name": "Keycloak Scope Check",
```
mikz: It is not really checking Scope, but Roles, right?

gateway/src/apicast/policy/keycloak_scope_check/apicast-policy.json
```diff
@@ -0,0 +1,63 @@
...
+  "summary": "Adds scope check with Keycloak.",
+  "description": [
+    "This policy adds scope check with Keycloak.\n",
+    "This policy exposes realm roles and client roles in the access token."
```
mikz: This policy does not expose anything, but verifies those roles, right?

gateway/src/apicast/policy/keycloak_scope_check/apicast-policy.json
```diff
@@ -0,0 +1,63 @@
...
+                  "client": {
+                    "type": "string",
+                    "description": "Client of the role"
+                  }
+                }
+              }
+            },
+            "resource": {
+              "type": "string",
+              "description": "Resource controlled by scope. Specify the path(like '/aaa') of the resource. Wildcard({wildcard}) can be used."
```
mikz: I think this should say it is the same format as Mapping Rules. And that it matches from the beginning of the string and to make an exact match you need to use `$` at the end.

gateway/src/apicast/policy/keycloak_scope_check/keycloak_scope_check.lua
y-tabata: Changed.

gateway/src/apicast/policy/keycloak_scope_check/keycloak_scope_check.lua
y-tabata: Changed.

gateway/src/apicast/policy/keycloak_scope_check/apicast-policy.json
y-tabata: Changed.

gateway/src/apicast/policy/keycloak_scope_check/apicast-policy.json
y-tabata: Changed.

gateway/src/apicast/policy/keycloak_scope_check/apicast-policy.json
mikz: @y-tabata this looks like one of the last blockers. This policy does not really do anything with OAuth Scopes as the name might imply.

y-tabata: Since I think "role check" means only to check the particular roles are existing or not in the jwt and "scope check" means to check it has enough roles to access the particular resources or not, I believe "scope check" is better.

mikz: But "scope" is a defined term in OIDC/OAuth terminology.  It is what Client is asking to access from the User. I'd try to not overload the term and give users false impressions this is related.

What about "Protect resources with Keycloak User and Realm Roles" ?

gateway/src/apicast/policy/keycloak_scope_check/keycloak_scope_check.lua
```diff
@@ -0,0 +1,163 @@
...
+--         "<client_role_A>", "<client_role_B>"
+--       ]
+--     }
+--   }
+-- }
+--
+-- And you need to specify the "client_roles" in this policy as follows:
+--
+-- "client_roles": [
+--   { "name": "<client_role_A>", "client": "<client_A>" },
```
mikz: @y-tabata what are the implications of "client" being static in the configuration? Won't every client have there different id ? 
Won't you need this to be the actual client id from the JWT?

gateway/src/apicast/policy/keycloak_scope_check/apicast-policy.json
y-tabata: According to [the document](https://www.keycloak.org/docs/4.0/server_admin/index.html#_role_scope_mappings), I understand that the "roles" in the access token is itself the "scope" which the user permitted to the client, so what we checks is the "scope".
How do you think?

> What about "Protect resources with Keycloak User and Realm Roles" ?

Your suggestion is not changing the policy name to "Keycloak Roles Check", but changing the summary like that?

gateway/src/apicast/policy/keycloak_scope_check/keycloak_scope_check.lua
y-tabata: The "client" means the property name of the json schema. The actual client is "<client_A>" and "<client_B>".

mikz: Yes, but my point is that "client" value (client_A, client_B) will be dynamic, correct?

If for example you have a JWT from a Client "foobar" then this JWT might have `"resource_access": { "foobar": { "roles": ["finance"] }}`.
Now imagine you have 10000 clients. You don't want to maintain 10000 lines in the policy configuration saying that each client requires role finance. 

Will the client name in the "resource_access" refer to the "client id"  in the `azp`/`aud` claim?

gateway/src/apicast/policy/keycloak_scope_check/apicast-policy.json
mikz: My suggestion would be to change the policy name to "Protect resources with Keycloak User and Realm Roles", but maybe that is too long to pass the validation.

The linked document explains how to limit "Scope" of the "Client" has to "User"'s "Roles".
It explains how a Client can have just limited Scope of access. It controls what roles will be given to the access token. 

This policy would still operate only on Roles.

nmasse-itix: I agree, we should use "role" and not "scope".

The scope is the mean for the client (the consuming application) to ask to use the role of a user, or when dealing with the client credentials flow, the role of the client itself. 

But, there are also implicit roles (you do not have to set a scope to use it) and they are present in the token along with the explicit roles: https://www.keycloak.org/docs/3.3/server_admin/topics/roles/realm-roles.html

So, I would definitely prefer "role" and not "scope".

gateway/src/apicast/policy/keycloak_scope_check/keycloak_scope_check.lua
nmasse-itix: I think you raised an interesting question. 

A client can ask to use one of its client roles or another client role (see https://www.keycloak.org/docs/3.3/server_admin/topics/roles/client-roles.html) 

If you have two clients (app1, app2) with a role defined in each (app1 defines roleapp1, app2 defines roleapp2) and a user has the two roles, the access token would contain:
```
    "realm_access": {
    },
    "resource_access": {
        "app1": {
            "roles": [
                "roleapp1"
            ]
        },
        "app2": {
            "roles": [
                "roleapp2"
            ]
        }
    },
```

With this given configuration, there is not much room for interpretation: if I want to restrict access to GET /a to appa/roleappa and GET /b to appb/roleappb, the code is straightforward.

But if both apps define the same role ? This might be a perfectly valid expectation from the customer and especially with the PSD2 regulation. 

Given an API with two endpoints:
 - `GetAccountDetails`
 - `MoneyTransfer`

The bank might want to make sure:
1) the user is entitled to do bank transfers globally
2) the app has been granted to access the `MoneyTransfer` endpoint
3) the user is entitled to do a bank transfer on this specific application

In RH-SSO, this scenario is achieved as such:
1) is a realm role or client role of a well known client
2) is scope restriction (=> "Full scope allowed" is unchecked in the client config)
3) is a client role in each client that is given on a per user basis

In this configuration, the rules to access MoneyTransfer would be:
- the token has the realm role `money_transfer` (or the client role `bank/money_transfer`)
- the token has the client role `money_transfer` **of the current client**

So, I think the policy should allow both use cases: 
1. checking that the token issued for a client (`appA`) has a defined client role of a well known client (`appB/roleB`)
2. checking that the token issued for a client (`appX`) has a defined client role of itself (`appX/role`)

Suggestion on how to configure the policy for both use cases: 
1. `"client_roles": [ { "name": "roleB", "client": "appB" } ]`
2. `"client_roles": [ { "name": "role" } ]`

I tried to be as clear as possible. Let me know if not. 😄

gateway/src/apicast/policy/keycloak_scope_check/apicast-policy.json
y-tabata: I understand "role" is proper in Keycloak.
Thank you @mikz @nmasse-itix!

CHANGELOG.md
```diff
@@ -22,6 +22,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - New `ssl_certificate` phase allows policies to provide certificate to terminate HTTPS connection [PR #622](https://github.com/3scale/apicast/pull/622).
 - Configurable `auth_type` for the token introspection policy [PR #755](https://github.com/3scale/apicast/pull/755)
 - `TimerTask` module to execute recurrent tasks that can be cancelled [PR #782](https://github.com/3scale/apicast/pull/782), [#784](https://github.com/3scale/apicast/pull/784)
+- Scope Check policy [PR #773](https://github.com/3scale/apicast/pull/773)
+- Keycloak Scope Check policy [PR #773](https://github.com/3scale/apicast/pull/773)
```
mikz: Looks like this needs a cleanup.

gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
```diff
@@ -114,7 +114,10 @@ local function match_client_roles(scope, jwt)
     local client = jwt.resource_access[role.client]
 
     if not client then
-      return false
+      client = jwt.resource_access[jwt.aud]
```
mikz: I think this should be described a bit more by comments, because it is not clear if it is correct.

Is it intentional, that when configuration says `client = "foobar"`, but there are no allowed roles for that name, it is going to fall back to `jwt.aud` ? Is it desired it is going to try both?

Is there more scripting we want do do? What about using Liquid?

y-tabata: OK. 

Liquid sounds good.

If a user want to use a client application as the `client`, specify the `client` like `{{jwt.aud}}` or `{{jwt.azp}}`.

In this way, the user can select the `client` more flexibly, and this policy can detect the `client` missing.

How do you think?

gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
```diff
@@ -0,0 +1,166 @@
...
+--       "roles": [
+--         "<client_role_A>", "<client_role_B>"
+--       ]
+--     }
+--   }
+-- }
+--
+-- And you need to specify the "client_roles" in this policy as follows:
+--
+-- "client_roles": [
```
mikz: Shouldn't this be called `resource_roles` ?

gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
mikz: If we do liquid, then it should be working in both the client and realm roles. And for their values too. 

So following options are possible:
```json
{
"realm_roles": [
 { "name": "foo", "name_type": "plain" },
 { "name": "{{ jwt.aud }}", "name_type": "liquid" }
],
"client_roles": [
 { "name": "foo", "name_type": "plain", "client": "somename", "client_type": "plain" },
 { "name": "{{ jwt.aud }}", "name_type": "liquid", "client": "{{ jwt.azp }}", "client_type": "liquid" },
]
}
```

Later we can make this generic enough to require any value from JWT by using patterns to traverse the JWT object. But we should wait with that until we have more use cases and requirements.

For example:

```json
"whitelist": [ 
  { "path": "resource_roles.{{ jwt.azp }}.roles", "value": "{{ jwt.azp }}" }  
]
```

gateway/src/apicast/policy/keycloak_scope_check/keycloak_scope_check.lua
mikz: I'd like to have more explanation of the use case in the integration test for this.
IMO we can cover the 2nd point by https://github.com/3scale/apicast/pull/773#discussion_r198058559. Using liquid we can compare any value in the JWT to the roles.

But the first one is quite hard to do dynamically. How that would work with hundreds of clients @nmasse-itix  ? I suppose you don't want to configure this in the gateway for every new client.

nmasse-itix: That's true, the first one is not meant to be dynamic. 

If you plan to use client roles this way, it should be well known clients roles, such as `realm-management/realm-admin` (for example to know if the user is administrator of the SSO Realm).

Right now, I cannot find a use case for referencing dynamically the role of **another** client. 

I agree that a design document would be helpful.

gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
nmasse-itix: I think `resource_roles` would be confusing since there is no resource role concept in RH-SSO. To add an item to the `resource_access` field of the access token, the administrator adds a *client role* to the *user roles*. 

I would stick to `client_roles`.

mikz: I see. It is quite confusing for me because the property is called `resource_access` and it has no relation to `client_roles`. 

I know it is quite lengthy but `resource_access_client_roles` would work for me pretty well. But if I'm the only one then lets leave it :)

gateway/src/apicast/policy/keycloak_scope_check/keycloak_scope_check.lua
mikz: @nmasse-itix I guess that with liquid support (https://github.com/3scale/apicast/pull/773#discussion_r198069574) we could use capture parameters from the mapping rules (or some other policy) as required role:


URL: `POST /MoneyTransfer/{client}`

```json
{
"client_roles": [
 { "name": "{{ usage.client }}", "name_type": "liquid", "client": "{{ usage.client }}", "client_type": "liquid" },
]
}
```

None of it is implemented yet, but at least we could have a path forward. 
@nmasse-itix do you think this information about another client would be available in the request ? URL, Headers, JWT, ...
If it would be somewhere in the request, we could match it and expose it for the scope verification.

y-tabata:
I added some examples to README.md in order to clarify the specification and help to understand and use this policy.

gateway/src/apicast/policy/keycloak_role_check/apicast-policy.json
```diff
@@ -0,0 +1,92 @@
...
+              "description": "Resource controlled by role. This is the same format as Mapping Rules. This matches from the beginning of the string and to make an exact match you need to use '$' at the end."
+            },
+            "resource_type": {
+              "description": "How to evaluate 'resource'",
+              "$ref": "#/definitions/value_type"
+            }
+          }
+        }
+      },
+      "type": {
```
davidor: I'm a bit confused by this. This type is global so it applies to all the scopes defined.
What if we want to whitelist some and blacklist others?

gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
```diff
@@ -0,0 +1,195 @@
...
+        return true
+      end
+    end
+
+  end
+
+  return false
+end
+
+local function auth_failed(service)
```
davidor: We should probably reuse the method that we have in the `Errors` module instead:
https://github.com/3scale/apicast/blob/a95ebddd2bc25e953fb8096ce181cd5123ec20dd/gateway/src/apicast/errors.lua#L12

gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
```diff
@@ -0,0 +1,195 @@
...
+
+    scope.resource_template_string = TemplateString.new(
+      scope.resource, scope.resource_type or default_type)
+
+  end
+end
+
+function _M.new(config)
+  local self = new()
+  self.config = config or {}
```
davidor: I don't think this is right.
`self.config` is assigned here, but it's never used, and then the code assumes that `config.type` and `config.scopes` exist.

gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
```diff
@@ -0,0 +1,195 @@
...
+
+  for _, scope in ipairs(scopes) do
+
+    local resource = scope.resource_template_string:render(context)
+
+    local mapping_rule = MappingRule.from_proxy_rule({
+      http_method = 'ANY',
+      pattern = resource,
+      querystring_parameters = {},
+      metric_system_name = 'hits'
```
davidor: 'hits' is not used anywhere in this code, right?

If that's the case, we should add a comment here clarifying that the name of the metric is irrelevant here, but we need to send it so MappingRule does not raise.

In the future, we should evaluate whether that `error()` in `MappinRule.new()` is really needed.

gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
```diff
@@ -0,0 +1,195 @@
...
+    local client_in_token = context.jwt.resource_access[client]
+
+    if not client_in_token then
+      ngx.log(ngx.DEBUG, "Client '", client, "' was not found in the access token.")
+      return false
+    end
+
+    local name = role.name_template_string:render(context)
+
+    if not check_roles_in_token(name, client_in_token.roles) then
```
davidor: Will `client_in_token.roles` always be != nil?

The function `check_roles_in_token()` expects a table and it will crash with nil. Maybe we should use `client_in_token.roles or {}` to be sure?

gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
y-tabata: Changed.

gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
y-tabata: Fixed.

gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
y-tabata: Added a comment.

gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
```diff
@@ -0,0 +1,195 @@
...
+  end
+
+  return false
+end
+
+local function match_realm_roles(scope, context)
+  if not scope.realm_roles then return true end
+
+  for _, role in ipairs(scope.realm_roles) do
+    if not context.jwt.realm_access then
```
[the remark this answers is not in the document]
y-tabata: Changed.

gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
y-tabata: According to Keycloak specification, `client_in_token.roles` always be != nil.
However, just to be on the safe side, I changed this.

spec/policy/keycloak_role_check/keycloak_role_check_spec.lua
```diff
@@ -0,0 +1,572 @@
+local KeycloakRoleCheckPolicy = require('apicast.policy.keycloak_role_check')
+
+describe('Keycloak Role check policy', function()
+
+  before_each(function()
+    ngx.header = {}
+  end)
+
+  describe('.access', function()
```
davidor: I think we should try to write `describe` and `it` comments in a way that they form a correct sentence. In my opinion, that would make the specs easier to follow and it would express more clearly the expected behavior of the policy.

Take the specs for the CORS policy for example:
https://github.com/3scale/apicast/blob/a95ebddd2bc25e953fb8096ce181cd5123ec20dd/spec/policy/cors/cors_spec.lua

For the specs of the `.rewrite` function, if we read the `describe` and `it` clauses together we get:
1) when the request is a CORS preflight, it exits with status code 204.
2) when the request is not a CORS preflight, it does not exit.

I don't think this should be a blocker for merging this PR. However, I think this is something we should take into account for future PRs.

gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
```diff
@@ -0,0 +1,194 @@
...
+
+local ipairs = ipairs
+local MappingRule = require('apicast.mapping_rule')
+local TemplateString = require('apicast.template_string')
+local errors = require('apicast.errors')
+local default_type = 'plain'
+
+local new = _M.new
+
+local function build_templates(scopes)
```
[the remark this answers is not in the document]
mikz: @y-tabata this will need to be fixed to pass CodeClimate checks. Either simplifying the method or splitting it into smaller ones.

`TemplateString.new(client_role.name, client_role.name_type or default_type)` is there repeated 3 times and can be extracted into a method.

spec/policy/keycloak_role_check/keycloak_role_check_spec.lua
y-tabata: Improved.

gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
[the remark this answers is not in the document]
y-tabata: Fixed.

davidor:
@y-tabata Thanks for fixing the codeclimate issues :+1:

PR #774, opened 2018-06-19
Fix reporting only 50% of calls to 3scale backend [THREESCALE-1080]

mikz:
Due to a bug in credentials juggling APIcast would send raw JWT to 3scale backend instead of just app_id.
That token is of course not understood and the reporting call rejected.

The fix changes how credentials are moved around to reduce risk of future issues.

Fixes [THREESCALE-1080](https://issues.jboss.org/browse/THREESCALE-1080)

@3scale/qe you might want to incorporate some tests that test metrics reporting accuracy to catch this kind of issues.

### DIFF
--- a/t/apicast-oidc.t
+++ b/t/apicast-oidc.t
@@ -1,64 +1,53 @@
 use lib 't';
-use Test::APIcast 'no_plan';
+use Test::APIcast::Blackbox 'no_plan';
 
-$ENV{TEST_NGINX_LUA_PATH} = "$Test::APIcast::spec/?.lua;$ENV{TEST_NGINX_LUA_PATH}";
-$ENV{TEST_NGINX_REDIS_HOST} ||= $ENV{REDIS_HOST} || "127.0.0.1";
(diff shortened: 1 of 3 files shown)


t/apicast-oidc.t
```diff
@@ -1,72 +1,113 @@
...
   iss => 'https://example.com/auth/realms/apicast',
-  exp => time + 10 }, key => \$::rsa, alg => 'RS256');
+  exp => time + 10 }, key => \$::private_key, alg => 'RS256');
 "Authorization: Bearer $jwt"
 --- no_error_log
 [error]
+
+
+
+=== TEST 2: Report calls to 3scale backend
```
davidor: I think this test is doing the same as the one above but making 2 requests instead of one

mikz: Exactly. I wanted it to be super obvious. Not sure if setting `repeat_each(2)` would be obvious enough.

davidor: In that case I would add a comment. Figuring out why this is being tested twice can be difficult without having some context like the PR title.

mikz: Good point 👍

PR #775, opened 2018-06-19
[proxy] refactor post_post action

mikz:
Extracted from https://github.com/3scale/apicast/pull/774

Refactors `proxy:post_action` to ease changes in #774. Proxy post_action should receive a context and be called from just one point - the policy.

### DIFF
--- a/gateway/src/apicast/proxy.lua
+++ b/gateway/src/apicast/proxy.lua
@@ -26,18 +26,18 @@ local setmetatable = setmetatable
 local ipairs = ipairs
 local encode_args = ngx.encode_args
 local resty_resolver = require 'resty.resolver'
-local semaphore = require('ngx.semaphore')
+local ngx_semaphore = require('ngx.semaphore')
 local backend_client = require('apicast.backend_client')
 local http_ng_ngx = require('resty.http_ng.backend.ngx')
-local timers = semaphore.new(tonumber(env.get('APICAST_REPORTING_THREADS') or 0))
+local timers = ngx_semaphore.new(tonumber(env.get('APICAST_REPORTING_THREADS') or 0))
 
 local empty = {}
 
 local response_codes = env.enabled('APICAST_RESPONSE_CODES')
 
-local using_post_action = response_codes or timers:count() < 1
+local using_sync_post_action = response_codes or timers:count() < 1
 
-if not using_post_action then
+if not using_sync_post_action then
   ngx.log(ngx.WARN, 'using experimental asynchronous reporting threads: ', timers:count())
 end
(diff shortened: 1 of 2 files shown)


gateway/src/apicast/proxy.lua
```diff
@@ -313,65 +309,63 @@ local function response_codes_data()
...
 
   local ok, err = timers:wait(10)
 
   if ok then
     -- TODO: try to do this in different phase and use semaphore to limit number of background threads
     -- TODO: Also it is possible to use sets in shared memory to enqueue work
-    local formatted_usage = format_usage(self.usage)
-    ngx.timer.at(0, post_action, self, cached_key, backend, formatted_usage, self.credentials, response_codes_data())
+
+    ngx.timer.at(0, handle_timer_post_action, self, timers, cached_key, backend, formatted_usage, credentials, response_codes_data())
```
mikz: @davidor looks like this is missing `self.extra_params_backend_authrep`

davidor: Good catch :+1: 

This means that the referrer filter policy does not work when using the experimental async reporting. I'll add a test and fix the issue.

davidor:
@mikz The integration tests were failing because the detection of whether to skip the post action was not done correctly. I've pushed a commit with the fix.

gateway/src/apicast/proxy.lua
```diff
@@ -350,8 +350,9 @@ end
 function _M:post_action(_)
   local cached_key = ngx.var.cached_key
 
-  if not cached_key or cached_key == "null" or cached_key ~= '' then
+  if not cached_key or cached_key == "null" or cached_key == '' then
       ngx.log(ngx.INFO, '[async] skipping after action, no cached key')
+      return
```
mikz: 😂

gateway/src/apicast/proxy.lua
```diff
@@ -313,65 +309,64 @@ local function response_codes_data()
...
 end
 
-local function post_action(_, self, cached_key, backend, ...)
+local function handle_timer_post_action(_, self, semaphore, cached_key, backend, ...)
   local res = util.timer('backend post_action', backend.authrep, backend, ...)
 
-  if not using_post_action then
-    timers:post(1)
+  if not using_sync_post_action then
+    semaphore:post(1)
```
mikz: I think this is really important when designing the `resty.concurrent.task`.
It needs to release the semaphore before continuing with the handler (that might terminate the request and the processing with `ngx.exit`) (even though I'm not sure how that behaves when executed in a timer)

PR #779, opened 2018-06-20
[oidc] fix reporting to backend when using OIDC [THREESCALE-1080]

mikz:
Fixes [THREESCALE-1080](https://issues.jboss.org/browse/THREESCALE-1080)
Simplified backport of https://github.com/3scale/apicast/pull/774 for 3.2 release.

This simpler backport does not rely or passing `context` object from `post_action` phase to `proxy` object, but rather just updates `proxy.credentials` after they are transformed by `oauth:transform_credentials` method. https://github.com/3scale/apicast/pull/774 relies on not storing the credentials in proxy at all and rather storing them only in the `context` object. 
But for a purpose of a patch release it is better to make minimal changes and reduce risk of unintended changes of behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Reporting only 50% calls to 3scale backend when using OIDC [PR #779](https://github.com/3scale/apicast/pull/779)
+
 ## [3.2.0] - 2018-06-04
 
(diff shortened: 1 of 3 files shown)


davidor:
@mikz what do you mean by 'simplified' backport?

mikz:
@davidor it does not port the whole post_action refactoring and passing context. It just properly updates 'self.credentials'. I consider this a bit dirty, but simpler approach. But appropriate in this situation of making minimal changes in patch releases.

Will update the PR description.

Issue #760, opened 2018-06-12
Rate Limiting policy


mikz:
Rate limiting policy was discussed and implemented in: https://github.com/3scale/apicast/pull/648, https://github.com/3scale/apicast/pull/703, https://github.com/3scale/apicast/issues/713

This meta issue is meant to gather all requirements for making rate limiting policy really useful in the next release.

- [X] Liquid templating the keys
- [X] Use JWT object for rate limiting
- [x] Apply rate limit to just a subset of requests
- [x] Allow dynamic reloading of configuration

## Liquid templating the keys

Defined by https://github.com/3scale/apicast/issues/713 and implemented by https://github.com/3scale/apicast/pull/719

## Use JWT object for rate limiting

Defined by https://github.com/3scale/apicast/issues/713, implemented by https://github.com/3scale/apicast/pull/718

## Allow dynamic reloading of configuration

Discussed in https://github.com/3scale/apicast/issues/667, being implemented in https://github.com/3scale/apicast/pull/703 and https://github.com/3scale/apicast/pull/758.

## Apply rate limit to just a subset of requests

The use case is: applying different rate limit to POST and GET requests.

First option is using https://github.com/3scale/apicast/issues/744 to apply different rate limiting policies to different endpoints.

The alternative is we allow setting the policy limit by Liquid too. That could allow following definitions: 
```liquid
{% if request == 'GET' %}50{% else %}10{% endif %}
```
However, this approach can be very error prone and not really intuitive.

Another alternative would be to define "values" for the rate limiting keys. For example the rate limit key `{{ request.method }}` could have several values: `POST: 50`, `GET: 10`.
This can be error prone too and easy to miss some values.

y-tabata:
Regarding "Apply rate limit to just a subset of requests", since those who want to add conditions to the policy maybe want to add a lot of conditions to control very detailed case, I think we should reduce the overhead of adding conditions.

Regarding "Conditional policy", in my understanding, the more conditions increase, the more policy chains increase. The overhead is big, I think.

Regarding "Match phase", I don't understand this exactly, we can use this like nginx phase, right? If it is easy and intuitive to define conditions, we can apply this one.

Regarding alternative approaches, I think it can be error prone, too. If defining conditions, we should add another property dedicated to conditions.

mikz:
@y-tabata  thank you!

### Match phase

[...]

PR #781, opened 2018-06-22
[rate-limit] add condition to the key

y-tabata:
Closes "Apply rate limit to just a subset of requests" of #760.


y-tabata:
Close once.

PR #782, opened 2018-06-22
Add TimerTask

davidor:
This PR adds a new module that allows running periodic tasks that can be canceled.

### DIFF
--- a/gateway/src/resty/concurrent/timer_task.lua
+++ b/gateway/src/resty/concurrent/timer_task.lua
@@ -1,0 +1,94 @@
+local uuid = require 'resty.jit-uuid'
+local setmetatable = setmetatable
+local getmetatable = getmetatable
+local newproxy = newproxy
+local unpack = unpack
+
+local _M = {}
+
+local default_interval_seconds = 60
+
+_M.active_tasks = {}
+
+function _M.register_task(id)
+  _M.active_tasks[id] = true
+end
+
+function _M.unregister_task(id)
+  _M.active_tasks[id] = nil
+end
+
+function _M.task_is_active(id)
+  return _M.active_tasks[id] or false
+end
+
+local function generate_id()
+  return uuid.generate_v4()
+end
+
+local function mt(id)
+  -- Using 'newproxy' is needed to be able to define __gc().
+  -- With __gc we can make sure that the task will stop scheduling more work
+  -- after it has been garbage collected.
+  local proxy = newproxy(true)
+  local res_mt = getmetatable(proxy)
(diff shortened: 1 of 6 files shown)


gateway/src/resty/concurrent/timer_task.lua
```diff
@@ -0,0 +1,94 @@
...
+  self.args = options.args
+  self.interval = options.interval or default_interval_seconds
+  self.id = id
+
+  _M.register_task(id)
+
+  return self
+end
+
+local run_periodic, schedule_next, timer_execute
```
mikz: That is very weird style. Why not just define each function as local?

mikz: I see. They all depend on each other so they need to be defined this way.

spec/resty/concurrent/timer_task_spec.lua
```diff
@@ -0,0 +1,136 @@
...
+    local interval = 10
+
+    before_each(function()
+      ngx_timer_stub = stub(ngx_timer, 'at')
+    end)
+
+    describe('when the task is active', function()
+      it('runs the task', function()
+        local timer_task = TimerTask.new(func, { args = args, interval = interval })
+        local func_stub = stub(timer_task, 'task')
```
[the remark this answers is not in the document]
mikz: Codeclimate does not update comments after initial review?

gateway/src/resty/concurrent/timer_task.lua
davidor: Exactly.

gateway/src/resty/concurrent/timer_task.lua
```diff
@@ -0,0 +1,94 @@
...
+
+run_periodic = function(self)
+  if not _M.task_is_active(self.id) then return end
+
+  self.task(unpack(self.args))
```
mikz: Would be good to use https://github.com/3scale/apicast/pull/783/commits/4a5a8159602c5e1173e0a6dabb6529ee34f10551 to make this safe.

spec/resty/concurrent/timer_task_spec.lua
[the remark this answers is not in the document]
davidor: Looks like it writes another comment without deleting the old one.

gateway/src/resty/concurrent/timer_task.lua
davidor: Yes, I can do that after resty.concurrent is merged.

PR #786, opened 2018-06-22
Use TimerTask in the 3scale batcher policy

davidor:
This PR replaces the usages of `ngx.timer.every` with `TimerTask` instances.

This solves the issue of not being able to cancel timers. Recurrent tasks created using `TimerTask` can be canceled, but we cannot cancel timers created with `ngx.timer.every`. This was causing a bug, Timer would continue running even after a config reload.

### DIFF
--- a/gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
+++ b/gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
@@ -15,9 +15,19 @@ local ipairs = ipairs
 local default_auths_ttl = 10
 local default_batch_reports_seconds = 10
 
-local _M = policy.new('Caching policy')
+local _M, mt = policy.new('Caching policy')
 
 local new = _M.new
+
+mt.__gc = function(self)
+  -- Instances of this policy are garbage-collected when the config is
+  -- reloaded. We need to ensure that the TimerTask instance schedules another
+  -- run before that so we do not leave any pending reports.
+
+  if self.timer_task then
+    self.timer_task:cancel(true)
+  end
+end
 
 function _M.new(config)
   local self = new(config)
--- a/gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
+++ b/gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
@@ -8,6 +8,7 @@ local reporter = require('reporter')
 local Transaction = require('transaction')
 local http_ng_resty = require('resty.http_ng.backend.resty')
 local semaphore = require('ngx.semaphore')
+local TimerTask = require('resty.concurrent.timer_task')
 
 local ipairs = ipairs
 
@@ -30,9 +31,7 @@ function _M.new(config)
   self.batch_reports_seconds = config.batch_report_seconds or
                                default_batch_reports_seconds
 
-  self.report_timer_on = false
-
-  -- Semaphore used to ensure that only one timer is started per worker.
+  -- Semaphore used to ensure that only one TimerTask is started per worker.
   local semaphore_report_timer, err = semaphore.new(1)
   if not semaphore_report_timer then
     ngx.log(ngx.ERR, "Create semaphore failed: ", err)
@@ -70,7 +69,7 @@ local function set_flags_to_avoid_auths_in_apicast(context)
   context.skip_apicast_post_action = true
 end
 
-local function report(_, service_id, backend, reports_batcher)
+local function report(service_id, backend, reports_batcher)
   local reports = reports_batcher:get_all(service_id)
 
   if reports then
@@ -81,22 +80,33 @@ local function report(_, service_id, backend, reports_batcher)
   reporter.report(reports, service_id, backend, reports_batcher)
 end
 
--- This starts a timer on each worker.
--- Starting a timer on each worker means that there will be more calls to
+local function timer_task(self, service_id, backend)
+  local task = report
+
+  local task_options = {
+    args = { service_id, backend, self.reports_batcher },
+    interval = self.batch_reports_seconds
+  }
+
+  return TimerTask.new(task, task_options)
+end
+
+-- This starts a TimerTask on each worker.
+-- Starting a TimerTask on each worker means that there will be more calls to
 -- 3scale backend, and the config param 'batch_report_seconds' becomes
 -- more confusing because the reporting frequency will be affected by the
 -- number of APIcast workers.
--- If we started a timer just on one of the workers, it could die, and then,
+-- If we started a TimerTask just on one of the workers, it could die, and then,
 -- there would not be any reporting.
-local function ensure_report_timer_on(self, service_id, backend)
-  local check_timer = self.semaphore_report_timer:wait(0)
+local function ensure_timer_task_created(self, service_id, backend)
+  local check_timer_task = self.semaphore_report_timer:wait(0)
 
-  if check_timer then
-    if not self.report_timer_on then
-      ngx.timer.every(self.batch_reports_seconds, report,
-        service_id, backend, self.reports_batcher)
+  if check_timer_task then
+    if not self.timer_task then
+      self.timer_task = timer_task(self, service_id, backend)
 
-      self.report_timer_on = true
+      self.timer_task:execute()
+
       ngx.log(ngx.DEBUG, 'scheduled 3scale batcher report timer every ',
                          self.batch_reports_seconds, ' seconds')
     end
@@ -166,7 +176,7 @@ function _M:access(context)
   local credentials = context.credentials
   local transaction = Transaction.new(service_id, credentials, usage)
 
-  ensure_report_timer_on(self, service_id, backend)
+  ensure_timer_task_created(self, service_id, backend)
 
   local cached_auth = self.auths_cache:get(transaction)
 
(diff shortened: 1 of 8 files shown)


gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
```diff
@@ -81,27 +80,38 @@ local function report(_, service_id, backend, reports_batcher)
...
 -- 3scale backend, and the config param 'batch_report_seconds' becomes
 -- more confusing because the reporting frequency will be affected by the
 -- number of APIcast workers.
--- If we started a timer just on one of the workers, it could die, and then,
+-- If we started a TimerTask just on one of the workers, it could die, and then,
 -- there would not be any reporting.
-local function ensure_report_timer_on(self, service_id, backend)
-  local check_timer = self.semaphore_report_timer:wait(0)
+local function ensure_timer_task_created(self, service_id, backend)
+  local check_timer_task = self.semaphore_report_timer:wait(0)
```
mikz: Can we move this to `.new` ? This policy should not be initialized in `init` phase, right?

davidor: We can't because we don't have access to the service ID in `.new()`

mikz: Interesting. Maybe better to leave it for later, butI think we should think about changing the data structure to remove the service_id dependency. It should be as simple as create timer in new that gets reports from reports batcher. There should be no need for service_id because reports batcher instance is specific to this policy and timer, so it will ever contain only reports from this service. I see the reports batcher actually uses that service_id to do a lock on shmem. Hopefully we can solve that.

davidor: You're right. That part can be simplified a lot.

mikz: Could we generate some UUID instead of using service id ? Then every policy would take care just of its own data.

davidor: This is related to an idea that we discussed some time ago.

Right now, pending reports are stored in a shared dictionary and every instance of the policy creates a timer (per worker) to report all the pending reports for its service ID.

After introducing `TimerTask` I see it more clearly that it might be better to adopt a different strategy. We could store the pending reports in a table of the instance, as we know that there'll only be one instance of the policy for a given service ID per Apicast worker. Pros: no locks, simpler code. Cons: possibility of losing reports if a worker dies, which in practice I don't think it'll be an issue.

davidor: Let's leave this for a future PR.
In this one, I'd like to focus on switching to using `TimerTask` to be able to cancel timers.

mikz: Yep, definitely 👍

davidor:
@mikz There's one remaining issue. 

My idea was to include a `__mt` method in the policy and report to backend any pending transactions there. However, it looks like we can't make network calls when garbage-collecting. I don't see an easy way to do that without introducing a finalizer that does not depend on `collectgarbage()` for policies.

In practice, this means that reports can be left there when the policy is garbage collected and there are pending transactions for its service and that service is not included in the config when it is reloaded. Also, when APIcast is stopped.

gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
```diff
@@ -160,13 +180,15 @@ end
 -- calls to backend.
 function _M:access(context)
   local backend = backend_client:new(context.service, http_ng_resty)
+self.backend = backend
   local usage = context.usage
   local service = context.service
   local service_id = service.id
+  self.service_id = service_id
```
mikz: Why it is being assigned to self ?

gateway/src/resty/concurrent/timer_task.lua
```diff
@@ -89,6 +89,13 @@ function _M:execute(run_now)
   run_periodic(run_now or false, self.id, self.task, self.args, self.interval)
 end
 
+--- Schedule one task.
+function _M:schedule_one(delay)
+  -- Need to wrap the task in a function that discards the first param (the
+  -- "premature" one sent by ngx.timer.at)
+  ngx.timer.at(delay or 0, function(_, ...) self.task(...) end, unpack(self.args))
```
mikz: This is going to allocate a function for each call. Is that necessary?

gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
davidor: Sorry, this is a leftover from some test. Same with `self.backend`.

spec/resty/concurrent/timer_task_spec.lua
```diff
@@ -114,6 +114,10 @@ describe('TimerTask', function()
         timer_task:execute(true)
 
         assert.stub(ngx_timer_stub).was_called()
+
+        -- Can't check all the arguments of ngx.timer.at because it calls an
+        -- private function but at least we can check the interval (first arg)
+        assert.equals(interval, ngx_timer_stub.calls[1].vals[1])
```
mikz: I think you can change this to use matchers and verify it was called with some function (`matcher.function`) and exact second parameter.

gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
```diff
@@ -8,16 +8,27 @@ local reporter = require('reporter')
...
 local new = _M.new
 
+mt.__gc = function(self)
+  -- Instances of this policy are garbage-collected when the config is
+  -- reloaded. We need to ensure that the TimerTask instance schedules another
+  -- run before that so we do not leave any pending reports.
+
+  if self.timer_task then
+    self.timer_task:schedule_one(1)
+  end
```
mikz: And we will rely on cancelling the timer by __gc callbacks ?

gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
```diff
@@ -8,16 +8,27 @@ local reporter = require('reporter')
...
 
 local new = _M.new
 
+mt.__gc = function(self)
+  -- Instances of this policy are garbage-collected when the config is
+  -- reloaded. We need to ensure that the TimerTask instance schedules another
+  -- run before that so we do not leave any pending reports.
+
+  if self.timer_task then
+    self.timer_task:schedule_one(1)
```
mikz: Why delay 1 and not 0 ?

gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
davidor: That works, but let's cancel the timer_task here to be explicit about it.

gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
davidor: :)

From the point of view of the policy, I think it does not matter. However, busted crashes with a segfault when this is 0. Not sure why exactly. It might be related to how busted collects the garbage after all the tests. It does not crash with a delay of 0.1.

mikz: Ok, then this is definitely worth a comment :-D

spec/resty/concurrent/timer_task_spec.lua
davidor: But the function is private, so I don't think we can do that.

mikz: Matcher can match the type like:

```lua
assert.spy(ngx_timer_stub).was_called_with(matcher.is_function(), 1)
```

http://olivinelabs.com/busted/#matchers

gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
davidor: Fixed.

gateway/src/resty/concurrent/timer_task.lua
davidor: Does no longer apply. This method has been replaced.

gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
davidor: Done :+1:

gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
davidor: In the end I changed this and implemented it as we discussed off-line: https://github.com/3scale/apicast/pull/786/commits/ba883c101ee8990af50cb56cd52784d74df00743

spec/resty/concurrent/timer_task_spec.lua
davidor: I'm familiar with matchers, but I misunderstood you.

I thought that you wanted to check that ngx_timer_stub was called with a specific function and I was saying that it was not possible because it's a private one. Now I see that you suggested just to check that it's a function no matter which one.

davidor: I added a check for this :+1:

PR #790, opened 2018-06-26
Workaround to make __gc work on tables

davidor:
In LuaJIT and Lua 5.1, the `__gc` metamethod does not work with tables, it only works with "userdata". This PR introduces a module that implements a workaround to make it work with tables.

### DIFF
--- a/gateway/src/apicast/gc.lua
+++ b/gateway/src/apicast/gc.lua
@@ -1,0 +1,94 @@
+-- In LuaJIT and Lua 5.1, the __gc metamethod does not work in tables, it only
+-- works in "userdata". This module introduces a workaround to make it work
+-- with tables.
+
+local rawgetmetatable = debug.getmetatable
+local getmetatable = getmetatable
+local setmetatable = setmetatable
+local newproxy = newproxy
+local ipairs = ipairs
+local pairs = pairs
+local pcall = pcall
+local table = table
+local unpack = unpack
+local error = error
+local tostring = tostring
+
+local _M = {}
+
+local function original_table(proxy)
+  return rawgetmetatable(proxy).__table
+end
+
+local function __gc(proxy)
+  local t = original_table(proxy)
+  local mt = getmetatable(proxy)
+
+  if mt and mt.__gc then mt.__gc(t) end
+end
+
+local function __tostring(proxy)
+  return tostring(original_table(proxy))
+end
+
+local function __call(proxy, ...)
+  local t = original_table(proxy)
+
+  -- Try to run __call() and if it's not possible, try to run it in a way that
+  -- it returns a meaningful error.
+  local ret = { pcall(t, ...) }
+  local ok = table.remove(ret, 1)
+
+  if ok then
+    return unpack(ret)
+  else
+    error(ret[1], 2)
+  end
+end
+
+local function __len(proxy)
+  return #(original_table(proxy))
+end
+
+local function __ipairs(proxy)
+  return ipairs(original_table(proxy))
+end
+
+local function __pairs(proxy)
+  return pairs(original_table(proxy))
+end
+
+--- Set a __gc metamethod in a table
+-- @tparam table t A table
+-- @tparam table metatable A table that will be used as a metatable. It needs
+--  to define __gc.
+function _M.set_metatable_gc(t, metatable)
+  setmetatable(t, metatable)
+
+  -- newproxy() returns a userdata instance
+  local proxy = newproxy(true)
+
+  -- We are going to define a metatable in the userdata instance to make it act
+  -- like a table. To do that, we'll just define the metamethods a table should
+  -- respond to.
+  local mt = getmetatable(proxy)
+
+  mt.__gc = __gc
+
+  mt.__index = t
+  mt.__newindex = t
+  mt.__table = t
+
+  mt.__call = __call
+  mt.__len = __len
+  mt.__ipairs = __ipairs
+  mt.__pairs = __pairs
+  mt.__tostring = __tostring
+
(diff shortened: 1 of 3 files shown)


spec/gc_spec.lua
```diff
@@ -0,0 +1,96 @@
+local GC = require('apicast.gc')
+
+describe('GC', function()
+  describe('.enable', function()
+    it('enables GC on a table', function()
+      local table = { 1, 2, 3 }
+      setmetatable(table, { __gc = function() end })
+      stub(getmetatable(table), '__gc')
+
+      local table_with_gc = GC.enable(table)
```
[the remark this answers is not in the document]
davidor: It's OK. We need to define it so it's garbage collected later.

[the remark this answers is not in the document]
davidor: I approved this one manually on codeclimate.

gateway/src/apicast/gc.lua
```diff
@@ -0,0 +1,53 @@
...
+  -- like a table. To do that, we'll just define the metamethods a table should
+  -- respond to.
+  local mt = getmetatable(proxy)
+
+  mt.__gc = __gc
+  mt.__call = __call
+
+  mt.__index = table
+  mt.__newindex = table
+  mt.__table = table
```
mikz: I think it would be nice to also hide the this metatable by:

```lua
mt.__metatable = getmetatable(table)
```

Then `getmetatable(proxy)` will return the metatable returned by `getmetatable(table)`. 

This would not propagate if someone would change the metatable of `table` directly, but that can be avoided by changing the API a bit.

gateway/src/apicast/gc.lua
```diff
@@ -0,0 +1,53 @@
...
+local function __call(proxy, ...)
+  local t = getmetatable(proxy).__table
+  local mt = getmetatable(t)
+
+  if mt and mt.__call then
+    return mt.__call(...)
+  end
+end
+
+function _M.enable(table)
```
mikz: We could rename this library to `setmetatable_gc(table, mt)`. 

It could make more sense when using the `__metatable` key. 
All changes to GC'd objects metatables should be done with this method.

gateway/src/apicast/gc.lua
```diff
@@ -0,0 +1,53 @@
...
+  local mt = getmetatable(proxy)
+
+  mt.__gc = __gc
+  mt.__call = __call
+
+  mt.__index = table
+  mt.__newindex = table
+  mt.__table = table
+
+  mt.__len = function() return #table end
```
mikz: We should avoid allocating functions for each object.

These can be shared by (even when using `__metatable`):

```lua
local rawgetmetatable = debug.getmetatable
local getmetatable = getmetatable

local function __len(t)
  return #(rawgetmetatable(t).__table)
end
```

gateway/src/apicast/gc.lua
```diff
@@ -0,0 +1,53 @@
...
+
+local function __gc(proxy)
+  local t = getmetatable(proxy).__table
+  local mt = getmetatable(t)
+
+  if mt and mt.__gc then mt.__gc(t) end
+end
+
+local function __call(proxy, ...)
+  local t = getmetatable(proxy).__table
```
mikz: I think we should defined this as:

```lua
getmetatable(proxy).__table(...)
```

Because now it basically defines all tables callable.

Unfortunately that can yield rather cryptic error message: `attempt to call field '__table' (a nil value)`.

Unfortunately Lua tries to be helpful and the error message depends on how you call the table:
* `attempt to call local 'varname' (a table value)`
* `attempt to call a table value`

So replicating those messages might not be possible.

Based on how the function is stored (or if it is a field).
The only workaround I can imagine is using `pcall` like:

```lua
local function __call(self, ...)
  local mt = rawgetmetatable(self)
  local ret = { pcall(mt.__table, ...) }

  local ok = table.remove(ret, 1)

  if ok then
	  return unpack(ret)
  else
	  error(ret[1], 2)
  end
end
```

That would throw error: `ERROR: gc.lua:66: attempt to call a table value` (even with correct line pointing to the original caller).

spec/gc_spec.lua
```diff
@@ -0,0 +1,98 @@
...
+
+describe('GC', function()
+  describe('.enable', function()
+    it('enables GC on a table', function()
+      local test_table = { 1, 2, 3 }
+      setmetatable(test_table, { __gc = function() end })
+      stub(getmetatable(test_table), '__gc')
+
+      local table_with_gc = GC.enable(test_table)
+      table_with_gc = nil
```
mikz: We could avoid that CC error by `assert(table_with_gc)` right ?

gateway/src/apicast/gc.lua
davidor: I didn't know about `debug.getmetatable` :+1:

davidor:
@mikz I addressed your comments. In some cases, I adopted a slightly different solution. Can you take a look?

gateway/src/apicast/gc.lua
```diff
@@ -0,0 +1,100 @@
...
+
+local function __gc(proxy)
+  local t = original_table(proxy)
+  local mt = getmetatable(proxy)
+
+  if mt and mt.__gc then mt.__gc(t) end
+end
+
+local function __tostring(proxy)
+  local mt = getmetatable(proxy)
```
mikz: Why not calling `tostring(original_table(proxy))` ?

davidor: This was needed because I wasn't calling `setmetatable(t, metatable)` at the beginning of `set_metatable_gc`. I needed to do that, otherwise, __index would not work.
I fixed this in a new commit. Please check it. I'll squash it after review.

PR #791, opened 2018-06-26
TimerTask improvements

davidor:
This PR implements several improvements in the `TimerTask` module:

1) Adapts the module to use the new `GC` module introduced in the previous PR #790
2) Fixes a bug that prevents the task from being garbage collected. This is done by avoiding passing `self`.
3) Fixes a bug when calculating the delay when scheduling the next task.

### DIFF
--- a/gateway/src/resty/concurrent/timer_task.lua
+++ b/gateway/src/resty/concurrent/timer_task.lua
@@ -1,7 +1,6 @@
+local GC = require 'apicast.gc'
+
 local uuid = require 'resty.jit-uuid'
-local setmetatable = setmetatable
-local getmetatable = getmetatable
-local newproxy = newproxy
 local unpack = unpack
 
 local _M = {}
@@ -26,16 +25,14 @@ local function generate_id()
   return uuid.generate_v4()
 end
 
-local function mt(id)
-  -- Using 'newproxy' is needed to be able to define __gc().
-  -- With __gc we can make sure that the task will stop scheduling more work
-  -- after it has been garbage collected.
-  local proxy = newproxy(true)
-  local res_mt = getmetatable(proxy)
-  res_mt.__gc = function() _M.unregister_task(id) end
-  res_mt.__index = _M
-  return res_mt
+local function gc(self)
+  _M.unregister_task(self.id)
 end
+
+local mt = {
+  __gc = gc,
+  __index = _M
+}
 
 --- Initialize a TimerTask.
 -- @tparam function task The function to be run periodically
@@ -47,7 +44,7 @@ function _M.new(task, opts)
 
   local id = generate_id()
 
-  local self = setmetatable({}, mt(id))
+  local self = GC.set_metatable_gc({}, mt)
   self.task = task
   self.args = options.args
   self.interval = options.interval or default_interval_seconds
--- a/gateway/src/resty/concurrent/timer_task.lua
+++ b/gateway/src/resty/concurrent/timer_task.lua
@@ -57,25 +57,25 @@ end
 
 local run_periodic, schedule_next, timer_execute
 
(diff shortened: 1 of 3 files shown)


gateway/src/resty/concurrent/timer_task.lua
```diff
@@ -60,25 +56,32 @@ end
...
   if run_now then
-    self.task(unpack(self.args))
+    func(unpack(args))
   end
 
-  schedule_next(self)
+  local t_diff = ngx_now() - t_start
+
+  local delay_for_next = interval - t_diff
+  if delay_for_next < 0 then delay_for_next = 0 end
```
mikz: I think LuaJIT recommends using `math.max` instead of conditions (as they can't be  JITed).

http://wiki.luajit.org/Numerical-Computing-Performance-Guide

gateway/src/resty/concurrent/timer_task.lua
```diff
@@ -60,25 +56,32 @@ end
...
 
   if run_now then
-    self.task(unpack(self.args))
+    func(unpack(args))
   end
 
-  schedule_next(self)
+  local t_diff = ngx_now() - t_start
+
+  local delay_for_next = interval - t_diff
```
mikz: Is it really needed?
I think it is just fine to delay the execution. Documenting it is way easier than doing this. Also it could end up in situation when are two called immediately after each other because they took more than the delay.

davidor: I think from a user perspective is more natural to expect a consistent reporting every X seconds. But there are no guarantees anyway because some reports might take more than the interval specified in the config. I don't think it's too important, we can remove it to simplify this a bit.

mikz: I'd prefer it to be simpler. It can be documented pretty easily. 

For the record, `ngx.timer.every` behaves quite stupidly and will schedule timer every `N` seconds regardless how long the timer takes.

gateway/src/resty/concurrent/timer_task.lua
```diff
@@ -27,14 +27,10 @@ local function generate_id()
...
-  -- Using 'newproxy' is needed to be able to define __gc().
-  -- With __gc we can make sure that the task will stop scheduling more work
-  -- after it has been garbage collected.
-  local proxy = newproxy(true)
-  local res_mt = getmetatable(proxy)
-  res_mt.__gc = function() _M.unregister_task(id) end
-  res_mt.__index = _M
-  return res_mt
+  return {
+    __gc = function() _M.unregister_task(id) end,
```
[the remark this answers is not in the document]
davidor: I don't think so because it needs the 'id'.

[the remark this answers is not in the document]
mikz: But it can get the id from `self` no ? It will get the table as first argument, so it can take the id from there.

davidor:
Comments addressed, some of them no longer apply after removing https://github.com/3scale/apicast/pull/791#discussion_r198391501

gateway/src/resty/concurrent/timer_task.lua
```diff
@@ -26,15 +25,15 @@ local function generate_id()
...
-  local proxy = newproxy(true)
-  local res_mt = getmetatable(proxy)
-  res_mt.__gc = function() _M.unregister_task(id) end
-  res_mt.__index = _M
-  return res_mt
+local function gc(self)
+  _M.unregister_task(self.id)
+end
+
+local function mt()
```
mikz: Now there is no need to allocate a metatable per instance and can just share one, right?

davidor: :+1:

PR #797, opened 2018-06-28
[circleci] install native apicast for profiling

mikz:
Create new job that can run stapxx and other OpenResty profiling tools.
Lets create some automated profiling tasks that record artifacts, so we can track how much Lua code is JIT compiled, GC pressure, etc.


mikz:
Travis fails because https://issues.jboss.org is down.

PR #798, opened 2018-07-03
[Docs] Updating native install centos docs to a working version

alanmoran:
Updating the native install instructions for CentOS. Also including them as RHEL instructions too as they cover both. 

cc: @mayorova


mikz:
@alanmoran you'll need to rebase this PR on master to get it green.

alanmoran:
@mikz I'm happy to create the Dockerfile and I agree that's a much more maintainable idea. Should it cover everything from installing deps, lastest APIcast and running it? (with mock creds - using ARGS)

Where in APICast should this live? Inside `doc/` ?

mikz:
@alanmoran I'd say it should cover what would the readme cover. It should be just executable documentation that can be verified that works. It can also be a shell script and run that shell script inside Dockerfile. @mayorova created a Dockerfile that should work. It should be just matter of cleaning it up and commenting to act as a documentation. 

But if we go there and create a Dockerfile, then I wonder why we can't just refer to s2i-openresty Dockerfile ? 

Maybe we can cleanup/comment https://github.com/3scale/s2i-openresty/blob/master/Dockerfile and just refer to that ? It would be less things to maintain.

eguzki:
Is this PR still valid? Most of the steps are being implemented in https://github.com/3scale/APIcast/pull/1335 in a Dockerfile based on ubi8

PR #799, opened 2018-07-04
[busted] fix leaking tests

mikz:
Stubs and global env modifications would leak in between tests.

This PR introduces several protections against that and fixes running the tests in loop without automatic insulation from busted.

```shell
rover exec bin/busted  --repeat=100 --no-auto-insulate
rover exec bin/busted  --repeat=10 # less repetitions, because busted leaks memory
```

### DIFF
--- a/gateway/src/apicast/configuration_store.lua
+++ b/gateway/src/apicast/configuration_store.lua
@@ -18,7 +18,15 @@ if env.enabled('APICAST_PATH_ROUTING_ENABLED') then ngx.log(ngx.WARN, 'DEPRECATI
 
 local mt = { __index = _M, __tostring = function() return 'Configuration Store' end }
 
-function _M.new(cache_size)
+function _M.new(cache_size, options)
+  local path_routing
+
+  if options and options.path_routing ~= nil then
+    path_routing = options.path_routing
+  else
+    path_routing = _M.path_routing
+  end
+
   return setmetatable({
     -- services hashed by id, example: {
     --   ["16"] = service1
@@ -31,6 +39,8 @@ function _M.new(cache_size)
     --    { service2 }
     --  }
     cache = lrucache.new(cache_size or _M.cache_size),
+
+    path_routing = path_routing,
 
     cache_size = cache_size
 
@@ -117,7 +127,7 @@ function _M.store(self, config, ttl)
         local host = lower(hosts[j])
         local h = by_host[host]
 
-        if #(h) == 0 or _M.path_routing then
+        if #(h) == 0 or self.path_routing then
           insert(h, service)
         else
           ngx.log(ngx.WARN, 'skipping host ', host, ' for service ', id, ' already defined by service ', h[1].id)
--- a/spec/luassert_helper.lua
+++ b/spec/luassert_helper.lua
@@ -1,18 +1,30 @@
 local busted = require('busted')
-local assert = require("luassert")
+local assert = require('luassert.assert')
 local util = require ('luassert.util')
 local say = require('say')
 
 do -- set up reverting stubs
-  local snapshot
+  local state = require("luassert.state")
 
-  busted.before_each(function()
-    snapshot = assert:snapshot()
-  end)
+  local function snapshot()
+    --- creates a snapshot and adds it to a stack
+    state.snapshot()
+    return nil, true -- to not stop the chain
+  end
 
-  busted.after_each(function()
-    snapshot:revert()
-  end)
+  local function revert()
+    --- reverts state of the current snapshot and removes it from the stack
+    state.revert()
+    return nil, true
+  end
+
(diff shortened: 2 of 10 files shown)


spec/spec_helper.lua
```diff
@@ -54,3 +47,16 @@ end)
 busted.subscribe({ 'file', 'end' }, function ()
   collectgarbage()
 end)
+
+do
```
davidor: I think we need to explain what this does, which is memoize ffi.cdef calls.
I also think we need to explain why we need to do this. It's hard to know the reason to do this without any context.

gateway/src/apicast/configuration_store.lua
```diff
@@ -18,7 +18,15 @@ if env.enabled('APICAST_PATH_ROUTING_ENABLED') then ngx.log(ngx.WARN, 'DEPRECATI
 
 local mt = { __index = _M, __tostring = function() return 'Configuration Store' end }
 
-function _M.new(cache_size)
+function _M.new(cache_size, options)
+  local path_routing
+
+  if options then
```
davidor: I think this should be `if options and options.path_routing`.

spec/luassert_helper.lua
```diff
@@ -1,18 +1,28 @@
...
 do -- set up reverting stubs
-  local snapshot
+  local state = require("luassert.state")
 
-  busted.before_each(function()
-    snapshot = assert:snapshot()
-  end)
+  local function snapshot()
+    state.snapshot()
+    return nil, true
```
davidor: Why do all these methods need to return `nil, true` ?

spec/luassert_helper.lua
```diff
@@ -1,18 +1,28 @@
 local busted = require('busted')
-local assert = require("luassert")
+local assert = require('luassert.assert')
 local util = require ('luassert.util')
 local say = require('say')
 
 do -- set up reverting stubs
-  local snapshot
+  local state = require("luassert.state")
```
davidor: Before this change we were calling `require('luassert'):snapshot()`. Is this equivalent to `require('luassert.state').snapshot()` ? What are the differences?

spec/ngx_helper.lua
```diff
@@ -51,6 +59,19 @@ local function reset_ngx_state()
   reset_ngx_shared(ngx_shared_original)
 end
 
+local function verify()
```
davidor: This needs a comment. You could reuse the one you added in the commit.
I think it's important to explain both what this is doing and why.

spec/ngx_helper.lua
```diff
@@ -1,9 +1,17 @@
+require('resty.core')
```
davidor: Why wasn't this required until now?

spec/ngx_helper.lua
```diff
@@ -51,6 +59,19 @@ local function reset_ngx_state()
   reset_ngx_shared(ngx_shared_original)
 end
 
+local function verify()
+  local allowed = { var = true, header = true, ctx = true }
```
davidor: Add a comment explaining why these are allowed and the rest are not?

spec/spec_helper.lua
```diff
@@ -1,5 +1,47 @@
-require 'ngx_helper'
+do
```
davidor: Please include the commit description here.

spec/spec_helper.lua
```diff
@@ -1,5 +1,47 @@
...
+      if name == var then
+        return val
+      elseif not name then
+        break
+      end
+      i = i + 1
+    end
+  end
+
+  local busted = getlocal(6, 'busted')
```
davidor: 6?

spec/spec_helper.lua
mikz: Yep.

spec/luassert_helper.lua
mikz: Otherwise the callback chain gets cancelled. Going to write a comment / helper.

spec/luassert_helper.lua
mikz: Yep. `state.snapshot()` creates one and stores it on a stack, so you don't need to have a reference to it. You can just push/pop from the stack. 

With `:snapshot()` and `:revert()` it was increasingly hard to keep the state somewhere (like on a stack) and it uses luassert.state underneath anyway.

spec/ngx_helper.lua
mikz: This is not strictly necessary and will work without it because resty.core is required in the busted.lua. But I want this file to be meaningful standalone and for that I think it is necessary to require resty.core because it overrides ngx methods. So if it would be required at some later point it would not pass the verification.

spec/spec_helper.lua
mikz: :-D 6 layers of stack deep, mind blowing :) will write a comment

Issue #709, opened 2018-05-15
Support HTTP Proxy


mikz:
APIcast should support proxying outgoing HTTP(S) connections through HTTP(S) proxy.

The underlying support is there: https://github.com/pintsized/lua-resty-http/pull/112

JIRA https://issues.jboss.org/browse/THREESCALE-221

# Use Cases

APIcast deployed in network constrained environment where all outgoing traffic has to go through HTTP proxy.
Starting APIcast with `http_proxy` ENV variable should make it use HTTP proxy.
APIcast should be able to use at least [Basic Authentication](https://tools.ietf.org/html/rfc7617#section-2) to the HTTP(S) proxy ([RFC7231 4.3.6 CONNECT](https://tools.ietf.org/html/rfc7231#section-4.3.6), [RFC7235, Section 4.3 Proxy-Authenticate](https://tools.ietf.org/html/rfc7235#section-4.3) , [RFC7235, Section 4.4 Proxy-Authorization](https://tools.ietf.org/html/rfc7235#section-4.4))

# Concerns

Using keepalives with HTTPS proxy could require small patch to the lua-resty-http library (https://github.com/pintsized/lua-resty-http/pull/112#discussion_r168259963).

APIcast might need to use `lua-resty-http` instead of native upstream+balancer when using HTTP proxy. Fortunately this could be abstracted away by the `http_ng` library.

mikz:
@3scale/product this is now finished, but without Basic authentication. That requirement is not in JIRA. We probably won't have time to implement that in this release anyway. If we need support for Basic authentication, please open another ticket.

albgus:
So this appears to make traffic to the API backed go trough the proxy as well, and that seems weird.

Is there a way to use proxy only for the Service Management API and Account Management API?

albgus:
There is nothing standard about having a reverse proxy, which directs everything through another proxy, to finally reach the intended API backend.

Look in the THREESCALE-221 ticket, and specifically the sidecar proxy workaround. The issue needing solving is when the gateway is deployed into a internal network, and needs to go trough a corporate proxy to reach the internet and talk to the APIs of the 3scale SaaS. I would imagine that everyone in this situation wants the API Gateway to service API Backends that is also located in the internal networks and doesn't need to go trough the proxy.

mikz:
@albgus by "configuring a proxy server" I meant configuring any software to use a proxy server. 

[...]

PR #800, opened 2018-07-04
[resty.http_ng] use http/s proxy for backend and upstream connections 

mikz:
Fixes https://github.com/3scale/apicast/issues/709

Builds on top of https://github.com/3scale/apicast/pull/835

By configuring standard environment variables `http_proxy`, `HTTP_PROXY`, `no_proxy`, `NO_PROXY`, `http_proxy`, `HTTPS_PROXY` you can instruct APIcast to use those instead of direct connection.

Proxy does not support any authentication yet.

# Tests

- [x] test http_ng.backend.ngx
- [x] test http_ng.backend.ngx + SSL
- [x] test http_ng.backend.resty
- [x] test http_ng.backend.resty + SSL
- [x] test nginx upstream `ngx.exec(@upstream)`
- [x] test nginx upstream `ngx.exec(@upstream)` + SSL
- [x] test upstream policy 
- [x] test upstream policy + SSL
- [x] test downloading configuration
- [x] test downloading configuration + SSL
- [x] test forwarding request body

### DIFF
--- a/gateway/src/apicast/policy/upstream/upstream.lua
+++ b/gateway/src/apicast/policy/upstream/upstream.lua
@@ -22,7 +22,7 @@ local function init_config(config)
     local upstream, err = Upstream.new(rule.url)
 
     if upstream then
-      tab_insert(res, { regex = rule.regex, upstream = upstream })
+      tab_insert(res, { regex = rule.regex, url = rule.url })
     else
       ngx.log(ngx.WARN, 'failed to initialize upstream from url: ', rule.url, ' err: ', err)
     end
@@ -49,7 +49,8 @@ function _M:rewrite(context)
   for _, rule in ipairs(self.rules) do
     if match(req_uri, rule.regex) then
       ngx.log(ngx.DEBUG, 'upstream policy uri: ', req_uri, ' regex: ', rule.regex, ' match: true')
-      context[self] = rule.upstream
+      -- better to allocate new object for each request as it is going to get mutated
+      context[self] = Upstream.new(rule.url)
       break
     else
       ngx.log(ngx.DEBUG, 'upstream policy uri: ', req_uri, ' regex: ', rule.regex, ' match: false')
@@ -61,7 +62,6 @@ function _M:content(context)
   local upstream = context[self]
 
   if upstream then
-    upstream:set_request_host()
     upstream:call(context)
   else
     return nil, 'no upstream'
--- a/gateway/src/apicast/upstream.lua
+++ b/gateway/src/apicast/upstream.lua
@@ -2,7 +2,7 @@
 -- Abstracts how to forward traffic to upstream server.
 --- @usage
 --- local upstream = Upstream.new('http://example.com')
---- upstream:set_request_host() -- set Host header to 'example.com'
+--- upstream:rewrite_request() -- set Host header to 'example.com'
 --- -- store itself in `context` table for later use in balancer phase and call `ngx.exec`.
 --- upstream:call(context)
 
@@ -13,6 +13,9 @@ local str_format = string.format
 local resty_resolver = require('resty.resolver')
 local resty_url = require('resty.url')
 local core_base = require('resty.core.base')
+local http_proxy = require('apicast.http_proxy')
+local str_find = string.find
+local str_sub = string.sub
 local new_tab = core_base.new_tab
 
 local _M = {
@@ -20,20 +23,24 @@ local _M = {
 }
 
 local function proxy_pass(upstream)
-    local uri = upstream.uri
-
-    return str_format('%s://%s%s%s%s',
-            uri.scheme,
-            upstream.upstream_name,
-            uri.path or ngx.var.uri or '',
-            ngx.var.is_args or '',
-            ngx.var.query_string or '')
+    return str_format('%s://%s', upstream.uri.scheme, upstream.upstream_name)
 end
 
 local mt = {
     __index = _M
 }
 
+local function split_path(path)
+    if not path then return end
+
+    local start = str_find(path, '?', 1, true)
+
+    if start then
+        return str_sub(path, 1, start - 1), str_sub(path, start + 1)
+    else
+        return path
+    end
+end
 
 local function parse_url(url)
     local parsed, err = resty_url.split(url)
@@ -47,7 +54,8 @@ local function parse_url(url)
     uri.password = parsed[3]
     uri.host = parsed[4]
     uri.port = tonumber(parsed[5])
-    uri.path = parsed[6]
+
+    uri.path, uri.query = split_path(parsed[6])
 
     return uri
 end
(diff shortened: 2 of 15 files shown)


gateway/src/proxy.lua
```diff
@@ -0,0 +1,204 @@
...
+    if match.method == 'CONNECT' then
+        local connect = ngx_re_match(match.uri, [[^(?<host>[^:]+):(?<port>\d+)$]])
+
+        proxy_connect_request(client, connect, upstream)
+    else
+        proxy_http_request(client, match.method, match.uri, match.http, upstream)
+    end
+end
+
+function _M.preread()
```
davidor: Is this needed to override the default behavior?

mikz: Nope. I'll remove this.

gateway/src/proxy.lua
```diff
@@ -0,0 +1,204 @@
...
+    local readline = client:receiveuntil(cr_lf)
+    local data, partial
+    data, err, partial = readline()
+
+    local match = tab_new(1,3)
+    match, err = re_match(data, [[^(?<method>[A-Z]+)\s(?<uri>\S+)\s(?<http>\S+)$]], 'oj', nil, match)
+
+    local upstream = ngx.socket.tcp()
+
+    if match.method == 'CONNECT' then
```
davidor: We should check whether there was an error first. Otherwise, this could fail.

davidor: Same in the `ngx_re_match` below.

gateway/src/resty/http_ng/backend/async_resty.lua
```diff
@@ -40,7 +43,15 @@ _M.async = function(request)
     end
   end
 
-  local ok, err = httpc:connect(host, port)
+
+  local proxy_uri = httpc:get_proxy_uri(scheme, host)
```
davidor: Async resty is not used

gateway/src/apicast/policy/upstream/upstream.lua
```diff
@@ -22,7 +22,7 @@ local function init_config(config)
     local upstream, err = Upstream.new(rule.url)
 
     if upstream then
-      tab_insert(res, { regex = rule.regex, upstream = upstream })
+      tab_insert(res, { regex = rule.regex, upstream = upstream, url = rule.url })
```
davidor: I think that because of the change below we no longer need to store `upstream` here.

gateway/src/apicast/policy/upstream/upstream.lua
```diff
@@ -49,7 +49,8 @@ function _M:rewrite(context)
   for _, rule in ipairs(self.rules) do
     if match(req_uri, rule.regex) then
       ngx.log(ngx.DEBUG, 'upstream policy uri: ', req_uri, ' regex: ', rule.regex, ' match: true')
-      context[self] = rule.upstream
+      -- better to allocate new object for each request as it is going to get mutated
+      context[self] = Upstream.new(rule.url)
```
davidor: In this case, it doesn't matter much because we're just storing one object, but it might be a better practice to namespace what we store in the context: `context[self].url = ... `

davidor: Another option would be to store just the url instead of the Upstream instance.
The Upstream can be instantiated in `content()`. Not a big difference, though.

gateway/src/apicast/policy/upstream/upstream.lua
```diff
@@ -61,7 +62,7 @@ function _M:content(context)
   local upstream = context[self]
 
   if upstream then
-    upstream:set_request_host()
+    upstream:rewrite_request()
```
davidor: This is a bit confusing because the function in the call below, `upstream:call()`, already calls `rewrite_request()` at least in some cases.

gateway/src/apicast/upstream.lua
```diff
@@ -13,27 +13,34 @@ local str_format = string.format
...
-            ngx.var.is_args or '',
-            ngx.var.query_string or '')
+    return str_format('%s://%s', upstream.uri.scheme, upstream.upstream_name)
 end
 
 local mt = {
     __index = _M
 }
 
+local function split_path(path)
```
davidor: Not something for now, but to take into account in future refactorings. I think we're already doing this in other modules, so it might be a good idea to extract methods like this one to a `Request` module that can be reused between policies.

gateway/src/apicast/upstream.lua
```diff
@@ -13,27 +13,34 @@ local str_format = string.format
...
+    if not path then return end
+
+    local start = str_find(path, '?', 1, true)
```
davidor: Can the find + subs be replaced with a single "split by ?" ?

spec/upstream_spec.lua
```diff
@@ -66,29 +66,44 @@ describe('Upstream', function()
...
+        stub(ngx.req, 'set_uri', function(uri)
+            ngx.var.uri = uri
+        end)
+        stub(ngx.req, 'set_uri_args', function(args)
+            ngx.var.args = args
+            ngx.var.query_string = args
+        end)
+    end
+
+    describe(':rewrite_request', function()
```
davidor: I think we should add tests for this part of the code too:
```lua
    if uri.query then
        ngx.req.set_uri_args(uri.query)
    end
```

gateway/src/apicast/policy/upstream/upstream.lua
mikz: We will have to try to reduce the allocations in the future, so I'm trying to not create objects when it is not absolutely necessary. In this case doing a namespace would require another table.

I'm parsing the upstream here, so potential failure will show early in the rewrite phase before other calls are made (like auth).

gateway/src/apicast/upstream.lua
mikz: Yep, this is probably the most common helper function. We should offer performance optimized version and reuse it.

gateway/src/apicast/upstream.lua
mikz: I did some benchmarking, but I don't have the results and I think this was fully JIT compatible and faster.

gateway/src/apicast/policy/upstream/upstream.lua
mikz: Should be fixed now.

gateway/src/apicast/policy/upstream/upstream.lua
mikz: Cleaned up.

spec/upstream_spec.lua
mikz: Done.

gateway/src/apicast/upstream.lua
davidor: Agreed.

PR #809, opened 2018-07-11
[wip][3scale] rename APIcast policy to 3scale

mikz:
APIcast policy is a really poor name. It is 3scale integration, so it should be named 3scale.

# TODO

- [ ] test migration path

/cc @3scale/system we need to figure out how to handle this from the UI


davidor:
Maybe not in this PR, but we should also consider renaming all the policies that only work when using 3scale. We should identify them with the '3scale' prefix. We already did that for some of them: "3scale_batcher", "3scale_referrer", but we should check with @3scale/system the implications that this might have for them.

eloycoto:
Close this one due inactivity.

Issue #744, opened 2018-06-05
Apply policies conditionally


mikz:
# Use Cases

Change Upstream API on matching HTTP Request Header. (https://github.com/3scale/apicast/pull/429)
Two endpoints of an API need different policies (for example rate limits).

# Concerns

If every policy is left to implement own matching then are going to be left with a lots of duplication.
Also policy would need to be changed to implement matching.

# Proposals

## Conditional policy

As outlined in https://github.com/3scale/apicast/issues/173#issuecomment-393808026. This policy would be itself a policy chain and execute the chain only when the precondition matches.

That would allow us to implement all condition matching only in this policy and just nest desired actions within it.

## Match phase

Match phase would be some shared code that the policies would get by default. 
The APIcast policy schema should define default properties for the policy schema to have extra configuration properties available.

Match phase would be executed once per policy to see if it applies to the request. 

## Conditions per "rule"

Some policies can have several "rules" like: add header "foo", add header "bar", rate limit /test to 100 rps, etc.
Each of those rules might need different condition. Offering some standard definition of conditions in the APIcast policy schema would allow policies to define conditions for rules easily. This would have to be opt-in feature made by policy developers. For examples of such conditions see https://github.com/3scale/ostia/issues/17#issuecomment-396913897.

# Comparison

We need to do deeper comparison of both approaches and possibly PoCs before committing to something.

# Concerns

Defining conditions in JSON schema can be quite hard especially because missing oneOf support.
It might be possible to use Liquid somehow, but not really see how. But that would offer the best flexibility at least in the beginning.

davidor:
@mikz I implemented a first version of the 'Conditional policy' option. Check this branch: https://github.com/3scale/apicast/tree/conditional-policy 

It's just a first version without much documentation and there are some things that can be improved, but it shows how it would work, and has tests showing that it behaves as expected.

davidor:
I implemented another prototype just to have another option to evaluate. It's very similar to the other one but it uses the sandbox lib (https://github.com/apitools/sandbox.lua) to run lua code instead of liquid: https://github.com/3scale/apicast/tree/conditional-policy-lua-sandbox

I think that there are 2 different topics for discussion. One is how to evaluate the condition specified in the policy (liquid, lua code, etc.) and the other is where to apply the matching (new conditional policy, match phase, etc.)

davidor:
I pushed a new branch that achieves the same using a different approach: https://github.com/3scale/apicast/tree/policy-conditions

[...]

PR #812, opened 2018-07-12
Conditional policy

davidor:
Part of #744 

This implements the conditional policy as discussed in #744 

The conditional engine is very simple and it's not how the final version will look at all. However, I decided to open this PR in case it helps unblock some of the work going on in [ostia](https://github.com/3scale/ostia). 

@mikz let me know if you'd like to have this first version merged soon. Otherwise, I'll start working on a more complex conditional engine (wip branch here: https://github.com/3scale/apicast/tree/lpeg-grammar-condition-engine)

### DIFF
--- a/gateway/src/apicast/policy/conditional/apicast-policy.json
+++ b/gateway/src/apicast/policy/conditional/apicast-policy.json
@@ -1,0 +1,26 @@
+{
+  "$schema": "http://apicast.io/policy-v1/schema#manifest#",
+  "name": "Conditional",
+  "summary": "Executes a policy chain conditionally",
+  "description": [
+    "Evaluates a condition, and when it's true, it calls its policy chain."
+  ],
+  "version": "builtin",
+  "configuration": {
+    "type": "object",
+    "properties": {
+      "condition": {
+        "description": "condition to be evaluated",
+        "type": "string"
+      },
+      "policy_chain": {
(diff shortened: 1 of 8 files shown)


gateway/src/apicast/policy/conditional/apicast-policy.json
```diff
@@ -0,0 +1,26 @@
...
+    "properties": {
+      "condition": {
+        "description": "condition to be evaluated",
+        "type": "string"
+      },
+      "policy_chain": {
+        "description": "the policy chain to execute when the condition is true",
+        "type": "array",
+        "items": {
+          "type": "object"
```
davidor: I know this is not exactly what we need.

t/apicast-policy-conditional.t
```diff
@@ -0,0 +1,105 @@
...
+In order to test this, we define a conditional policy that only runs the
+phase_logger policy when the request path is /log.
+We know that the policy outputs "running phase: some_phase" for each of the
+phases it runs, so we can use that to verify it was executed.
+--- configuration
+{
+  "services": [
+    {
+      "id": 42,
+      "backend_version":  1,
```
mikz: We should not need those when APIcast (3scale) policy is not active, right?

davidor: Right :+1: 
Fixed.

PR #814, opened 2018-07-18
JSON conditions for conditional policy

### LINKED ISSUE #744: Apply policies conditionally

davidor:
Closes #744 

This PR defines the conditions supported in the conditional policy. It also includes an engine that evaluates them.

The kinds of conditions supported are very basic (`==`, `!=`, `and`, `or`) and they can be expressed in JSON.

The alternative was to define a custom DSL. I implemented a prototype in this branch: https://github.com/3scale/apicast/tree/lpeg-grammar-condition-engine . This approach would allow us to express more complex conditions. However, the implementation is more complicated, and it would tie us to lpeg. The solution implemented in this PR should be enough for our current needs so it looks like a better option.

For now this only support `==` and `!=`. We can add more operations like `starts_with` or `match_exactly`, later. I'd rather agree on the approach and the basics first.


gateway/src/apicast/policy/conditional/condition.lua
```diff
@@ -0,0 +1,54 @@
...
+
+  assert(supported_combine_ops[self.combine_op], 'Unsupported combine op')
+
+  return self
+end
+
+function _M:evaluate(context)
+  if #self.operations == 0 then return true end
+
+  if self.combine_op == 'and' then
```
mikz: This can be done at initialization time, right? 

It would allow for better JIT if there would be no branching in this method.

gateway/src/apicast/policy/conditional/operation.lua
```diff
@@ -0,0 +1,53 @@
...
+  self.templated_right = create_template(right, right_type)
+
+  return self
+end
+
+function _M:evaluate(context)
+  local left_value = self.templated_left:render(context)
+  local right_value = self.templated_right:render(context)
+
+  if self.op == '==' then
```
mikz: I think we can have a dispatch table for operations like:


```lua

local operations = {
  ['=='] = function(left, right) return left == right end
}
```

And assign the comparator function during initialization for better JIT.

spec/policy/conditional/operation_spec.lua
```diff
@@ -0,0 +1,45 @@
...
+    it('evaluates values as plain text by default', function()
+      assert.is_true(Operation.new('1', nil, '==', '1', nil):evaluate({}))
+      assert.is_false(Operation.new('1', nil, '==', '2', nil):evaluate({}))
+    end)
+
+    it('evaluates liquid when indicated in the types', function()
+      local context = { var_1 = '1', var_2 = '2' }
+
+      local res_true = Operation.new(
+        '{{ var_1 }}', 'liquid', '==', '1', 'plain'
```
mikz: Are we going to handle types somehow? What if the variable comes from nginx and is actually a number ?

gateway/src/apicast/policy/conditional/condition.lua
davidor: Right :+1: 
Fixed.

gateway/src/apicast/policy/conditional/operation.lua
davidor: Fixed the same way as the one above :+1:

spec/policy/conditional/operation_spec.lua
davidor: In that example, even if `var_1` was a number, the comparison would end up being between 2 strings, because `LiquidTemplateString:render()` returns a string.

I included a test that checks this. Also, I added `tostring()` in the `==` and `!=` operations in `Operation` to make this explicit.

PR #815, opened 2018-07-20
[dev] do not set timer_resolution in development mode

mikz:
Setting `timer_resolution` in nginx configuration will result in logging ~7 lines every 100ms (or other timer_resolution interval) in debug log. That makes any debugging impossible.

Disabling this in development is reasonable as it is just a performance  optimization for production environment.

### DIFF
--- a/gateway/src/apicast/linked_list.lua
+++ b/gateway/src/apicast/linked_list.lua
@@ -10,7 +10,7 @@ local noop = function() end
 
 local empty_t = setmetatable({}, { __newindex = noop })
 local __index = function(t,k)
-    return t.current[k] or t.next[k]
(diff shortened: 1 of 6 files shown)


gateway/src/apicast/linked_list.lua
```diff
@@ -10,7 +10,7 @@ local noop = function() end
 
 local empty_t = setmetatable({}, { __newindex = noop })
 local __index = function(t,k)
-    return t.current[k] or t.next[k]
+    if t.current[k] == nil then return t.next[k] else return t.current[k] end
```
davidor: Did you find any bugs caused by this?

mikz: Yep. I could not override the `timer_resolution` with `false`. So I bundled it into one PR.

PR #816, opened 2018-07-20
[wip] upstream module

mikz:
TBD

abstration that handles where to proxy traffic

### DIFF
--- a/gateway/src/apicast/upstream.lua
+++ b/gateway/src/apicast/upstream.lua
@@ -1,0 +1,114 @@
+local setmetatable = setmetatable
+local tonumber = tonumber
+local str_format = string.format
+
+local resty_resolver = require('resty.resolver')
+local resty_url = require('resty.url')
+local core_base = require('resty.core.base')
+local new_tab = core_base.new_tab
+
+local _M = {
+
+}
+
+local function proxy_pass(upstream)
+    local uri = upstream.uri
+
+    return str_format('%s://%s%s%s%s',
+            uri.scheme,
+            upstream.upstream_name,
+            uri.path or ngx.var.uri,
+            ngx.var.is_args,
+            ngx.var.query_string or '')
+end
+
+local mt = {
+    __index = _M
+}
+
+
+local function parse_url(url)
+    local parsed, err = resty_url.split(url)
+
+    if err then return nil, err end
+
+    local uri = new_tab(0, 6)
+
+    uri.scheme = parsed[1]
+    uri.user = parsed[2]
+    uri.password = parsed[3]
+    uri.host = parsed[4]
+    uri.port = tonumber(parsed[5])
+    uri.path = parsed[6]
+
+    return uri
+end
+
+function _M.new(url)
+    local uri, err = parse_url(url)
+
+    if err then
+        return nil, 'invalid upstream'
+    end
+
+    return setmetatable({
+        uri = uri,
+        location_name = '@upstream',
+        upstream_name = 'upstream',
+        resolver = resty_resolver,
+    }, mt)
+end
+
+function _M:resolve(servers)
+    local resolver = self.resolver
+    local uri = self.uri
+
+    if servers then
+        self.servers = servers
+        return servers
+    end
+
+    if not resolver or not uri then return nil, 'not initialized' end
+
+    local res, err = resolver:instance():get_servers(uri.host, uri)
+
+    if err then
+        return nil, err
+    end
+
+    self.servers = res
+
+    return res
+end
+
+function _M:port()
+    if not self or not self.uri then
+        return nil, 'not initialized'
+    end
+
+    return self.uri.port or resty_url.default_port(self.uri.scheme)
+end
+
+function _M:rewrite(host)
+    ngx.req.set_header('Host', host or self.uri.host)
+end
+
+local function exec(self)
+    ngx.var.proxy_pass = proxy_pass(self)
+
+    if self.location_name then
+        ngx.exec(self.location_name)
+    end
+end
+
+function _M:call(context)
+    if ngx.headers_sent then return nil, 'response sent already' end
+
+    if not self.servers then self:resolve() end
+
+    context[self.upstream_name] = self
+
+    return exec(self)
+end
+
+return _M
--- a/spec/luassert_helper.lua
+++ b/spec/luassert_helper.lua
@@ -41,6 +41,7 @@ end
 do -- adding assert.returns_error(error_text, ok, ret) : assert.returns_error('not initialized', _M:call())
   local tostring = tostring
   local tonumber = tonumber
+  local s = require('say')
 
   local function returns_error(state, arguments, level)
     local argc = arguments.n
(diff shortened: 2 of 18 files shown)


t/apicast-policy-upstream.t
```diff
@@ -311,13 +311,13 @@ Upstream policy should work if used standalone without apicast policy.
...
      content_by_lua_block {
-       require('luassert').are.equal('GET /a_path? HTTP/1.1',
+       require('luassert').are.equal('GET /a_path?query HTTP/1.1',
                                      ngx.var.request)
        ngx.say('yay, api backend');
      }
   }
 --- request
-GET /a_path?
+GET /a_path?query
```
mikz: We will not preserve the `?` when there is no query string.

spec/luassert_helper.lua
```diff
@@ -58,4 +59,72 @@ do -- adding assert.returns_error(error_text, ok, ret) : assert.returns_error('n
...
+
+        if not same then
+          crumbs = crumbs or { }
+          table.insert(crumbs,  k)
+          return false, crumbs
+        end
+      end
+
+      return true
+    elseif left == right then
```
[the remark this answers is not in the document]
mikz: Fixed.

gateway/src/apicast/upstream.lua
```diff
@@ -0,0 +1,135 @@
...
+--- Rewrite request Host header to what is provided in the argument or in the URL.
+--- @tparam string host Host header of the request
+function _M:rewrite(host)
+    ngx.req.set_header('Host', host or self.uri.host)
+end
+
+local function exec(self)
+    ngx.var.proxy_pass = proxy_pass(self)
+
+    if self.location_name then
```
davidor: I don't think this `if` is needed as `self.location_name` is always set.

mikz: It can be unset. And it is in the http client. Because that one does not do .exec (because it does its own).

spec/upstream_spec.lua
```diff
@@ -0,0 +1,152 @@
...
+        it('sets request Host header to the host passed as argument', function()
+            stub(ngx.req, 'set_header')
+
+            assert(Upstream.new('http://example.com/')):rewrite('localhost')
+
+            assert.spy(ngx.req.set_header).was_called_with('Host', 'localhost')
+        end)
+    end)
+
+    describe(':rewrite', function()
```
davidor: `:call`

gateway/src/apicast/upstream.lua
```diff
@@ -0,0 +1,135 @@
...
+function _M.new(url)
+    local uri, err = parse_url(url)
+
+    if err then
+        return nil, 'invalid upstream'
+    end
+
+    return setmetatable({
+        uri = uri,
+        location_name = '@upstream',
```
davidor: I think we should add a comment here saying that this is defined in `apicast.conf`.

spec/upstream_spec.lua
```diff
@@ -0,0 +1,152 @@
...
+        it('uses .resolver', function()
+            local upstream = Upstream.new(valid_url)
+            local resolver, instance, servers = {}, {}, {}
+            stub(resolver, 'instance').returns(instance)
+            stub(instance, 'get_servers').returns(servers)
+
+            upstream.resolver = resolver
+
+            assert.equal(servers, upstream:resolve())
+            assert.spy(instance.get_servers).was_called_with(instance, 'localhost', match.contains({ port = 8080 }))
```
davidor: I'd define `host` and `port` next to `valid_url` so it's clear where they come from.

spec/upstream_spec.lua
```diff
@@ -0,0 +1,152 @@
...
+    describe(':resolve', function()
+        it('uses .resolver', function()
+            local upstream = Upstream.new(valid_url)
+            local resolver, instance, servers = {}, {}, {}
+            stub(resolver, 'instance').returns(instance)
+            stub(instance, 'get_servers').returns(servers)
+
+            upstream.resolver = resolver
+
+            assert.equal(servers, upstream:resolve())
```
davidor: I'm not sure it's a good idea to test with servers = {}
If `upstream:resolve()` fails somehow and returns an empty table we won't notice.

spec/upstream_spec.lua
mikz: I'll define the url in tests that actually require this info.

spec/upstream_spec.lua
mikz: `.equal` is by reference, so it needs to be exactly the same table

but I'll add there a value to make that clearer

spec/upstream_spec.lua
```diff
@@ -0,0 +1,152 @@
...
+            local upstream = Upstream.new(valid_url)
+            local servers = { 'resolved servers instance' }
+            upstream.servers = servers
+
+            assert.equal(upstream.servers, upstream:resolve())
+            assert.equal(upstream.servers, servers)
+        end)
+    end)
+
+    describe(':port', function()
```
davidor: Should we define what this returns when a port is not defined and it cannot be inferred from the scheme?

spec/policy/upstream/upstream_spec.lua
```diff
@@ -64,90 +66,58 @@ describe('Upstream policy', function()
...
-        assert.stub(ngx.exec).was_called_with('@upstream')
-      end)
+        local ctx = { [policy] = upstream }
+        policy:content(ctx)
 
-      it('marks in the context that the upstream has changed', function()
-        local upstream = UpstreamPolicy.new({})
-        upstream:content(context)
-        assert.is_truthy(context.upstream_changed)
+        assert.spy(upstream.rewrite).was_called_with(upstream)
```
davidor: TIL I can use `assert.spy(...)` even when it was declared as a stub.

spec/upstream_spec.lua
mikz: Yep.

spec/policy/upstream/upstream_spec.lua
mikz: The difference should be that spy calls the original but stub does not.

gateway/src/apicast/upstream.lua
mikz: Yeo. The same for the upstream_name.

spec/upstream_spec.lua
mikz: 👁

davidor: We should test the case where `ngx.headers_sent` is true.

gateway/src/apicast/upstream.lua
```diff
@@ -0,0 +1,135 @@
...
+    if not self or not self.uri then
+        return nil, 'not initialized'
+    end
+
+    return self.uri.port or resty_url.default_port(self.uri.scheme)
+end
+
+--- Rewrite request Host header to what is provided in the argument or in the URL.
+--- @tparam string host Host header of the request
+function _M:rewrite(host)
```
davidor: This is not a policy so there's no need to call this method like a policy phase.

gateway/src/apicast/upstream.lua
```diff
@@ -0,0 +1,135 @@
...
+local function exec(self)
+    ngx.var.proxy_pass = proxy_pass(self)
+
+    if self.location_name then
+        ngx.exec(self.location_name)
+    end
+end
+
+--- Execute the upstream.
+function _M:call(context)
```
davidor: I'm a bit concerned about sharing the policy context with modules that are not policies.
I'll come back to this after I review everything to evaluate some other options.

gateway/src/apicast/upstream.lua
davidor: I see. Is it OK to set `ngx.var.proxy_pass` when `ngx.exec` is not going to be called?

spec/policy/upstream/upstream_spec.lua
```diff
@@ -64,90 +66,58 @@ describe('Upstream policy', function()
...
     describe('when the upstream has not been changed in previous phases', function()
       it('does not call the balancer', function()
-        stub(balancer, 'call')
-        local upstream = UpstreamPolicy.new()
+        stub(apicast_balancer, 'call')
+        local policy = UpstreamPolicy.new()
 
         for _, val in ipairs({ false, nil }) do
-          upstream:balancer({ upstream_changed = val })
+          policy:balancer({ upstream_changed = val })
```
davidor: `upstream_changed` is no longer used, so this should be changed.

spec/policy/upstream/upstream_spec.lua
```diff
@@ -64,90 +66,58 @@ describe('Upstream policy', function()
...
     describe('when the upstream has been changed in previous phases', function()
       it('calls the balancer', function()
-        stub(balancer, 'call')
-        local upstream = UpstreamPolicy.new()
-
-        upstream:balancer({ upstream_changed = true })
+        local policy = UpstreamPolicy.new()
 
-        assert.stub(balancer.call).was_called()
+        assert.equal(apicast_balancer.call, policy.balancer)
```
davidor: This tests that the functions are the same, but not that it was called as the `describe` says.

spec/upstream_spec.lua
mikz: Yep.

gateway/src/apicast/upstream.lua
mikz: `set_request_host` would be ok?

gateway/src/apicast/upstream.lua
mikz: I think other objects can work with context too. In this case it needs to set itself to the context before it calls exec. I think it would make for every weird API and create a lot of duplicity if we would want to control this outside this module.

gateway/src/apicast/upstream.lua
davidor: :+1:

gateway/src/apicast/upstream.lua
davidor: I saw that this is not called only with a policies context, it's also called with `ngx.ctx`.
So context can be any table. Can we document this so we don't forget?

PR #817, opened 2018-07-23
Set LOG_LEVEL param to required

PhilipGough:
When following the [openshift-guide](https://github.com/3scale/apicast/blob/master/doc/openshift-guide.md), running the `oc new-app` command without specifying a log level causes the apicast pods to crash due to invalid config.


PhilipGough:
@mikz crashes with the following error `bad --errlog-level option value:  at /usr/bin/resty line 315.`

I can set that in the template if needed. Was not sure if we would like a user to set the value explicitly. What should the default be?

PR #820, opened 2018-07-24
Add docs for the conditional policy

davidor:


davidor:
@jmprusi not for now. The only available operations are "==" and "!=". However, we could add operations like "matches" or "starts_with" that work with regexes, if there's a need for it. It should be pretty straightforward.

unleashed:
If we had `nand` (or `nor`, but harder) for combining conditions we could have a full algebra even though multiple evaluation of the same expressions could be an issue.

Extra operations would also be interesting, ie. `>=`.

I foresee we'll need more things, combining stuff with nands without associativity other than left-to-right or top-down can be hard, but this might be enough for now.

unleashed:
How about adding the possibility to _choose_ a policy as the one that will be passed responsibility to say yay/nay:
```json
"condition": {
  "external": {
    "name": "a_policy_that_talks_to_a_decision_engine",
    "parameters": "all"
  }
}
```

davidor:
I opened a new issue to keep track of the operations that we need to support according to what has been discussed here. Feel free to add more: #821

davidor:
@unleashed That doesn't really fit the policy model we currently have. 

Each policy needs to implement a function for each of the nginx phases it wants to run on. However, what you need is a method that evaluates a condition and returns a boolean.

What you could do is implement your own conditional policy. It would be the same as this one but you could choose to describe and evaluate the conditions as you wish.

jmprusi:
yes at some moment, having a policy that can use for ex. Open Policy Agent would be really cool :)

PR #823, opened 2018-07-25
Allow to place the batcher policy after APIcast policy in the chain

davidor:
The batcher policy was not working correctly when placed after the APIcast policy in the chain.
This PR fixes the issue.


davidor:
Rebased on top of master and solved conflicts in the changelog.

PR #824, opened 2018-07-26
CHANGELOG: add missing links to JIRA

davidor:


mikz:
That should give @3scale/qe info what is done.

@3scale/product this is what I meant when I mentioned it would be good to review the changelog and add links to JIRA. Also it would be to review now and see what does not have links to JIRA and is a new feature to see if we want to support it or no.

PR #827, opened 2018-07-30
"Rewrite URL captures" policy

davidor:
Closes https://github.com/3scale/apicast/issues/711 (first use case)

After the discussion in #711, I think that implementing the first use case of that issue with a new policy seems reasonable. We need something for the next release, although in the future, we might be able to do something better as we add more features to the conditional policy.

Regarding the name of the policy, I'm open to suggestions. I can't think of a good name for it :thinking:


t/apicast-policy-rewrite-url-captures.t
```diff
@@ -0,0 +1,187 @@
...
+      "backend_version":  1,
+      "backend_authentication_type": "service_token",
+      "backend_authentication_value": "token-value",
+      "proxy": {
+        "api_backend": "http://test:$TEST_NGINX_SERVER_PORT/",
+        "proxy_rules": [
+          { "pattern": "/", "http_method": "GET", "metric_system_name": "hits", "delta": 2 }
+        ],
+        "policy_chain": [
+          { "name": "apicast.policy.apicast" },
```
davidor: I'll change these tests so they use the Echo policy instead. I think that will simplify the code.

I'll keep just one with the APIcast policy just to make sure that there are not any problems when combining it with the new policy.

mikz: I'd love that 👍

gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
```diff
@@ -0,0 +1,115 @@
...
+  local matches = re_match(path, self.regex_rule, 'oj')
+
+  if not matches or #self.named_args ~= #matches then
+    return false
+  end
+
+  local replaced_template = replace_in_template(
+    self.named_args, matches, self.template)
+
+  local uri, raw_params = unpack(re_split(replaced_template, '\\?'))
```
mikz: `unpack` can't be JITed so this should not be in the hot path.

mikz: And instead of `'\\?'` you can use `[[\?]]`, but that might be just my preference.

gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  return res
+end
+
+local function replace_in_template(args, vals, template)
+  local res = template
+
+  for i, arg in ipairs(args) do
+    res = re_gsub(res, "{" .. arg .. "}", vals[i], 'oj')
```
mikz: This is allocating  3 strings in a loop. Better to use `string.format` and even better to do this during policy initialization.

gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
```diff
@@ -0,0 +1,115 @@
...
+
+-- Transforms a string representing the args of a query like:
+-- "a=1&b=2&c=3" into a table: { a = "1", b = "2", c = "3" }.
+local function string_params_to_table(string_params)
+  local res = {}
+
+  local params_split = re_split(string_params, '&')
+
+  for _, param in ipairs(params_split) do
+    local name, val = unpack(re_split(param, '='))
```
mikz: Another unpack in a loop. This will prevent JIT.

gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
```diff
@@ -0,0 +1,115 @@
...
+    'oj'
+  )
+end
+
+-- Transforms a string representing the args of a query like:
+-- "a=1&b=2&c=3" into a table: { a = "1", b = "2", c = "3" }.
+local function string_params_to_table(string_params)
+  local res = {}
+
+  local params_split = re_split(string_params, '&')
```
mikz: Maybe good to extract `re_split` into own function `split` that will always pass `oj` flags.

t/apicast-policy-rewrite-url-captures.t
davidor: I just realized that we can't do that because the Echo policy uses `ngx.say(ngx.var.request)`, so it just works with the original request.

mikz: So the upstream policy pointing to the echo endpoint? That is still simpler than APIcast + dependencies.

gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
davidor: Right, that way we won't forget to add `oj`. Fixed :+1:

gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
davidor: Fixed :+1:

gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
davidor: Fixed the `unpack`.
I don't have a strong preference on `\\?` vs `[[\?]]` so I changed it.

gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
davidor: Right. It turns out that we always use the result of this wrapped between "{" and "}" so I concatenate them in `extract_named_args` which is only called from `.new()`.

gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
```diff
@@ -0,0 +1,133 @@
...
+
+-- Transforms a string representing the args of a query like:
+-- "a=1&b=2&c=3" into a table: { a = "1", b = "2", c = "3" }.
+local function string_params_to_table(string_params)
+  local res = {}
+
+  local params_split = split(string_params, '&')
+
+  for _, param in ipairs(params_split) do
+    local parts = split(param, '=') -- avoid unpack, not jitted.
```
mikz: You might want to limit the number of parts you want by setting the max results parameter.

gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
```diff
@@ -0,0 +1,133 @@
...
+    local name = parts[1]
+    local val = parts[2]
+    res[name] = val
```
mikz: [LuaJIT performance guide](http://wiki.luajit.org/Numerical-Computing-Performance-Guide) says it is better not to create short lived temporary variables like these and just use the tuple:

```lua
res[parts[1]] = parts[2]
```

gateway/src/apicast/policy/rewrite_url_captures/rewrite_url_captures.lua
```diff
@@ -0,0 +1,74 @@
...
+end
+
+-- When a param in 'new_params' exist in the request, this function replaces
+-- its value. When it does not exist, it simply adds it.
+-- This function does not delete or modify the params in the query that do not
+-- appear in 'new_params'.
+local function set_query_params(new_params)
+  local query_params = QueryParams.new()
+
+  for arg, value in pairs(new_params) do
```
mikz: `pairs` is not JITed. Does it need to be key valued table? Integer indexed could be JITed.

gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
```diff
@@ -0,0 +1,133 @@
...
+
+  for i, arg in ipairs(args) do
+    res = re_gsub(res, arg, vals[i], 'oj')
+  end
+
+  return res
+end
+
+local function uri_and_params_from_template(template)
+  local parts = split(template, [[\?]]) -- avoid unpack, not jitted.
```
mikz: Here it might need to limit the results too.

gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
davidor: I think that defining those vars helps readability. I doubt this will have a noticeable impact.

davidor:
@mikz Regarding the name of the policy, do you have any suggestions?

t/apicast-policy-rewrite-url-captures.t
davidor: Using the echo policy requires more work because we are only interested in checking the path and the query args, but the echo returns more data.

In the end, I opted for deleting the APIcast policy from the chain and adding an upstream policy that redirects all the request to the upstream defining in the config. I think this is the simplest solution.

I kept one of the tests I had with the APIcast policy just to make sure that both policies can be combined without any problems.

gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
davidor: Done :+1:

davidor:
I addressed all your comments @mikz

gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
```diff
@@ -0,0 +1,140 @@
...
+local _M = {}
+
+local mt = { __index = _M }
+
+local function split(string, separator, max_matches)
+  return re_split(string, separator, 'oj', nil, max_matches)
+end
+
+-- Returns a list of named args extracted from a match_rule.
+-- For example, for the rule /{abc}/{def}?{ghi}=1, it returns this list:
```
mikz: So, do you think it makes sense to use just `{}` instead of `{{ }}` to match with the liquid syntax?

gateway/src/apicast/policy/rewrite_url_captures/rewrite_url_captures.lua
```diff
@@ -0,0 +1,73 @@
...
+
+local policy = require('apicast.policy')
+local _M = policy.new('Capture args policy')
+
+local new = _M.new
+
+function _M.new(config)
+  local self = new(config)
+
+  self.matchers = {}
```
mikz: Just for the future. Using `tab_new(#config.transformations)` would save table expansion because it would preallocate slots in the table for each matcher.

gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
davidor: I think `{}` is better in this case because it matches the same format we are using for the mapping rules.

mikz: Ok, thats fair point. And for the replacement ? Shouldn't we just pass it through liquid?

davidor: Not sure. That would only replace the code we have in `replace_in_template` which is not problematic, in my opinion.

mikz: Yep. I'm thinking about consistency. If push people everywhere to use liquid, then it would make sense to stick to it. I have no hard feelings one way or the other, just want to avoid confused people that can't use what they do everywhere else.

But it will not be hard to migrate. Just adding the `template_type`. So if you have no feelings one way or the other then lets merge this 👍

davidor: I'd rather leave this one as it is, at least for now :+1:

PR #835, opened 2018-08-06
[resty.http] http proxy supporrt for lua-resty-http

mikz:
Support http proxy when downloading configuration from system.

Following ENV variables can be set: `all_proxy`, `ALL_PROXY`, `http_proxy`, `HTTP_PROXY`, `https_proxy`, `HTTPS_PROXY`, `no_proxy`, `NO_PROXY`.


t/fixtures/proxy.lua
```diff
@@ -0,0 +1,222 @@
...
+
+local cr_lf = "\r\n"
+
+local http_ok = 'HTTP/1.1 200 OK' .. cr_lf
+
+
+ngx.log(ngx.INFO, 'proxy starting')
+
+local function send(socket, data)
+--    if ngx.ctx.upstream == socket then
```
davidor: Should we delete this?

t/fixtures/proxy.lua
```diff
@@ -0,0 +1,222 @@
...
+    while true do
+        data, err, partial = input:receive(4 * 1028) -- read any data
+
+        if data or partial then
+            send(output, data or partial)
+            ngx.log(ngx.DEBUG, 'sending data for proxy request: ', ngx.ctx.request)
+        end
+
+        if err == 'closed' then
+            -- ngx.log(ngx.DEBUG, "failed to read any data: ", err, ' socket: ', name)
```
davidor: :eyes:

t/fixtures/proxy.lua
```diff
@@ -0,0 +1,222 @@
...
+
+        if not match then
+            ngx.log(ngx.WARN, 'got invalid request: ', data)
+            return client:close()
+        end
+
+        local upstream = ngx.socket.tcp()
+
+        ngx.log(ngx.DEBUG, 'proxy request: ', data)
+        ngx.ctx.request = data
```
davidor: Why do we need to store this in the context?

t/fixtures/proxy.lua
```diff
@@ -0,0 +1,222 @@
...
+        data, err, partial = readline()
+
+        if data then
+            ngx.log(ngx.DEBUG, 'got line: ', data)
+        else
+            ngx.log(ngx.WARN, 'got partial: ', partial, ' err: ', err)
+            ngx.exit(ngx.OK)
+        end
+
+        local match = ngx_re_match(data, [[^(?<method>[A-Z]+)\s(?<uri>\S+)\s(?<http>\S+)$]], 'oj')
```
davidor: How about extracting the regex to a var?

t/fixtures/proxy.lua
```diff
@@ -0,0 +1,222 @@
...
+            return client:close()
+        end
+
+        local upstream = ngx.socket.tcp()
+
+        ngx.log(ngx.DEBUG, 'proxy request: ', data)
+        ngx.ctx.request = data
+
+        if match.method == 'CONNECT' then
+            local connect = ngx_re_match(match.uri, [[^(?<host>[^:]+):(?<port>\d+)$]])
```
davidor: forgot `oj` ? 
Maybe we should extract `ngx_re_match` like we did in other PRs to avoid forgetting to set the options?

t/fixtures/proxy.lua
```diff
@@ -0,0 +1,222 @@
...
+
+local function proxy_http_request(sock, method, url, http_version, upstream)
+    local uri = assert(http.parse_uri(nil, url))
+    local port = uri[3]
+    local host = uri[2]
+    local ok, err = upstream:connect(host, port)
+
+    if ok then
+        ngx.log(ngx.DEBUG, 'connected to upstream ', host, ':', port)
+        ngx.ctx.upstream = upstream
```
davidor: I'd document why we need to store this info in the ctx.

t/fixtures/proxy.lua
```diff
@@ -0,0 +1,222 @@
...
+        ngx.log(ngx.DEBUG, 'got header line: ', header_line)
+    until header_line == cr_lf
+
+    send(sock, http_ok)
+    send(sock, cr_lf)
+
+    return pipe_stream(sock, upstream)
+end
+
+return function()
```
davidor: Understanding what this function does is not complicated. However, as it's the only method exposed by the module, I'd document it. I think it will help communicate the purpose of this module.

t/fixtures/proxy.lua
```diff
@@ -0,0 +1,222 @@
...
+        end
+
+        local match = ngx_re_match(data, [[^(?<method>[A-Z]+)\s(?<uri>\S+)\s(?<http>\S+)$]], 'oj')
+
+        if not match then
+            ngx.log(ngx.WARN, 'got invalid request: ', data)
+            return client:close()
+        end
+
+        local upstream = ngx.socket.tcp()
```
davidor: Where is this closed?

t/fixtures/proxy.lua
mikz: It is testing/debugging code. I rather keep it because it is easy to uncomment when debugging.

t/fixtures/proxy.lua
mikz: Will describe it in the comment, but basically so we can print `CONNECT host:port` later when reusing keepalive connection so the test can match that.

t/fixtures/proxy.lua
mikz: Why? It is not reused. I prefer having it inline because it is easy to see what the regex is supposed to match.

t/fixtures/proxy.lua
mikz: Fixe the `oj`.

davidor: :+1: Forgot for a moment that this is in `fixtures`

t/fixtures/proxy.lua
davidor: Not a big deal. Just personal preference.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Conditional policy. This policy includes a condition and a policy chain, and only executes the chain when the condition is true [PR #812](https://github.com/3scale/apicast/pull/812), [PR #814](https://github.com/3scale/apicast/pull/814), [PR #820](https://github.com/3scale/apicast/pull/820)
 - Request headers are now exposed in the context available when evaluating Liquid [PR #819](https://github.com/3scale/apicast/pull/819)
 - Rewrite URL captures policy. This policy captures arguments in a URL and rewrites the URL using them [PR #827](https://github.com/3scale/apicast/pull/827), [THREESCALE-1139](https://issues.jboss.org/browse/THREESCALE-1139)
+- Support for HTTP Proxy [PR #835](https://github.com/3scale/apicast/pull/835)
```
davidor: Link to Jira issue: https://issues.jboss.org/browse/THREESCALE-221

mikz: Done.

PR #839, opened 2018-08-10
Add conditions to rate-limit policy

davidor:
Closes the last item of the TODO list in https://github.com/3scale/apicast/issues/760

### DIFF
--- a/gateway/src/apicast/policy/rate_limit/apicast-policy.json
+++ b/gateway/src/apicast/policy/rate_limit/apicast-policy.json
@@ -54,6 +54,72 @@
           "enum": ["log"],
           "description": "Let the request go through and only output logs"
         }]
+      },
+      "operation": {
+        "type": "object",
+        "$id": "#/definitions/operation",
+        "properties": {
+          "left": {
+            "type": "string"
+          },
+          "op": {
+            "type": "string",
+            "enum": ["==", "!="]
+          },
+          "right": {
+            "type": "string"
+          },
+          "left_type": {
+            "description": "How to evaluate 'left'",
+            "type": "string",
+            "default": "plain",
+            "oneOf": [
+              {
+                "enum": ["plain"],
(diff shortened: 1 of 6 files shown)


gateway/src/apicast/policy/rate_limit/apicast-policy.json
```diff
@@ -54,6 +54,72 @@
           "enum": ["log"],
           "description": "Let the request go through and only output logs"
         }]
+      },
+      "operation": {
+        "type": "object",
+        "$id": "#/definitions/operation",
```
mikz: Can we define those in the policy manifest spec? And then just refer to those?

davidor: You mean `manifest-schema.json` ?

mikz: Yes. It could make the validation harder, but sharing those definitions would be good in my opinion.

davidor: Yes, it would avoid duplication, but I'm not sure if the tooling we have for validation will work correctly. I'll try it.

t/apicast-policy-rate-limit.t
```diff
@@ -1227,3 +1227,75 @@ and rejected properly.
...
+            }
+          }
+        }
+      }
+    })
+  }
+  lua_shared_dict limiter 1m;
+--- config
+  include $TEST_NGINX_APICAST_CONFIG;
+  resolver $TEST_NGINX_RESOLVER;
```
mikz: No need for resolver.

davidor: Right, copy-pasted that before I saw your new PR :D

davidor: Fixed :+1:

PR #843, opened 2018-08-14
[count-inc] Add decrement logic to the incoming function

### LINKED ISSUE #830: [rate-limit] The cached count value differs according to the order of limiters.
This depends on the `lua-resty-limit-traffic`.

https://github.com/openresty/lua-resty-limit-traffic/blob/ef080731b80fe00a385a4aed8a7f7f575d01c152/lib/resty/limit/traffic.lua#L26-L29

For example, we have 2 limiters (Limiter1 and Limiter2).
If the Limiter1 is rejected, it is not committed.
If the Limiter2 is rejected, it is committed.

We should call the `uncommit` in the above code.
Maybe the `connections` limiters are affected by this.

How to fix this is, first to submit a patch to the `lua-resty-limit-traffic`, and if no response, then to fork and to use the forked version, right?

y-tabata:
Closes #830.

### DIFF
--- a/gateway/src/resty/limit/count-inc.lua
+++ b/gateway/src/resty/limit/count-inc.lua
@@ -33,6 +33,17 @@ function _M.incoming(self, key, commit)
         if not count then
             return nil, err
         end
+
+        if count > limit then
+            count, err = dict:incr(key, -1)
+
+            if not count then
+                return nil, err
+            end
+
+            return nil, "rejected"
+        end
+
     else
         count = (dict:get(key) or 0) + 1
     end
--- a/spec/policy/rate_limit/rate_limit_spec.lua
+++ b/spec/policy/rate_limit/rate_limit_spec.lua
@@ -194,7 +194,7 @@ describe('Rate limit policy', function()
           assert(rate_limit_policy:access(context))
           assert.returns_error('limits exceeded', rate_limit_policy:access(context))
 
-          assert.equal('2', redis:get('11110_fixed_window_test3'))
+          assert.equal('1', redis:get('11110_fixed_window_test3'))
           assert.spy(ngx.exit).was_called_with(429)
         end)
 
@@ -211,7 +211,7 @@ describe('Rate limit policy', function()
           assert(rate_limit_policy:access(ctx))
           assert.returns_error('limits exceeded', rate_limit_policy:access(ctx))
 
-          assert.equal('2', redis:get('11110_fixed_window_test3'))
+          assert.equal('1', redis:get('11110_fixed_window_test3'))
           assert.spy(ngx.exit).was_called_with(429)
         end)
 
@@ -232,7 +232,7 @@ describe('Rate limit policy', function()
           assert(rate_limit_policy:access(context))
           assert.returns_error('limits exceeded', rate_limit_policy:access(context))
 
-          assert.equal('2', redis:get('11110_fixed_window_' .. test_host))
+          assert.equal('1', redis:get('11110_fixed_window_' .. test_host))
           assert.spy(ngx.exit).was_called_with(429)
         end)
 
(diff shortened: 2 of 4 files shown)


spec/configuration_loader/remote_v1_spec.lua
```diff
@@ -1,29 +1,68 @@
-local loader = require 'apicast.configuration_loader.remote_v1'
+local _M = require 'apicast.configuration_loader.remote_v1'
```
davidor: This commit is not related to the rest of the PR, so it should be in a different one.

spec/policy/rate_limit/rate_limit_spec.lua
```diff
@@ -194,7 +194,7 @@ describe('Rate limit policy', function()
           assert(rate_limit_policy:access(context))
           assert.returns_error('limits exceeded', rate_limit_policy:access(context))
 
-          assert.equal('2', redis:get('11110_fixed_window_test3'))
+          assert.equal('1', redis:get('11110_fixed_window_test3'))
```
davidor: I don't fully understand the purpose of this change.
Before, we increased the number of hits even when going over limits. Now we force the value to be at most equal to the limit. Why do we need to do this?

y-tabata: > Before, we increased the number of hits even when going over limits.

First, we should not increase the number of hits when going over limits, because we should count the number the client calls the API backend.

The previous behavior is the below:

Case 1:
```json
fixed_window_limiters : [
  { "key" : { "name" : "a"}, "count" : 1, "window" : 10 },
  { "key" : { "name" : "b"}, "count" : 2, "window" : 10 }
]
```

After 2 API requests, 
the number of hits 'a' is '1',
the number of hits 'b' is '1'.
-> correct


Case 2:
```json
fixed_window_limiters : [
  { "key" : { "name" : "a"}, "count" : 2, "window" : 10 },
  { "key" : { "name" : "b"}, "count" : 1, "window" : 10 }
]
```

After 2 API requests,
the number of hits 'a' is '1',
the number of hits 'b' is '2'.
-> wrong

spec/configuration_loader/remote_v1_spec.lua
y-tabata: Changed.

gateway/src/resty/limit/count-inc.lua
```diff
@@ -33,6 +33,17 @@ function _M.incoming(self, key, commit)
         if not count then
             return nil, err
         end
+
+        if count > limit then
+            count, err = dict:incr(key, -1)
```
mikz: I believe this should be done by the `.uncommit` that is being called by `resty.limit.traffic`.

y-tabata: @mikz you mean `.uncommit` needs to be called here with the condition of `i == n`, right?
https://github.com/openresty/lua-resty-limit-traffic/blob/master/lib/resty/limit/traffic.lua#L27-L29

However `resty.limit.conn` includes this.

So we should choose which file we edit, `resty.limit.traffic` and `resty.limit.conn`, or `resty.limit.count-inc`.

mikz: @y-tabata I think we don't need to change any resty.limit library. We have to use it correctly.

`resty.limit.traffic` calls `:incoming(key, false)` until the last limiter which is called as `:incoming(key, true)`. If any of those return an error, it is fine to just exit. Nothing was committed to the database.

When all of them succeed, it tries to call again `:incoming(key, true)`, but in this case, if any of them fails then call `:uncommit(key)` for all limiters that it already called `:incoming`.

So I'd suspect our limiters are not rolled back because they are not wrapped in one `resty.limit.traffic`.

y-tabata: > resty.limit.traffic calls :incoming(key, false) until the last limiter which is called as :incoming(key, true). If any of those return an error, it is fine to just exit. Nothing was committed to the database.

Using the current `count-inc`, I think that when the last limiter, which is called as :incoming(key, true), only returns nil and "rejected", the last one is committed to the database.

mikz: Ok. I got it now. Implemented a fix to the upstream library: https://github.com/openresty/lua-resty-limit-traffic/pull/34/commits/c53f2240e953a9656580dbafcc142363ab063100

I think we can just copy that here.

PR #845, opened 2018-08-14
Fix bug when downloading configs that contain Liquid

### LINKED ISSUE #828: Liquid tags are not working properly and can cause failure of the Apicast
When using liquid tags in the policies for example: `url_rewrite` or `header_policy` apicast is not able to resolve this liquid values.

Here are the logs:
- Apicast: http://pastebin.test.redhat.com/625140
- with debug: http://pastebin.test.redhat.com/625141

##### Version

Built from the master branch. 

##### Steps To Reproduce

1. Create new service with headers policy
2. Set value for a header for example: `Service_id {{ service.id }}`
3. Save and get an error: 500 Internal server error
4. Each other request will cause the `failed to acquire lock to lazy load`

Here is a sandbox config: http://pastebin.test.redhat.com/625156 (Service with id 2)

##### Additional Information

It was staging gateway with lazy loading.
Configuration: 2.2-GA with apicast built from the master (30-07-2018)

**Source Commit:** Merge pull request #820 from 3scale/conditional-policy-docs ab474ec authored by David Ortiz

The issue has been discussed with @davidor on Slack.

davidor:
We were parsing the service just to get the OIDC issuer endpoint. That's unnecessary because we can get easily get it from the decoded JSON.

Also, parsing the service was causing a bug. An error was raised with services containing a policy with fields to be interpreted by Liquid (have '{{' and '}}'). (Closes #828 )


gateway/src/apicast/configuration_loader/remote_v2.lua
```diff
@@ -317,8 +317,10 @@ function _M:config(service, environment, version)
   if res.status == 200 then
     local proxy_config = cjson.decode(res.body).proxy_config
 
-    local config_service = configuration.parse_service(proxy_config.content)
-    local issuer, oidc_config = self:oidc_issuer_configuration(config_service)
+    local oidc_issuer_endpoint = proxy_config.content.proxy and
```
davidor: Note: when calling `parse_service`, the `configuration` module builds the policy chain, and when doing so, in the case of the headers policy, the configuration received in `new()` is modified because it adds a `template_string` field with an instance of `TemplateString`.

That's why this failed only when the service was parsed. Parsing the service is not a problem by itself, but later, when the proxy configs are encoded, it fails because of the template string.

An unexpected side-effect.

pestanko:
@davidor is there any estimate when this will be merged? (I suppose next week?)

gateway/src/apicast/policy/headers/headers.lua
```diff
@@ -5,6 +5,8 @@
 -- Similarly, this policy also allows to add, modify, and delete the headers
 -- included in the response.
 
+local tablex = require('pl.tablex')
```
mikz: I really don't want to depend on penlight in policies. OpenResty provides table.clone that does shallow clone. 

Also, if we are doing this, then it should be pushed from the top, not done in each policy.

gateway/src/apicast/configuration_loader/remote_v2.lua
mikz: We will need an integration test to reproduce this failure.

And I wonder why it crashes when we add an instance of TemplateString.

davidor: I'll add an integration test.

Regarding the `TemplateString` exception, here's a minimal case to reproduce it:
```lua
local cjson = require 'cjson'
local Liquid = require 'liquid'
local LiquidTemplate = Liquid.Template

local template = LiquidTemplate:parse('{{ service.id }}')

cjson.encode(template) -- Error: Cannot serialise table: table key must be a number or string
```

mikz: I see, the configuration loader tries to encode back the table into string:

https://github.com/3scale/apicast/blob/635590f8dedd895cbba4288faaecc13854967781/gateway/src/apicast/configuration_loader/remote_v2.lua#L128

Then we could do the deepclone just before parsing the service:
https://github.com/3scale/apicast/blob/635590f8dedd895cbba4288faaecc13854967781/gateway/src/apicast/configuration_loader/remote_v2.lua#L119

davidor: That would work too :+1:

gateway/src/apicast/policy/headers/headers.lua
davidor: OK. I suggest we solve the problem in the loader first and leave the immutable config for a future PR.

gateway/src/apicast/configuration_loader/remote_v2.lua
davidor: At first I thought that there was no point in parsing the service, because the information can be easily get from the JSON. But maybe parsing the service is better in case we're performing some validations there.

gateway/src/apicast/policy/headers/headers.lua
mikz: Agreed 👍

PR #847, opened 2018-08-20
Fix bug when downloading configs that contain Liquid (portal endpoint with path)

### LINKED ISSUE #828: Liquid tags are not working properly and can cause failure of the Apicast

davidor:
Closes #828 

This should have been part of #845 . It's the same bug but triggered when `THREESCALE_PORTAL_ENDPOINT` has a path.


davidor:
Added a changelog entry and deleted a `---ONLY` that I left in one of the integration tests.

PR #849, opened 2018-08-22
Liquid context debugging policy

### LINKED ISSUE #838: expose objects available for liquid expressions and policies
It will be great to be able to see values stored in objects available for policy authors and used in liquid extensions. I mean these objects: 
 - line 5-8 https://github.com/3scale/apicast/blob/master/gateway/src/apicast/policy/ngx_variable.lua
 - plus JWT object
 - any other available in future

I think it will be great to have this in echo policy so user add echo policy into policy chain and user can create liquid expressions or write new policy based on values stored in exposed object.

Also I think it will be great to group all available objects to one root object (i.e. window or document object in Javascript in browsers).

davidor:
Closes #838

### DIFF
--- a/gateway/src/apicast/policy/liquid_context_debug/context_content.lua
+++ b/gateway/src/apicast/policy/liquid_context_debug/context_content.lua
@@ -1,0 +1,97 @@
+local type = type
+local pairs = pairs
+local tostring = tostring
+
+local _M = {}
+
+-- The context usually contains a lot of information. For example, it includes
+-- the whole service configuration. Also, some of the values are objects that
+-- we can't really use when evaluating liquid, like functions.
+-- That's why we define only a few types of keys and values to return.
+local accepted_types_for_keys = {
+  string = true,
+  number = true,
+}
+
+local accepted_types_for_values = {
+  string = true,
+  number = true,
+  table = true
+}
+
+-- In the context, there might be large integers as keys. This can be a problem
+-- when converting to JSON. Imagine an entry of the table like this:
+-- { [10000] = 'something' }. To convert that to JSON, we need an array of 1000
+-- positions with only one occupied. With arrays like these, 'cjson' raises an
+-- "Excessively sparse arrays" error:
+-- https://github.com/efelix/lua-cjson/blob/4f27182acabc435fcc220fdb710ddfaf4e648c86/README#L140
+-- For example, there is a lrucache instance with a table that maps service ids
+-- (numbers) to hosts. So we can have something like { [123456] = "127.0.0.1" }.
+-- This can be limited using a global setting of the 'cjson' library. However,
+-- we're going to implement an add-hoc solution here so we don't affect the
+-- other modules. We're going to convert those large ints to strings as a
+-- workaround.
+local max_integer_key = 1000
+
+local function key(k)
+  if type(k) ~= 'number' then return k end
+
+  if k <= max_integer_key then
+    return k
+  else
+    return tostring(k)
+  end
+end
+
+local value_from
+
+local ident = function(...) return ... end
+local value_from_fun = {
+  string = ident, number = ident,
+  table = function(table)
+    local res = {}
+    for k, v in pairs(table) do
+      local wanted_types = accepted_types_for_keys[type(k)] and
+                           accepted_types_for_values[type(v)]
+
+      if wanted_types then
+        res[key(k)] = value_from(v)
+      end
+    end
+
+    return res
+  end
+}
+
+value_from = function(object)
+  local fun = value_from_fun[type(object)]
+  if fun then return fun(object) end
+end
+
+local function add_content(object, acc)
+  if type(object) ~= 'table' then return nil end
+
+  -- The context is a list where each element has a "current" and a "next".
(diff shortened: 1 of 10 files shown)


gateway/src/apicast/policy/liquid_context_debug/context_content.lua
```diff
@@ -0,0 +1,68 @@
...
+}
+
+local function value_from(object)
+  if type(object) == 'string' or type(object) == 'number' then
+    return object
+  end
+
+  if type(object) ~= 'table' then return nil end
+
+  local res = {}
```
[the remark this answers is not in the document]
davidor: It looks a bit better :+1:

gateway/src/apicast/policy/liquid_context_debug/liquid_context_debug.lua
```diff
@@ -0,0 +1,30 @@
...
+  -- There is a lrucache instance with a table that maps service ids (numbers)
+  -- to hosts. So we can have something like { [123456] = "127.0.0.1" }. This
+  -- is a problem when encoding the table into JSON because it's a table with
+  -- many positions but only one != null. When converting this, cjson raises
+  -- an "Excessively sparse arrays" error:
+  -- https://github.com/efelix/lua-cjson/blob/4f27182acabc435fcc220fdb710ddfaf4e648c86/README#L140
+  -- With this call we just convert that number into a string. Not a big deal
+  -- since nobody is going to use those fields from the lrucache instance in
+  -- liquid.
+  cjson.encode_sparse_array(true)
```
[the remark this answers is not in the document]
davidor: Right, I didn't notice. It wasn't my intention to set this global setting there.

I'll just probably implement some ad-hoc solution in `ContextContent`. If the key is an integer and is greater than X then convert it to string.

gateway/src/apicast/policy/liquid_context_debug/liquid_context_debug.lua
```diff
@@ -0,0 +1,30 @@
...
+
+local new = _M.new
+
+function _M.new(config)
+  local self = new(config)
+  return self
+end
+
+function _M.content(_, context)
+  local content = context_content.from(context)
```
mikz: We also should add the ngx_variable, right?

davidor: You're right. And to do that I needed to modify `ngx_variable` so it behaves as a context (linked list): https://github.com/3scale/apicast/pull/849/commits/822345df1bb601fd5bf00c1e799af2f045e92cb0

gateway/src/apicast/policy/liquid_context_debug/liquid_context_debug.lua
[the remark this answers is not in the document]
davidor: Here's how I implemented this: https://github.com/3scale/apicast/pull/849/files#diff-d126ed9fb83c6261c4bb9fefa80fd029R22

gateway/src/apicast/policy/liquid_context_debug/liquid_context_debug.lua
davidor: Also, the policy takes that into account now: https://github.com/3scale/apicast/pull/849/commits/98d3d8a946656f12e1d0a14b7bbfa72352fba58c

gateway/src/apicast/policy/liquid_context_debug/context_content.lua
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  return res
+end
+
+local function add_content(object, acc)
+  if type(object) ~= 'table' then return nil end
+
+  -- The context is a list where each element has a "current" and a "next".
+  local current = object.current
```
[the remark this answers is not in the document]
davidor: That's a good idea, but I'd rather leave it for a future PR as it involves adding the functionality in the `LinkedList` module, test it, and check whether we can use it in some of the existing callers.

gateway/src/apicast/policy/liquid_context_debug/context_content.lua
```diff
@@ -0,0 +1,97 @@
...
+-- positions with only one occupied. With arrays like these, 'cjson' raises an
+-- "Excessively sparse arrays" error:
+-- https://github.com/efelix/lua-cjson/blob/4f27182acabc435fcc220fdb710ddfaf4e648c86/README#L140
+-- For example, there is a lrucache instance with a table that maps service ids
+-- (numbers) to hosts. So we can have something like { [123456] = "127.0.0.1" }.
+-- This can be limited using a global setting of the 'cjson' library. However,
+-- we're going to implement an add-hoc solution here so we don't affect the
+-- other modules. We're going to convert those large ints to strings as a
+-- workaround.
+local max_integer_key = 10
```
mikz: I think arrays with 10 elements are common. Maybe better to set this to a something like 1000 ?

davidor: OK. I'll raise the limit to 1k. I think that in practice large service IDs are going to be the problem.

PR #850, opened 2018-08-23
[OIDC] support JWK discovery through OIDC Discovery process

mikz:
Load all JWK serialized keys offered by the OIDC Discovery endpoint.
Only RSA is supported at the moment.

This is internal change - how we load signing keys from the IDP. Previously we used Keycloak/RHSSO specific method: loading "public_key" from the issuer endpoint. That is not compatible with OIDC spec, so this PR introduces loading JWK keys from the "jwks_uri" offered by the OIDC Discovery.

It uses OpenSSL via FFI, so there can be some potential memory access issues leading to segfaults/memory leaks. However, I run out unit test suite 10000 times just to verify it does not crash and the memory usage does not increase.

This also unifies how we load OIDC configuration and improves its internal format. OIDC configuration is loaded even when using JSON configuration file (fixes #
https://github.com/3scale/apicast/issues/654), but only when not already provided.
JSON configurations with "oidc" configuration are no longer valid, as the internal structure changed. 

# TODO
- [x] stress test the JWK conversion for memory leaks / segfaults
- [x]  unit test the OIDC discovery process
- [x] cleanup the OIDC configuration object 
- [x] unify loading the OIDC configuration


mikz:
@3scale/product this implements JWK originally planned for 3.4 release. Would be good to know to which JIRA to link it in the Changelog.

spec/resty/oidc/discovery_spec.lua
```diff
@@ -0,0 +1,103 @@
...
+    before_each(function() discovery = _M.new(test_backend) end)
+
+    describe('.new', function()
+        it('has .http_client', function()
+            assert(_M.new().http_client,'has .http_client')
+        end)
+    end)
+
+    describe(':openid_configuration(endpoint)', function()
+        it('loads configuration from the discovery endpoint', function()
```
davidor: We should also test the error cases

spec/resty/oidc/discovery_spec.lua
```diff
@@ -0,0 +1,103 @@
...
+lVMCGXQ81OB73muc9ovR68Ajzn8KzpU/qegh8iHwk+SQvJxIIvgNJCJTC6BWnwS9
+Bw2ns0fQOZZRjWFRVh8BjkVdqa4vCAb6zw8hpR1y9uSNG+fqUAPHy5IYQaD8k8QX
+0obxJ0fld61fH+Wr3ENpn9YZWYBcKvnwLm2bvxqmNVBzW4rhGEZb9mf+KrSagD5G
+UwIDAQAB
+-----END PUBLIC KEY-----
+]],
+            } }, keys)
+        end)
+
+        it('can process ForgeRock response', function()
```
davidor: Is Forgerock different somehow or just an example?

mikz: It has some keys with ciphers we don't support and possibly some properties can be different. 

I wanted to keep it "ForgeRock" so it is obvious we support ForgeRock and it was copied from ForgeRock. 

Over time we can add tests for edge cases and describe them like "does not crash on unknown ciphers". But for now I wanted to say this is ForgeRock and it works.

gateway/src/resty/oidc/discovery.lua
```diff
@@ -0,0 +1,103 @@
...
+
+    if not config then
+        ngx.log(oidc_log_level, 'invalid OIDC Provider, expected application/json got:  ', res.headers.content_type, ' body: ', res.body)
+        return nil, 'invalid JSON'
+    end
+
+    return config
+end
+
+function _M:jwks(configuration)
```
davidor: Maybe add a comment to make it clear that apart from extracting keys this method converts them to the PEM format?

spec/configuration_loader/oidc_spec.lua
```diff
@@ -0,0 +1,58 @@
...
+local loader = require 'apicast.configuration_loader.oidc'
+local cjson = require('cjson')
+
+describe('OIDC Configuration loader', function()
+  describe('.call', function()
+    local test_backend
+    before_each(function() test_backend = test_backend_client.new() end)
+    before_each(function() loader.discovery.http_client.backend = test_backend end)
+
+    it('ignores config', function()
```
davidor: ignores empty config?

mikz: Yep.

mikz: Fixed.

PR #856, opened 2018-08-27
Logging policy

davidor:
Allows enabling/disabling access logs per service, so addresses #712 partially.

### DIFF
--- a/gateway/conf/nginx.conf.liquid
+++ b/gateway/conf/nginx.conf.liquid
@@ -120,7 +120,8 @@ http {
 
   server {
 
-    access_log {{ access_log_file | default: "/dev/stdout" }} time;
+    set $access_logs_enabled '1';
+    access_log {{ access_log_file | default: "/dev/stdout" }} time if=$access_logs_enabled;
 
     {%- assign http_port = port.apicast | default: 8080 %}
     {%- assign https_port = env.APICAST_HTTPS_PORT %}
--- a/gateway/src/apicast/policy/logging/logging.lua
+++ b/gateway/src/apicast/policy/logging/logging.lua
@@ -1,0 +1,40 @@
+--- Logging policy
+
+local _M  = require('apicast.policy').new('Logging Policy')
+
+local new = _M.new
+
+local default_enable_access_logs = true
+
+-- Defined in ngx.conf.liquid and used in the 'access_logs' directive.
+local ngx_var_access_logs_enabled = 'access_logs_enabled'
+
+-- Returns the value for the ngx var above from a boolean that indicates
+-- whether access logs are enabled or not.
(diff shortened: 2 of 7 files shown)


gateway/src/apicast/policy/logging/logging.lua
```diff
@@ -0,0 +1,37 @@
...
+
+local _M  = require('apicast.policy').new('Logging Policy')
+
+local new = _M.new
+
+local default_enable_access_logs = true
+
+-- Returns the value for ngx.var.enable_acess_logs
+local function enable_access_log_value(enabled)
+  return (enabled and '1') or '0'
```
mikz: Better to use dispatch table for this. Conditionals are not JITable.

gateway/conf/nginx.conf.liquid
```diff
@@ -120,7 +120,8 @@ http {
 
   server {
 
-    access_log {{ access_log_file | default: "/dev/stdout" }} time;
+    set $enable_access_logs '1';
```
mikz: I'd rather name it `access_logs_enabled` so we can later have `access_logs_format` etc.

gateway/src/apicast/policy/logging/logging.lua
davidor: In this case, it doesn't matter much, because it's only called from `new()`.
But I changed it because as a general rule we should do this, and I tend to forget about this :sweat_smile:

gateway/conf/nginx.conf.liquid
davidor: Sounds good :+1:

gateway/conf/nginx.conf.liquid
```diff
@@ -120,7 +120,8 @@ http {
 
   server {
 
-    access_log {{ access_log_file | default: "/dev/stdout" }} time;
+    set $access_logs_enabled '1';
```
mikz: Are you sure about the quotes?

davidor: I think it's correct. From the docs:
`The if parameter (1.7.0) enables conditional logging. A request will not be logged if the condition evaluates to “0” or an empty string.`

mikz: Yes. But what I meant is that `set` value should not be quoted. Now the value is `'1'`, not `1`.
It is a bit confusing (even though it will still work).

mikz: Ok. There is some magic going on. 

Both these are the same value (string):

```nginx
    set $access_log_enabled '1';
    set $access_log_enabled_num 1;
```

I thought we had a PR to strip `''` from some of the values because it was generating quotes. My bad I guess.

PR #857, opened 2018-08-28
[cli] fix relative access/error log paths [THREESCALE-1090]

mikz:
Expand the access/error log paths to absolue paths from the current directory.

Fixes https://issues.jboss.org/browse/THREESCALE-1090


mikz:
I think this PR broke passing `syslog:`  as a log name. Will make a patch with a test to verify syslog works.

PR #858, opened 2018-08-28
[cli] fix logging to syslog

mikz:
Part of https://github.com/3scale/apicast/issues/712
Fixes issue introduced by https://github.com/3scale/apicast/pull/857

Log path starting with `syslog:` or `memory:` is not relative path, but
has to be passed to nginx unchanged.

For how to configure syslog see nginx documentation: http://nginx.org/en/docs/syslog.html

### DIFF
--- a/t/apicast-syslog.t
+++ b/t/apicast-syslog.t
@@ -1,0 +1,57 @@
+use lib 't';
+use Test::APIcast::Blackbox 'no_plan';
+
+repeat_each(3);
+
+run_tests();
+
+__DATA__
+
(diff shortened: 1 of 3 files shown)


t/apicast-syslog.t
```diff
@@ -0,0 +1,57 @@
+use lib 't';
+use Test::APIcast::Blackbox 'no_plan';
+
+repeat_each(3);
```
davidor: Do we need 3 for these?

mikz: 2 probably would do, but I feel better with 3

PR #863, opened 2018-08-31
release: 3.3.0-beta1

davidor:


davidor:
Travis fails because the changelog links to the `3.3.0-beta1` which will be created after the merge.

PR #864, opened 2018-09-03
[oidc] JWT validation according to the RFC 7523

mikz:
https://tools.ietf.org/html/rfc7523#section-3

So for example, `nbf` is not required. Neither is `iat`. On the other hand `sub` claim is required now.
Discovered in https://issues.jboss.org/browse/THREESCALE-583?focusedCommentId=13627302&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13627302

### DIFF
--- a/gateway/src/apicast/oauth/oidc.lua
+++ b/gateway/src/apicast/oauth/oidc.lua
@@ -41,11 +41,33 @@ function _M.new(service)
     keys = oidc.keys or empty,
     clock = ngx_now,
     alg_whitelist = util.to_hash(config.id_token_signing_alg_values_supported),
+    -- https://tools.ietf.org/html/rfc7523#section-3
     jwt_claims = {
-      nbf = jwt_validators.is_not_before(),
-      exp = jwt_validators.is_not_expired(),
-      aud = jwt_validators.required(),
+      -- 1. The JWT MUST contain an "iss" (issuer) claim that contains a
+      -- unique identifier for the entity that issued the JWT.
       iss = jwt_validators.equals_any_of({ issuer }),
+
+      -- 2. The JWT MUST contain a "sub" (subject) claim identifying the
+      -- principal that is the subject of the JWT.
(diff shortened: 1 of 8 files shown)


gateway/src/apicast/oauth/oidc.lua
```diff
@@ -41,11 +41,33 @@ function _M.new(service)
     keys = oidc.keys or empty,
     clock = ngx_now,
     alg_whitelist = util.to_hash(config.id_token_signing_alg_values_supported),
+    -- https://tools.ietf.org/html/rfc7523#section-3
```
davidor: We don't need to specify anything about 7?
```
   7.   The JWT MAY contain a "jti" (JWT ID) claim that provides a
        unique identifier for the token.  The authorization server MAY
        ensure that JWTs are not replayed by maintaining the set of used
        "jti" values for the length of time for which the JWT would be
        considered valid based on the applicable "exp" instant.
```

mikz: We don't do anything about it, so I don't think so.

spec/oauth/oidc_spec.lua
```diff
@@ -83,21 +83,41 @@ describe('OIDC', function()
...
+          nbf = ngx.now() + 1,
           exp = ngx.now() + 10,
         },
       })
 
       local credentials, _, _, err = oidc:transform_credentials({ access_token = access_token })
 
-      assert.falsy(credentials)
-      assert.truthy(err)
+      assert.returns_error([['nbf' claim not valid until Thu, 11 Aug 2005 07:13:12 GMT]], credentials, err)
```
davidor: If possible, would be better to extract that `1123744391` below into its own var and transform it here. So the relationship between this date and `ngx.now` is more obvious.

mikz: Fixed.

PR #866, opened 2018-09-03
release: 3.3.0-beta2

davidor:


davidor:
Travis fails because the changelog links to the `v3.3.0-beta2` which will be created after the merge.

PR #874, opened 2018-09-05
[t] test empty APICAST_LOG_LEVEL

mikz:
Regression test for https://github.com/3scale/apicast/pull/868

### DIFF
--- a/t/apicast-bin.t
+++ b/t/apicast-bin.t
@@ -1,0 +1,17 @@
+use lib 't';
+use Test::APIcast::Blackbox 'no_plan';
+
+run_tests();
+
+__DATA__
+
+=== TEST 1: Empty APICAST_LOG_LEVEL does not crash APIcast.
+--- env eval
+(
+  'APICAST_LOG_LEVEL' => '',
+)
+
+--- configuration fixture=echo.json
+--- request
+GET /
+--- error_code: 200


t/apicast-bin.t
```diff
@@ -0,0 +1,21 @@
+
+use lib 't';
+use Test::APIcast::Blackbox 'no_plan';
+
+# Can't run twice because one of the test checks the contents of the cache, and
```
davidor: :eyes:

t/apicast-bin.t
```diff
@@ -0,0 +1,21 @@
...
+# those change between runs (cache miss in first run, cache hit in second).
+
+run_tests();
+
+__DATA__
+
+=== TEST 1: Empty APICAST_LOG_LEVEL is ignored.
```
davidor: Should we say that the default applies rather than it is ignored?

t/apicast-bin.t
mikz: :-D  Copy 🍝

t/apicast-bin.t
mikz: Then we should test the default and that's harder than just verifying it works.

mikz: I'll check it to "does not crash".

davidor: :+1:

PR #886, opened 2018-09-06
Clarify order or policies that run on the content phase

### LINKED ISSUE #883: apicast debug policy specification
I've used 
```
User-Agent: APIcast/3.3.0-beta2 (Linux; x64; env:sandbox) APIcast/3.3.0-beta2
```
and I've realized after @davidor advise that *debug policy* should be in policy chain before *apicast policy*. 
It will be great to have this info about *debug policy* placement in policy chain in *debug policy* specification(https://github.com/3scale/apicast/blob/master/gateway/src/apicast/policy/liquid_context_debug/apicast-policy.json) , so user can read it in UI.

Also if there is any other policy like *upstream policy* which should be in policy chain before *apicast policy* it should be written also in its description.

davidor:
Closes #883

### DIFF
--- a/doc/policies.md
+++ b/doc/policies.md
@@ -71,6 +71,18 @@ mapping rules of APIcast will be applied against the rewritten path.
 However, if the URL policy appears after the APIcast one, the mapping rules
 will be applied against the original path.
 
+Regarding policies that run on the `content` phase, it is important to bear in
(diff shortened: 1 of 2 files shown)


davidor:
@mikz I addressed your comments.

PR #888, opened 2018-09-07
Expose HTTP method to liquid

PhilipGough:
As part of https://github.com/3scale/ostia/issues/17, this feature will expose the HTTP method to liquid templating. 

Feature is requested because it allows us to apply rate limit policies based conditionals which examine the request method.


gateway/src/apicast/policy/ngx_variable.lua
```diff
@@ -7,7 +7,8 @@ local function context_values()
     uri = ngx.var.uri,
     host = ngx.var.host,
     remote_addr = ngx.var.remote_addr,
-    headers = ngx.req.get_headers()
+    headers = ngx.req.get_headers(),
+    http_method = ngx.req.get_method()
```
mikz: Would be nice to keep `,` at the end, so the next change does not have to change the previous line.

PR #895, opened 2018-09-13
Fix bug when using the liquid debug policy with a config that contains Liquid templates

### LINKED ISSUE #882: if there is liquid expression after debbug policy in policy chain it returns 500
I've configured policy chain like this:
```
{
 "policies_config": [
  {
   "name": "liquid_context_debug",
   "version": "builtin",
   "configuration": {},
   "enabled": true
  },
  {
   "name": "apicast",
   "version": "builtin",
   "configuration": {},
   "enabled": true
  },
  {
   "name": "headers",
   "version": "builtin",
   "configuration": {
    "response": [
     {
      "op": "push",
      "header": "X-HEADER",
      "value_type": "liquid",
      "value": "{{ jwt.iss }}"
     }
    ],
    "request": [
     {
      "op": "push",
      "header": "X-HEADER",
      "value_type": "liquid",
      "value": "{{ jwt.iss }}"
     }
    ]
   },
   "enabled": true
  }
 ]
}
```

I see in apicast log:
```
2018/09/05 14:45:04 [debug] 22#22: *14791 [lua] policy_chain.lua:155: policy chain execute phase: content, policy: Liquid context debug, i: 1
2018/09/05 14:45:04 [error] 22#22: *14791 lua entry thread aborted: runtime error: stack overflow
stack traceback:
coroutine 0:
	[C]: in function 'pairs'
	.../apicast/policy/liquid_context_debug/context_content.lua:53: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
	.../apicast/policy/liquid_context_debug/context_content.lua:58: in function 'value_from'
... while sending to client, client: 10.130.0.1, server: _, request: "GET 
```

We've found with @davidor that if there is liquid expression in *Modify header policy* which is after *Debug policy* it returns return code == 500:
```
<center><h1>500 Internal Server Error</h1></center>
<hr><center>openresty/1.13.6.2</center>
```
If there is plain text expression in *Modify header policy*, there is response with all the debug info and return code == 200.

I've used 
```
User-Agent: APIcast/3.3.0-beta2 (Linux; x64; env:sandbox) APIcast/3.3.0-beta2
```

davidor:
Closes #882

### DIFF
--- a/gateway/src/apicast/policy/liquid_context_debug/context_content.lua
+++ b/gateway/src/apicast/policy/liquid_context_debug/context_content.lua
@@ -33,12 +33,9 @@ local accepted_types_for_values = {
 -- workaround.
 local max_integer_key = 1000
 
--- Liquid.Template instances of the 'liquid-lua' library that we use store the
--- context when they render a template. Specifically,
--- template_instance.interpreter.interpretercontext contains the context.
--- In order to avoid entering in a infinite loop when parsing the context,
--- we just discard keys with this name.
-local liquid_template_context_key = 'interpretercontext'
+local avoid_circular_refs_msg = {
+  already_seen = 'ALREADY SEEN - NOT CALCULATING AGAIN TO AVOID CIRCULAR REFS'
+}
 
 local function key(k)
   if type(k) ~= 'number' then return k end
@@ -55,14 +52,17 @@ local value_from
 local ident = function(...) return ... end
 local value_from_fun = {
   string = ident, number = ident,
-  table = function(table)
+  table = function(table, already_seen)
+    if already_seen[table] then return avoid_circular_refs_msg end
+    already_seen[table] = true
+
     local res = {}
     for k, v in pairs(table) do
       local wanted_types = accepted_types_for_keys[type(k)] and
                            accepted_types_for_values[type(v)]
 
-      if wanted_types and k ~= liquid_template_context_key then
-        res[key(k)] = value_from(v)
+      if wanted_types then
+        res[key(k)] = value_from(v, already_seen)
       end
     end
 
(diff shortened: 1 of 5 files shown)


gateway/src/apicast/policy/liquid_context_debug/context_content.lua
```diff
@@ -54,7 +61,7 @@ local value_from_fun = {
       local wanted_types = accepted_types_for_keys[type(k)] and
                            accepted_types_for_values[type(v)]
 
-      if wanted_types then
+      if wanted_types and k ~= liquid_template_context_key then
```
mikz: What about hiding the whole TemplateString? 

Also, I think we will need some protection from recursive structures in general. It should be as easy as a table that gets passed to record all elements we've seen already.

davidor: Hiding TemplateString instances is not so straightforward because every policy could give them a different name.

mikz: We can do that by checking the metatable of the object.

davidor: Regarding what you mentioned about avoid recursive structures in general, I agree. However, that alone would not solve the specific case of the liquid templates. It's a bit tricky. The context stored in the template string instance could be different from the context that we have in the liquid context debug policy. The context in the template string instance is stored when we call render() on the template. The context at that moment could be different from any other context received in other policies, including the context debug one, because any policy can modify the context in any of the phases.

PR #899, opened 2018-09-14
release: 3.3.0-cr1

davidor:


davidor:
Travis fails because the changelog links to the `v3.3.0-cr1` which will be created after merging.

PR #900, opened 2018-09-14
Merge 3.3-stable into master 

davidor:
To keep master up-to-date with the latest 3.3 changes.


davidor:
Travis fails because the changelog links to the v3.3.0-cr1 which will be created after merging.

Issue #745, opened 2018-06-05
Basic prometheus metrics


jmprusi:
APIcast ships with prometheus support, but only exposes the `nginx_metric_errors_total` metric.

I would like to propose some basic metrics to be added to the APIcast base policy:

### Counters

* Request:
  * Total
  * Request 2xx
  * Request 4xx
  * Request 5xx

* Connections:
  * Read
  * Write
  * Wait
  * Open
  
* Nginx Error Log
* Free Dictionary Space 

* Threescale (fetching config):
  * Update
  * Reload

### Histogram

* Request latency

Some of them where already added to apicast-cloud-hosted: https://github.com/3scale/apicast-cloud-hosted/pull/5/files#diff-047b1780b0ffeb4eba7b3d05beb76d5e

What do you think? Any other metrics to add?

andrewdavidmackenzie:
Reading between the lines, it sounds like there's a slight difference in *where* the metrics are implemented.

Joaquim lists a list of metrics, and suggests doing in apicast policy.

Michal lists a set that are related to apicast policy operations and would make sense to do there.

But not clear about how to do the ones that are not apicast policy related?

BTW: I think I saw people asking for "#BytesTransferred" also in sme-apis.

andrewdavidmackenzie:
OK, cool.

My main point was that it sounded to me like Joaquim was asking for some metrics that are not related to any specific policy, but the underlying NGINX?
(Free dictionary space etc....)

mikz:
Yes. And we can expose them in some other policy or make every policy responsible for monitoring own free space. But we will need some global non apicast/3scale metrics anyway, so probably better to shove it to some nginx metrics policy.

andrewdavidmackenzie:
👍

jmprusi:
@mikz Yes, it makes sense to have specific apicast policy metrics (related mostly to threescale, and operation mode) and then have another policy for basic metrics.

davidor:
Ping @3scale/product 
Can you provide your input for this feature and decide whether it should be part of the next release?
Thanks.

mikz:
@davidor I think this is not necessary for the next release, but it is possible will be there done by ostia team.

andrewdavidmackenzie:
This was raised last week by Product as a last minute request.
If @MarkCheshire can get us a simple list, ASAP, then we said we'd consider it.
But I'd say it's at the bottom of the priority list and we shouldn't delay the release for this.

MarkCheshire:
[...]

PR #902, opened 2018-09-17
Add Prometheus metrics for the 3scale batching policy

davidor:
This PR adds a couple of metrics for the 3scale batching policy:

```
# HELP batching_policy_auths_cache_hits Hits in the auths cache of the 3scale batching policy
# TYPE batching_policy_auths_cache_hits counter
batching_policy_auths_cache_hits 2
# HELP batching_policy_auths_cache_misses Misses in the auths cache of the 3scale batching policy
# TYPE batching_policy_auths_cache_misses counter
batching_policy_auths_cache_misses 1
```

Part of #745

### DIFF
--- a/gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
+++ b/gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
@@ -10,6 +10,8 @@ local http_ng_resty = require('resty.http_ng.backend.resty')
 local semaphore = require('ngx.semaphore')
 local TimerTask = require('resty.concurrent.timer_task')
 
+local metrics = require('apicast.policy.3scale_batcher.metrics')
+
 local ipairs = ipairs
 
 local default_auths_ttl = 10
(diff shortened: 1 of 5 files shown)


gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
```diff
@@ -198,8 +208,12 @@ function _M:access(context)
   ensure_timer_task_created(self, service_id, backend)
 
   local cached_auth = self.auths_cache:get(transaction)
+  local auth_is_cached = (cached_auth and true) or false
+  metrics.update_cache_counters(auth_is_cached)
```
mikz: Not sure if it would be any clearer, but I'd go for:

```lua
metrics.update_cache_counters(cached_auth)
```

And there:

```lua
function _M.update_cache_counters(cached_auth)
   func_update_counters[not not cached_auth]()
 end
```

No need to change anything though.

davidor: Just a matter of preference over a small detail, but I prefer my solution because it does not have the `not not`.

PR #904, opened 2018-09-19
Expose OIDC as standalone policy

mikz:
We would like to use OIDC without 3scale, so we need a standalone OIDC policy for authentication.

Requirement for https://github.com/3scale/ostia/issues/14

/cc @3scale/ostia

### DIFF
--- a/gateway/src/apicast/oauth/oidc.lua
+++ b/gateway/src/apicast/oauth/oidc.lua
@@ -8,6 +8,8 @@ local setmetatable = setmetatable
 local ngx_now = ngx.now
 local format = string.format
 local type = type
+local tostring = tostring
+local assert = assert
 
 local _M = {
   cache_size = 10000,
@@ -28,20 +30,18 @@ local mt = {
 
 local empty = {}
 
-function _M.new(service)
-  local oidc = service.oidc or empty
-
-  local issuer = oidc.issuer or ""
+function _M.new(oidc_config)
+  local oidc = oidc_config or empty
+  local issuer = oidc.issuer
   local config = oidc.config or empty
   local alg_values = config.id_token_signing_alg_values_supported or empty
 
   local err
-  if #issuer == 0 or #alg_values == 0 then
+  if not issuer or #alg_values == 0 then
     err = 'missing OIDC configuration'
   end
 
   return setmetatable({
-    service = service,
     config = config,
     issuer = issuer,
     keys = oidc.keys or empty,
@@ -51,7 +51,7 @@ function _M.new(service)
     jwt_claims = {
       -- 1. The JWT MUST contain an "iss" (issuer) claim that contains a
       -- unique identifier for the entity that issued the JWT.
-      iss = jwt_validators.equals_any_of({ issuer }),
+      iss = jwt_validators.chain(jwt_validators.required(), issuer and jwt_validators.equals_any_of({ issuer })),
 
       -- 2. The JWT MUST contain a "sub" (subject) claim identifying the
       -- principal that is the subject of the JWT.
@@ -74,6 +74,9 @@ function _M.new(service)
       -- 6. The JWT MAY contain an "iat" (issued at) claim that identifies
       -- the time at which the JWT was issued.
       iat = jwt_validators.opt_greater_than(0),
+
+      -- This is keycloak-specific. Its tokens have a 'typ' and we need to verify
+      typ = jwt_validators.opt_equals_any_of({ 'Bearer' }),
     },
   }, mt), err
 end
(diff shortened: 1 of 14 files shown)


gateway/src/resty/oidc/discovery.lua
```diff
@@ -15,6 +15,9 @@ local _M = { }
 
 local mt = { __index = _M }
 
+-- to be overridden in tests
```
davidor: Why is this needed? Can't we just use the `http_backend` that `.new()` receives?

gateway/src/apicast/oauth/oidc.lua
```diff
@@ -51,7 +51,7 @@ function _M.new(service)
     jwt_claims = {
       -- 1. The JWT MUST contain an "iss" (issuer) claim that contains a
       -- unique identifier for the entity that issued the JWT.
-      iss = jwt_validators.equals_any_of({ issuer }),
+      iss = issuer and jwt_validators.equals_any_of({ issuer }) or jwt_validators.required(),
```
davidor: Why is `or jwt_validators.required()` needed?

gateway/src/apicast/oauth/oidc.lua
```diff
@@ -92,64 +95,99 @@ end
...
 
-  local jwt_obj = cache:get(cache_key)
+  jwt.token = token
 
-  if jwt_obj then
+  return setmetatable(jwt, jwt_mt)
+end
+
+function _M:parse(jwt, cache_key)
+  local cached = self.cache:get(cache_key or jwt)
```
davidor: Why `cache_key or jwt` ? Shouldn't it be just `cache_key` ?

gateway/src/apicast/policy/oidc_authentication/oidc_authentication.lua
```diff
@@ -0,0 +1,91 @@
+-- This is a oidc_authentication description.
```
davidor: Interesting :eyes:

gateway/src/apicast/policy/oidc_authentication/oidc_authentication.lua
```diff
@@ -0,0 +1,91 @@
...
+local OIDC = require('apicast.oauth.oidc')
+local oidc_discovery = require('resty.oidc.discovery')
+local http_authorization = require('resty.http_authorization')
+local resty_url = require('resty.url')
+local policy = require('apicast.policy')
+local _M = policy.new('oidc_authentication')
+
+local tostring = tostring
+
+_M.cache_size = 100
```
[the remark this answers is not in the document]
mikz: Other modules to this too. I think it is reasonable. If you want to you could write a policy that in the init phase bumps cache sizes of other policies.

gateway/src/apicast/oauth/oidc.lua
mikz: Because `cache_key` can be `nil`. So the token itself is its cache key.

gateway/src/apicast/oauth/oidc.lua
mikz: Basically, I want to validate the issuer, if there is some, or just require the field.

gateway/src/resty/oidc/discovery.lua
mikz: The issue is that when you use this for example in the policy initializer, then the policy should accept this too. Pushing testing client through all levels of the stack seems wrong.

gateway/src/apicast/policy/oidc_authentication/oidc_authentication.lua
```diff
@@ -0,0 +1,91 @@
...
+end
+
+local new = _M.new
+--- Initialize a oidc_authentication
+-- @tparam[opt] table config Policy configuration.
+function _M.new(config)
+  local self = new(config)
+
+  self.issuer_endpoint = valid_issuer_endpoint(config and config.issuer_endpoint)
+  self.discovery = oidc_discovery.new(self.http_backend)
```
davidor: Isn't `self.http_backend` always `nil` at this point?

mikz: Yes, unless you set it on the module level. `self` already has the metatable pointing to _M.

spec/policy/oidc_authentication/oidc_authentication_spec.lua
```diff
@@ -0,0 +1,185 @@
...
+    iss = 'http://example.com/issuer',
+    sub = 'some',
+    aud = 'one',
+    exp = ngx.now() + 3600,
+  },
+}, { __tostring = function(jwt) return JWT:sign(rsa.private, jwt) end })
+
+describe('oidc_authentication policy', function()
+  describe('.new', function()
+    it('works without configuration', function()
```
davidor: For these 2 cases with empty or nil config, it might be useful to check that `rewrite` and `access` do not raise unexpected errors.

spec/policy/oidc_authentication/oidc_authentication_spec.lua
```diff
@@ -0,0 +1,185 @@
...
+
+      stub(oidc_discovery, 'http_backend', test_backend)
+
+      test_backend
+              .expect{ url = issuer_endpoint .. "/.well-known/openid-configuration" }
+              .respond_with{ status = 200, headers = { content_type = 'application/json;charset=UTF-8' },
+                             body = [[
+            {
+              "issuer": "https://example.com/issuer",
+              "jwks_uri": "https://example.com/jwks",
```
davidor: I think that extracting this one to a local var used in both mocks would nicely show how they're related.

spec/policy/oidc_authentication/oidc_authentication_spec.lua
```diff
@@ -0,0 +1,185 @@
...
+      local context = { }
+
+      ngx.var.http_authorization = 'Bearer ' .. tostring(access_token)
+
+      policy.oidc.alg_whitelist = { RS256 = true }
+      policy.oidc.keys = { [access_token.header.kid] = { pem = rsa.pub } }
+
+      policy:rewrite(context)
+
+      assert.contains({ payload = { iss = 'http://example.com/issuer' } }, context.jwt)
```
davidor: Shouldn't this return the other claims? I don't see why it only returns the `iss`.

mikz: It does. But the `assert.contains` verifies just subset. I could extend it to all the claims.

gateway/src/apicast/policy/oidc_authentication/oidc_authentication.lua
davidor: :+1:

gateway/src/resty/oidc/discovery.lua
davidor: If this is only for testing, another option would be to modify `self.backend` in the tests. Not sure if better.

I get your point of having to pass this through multiple objects.

:+1:

mikz: Actually, this got me thinking. I think we should offer this in http_ng, so we can stub it globally in the whole test suite.

spec/policy/oidc_authentication/oidc_authentication_spec.lua
davidor: I'd do that. It's a bit confusing in my opinion right now.
How about `assert.same(access_token.payload, context.jwt)` ?

gateway/src/apicast/oauth/oidc.lua
davidor: Hmm so is it OK if `iss` is there but not included in `issuer` ? :thinking:

mikz: Well, that's the thing. If there is no issuer in the configuration, what to do?

I decided to just require any value for `iss`.

spec/policy/oidc_authentication/oidc_authentication_spec.lua
```diff
@@ -0,0 +1,185 @@
...
+      _M.new{ issuer_endpoint = issuer_endpoint }
+    end)
+  end)
+
+  describe(':rewrite', function()
+    before_each(function()
+      ngx.var = {}
+    end)
+
+    it('parses access token', function()
```
davidor: I'd also mention that it stores the parsed token in the context.
I think this is important because other policies might want to check if there's already a parsed jwt in there.

davidor:
Shouldn't this policy have an `apicast-policy.json` ?

mikz:
That would show it in the UI. I'm not sure we are ready for that. What would happen when a customer would add this policy to existing OIDC service? IMO we should think a bit more how that should work before exposing it.

Issue #903, opened 2018-09-19
[policy.3scale] describe how private endpoint URL behaves when it has a path


mikz:
We should clearly define the behaviour before merging https://github.com/3scale/porta/pull/3

The following use cases come in mind:

* `http://example.com`
* `http://example.com/`
* `http://example.com/path`
* `http://example.com/path/`
* `http://example.com/path/foo`
* `http://example.com/path/foo/`
* `http://example.com/path?test=value` (query param not supported)

With the following requests:

* `/`
* `/foo`

PR #905, opened 2018-09-19
[upstream] support path in the upstream url

mikz:
Route traffic to that path + current url.

Closes https://github.com/3scale/apicast/issues/903

### DIFF
--- a/t/apicast-policy-upstream.t
+++ b/t/apicast-policy-upstream.t
@@ -215,16 +215,12 @@ yay, api backend
 }
 --- upstream
   location /path_in_the_rule {
-     content_by_lua_block {
-       require('luassert').are.equal('GET /path_in_the_rule?user_key=uk&a_param=a_value HTTP/1.1',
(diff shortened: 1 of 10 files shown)


t/apicast-policy-upstream.t
```diff
@@ -215,16 +215,12 @@ yay, api backend
...
-                                     ngx.var.request)
-       ngx.say('yay, api backend');
-     }
+     echo $request;
   }
 --- request
 GET /some_path?user_key=uk&a_param=a_value
 --- response_body
-yay, api backend
+GET /path_in_the_rule/some_path?user_key=uk&a_param=a_value HTTP/1.1
```
mikz: Technically this is a breaking change as it used to be `GET /path_in_the_rule`.

But I was thinking that the upstream policy needs some improvements anyway. For example, stripping the path that it matched from the URL etc. So we can make general improvements to that policy and make it configurable how it operates on URI paths.

PR #907, opened 2018-09-21
IP check policy

### LINKED ISSUE #906: Add a policy to whitelist and blacklist IPs
This policy should accept or deny requests based on a list of whitelisted or blacklisted IPs.

davidor:
Closes #906

### DIFF
--- a/gateway/src/apicast/policy/ip_check/client_ip.lua
+++ b/gateway/src/apicast/policy/ip_check/client_ip.lua
@@ -1,0 +1,40 @@
+local ipairs = ipairs
+local re = require('ngx.re')
+
+local _M = {}
+
+local function last_caller_ip()
+  return ngx.var.remote_addr
+end
+
+local function ip_from_x_real_ip_header()
+  return ngx.req.get_headers()['X-Real-IP']
+end
+
+local function ip_from_x_forwarded_for_header()
+  local forwarded_for = ngx.req.get_headers()['X-Forwarded-For']
+
+  if not forwarded_for or forwarded_for == "" then
+    return nil
+  end
+
+  return re.split(forwarded_for, ',', 'oj')[1]
+end
+
+local get_ip_func = {
+  last_caller = last_caller_ip,
+  ["X-Real-IP"] = ip_from_x_real_ip_header,
+  ["X-Forwarded-For"] = ip_from_x_forwarded_for_header
+}
+
+function _M.get_from(sources)
+  for _, source in ipairs(sources or {}) do
+    local func = get_ip_func[source]
+
+    local ip = func and func()
+
+    if ip then return ip end
+  end
+end
+
+return _M
--- a/gateway/src/apicast/policy/ip_check/ip_check.lua
+++ b/gateway/src/apicast/policy/ip_check/ip_check.lua
@@ -1,4 +1,5 @@
 local iputils = require("resty.iputils")
+local ClientIP = require('apicast.policy.ip_check.client_ip')
 
 local policy = require('apicast.policy')
 local _M = policy.new('IP check policy')
@@ -6,17 +7,14 @@ local _M = policy.new('IP check policy')
 local new = _M.new
 
 local default_err_msg = 'IP address not allowed'
+local default_client_ip_sources = { 'last_caller' }
 
 local function parse_cidrs(cidrs)
   return iputils.parse_cidrs(cidrs or {})
 end
 
-local function request_ip_addr()
-  return ngx.var.remote_addr
-end
-
-local function req_ip_in_range(cidrs)
-  return iputils.ip_in_cidrs(request_ip_addr(), cidrs)
+local function ip_in_range(ip, cidrs)
+  return iputils.ip_in_cidrs(ip, cidrs)
 end
 
 local function deny_request(error_msg)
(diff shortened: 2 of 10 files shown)


gateway/src/apicast/policy/ip_check/ip_check.lua
```diff
@@ -0,0 +1,75 @@
+local iputils = require("resty.iputils")
```
mikz: We should move this into `production` group in the Roverfile:

https://github.com/3scale/apicast/blob/6edf4a088527e2528e30113b75fcf056e4ed1474/gateway/Roverfile#L14

gateway/src/apicast/policy/ip_check/ip_check.lua
```diff
@@ -0,0 +1,75 @@
...
+end
+
+local noop = function() end
+
+local function check_ips_function(blacklisted_ips, whitelisted_ips)
+  local blacklist = #blacklisted_ips > 0
+  local whitelist = #whitelisted_ips > 0
+
+  if blacklist and whitelist then
+    log_mutually_exclusive_err()
```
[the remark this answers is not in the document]
davidor: I didn't think about the use case that you mentioned.

I was thinking about cases where the 2 sets of IPs do not overlap. For example, blacklist '1.2.3.0/28' and whitelist '2.3.4.0/28'. What should we do with an IP like '3.3.3.3' ? It's not blacklisted, but not whitelisted either. So we'd need to complicate things a bit to define what happens in scenarios like that one.

gateway/src/apicast/policy/ip_check/ip_check.lua
davidor: I didn't notice this one is in the "development" group :+1:

gateway/src/apicast/policy/ip_check/ip_check.lua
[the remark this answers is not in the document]
davidor: It's not clear to me why whitelisting or blacklisting should take precedence over the other.

I think the best option is to add a "type" attr in the policy config, as you suggested. That would avoid the "precedence" problem. Also, it would allow whitelisting a group of IPs and blacklisting one in that group by chaining 2 instances of the policy.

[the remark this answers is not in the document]
mikz: Agreed 👍

gateway/src/apicast/policy/ip_check/ip_check.lua
davidor: Fixed :+1:

gateway/src/apicast/policy/ip_check/ip_check.lua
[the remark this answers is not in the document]
davidor: Done :+1:

gateway/src/apicast/policy/ip_check/ip_check.lua
```diff
@@ -0,0 +1,75 @@
...
+local _M = policy.new('IP check policy')
+
+local new = _M.new
+
+local function parse_cidrs(cidrs)
+  return iputils.parse_cidrs(cidrs or {})
+end
+
+local function request_ip_addr()
+  return ngx.var.remote_addr
```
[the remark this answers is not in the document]
davidor: You are right.
I've extracted the logic of getting the client IP to a separated module (see 3 latest commits). Let me know if you agree with the approach taken.

gateway/src/apicast/policy/ip_check/ip_check.lua
```diff
@@ -0,0 +1,75 @@
...
+end
+
+local function req_ip_in_range(cidrs)
+  return iputils.ip_in_cidrs(request_ip_addr(), cidrs)
+end
+
+local function deny_request()
+  ngx.status = ngx.HTTP_FORBIDDEN
+  ngx.say('IP address not allowed')
```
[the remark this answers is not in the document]
davidor: For now, I've made the message configurable.
I think that supporting different formats affects several policies and should be done in a separate PR.

gateway/src/apicast/policy/ip_check/client_ip.lua
```diff
@@ -0,0 +1,40 @@
...
+end
+
+local function ip_from_x_forwarded_for_header()
+  local forwarded_for = ngx.req.get_headers()['X-Forwarded-For']
+
+  if not forwarded_for or forwarded_for == "" then
+    return nil
+  end
+
+  return re.split(forwarded_for, ',', 'oj')[1]
```
mikz: This is more complicated than just taking the taking the first one.

The realip module http://nginx.org/en/docs/http/ngx_http_realip_module.html exposes some "internal" ranges, so they don't get considered to be client IPs.

I hope this is enough for the OpenShift case, as for example our echo-api (running on OpenShift) gets following headers:

```
X-Forwarded-For: 46.135.85.38, 10.0.103.238
Forwarded: for=10.0.103.238;host=echo-api.3scale.net;proto=http
```

gateway/src/apicast/policy/ip_check/client_ip.lua
```diff
@@ -0,0 +1,40 @@
...
+local function last_caller_ip()
+  return ngx.var.remote_addr
+end
+
+local function ip_from_x_real_ip_header()
+  return ngx.req.get_headers()['X-Real-IP']
+end
+
+local function ip_from_x_forwarded_for_header()
+  local forwarded_for = ngx.req.get_headers()['X-Forwarded-For']
```
mikz: There is a posiblity that `get_headers` will be called twice. IMO we should avoid that by passing it from outside.

I've been thinking about this for some time and I think we could at some point extract "request" object to the context. This object would lazy load headers when accessed and cache them.

There is a tradeoff between: loading headers always even when not needed, or twice when using both realip and forwarded for, or complicating the code to load headers only when needed.

gateway/src/apicast/policy/ip_check/ip_check.lua
```diff
@@ -0,0 +1,68 @@
...
+  self.client_ip_sources = conf.client_ip_sources or default_client_ip_sources
+
+  local check_type = conf.check_type
+  self.check_client_ip = check_ip_function[check_type] or noop
+
+  return self
+end
+
+function _M:access()
+  local client_ip = ClientIP.get_from(self.client_ip_sources)
```
mikz: Might be good to cache the `get_from` function in a local variable for the JIT compiler.

But that is something we should do on many other places of the codebase, so no changes needed.

davidor: :+1:

I think it's better to make this kind of changes after measuring the perf penalty they add.

Issue #908, opened 2018-09-24
Staging Apicast is not able to recover when there is an error in the liquid


pestanko:
[provide a description of the issue]

##### Version

3scale 2.3 CR1

##### Steps To Reproduce

1. Create an API
2. Configure for example query rewrite
3. set new arg with value: ` {{ now() }}`
4. Make a request - the request will fail with 500
5. Change the value to `{{ now }}`
6. See the error: Execution timeout (due being not able to acquire lock)

##### Current Result

```
2018/09/24 11:20:40 [error] 21#21: *86 lua entry thread aborted: runtime error: /usr/local/share/lua/5.1/liquid.lua:1115: Invalid syntax: parsing stoped at: line 1 cols: 7 current_token is Token(LPAREN,()
--
  | info: expect var_end '}}' or '-}}'
  | stack traceback:
  | coroutine 0:
  | [C]: in function 'error'
  | /usr/local/share/lua/5.1/liquid.lua:1115: in function 'raise_error'
  | /usr/local/share/lua/5.1/liquid.lua:1638: in function 'var'
  | /usr/local/share/lua/5.1/liquid.lua:1219: in function 'block'
  | /usr/local/share/lua/5.1/liquid.lua:1176: in function 'document'
  | /opt/app-root/src/src/apicast/template_string.lua:75: in function 'liquid_parser'
  | /opt/app-root/src/src/apicast/template_string.lua:89: in function 'new'
  | ...t/src/src/apicast/policy/url_rewriting/url_rewriting.lua:64: in function 'build_template'
  | ...t/src/src/apicast/policy/url_rewriting/url_rewriting.lua:98: in function 'load_policy'
  | /opt/app-root/src/src/apicast/configuration.lua:69: in function 'build_policy_chain'
  | /opt/app-root/src/src/apicast/configuration.lua:93: in function 'parse_service'
  | ...-root/src/src/apicast/configuration_loader/remote_v2.lua:125: in function 'index'
  | ...-root/src/src/apicast/configuration_loader/remote_v2.lua:145: in function 'call'
  | /opt/app-root/src/src/apicast/configuration_loader.lua:54: in function 'load'
  | /opt/app-root/src/src/apicast/configuration_loader.lua:214: in function 'rewrite'
  | ...apicast/policy/load_configuration/load_configuration.lua:33: in function <...apicast/policy/load_configuration/load_configuration.lua:31>
[...]

PR #854, opened 2018-08-24
[policy_chain] skip invalid policy configurations

mikz:
Invalid policies in the chain are ignored.

PR #911, opened 2018-09-25
Handle properly configs with invalid Liquid templates

### LINKED ISSUE #909: Some liquid functions are not working - especially the time functions
All of the time functions are returning the empty string:

**Functions:** `time`, `today`, `http_time`, `cookie_time`, `localtime`, `now`, `utctime`

##### Version

3scale 2.3-CR1

##### Steps To Reproduce

1. Create application
2. add URL rewrite policy
3. add rules for any of these functions with liquid
  Example: `arg` = `{{ now }}`
4. Make a request and response should look like:
```
"args": "user_key=5941428bd561fa5a91489f2c36ed8cb3&arg=",
```

##### Current Result
```json
{
  "method": "GET",
  "path": "/test/path",
  "args": "user_key=5941428bd561fa5a91489f2c36ed8cb3&arg=",
  "body": "",
...
}
```
##### Expected Result
```json
{
  "method": "GET",
  "path": "/test/path",
  "args": "user_key=5941428bd561fa5a91489f2c36ed8cb3&arg=TIMESTAMP",
  "body": "",
...
}
```

davidor:
Closes partially #908

### DIFF
--- a/gateway/src/apicast/policy_chain.lua
+++ b/gateway/src/apicast/policy_chain.lua
@@ -14,6 +14,7 @@ local require = require
 local insert = table.insert
 local sub = string.sub
 local format = string.format
+local pcall = pcall
 local noop = function() end
 
 require('apicast.loader')
@@ -87,7 +88,13 @@ function _M.load_policy(module, version, ...)
         local mod, err = policy_loader:pcall(module, version or 'builtin')
 
         if mod then
-            return mod.new(...)
+            local new_policy_ok, policy, new_err = pcall(mod.new, ...)
+            if new_policy_ok then
+                return policy, new_err
+            else
+                ngx.log(ngx.ERR, 'Policy ', module, ' crashed in .new(). It will be ignored.')
+                return nil, policy
+            end
         else
             return nil, err
         end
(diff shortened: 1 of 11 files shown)


davidor:
You are right, it's #908

davidor:
And as you mentioned we should probably make this generic and not only liquid-specific.

mikz:
@davidor yes, something along the lines of https://github.com/3scale/apicast/pull/854

That one, unfortunately, missed the failure in `.new`.

gateway/src/apicast/policy_chain.lua
```diff
@@ -87,7 +88,12 @@ function _M.load_policy(module, version, ...)
         local mod, err = policy_loader:pcall(module, version or 'builtin')
 
         if mod then
-            return mod.new(...)
+            local new_policy_ok, res = pcall(mod.new, ...)
```
mikz: It might be good to expect a second return value `err`. It is possible a policy could `return nil, 'error'`.

davidor: Fixed :+1:

PR #912, opened 2018-09-25
release: 3.3.0-cr2

davidor:


davidor:
Travis fails because the changelog links to the v3.3.0-cr2 tag which will be created after merging.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [3.3.0-cr2] - 2018-09-25
```
mikz: Isn't it rc2?

davidor: :man_facepalming: 
Copy-pasted from rc1 which is wrong.

davidor: let's keep 'rc' for 3.3.0 at least so we don't have to change the existing rc1 tag.

PR #917, opened 2018-10-01
metrics: rename backend_response to threescale_backend_response

davidor:
To clearly distinguish it from the upstream API as suggested in https://github.com/3scale/apicast/issues/745#issuecomment-419365339

### DIFF
--- a/gateway/src/apicast/metrics/3scale_backend_status.lua
+++ b/gateway/src/apicast/metrics/3scale_backend_status.lua
@@ -7,7 +7,7 @@ local _M = {}
 
 local backend_response_metric = prometheus(
   'counter',
-  'backend_response',
+  'threescale_backend_response',
   "Response status codes from 3scale's backend",
   { 'status' }
 )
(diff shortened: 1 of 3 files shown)


gateway/src/apicast/metrics/3scale_backend_status.lua
```diff
@@ -7,7 +7,7 @@ local _M = {}
 
 local backend_response_metric = prometheus(
   'counter',
-  'backend_response',
+  'threescale_backend_response',
```
mikz: Can they start with a number? `threescale` should be the last resort how to write`3scale`

davidor: no, that's why I chose `threescale` :D

mikz: That does not sound that bad. I'm fine with both.

davidor: I'll add the `apicast` prefix. I think it makes sense if we take into account that other apps can export metrics to the same Prometheus instance. Also, it's recommended in the Prometheus guidelines: https://prometheus.io/docs/practices/naming/

mikz: 👍

PR #918, opened 2018-10-02
Add Prometheus metrics for the upstream

davidor:
Part of #745

### DIFF
--- a/gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
+++ b/gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
@@ -9,6 +9,8 @@ local select = select
 local find = string.find
 local pairs = pairs
 
+local upstream_metrics = require('apicast.metrics.upstream')
+
 local new = _M.new
 
 local log_levels_list = {
@@ -108,6 +110,10 @@ function _M:metrics()
   for name,dict in pairs(ngx.shared) do
     metrics_updater.set(shdict_free_space_metric, dict:free_space(), name)
   end
+end
+
+function _M.log()
+  upstream_metrics.report(ngx.var.upstream_status, ngx.var.upstream_response_time)
 end
 
 return _M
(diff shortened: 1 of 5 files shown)


gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
```diff
@@ -110,4 +112,8 @@ function _M:metrics()
   end
 end
 
+function _M.log()
+  upstream_metrics.report(ngx.var.upstream_status, ngx.var.upstream_response_time)
```
mikz: I think we should use labels to differentiate different upstreams. Having just one metric for all upstreams won't make it much useful as you won't see which one it is taking longer or failing.

davidor: I think it can be helpful to check if there's a noticeable mismatch between apicast vs upstream status codes.

I agree that it would be more useful to have a label for the upstreams. However, I'm not sure about the performance impact that it might have. The number of upstreams could be very large and the Prometheus docs recommend against using labels that can have many different values. See the caution note here: https://prometheus.io/docs/practices/naming/#labels . Also: https://prometheus.io/docs/practices/instrumentation/#do-not-overuse-labels

mikz: @davidor btw upstream_status can be missing if any policy terminates the request.
We have this code in APIcast Cloud Hosted to measure the difference between upstream_status and ngx.status: https://github.com/3scale/apicast-cloud-hosted/blob/e5da0fc2f9df0d0ea89339ab8bfab3d27d169f78/apicast/policies/cloud_hosted.balancer_blacklist/0.1/balancer_blacklist.lua#L104-L117

The prometheus docs say that we should avoid cardinality higher than 100. I don't think we could get there. If we just store the host and port as a label it should be fine. 

But also it is something that can be done later.

davidor: I added specs and modified the code to make sure that we cover the cases where the status or the latency are nil or empty.
https://github.com/3scale/apicast/pull/918/commits/de684be62e320275890c20af8e8bc6c338f4db32

PR #919, opened 2018-10-03
Metrics: add the type of 3scale backend call (auth, authrep, report)

davidor:
Part of https://github.com/3scale/apicast/issues/745


gateway/src/apicast/metrics/3scale_backend_calls.lua
```diff
@@ -0,0 +1,36 @@
...
+local valid_endpoint = {
+  auth = true, authrep = true, report = true
+}
+
+function _M.report(endpoint, status)
+  if threescale_backend_call then
+    if valid_endpoint[endpoint] then
+      threescale_backend_call:inc(1, { endpoint, label_for_status(status) })
+    else
+      ngx.log(ngx.WARN, 'Invalid 3scale backend endpoint endpoint when updating metrics')
```
mikz: Do we really care about this?

spec/backend_client_spec.lua
```diff
@@ -2,14 +2,18 @@ local _M = require('apicast.backend_client')
...
 
 describe('backend client', function()
 
   local test_backend
   local options_header_oauth = 'rejection_reason_header=1'
   local options_header_no_oauth = 'rejection_reason_header=1&no_body=1'
 
-  before_each(function() test_backend = test_backend_client.new() end)
+  before_each(function()
+    test_backend = test_backend_client.new()
```
mikz: This should be not necessary with the changes from https://github.com/3scale/apicast/pull/904/commits/576fa67570358ffdeba4381b5693464b3882b33c

spec/backend_client_spec.lua
```diff
@@ -90,6 +94,17 @@ describe('backend client', function()
 
       assert.equal(200, res.status)
     end)
+
+    it('reports the call with the status', function()
+      local service = configuration.parse_service({ id = '42' })
+      local status = 200
+      stub(test_backend, 'send', function() return { status = status } end)
```
mikz: Why this instead of the `test_backend.expect` ?

davidor: Because in this case, we don't really care about what the backend client receives, we just want it to return a 200 so we can report it.

mikz: Would something like `test_backend.expect { }.respond_with{status = 200}` work?

spec/metrics/3scale_backend_calls_spec.lua
```diff
@@ -0,0 +1,73 @@
+describe('backend calls metrics', function()
+  describe('report', function()
+    local threescale_backend_calls_metrics
+    local test_counter = { inc = function() end }
+
+    before_each(function()
+      stub(test_counter, 'inc')
+      local Prometheus = require('apicast.prometheus')
+      getmetatable(Prometheus).__call = function(_, type)
```
mikz: Doing this dance in tests is so ugly and error-prone, that I think this could be an excellent time to fix this interface.

spec/metrics/3scale_backend_calls_spec.lua
```diff
@@ -0,0 +1,73 @@
...
+        if type == 'counter' then
+          return test_counter
+        end
+      end
+
+      package.loaded['apicast.metrics.3scale_backend_calls'] = nil
```
mikz: Similar comment as above. If we have to do this, looks like something is broken in our interface.

spec/metrics/3scale_backend_calls_spec.lua
davidor: Agreed. I thought the same because it's the second time we need to write something like this.
I was thinking about whether to modify it in this PR or the next one.

mikz: I'm good either way.

gateway/src/apicast/metrics/3scale_backend_calls.lua
davidor: What about the possibility of having invalid information in Prometheus?

mikz: But what would be that invalid information? Someone part of code calling the wrong endpoint on the backend. I don't consider that invalid metric. If that call happened, then it should be measured and reported.

davidor: Fair enough.

spec/metrics/3scale_backend_calls_spec.lua
davidor: Let's address this one separately because it will change many things not directly related to this PR.

PR #926, opened 2018-10-05
Standalone configuration

### LINKED ISSUE #795: APIcast standalone configuration
Right now the APIcast configuration is tightly coupled to the 3scale data model with Service and Proxy objects.

To be a useful standalone gateway APIcast should be usable general purpose gateway without those concepts. As mentioned in https://github.com/3scale/apicast/issues/173#issuecomment-393808026 we might need to re-envision how APIcast configuration/data model would look like and then map 3scale concepts to it. With features like TLS termination or TCP proxying we need to introduce concepts of a Port/Server and get rid of routing only by hostname. Also the internal context object passed to policies will change as there might not be "service" or it might have different structure.

Proposals should be posted as comments.
Please note that this is not about the serialization format or how to deliver the configuration (json, toml, grpc, file), but about it's data model and core concepts.

## Requirements

* Start listening on more ports, for example: HTTP + HTTPS, HTTP + proxy-protocol, TCP, UDP
* Ability to route incoming traffic by its metadata:
  - HTTP request by HTTP request data (path, headers, host)
  - TCP/UDP request by port/remote address/other data available from the request.
  - TLS request by TLS metadata (like SNI).
* Ability to redirect request to another service/policy chain

### Scenario 1

Current APIcast configuration. Echo API, Fake Backend, APIcast, Management API, Metrics API.

### Scenario 2

Path based routing (https://github.com/3scale/apicast/issues/173#issue-193484370). API composition.
Internal Service Flights and Service Tickets are exposed as `/flights` and `/tickets` on public API `api.example.com`.
But they are also reachable internally on hostnames `flights.internal` and `tickets.internal`.

### Scenario 3

HTTP Request header based routing. Requests having Cookie `env=staging` should be router to staging service, and everyone else to production.

### Scenario 4

Sharing policy chain between staging and production. A public API is exposed as `api.example.com` and points to upstream `api.prod.internal`. It has defined a policy chain doing CORS and URL rewriting. It's staging deployment is exposed as `api.staging.example.com` pointing to upstream `api.stag.internal`. If production configuration is changed it should be reflected in the staging configuration too.

Nice to have: explore both options: deploy staging and production together in one config, define production and staging in different configs 

### Scenario 5

API available on multiple hostnames. One API is available both as `flights.api.example.com` and `flights.internal` pointing to the same upstream (`flights.internal`). It should use the same policy chain.

### Scenario 6

2 independent public APIs are each pointing to unique upstream. CORS needs to be configured on both of them the same way.

As a nice to have: plus each of those APIs has own unique URL rewriting.

### Scenario 7

An API is running on a wildcard domain `*.example.com` using upstream `api.internal`. There is also another API running on `*-admin.example.com` pointing to `admin.internal` upstream. Plus there is an exception for `michal-admin.example.com` pointing to `admin.staging.internal`.

### Scenario 8

APIcast should start listening on 53/tcp, 53/udp, and 853/tcp+tls and use "dns" policy.

### Scenario 9

Two public APIs are authenticating by mutual TLS using different CA certificates. 
Each API needs to provide own way of verifying the client certificate and serve own certificate to the client. Routing should be done by using SNI as they don't have a unique IP address.
Nice to have: One API is using TLS 1.2 only and the other TLS 1.3 only.

### Scenario 10

APIcast can be started in different environments (staging, production, development, testing) by supplying an environment lua file that can define things like policy chains, ports APIcast starts on, caching TTL. 
This technique can be used to parameterize the main configuration file for each environment without needing to change the main configuration (for development for example).

### Scenario 11

Minimal viable configuration that proxies all requests to an upstream.

### Scenario 12

In a microservices architecture we need APIcast to have the ability to auto-discover new APIs (upstreams) deployed within the same namespace. This could be done in a number of ways:
- The API services are pushed into the local registry
  - APIcast polls that registry to discover new API services.
  - The new API services are pushed to APIcast or triggers a call from APIcast to pull the new API services.
- The last option would be to use APIs.json to expose each of the new APIs. Less preferable as we already have the registry in Kubernetes here.

APIcast should then generate a proxy for the _"new"_ API service. The entrypoint for each namespace is **always** the gateway. Policies should be configured in exactly the same manner as in any other architecture. Scenario 5 would be a very typical use case in a microservices architecture where there is communication between microservices. _"Should those requests be rate limited?"_ We need to think about how to define rate limits based on internal or external requests. Scenario 6 would apply in this context for both external and internal requests. each namespace could be exposed on different subdomains and therefore traffic between namespace would invoke CORS. Scenario 10 in a microservices architecture would work slightly differently. The configuration should be defined by the environment and how APIcast gets that configuration should be the same in every environment. The configuration available in each environment should be scoped by environment.

mikz:
Fixes #795 

Implements "standalone" configuration, that does not depend on 3scale and can be human-authored YAML document describing internal services and routing between them.

Please note this is just first version and it will be refined as we start using it more.


gateway/src/apicast/policy/standalone/apicast-policy.json
```diff
@@ -0,0 +1,13 @@
+{
```
davidor: Do we need to expose this one?
I thought this one would be used transparently just when starting APIcast with `APICAST_CONFIGURATION`.

mikz: Not needed 👍 It was generated by `bin/apicast generate policy`

gateway/config/standalone.lua
```diff
@@ -0,0 +1,22 @@
+local PolicyChain = require('apicast.policy_chain')
+
+local standalone = PolicyChain.load_policy(
+        'apicast.policy.standalone',
+        'builtin',
+        { url = context.configuration })
+
+if arg then -- running CLI to generate nginx config
```
davidor: Where does this `arg` come from?

davidor: Oh right, lua's arg.

mikz: Yep. It is a global variable when executed through CLI.

spec/policy/standalone/configuration_spec.lua
```diff
@@ -0,0 +1,21 @@
...
+
+        it('does not accept https', function()
+            assert.returns_error('scheme not supported', _M.new('https://example.com'))
+        end)
+
+        it('does not accept invalid URL', function()
+            assert.returns_error('missing scheme', _M.new('invalid url'))
+        end)
+    end)
+end)
```
davidor: I think we should also add tests for `.load()`.

Testing the error with an invalid scheme should be easy, and we could create a simple yaml in the fixtures to make sure that it returns the correct decoded lua table.

gateway/src/apicast/policy/standalone/standalone.lua
```diff
@@ -0,0 +1,364 @@
...
+    service.policy_chain[phase](service.policy_chain, ...)
+  end
+end
+
+function _M:init(...)
+  if self then -- initializing policy instance
+    local config, err = self:load_configuration(self)
+
+    if config then
+      -- TODO: we need to run this because some policies are "internal" and not being
```
davidor: I think we had a similar problem in `Executor` and figure out a way to run every phase only once.

gateway/src/apicast/policy/standalone/standalone.lua
```diff
@@ -0,0 +1,364 @@
...
+    }, Destination_mt)
+  end
+end
+
+local Condition = { }
+do
+
+  local Condition_mt = { __index = Condition }
+
+  local operations = {
```
davidor: In a future iteration, we should think about how to reuse what we already did for the Conditional policy or add there whatever is missing to support what you introduced here.

gateway/src/apicast/policy/standalone/standalone.lua
```diff
@@ -0,0 +1,364 @@
...
+local Service = { }
+
+local function build_services(configuration)
+  return build_objects(Service.new, configuration.internal)
+end
+
+do
+  local Service_mt = { __index = Service }
+
+  local function build_policy_chain(policies)
```
davidor: It would make sense to have this method in the `PolicyChain` module.

gateway/http.d/standalone.conf.liquid
```diff
@@ -0,0 +1,58 @@
...
+
+    {% if server.tls.protocols -%}
+        ssl_protocols {{ server.tls.protocols | join: " " }};
+    {%- endif %}
+
+    {%- capture server_name -%}{{ server.name | default: 'default' }}{%- endcapture %}
+    set $port_name {{ server_name }};
+    server_name {{ server_name }};
+
+    set $ctx_ref -1;
```
davidor: Why is this needed?

examples/configuration/standalone.yml
```diff
@@ -0,0 +1,171 @@
...
+    protocol: http2 # | spdy | http
+    tls: true
+  - port: 8444
+    name: default
+    protocol: http2 # | spdy | http
+    proxy_protocol: true
+    tls:
+      protocols: TLSv1.3
+      # those two could be the defaults as policies have ssl_certificate phase
+      certificate: /var/run/secret/apicast/certificate.crt
```
davidor: Should we add a note somewhere saying that these are not supported yet? Or are they?

davidor:
I agree with the approach taken :+1: 

My comments are just about minor things. I also think that we need more unit and integration tests.

As you mentioned in your original comment, this is just a first version, so no need to fix everything now.

gateway/http.d/standalone.conf.liquid
mikz: It is used by `require('resty.ctx').stash()` to share the `ngx.ctx` between normal phases and post_action. apicast.conf does the same.

examples/configuration/standalone.yml
mikz: Good point. Not yet.

gateway/src/apicast/policy/standalone/standalone.lua
mikz: Extracted what could be extracted into the `PolicyChain`. But still some had to stay because it depends on the structure of the standalone policy configuration.

spec/policy/standalone/configuration_spec.lua
mikz: Done.

examples/configuration/standalone.yml
mikz: Fixed, supported now.

gateway/src/apicast/policy/standalone/standalone.lua
mikz: It was a bit different problem I think.

Executor does the following:

* call `:init` and initialize all policies in the current chain
* use `PolicyLoader:get_all()` to load "all" policies
* call `:init` on the remaining policies

Now  `PolicyLoader:get_all()` does not return all the policies. Some don't have `apicast-policy.json` file, so they are not discovered.

So the problem is that the Executor is going to call the `:init` phase on all policies that it can find but does not expose the information if the phase was already executed. So standalone can't really know if it was already executed or not.

There are two options I see:

* expose information about what phase was executed on what policy in the Executor
* standalone policy should not trigger `:init` on policies that can be found by `PolicyLoader` and assume it was done by the Executor module


Wdyt?

gateway/src/apicast/policy/standalone/standalone.lua
mikz: Agreed. Wanted to have this here, for now, to see how it evolves. But in the future there should be just one way to do this.

gateway/src/apicast/policy/standalone/standalone.lua
davidor: I prefer the first one because I think it's more explicit, but I think that both options should be OK.

mikz: So after some digging, I realised the problem is a bit deeper.

Executor keeps track of what it executed, but it can't keep track of what nested policy chains executed. Let's say Executor has 3 policies and the 2nd is itself a policy chain with 3 more policies. The Executor would be aware only of the first level. 

We would have to figure out how to collect all policies from the executor without calling them.


Maybe I'd rather postpone this until we get more information and see more places where it breaks.

davidor: Ah, I didn't think about the nested chains scenario.
OK to postpone it :+1: No need to solve this for a first version.

Issue #335, opened 2017-03-24
Rate limiting headers and status code


ErwinSteffens:
For now we always get a ‘403 Forbidden’ error with body ‘Authentication failed’. We would like to return a more specific error status code like ‘429 Too Many Requests’. 

Also we would like to add ‘X-Rate-Limit’ headers like in the twitter API: https://dev.twitter.com/rest/public/rate-limiting. 

Is this possible to create with some custom scripting?

mikz:
The rate limiting is done by 3scale Service Management API. 

When Rate Limit is reached it returns response like:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<status>
  <authorized>false</authorized>
  <reason>usage limits are exceeded</reason>
  <plan>test</plan>
  <usage_reports>
    <usage_report metric="test" period="minute">
      <period_start>2017-03-24 11:21:00 +0000</period_start>
      <period_end>2017-03-24 11:22:00 +0000</period_end>
      <max_value>10</max_value>
      <current_value>10</current_value>
    </usage_report>
  </usage_reports>
</status>
```

Now I'm not sure how many elements there will be if multiple metrics are hit.

But I guess it would be feasible to extract this information and report it to the client. The issue is parsing XML. There are some libraries but nothing really great. 

We eventually would like this feature to be in the product.

ErwinSteffens:
Ok, I think we will build something based on this. 

We can share it back as an example when finished, but I think it will be based on a single metric.

mikz:
Let me know if you make any progress. Don't try LuaXML - it does not work well with OpenResty.
Using some FFI bindings for libexpat is kind of ok, but not great. I'd suggest using https://github.com/Phrogz/SLAXML if the performance is not so terrible as it is pure lua.

If you feel like try to open an PR and I'll try to help you to get it into master. 

Because the whole process is asynchronous, it has to be preparing the response for the next request.
There is shared memory for storing the last response, so adding there more metadata could work. 

I guess the hardest part is to define how it is supposed to work (including multiple metrics to prevent some nasty accidents) and test it.

mikz:
So we are playing with the idea of introducing `Retry-After` header which is in the HTTP spec.
It says in what time client should retry the call. This would fit the rate limiting story.

The algorithm would be following:

* from all the limits that the authorise call exceeded
* take the greatest granularity one
* send in Retry-After the timestamp when the limit will reset

Would that be enough for you?  @ErwinSteffens 

[...]

PR #929, opened 2018-10-10
Add "Retry-After" header

davidor:
Part of #335 

This PR introduces a "Retry-After" header in the APIcast response.
When rate-limited by the 3scale backend, APIcast returns the number of seconds that the caller needs to wait before it's authorized again in the "Retry-After" header.

### DIFF
--- a/gateway/src/apicast/backend_client.lua
+++ b/gateway/src/apicast/backend_client.lua
@@ -158,15 +158,21 @@ local function create_token_path(service_id)
 end
 
 -- Returns the authorize options that 3scale backend accepts. Those options
--- are specified via headers. Right now there are 2:
+-- are specified via headers:
 --  - rejection_reason_header: asks backend to return why a call is denied
 --    (limits exceeded, application key invalid, etc.)
---  - no_nody: when enabled, backend will not return a response body. The
+--  - no_body: when enabled, backend will not return a response body. The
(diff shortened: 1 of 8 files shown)


gateway/src/apicast/backend_client.lua
```diff
@@ -158,15 +158,21 @@ local function create_token_path(service_id)
...
 --  - rejection_reason_header: asks backend to return why a call is denied
 --    (limits exceeded, application key invalid, etc.)
---  - no_nody: when enabled, backend will not return a response body. The
+--  - no_body: when enabled, backend will not return a response body. The
 --    body has many information like metrics, limits, etc. This information is
 --    parsed only when using oauth. By enabling this option will save some work
 --    to the 3scale backend and reduce network traffic.
+--  - limit_headers: when enabled and the request is rate-limited, backend
+--    returns the number of seconds remaining until the limit expires. It
+--    returns -1 when there are no limits. With this header, backend returns
```
ioggstream: @davidor https://tools.ietf.org/html/rfc7231#section-7.1.3 
```A delay-seconds value is a non-negative decimal integer, representing
   time in seconds.``` 

In https://github.com/3scale/apisonator/blob/master/docs/extensions.md#limit_headers-boolean I'd add a notice on that.

PR #930, opened 2018-10-15
Add Prometheus metric for request response times

davidor:
Part of #745 
It's the only metric missing from the list in the issue.

### DIFF
--- a/gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
+++ b/gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
@@ -73,6 +73,12 @@ local http_connections_metric =  prometheus('gauge', 'nginx_http_connections', '
 local shdict_capacity_metric = prometheus('gauge', 'openresty_shdict_capacity', 'OpenResty shared dictionary capacity', {'dict'})
 local shdict_free_space_metric = prometheus('gauge', 'openresty_shdict_free_space', 'OpenResty shared dictionary free space', {'dict'})
 
+local response_times = prometheus(
+  'histogram',
+  'total_response_time_seconds',
+  'Time needed to sent a response to the client (in seconds).'
+)
+
 function _M.init()
   errlog.set_filter_level(filter_level())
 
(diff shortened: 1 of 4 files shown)


gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
```diff
@@ -73,6 +73,12 @@ local http_connections_metric =  prometheus('gauge', 'nginx_http_connections', '
 local shdict_capacity_metric = prometheus('gauge', 'openresty_shdict_capacity', 'OpenResty shared dictionary capacity', {'dict'})
 local shdict_free_space_metric = prometheus('gauge', 'openresty_shdict_free_space', 'OpenResty shared dictionary free space', {'dict'})
 
+local response_times = prometheus(
+  'histogram',
+  'response_time_seconds',
```
mikz: I think we will have more "response time" metrics, so we should name them well. In this case I'd mention it is "total" response time, because in the future we will have "upstream" response time metric for sure.

gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
```diff
@@ -73,6 +73,12 @@ local http_connections_metric =  prometheus('gauge', 'nginx_http_connections', '
...
+  'Request response times in seconds'
```
mikz: I think we can expand the description into something more human-readable like: "How many seconds it took to send a response to the client".

gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
davidor: Fixed :+1: 
I also adapted the name of the metric that tracks upstream response times to follow the same format. It was introduced after the latest release, so it's not a breaking change.

gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
davidor: Fixed :+1:

gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
```diff
@@ -112,8 +118,17 @@ function _M:metrics()
...
+  local resp_time = tonumber(ngx.var.request_time)
+
+  if resp_time and response_times then
+    response_times:observe(resp_time)
+  end
+end
+
 function _M.log()
   upstream_metrics.report(ngx.var.upstream_status, ngx.var.upstream_response_time)
+  report_req_response_time()
```
mikz: Does this include post action time? I guess it does right? Then it might not be exactly correct, because client got the response before post action finished.

PR #940, opened 2018-10-22
[THREESCALE-1289] Fix error when loading a config with only some services using OIDC

davidor:
This PR fixes properly the bug that we tried to fix in https://github.com/3scale/apicast/pull/893

The bug was still there because the integration tests that load the configuration on boot mock the loader v2 module. In the previous PR, we fixed the bug in the remote loader v2. Unfortunately, the same bug was present also in the OIDC loader module and we didn't have tests for that.

In this PR, I've added integration tests that properly test the whole loading flow including the OIDC loader module.

Ref: https://issues.jboss.org/browse/THREESCALE-1289


t/configuration-loading-with-oidc.t
```diff
@@ -0,0 +1,232 @@
+use lib 't';
+use Test::APIcast 'no_plan';
```
mikz: Can't it be a blackbox test?

davidor:
I fixed the codeclimate checks and also adapted a couple of tests of the management endpoint (https://github.com/3scale/apicast/pull/940/commits/bd60b9c1335023216775896395ba153c1d4274dd)

t/configuration-loading-with-oidc.t
davidor: I based these tests on the ones we already have that test the configuration loading.
I'm not sure if a Blackbox test would work in this case because we'd need to define all those locations to get the configuration and there wouldn't be a "---configuration" directive.

mikz: It would work with lazy loading the configuration. 

My point is that Blackbox is 100% real APIcast. This is somewhat similar to APIcast but not really. And we really should aim to not add more of these tests as they are not compatible with the liquid templating of the nginx config.

PR #943, opened 2018-10-24
release: 3.4.0-beta1

davidor:


davidor:
Travis fails because the readme points to a tag that will be created after the merge.

PR #950, opened 2018-11-08
Renew expired certs in mutual ssl tests

davidor:


davidor:
Changed this to load the certs and keys from the file system :+1:

PR #951, opened 2018-11-13
release: 3.4.0-rc1

davidor:


davidor:
Travis build fails because the `3.4.0-rc1` tag does no exist yet.

PR #954, opened 2018-11-16
[THREESCALE-1547] Fix default creds policy

davidor:
The default credentials policy was not working correctly because it stored the default credentials in the service instead of the policies context.
Ref: https://issues.jboss.org/browse/THREESCALE-1547

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed 
+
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed 
+
+- Fix bug in the Default credentials policy. It was using the default credentials in some cases where it should not [PR #954](https://github.com/3scale/apicast/pull/954), [THREESCALE-1547](https://issues.jboss.org/browse/THREESCALE-1547)
```
y-tabata: Since THREESCALE-1547 is the internal ticket, we cannot see it.
Is it no problem to write this kind of private information in the change log?

PR #955, opened 2018-11-16
release: 3.4.0-rc2

davidor:


davidor:
Travis fails because there are some links to the Openshift project that are no longer valid.

PR #959, opened 2018-11-21
Merge 3.4-stable to master

davidor:
Includes #954 and #955


davidor:
Travis fails because broken we have some broken links in the docs, but that's already been fixed in master.

PR #964, opened 2018-11-30
Extract the main APIcast nginx configuration into own file

mikz:
Having all APIcast specific configuration in one file included by liquid allows us to replace that file with different server blocks etc. For example that is used by the standalone mode: #926.

Depends on #963.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -4,3 +4,5 @@ busted = {std = "+busted", globals = { 'fixture' } }
 files["**/spec/**/*_spec.lua"] = busted
 
 globals = { 'rawlen' }
+
+files['gateway/config/*.lua'] = { globals = { 'context' } }
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,4 +1,14 @@
-std = 'ngx_lua+lua52' -- lua52 has table.pack
+local standards = require "luacheck.standards"
+
+-- stds global is provided by luacheck
+-- https://luacheck.readthedocs.io/en/stable/config.html#custom-sets-of-globals
+stds.ngx = {
+    read_globals = {
+        coroutine = standards.def_fields('_yield', '_create', '_resume')
+    }
+}
+
+std = 'ngx_lua+lua52+ngx' -- lua52 has table.pack
 
 busted = {std = "+busted", globals = { 'fixture' } }
 files["**/spec/**/*_spec.lua"] = busted
--- a/gateway/conf/nginx.conf.liquid
+++ b/gateway/conf/nginx.conf.liquid
@@ -46,7 +46,6 @@ http {
   lua_socket_pool_size {{ lua_socket_pool_size | default: 512 }};
   server_names_hash_bucket_size 128;
 
-  log_format time '[$time_local] $host:$server_port $remote_addr:$remote_port "$request" $status $body_bytes_sent ($request_time) $post_action_impact';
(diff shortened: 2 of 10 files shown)


.luacheckrc
```diff
@@ -1,6 +1,16 @@
-std = 'ngx_lua+lua52' -- lua52 has table.pack
+local standards = require "luacheck.standards"
+
+stds.ngx = {
```
davidor: Where does this `stds` table come from?

mikz: Provided by luacheck.

https://luacheck.readthedocs.io/en/stable/config.html#custom-sets-of-globals

Should I link it in the file?

davidor: I think we should

gateway/conf/nginx.conf.liquid
```diff
@@ -69,119 +68,19 @@ http {
   {%- endcapture -%}
 
   {% for file in "http.d/*.conf" | filesystem %}
+    ## include {{ file }}
```
davidor: Why is this needed?

gateway/src/apicast/cli/template.lua
```diff
@@ -77,27 +77,42 @@ local function nginx_prefix()
...
+end
+
 function _M:interpret(str)
     local interpreter = build_interpreter(str)
 
     local context = self.context
     local filesystem = self.filesystem
     local filter_set = FilterSet:new()
-    local resource_limit = ResourceLimit:new(nil, 1000, nil)
+    local resource_limit = ResourceLimit:new(100000000, 100000, nil)
```
davidor: Did you find a problem with the previous limits?

gateway/src/apicast/cli/filesystem.lua
```diff
@@ -9,9 +9,9 @@ local pl_path_dir = pl_path.dir
 local pl_path_join = pl_path.join
 local abspath = pl_path.abspath
 local pcall = pcall
-local co_yield = coroutine.yield
-local co_create = coroutine.create
-local co_resume = coroutine.resume
+local co_yield = coroutine._yield
```
davidor: I think it would be good to add a comment that clarifies why we're using the original methods here.

gateway/conf/nginx.conf.liquid
mikz: It is not really needed, but it is useful to see the file that was included in the rendered template.

gateway/src/apicast/cli/template.lua
mikz: Yep, including files was hitting some of the limits. Maybe I should set something smaller and more reasonable. These are really ridiculously high.

gateway/src/apicast/cli/filesystem.lua
mikz: Yeah, I'll write a comment.

gateway/conf/nginx.conf.liquid

gateway/src/apicast/cli/filesystem.lua
mikz: In the end, this is not necessary. It would have to be fixed in the penlight library anyway.

gateway/src/apicast/cli/template.lua
mikz: In the end it is fine with the limits we had before.

PR #966, opened 2018-12-06
TLS Client Certificate validation policy

mikz:
Experiment with FFI and OpenSSL to validate client certificates.

- [x] package this as a customization for APIcast 3.4 (https://github.com/mikz/apicast-tls-validation-policy)
- [x] test expired certificate
- [x] test certificate use before its issue date (or not valid before)
- [x] `rover exec bin/busted spec/resty/openssl --repeat=10000 --no-auto-insulate` without memory leaks
- [x] policy unit tests
- [x] policy integration tests

In case of expired or not yet valid certificate nginx rejects the client certificate right away before applying any policies.


gateway/config/development.lua
```diff
@@ -1,21 +1,45 @@
...
-        "policy_chain": [
-          { "name": "apicast.policy.echo" }
-        ]
+        },
+        policy_chain = setmetatable({
+            { name = "apicast.policy.tls_validation",
+              configuration = {
+                  whitelist = setmetatable({
+                      { pem_certificate = [[
+-----BEGIN CERTIFICATE-----
```
mikz: This will be removed.

gateway/src/apicast/policy/tls_validation/apicast-policy.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "http://apicast.io/policy-v1/schema#manifest#",
+  "name": "tls_validation",
```
mikz: This will be improved.

mikz:
@y-tabata yes, this policy enables validation of TLS Client certificates used by clients when connecting to APIcast.

t/apicast-policy-tls_validation.t
```diff
@@ -0,0 +1,150 @@
...
+        ]
+    }
+  }]
+});
+--- test env
+proxy_ssl_certificate $TEST_NGINX_SERVER_ROOT/html/client.crt;
+proxy_ssl_certificate_key $TEST_NGINX_SERVER_ROOT/html/client.key;
+proxy_pass https://$server_addr:$apicast_port/t;
+proxy_set_header Host localhost;
+log_by_lua_block { collectgarbage() }
```
davidor: Why is this needed?

spec/policy/tls_validation/tls_validation_spec.lua
```diff
@@ -0,0 +1,75 @@
...
+    end)
+
+    it('rejects non whitelisted certificate', function()
+      ngx.var = { ssl_client_raw_cert = client }
+
+      local policy = _M.new({ whitelist = { { pem_certificate = server }}})
+
+      policy:access()
+
+      assert.stub(ngx.exit).was_called_with(400)
```
davidor: Would 403 be more appropriate?

t/apicast-policy-tls_validation.t
mikz: To run the Test::Nginx memory leak test mode. I was worried that there are some memory leaks in the OpenSSL wrapper, so I was running the memory leak checker.

spec/policy/tls_validation/tls_validation_spec.lua
mikz: I think I got the 400 from testing normal nginx certificate veritifation.

PR #973, opened 2019-01-02
fixed description in CORS policy.

palmieric:
I added further specification to the description of the field allow_origin in the CORS policy so that the users are aware that only one origin or '*' are allowed.


davidor:
Thanks @palmieric !

PR #976, opened 2019-01-08
[THREESCALE-1709] Routing policy

### LINKED ISSUE #972: New policy: routing
We need a new policy that allows us to select an upstream based on a header, the request path, a claim in a JWT, and a query argument.

We already have a policy that does this but only for the request path. The functionality will be similar.

I'll start working on this.

Ref: https://issues.jboss.org/browse/THREESCALE-1709

davidor:
Closes #972 

This PR adds a new policy, routing. It can select an upstream based on the request path, a request header, a request query argument, and a JWT claim.

I added a [README](https://github.com/3scale/apicast/pull/976/files#diff-4b14e371f001593fb9bdbc2891a5ef2c) that explains how the policy works.

Depends on #974 and #975 

Ref: https://issues.jboss.org/browse/THREESCALE-1709

### DIFF
--- a/gateway/src/apicast/policy/routing/README.md
+++ b/gateway/src/apicast/policy/routing/README.md
@@ -1,0 +1,316 @@
+# Routing policy
+
+- [**Description**](#description)
+- [**Rule that uses the request path**](#rule-that-uses-the-request-path)
+- [**Rule that uses a header**](#rule-that-uses-a-header)
+- [**Rule that uses a query argument**](#rule-that-uses-a-query-argument)
+- [**Rule that uses a jwt claim**](#rule-that-uses-a-jwt-claim)
+- [**Rule with several operations**](#rule-with-several-operations)
+- [**Combining rules**](#combining-rules)
+- [**Supported operations**](#supported-operations)
+- [**Liquid templating**](#liquid-templating)
+
+## Description
+
+This policy allows to modify the upstream (scheme, host, and port) of a request
+based on:
+
+- The request path
+- A header
+- A query argument
+- A jwt claim
+
+## Rule that uses the request path
+
+This is a configuration that routes to `http://example.com` when the path is
+`/accounts`:
+
+```json
+  {
+    "name": "routing",
+    "version": "builtin",
+    "configuration": {
+      "rules": [
+        {
+          "url": "http://example.com",
+          "condition": {
+            "operations": [
+              {
+                "thing_to_match": "path",
+                "op": "==",
+                "value": "/accounts"
+              }
+            ]
+          }
+        }
+      ]
+    }
+  }
+```
+
+## Rule that uses a header
+
+This is a configuration that routes to `http://example.com` when the value of
+the header `Test-Header` is `123`:
+
+```json
+  {
+    "name": "routing",
+    "version": "builtin",
+    "configuration": {
+      "rules": [
+        {
+          "url": "http://example.com",
+          "condition": {
+            "operations": [
+              {
+                "thing_to_match": "header",
+                "header_name": "Test-Header",
+                "op": "==",
+                "value": "123"
+              }
+            ]
+          }
+        }
+      ]
+    }
+  }
+```
+
+## Rule that uses a query argument
+
+This is a configuration that routes to `http://example.com` when the value of
+the query argument `test_query_arg` is `123`:
+
+```json
+  {
+    "name": "routing",
+    "version": "builtin",
+    "configuration": {
+      "rules": [
+        {
+          "url": "http://example.com",
+          "condition": {
+            "operations": [
+              {
+                "thing_to_match": "query_arg",
+                "query_arg_name": "test_query_arg",
+                "op": "==",
+                "value": "123"
+              }
+            ]
+          }
+        }
+      ]
+    }
+  }
+```
+
+## Rule that uses a jwt claim
+
+In order to be able to route based on the value of a jwt claim, there needs to
+be a policy in the chain that validates the jwt and stores it in the context
+that the policies share.
+
+This is a configuration that routes to `http://example.com` when the value of
+the jwt claim `test_claim` is `123`:
+
+```json
+  {
+    "name": "routing",
+    "version": "builtin",
+    "configuration": {
+      "rules": [
+        {
+          "url": "http://example.com",
+          "condition": {
+            "operations": [
+              {
+                "thing_to_match": "jwt_claim",
+                "jwt_claim_name": "test_claim",
+                "op": "==",
+                "value": "123"
+              }
+            ]
+          }
+        }
+      ]
+    }
+  }
+```
+
+## Rule with several operations
+
+Rules can have several operations and route to the given upstream only when all
+of them evaluate to true (using the 'and' `combine_op`), or when at least one of
+them evaluates to true (using the 'or' `combine_op`). The default value of
+`combine_op` is 'and'.
+
+This is a configuration that routes to `http://example.com` when the path of the
+request is `/accounts` and when the value of the header `Test-Header` is `123`:
+
+```json
+  {
+    "name": "routing",
+    "version": "builtin",
+    "configuration": {
+      "rules": [
+        {
+          "url": "http://example.com",
+          "condition": {
+            "combine_op": "and",
+            "operations": [
+              {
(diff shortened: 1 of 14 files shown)


gateway/src/apicast/policy/routing/request.lua
```diff
@@ -0,0 +1,35 @@
+local setmetatable = setmetatable
```
davidor: My idea is to use this module from other policies too.

I think we should instantiate it when the policies context is created for the current request, and store an instance of this module there. I think it would improve code organization, and also avoid reading several times the headers and the query arguments.

I'd rather do that in a different PR because it might affect many policies and need changes in some tests.

davidor:
@mikz I added support for conditions with 'or' and for liquid templating in the last two commits.

It's true that supporting those 2 things complicates a bit the config schema because it requires extra attributes. However, I think that those are 2 useful features and we should support them. I added them in 2 separate commits, so they can be easily reverted.

gateway/src/apicast/policy/routing/README.md
```diff
@@ -0,0 +1,316 @@
...
+    "configuration": {
+      "rules": [
+        {
+          "url": "http://example.com",
+          "condition": {
+            "operations": [
+              {
+                "thing_to_match": "jwt_claim",
+                "jwt_claim_name": "test_claim",
+                "op": "==",
```
mikz: We should consider array-like claims, for example, "scopes" claim that might need "includes" and "excludes" operation.

gateway/src/apicast/policy/routing/apicast-policy.json
```diff
@@ -0,0 +1,154 @@
...
+            ]
+          }
+        },
+        "required": [
+          "thing_to_match",
+          "op",
+          "value"
+        ],
+        "dependencies": {
+          "thing_to_match": {
```
mikz: That is a bit weird property name, no?

gateway/src/apicast/policy/routing/routing_operation.lua
```diff
@@ -0,0 +1,66 @@
...
+end
+
+function _M.new_op_with_path(op, value, value_type)
+  local eval_left_func = function(request) return request:get_uri() end
+  return new(eval_left_func, op, value, value_type)
+end
+
+function _M.new_op_with_header(header_name, op, value, value_type)
+  local eval_left_func = function(request)
+    return request:get_header(header_name)
```
mikz: Allocation function for each operation might be a bit resource intensive, no?
But I guess it might not be that bad when they are allocated only when they are instantiated and not on evaluation.

t/apicast-policy-routing.t
```diff
@@ -0,0 +1,2083 @@
...
+iss => 'https://example.com/auth/realms/apicast',
+exp => time + 3600 }, key => \$::rsa, alg => 'RS256', extra_headers=>{ kid => 'somekid' });
+"Authorization: Bearer $jwt"
+--- error_code: 200
+--- response_body
+yay, api backend
+--- no_error_log
+[error]
+
+=== TEST 22: the rule matches a jwt claim using "!=" and the condition is false
```
mikz: I think this integration test has way too many test cases to be readable and understandable for me.
Testing all possible options and their negations can be done in lower level tests line unit tests.

I'd rather leave this focused on high level operations, how it treats the http request, what Host header it passes, how it mangles request path etc.

gateway/src/apicast/policy/routing/apicast-policy.json
```diff
@@ -0,0 +1,154 @@
...
+      }
+    },
+    "properties": {
+      "rules": {
+        "description": "list of rules to be applied",
+        "type": "array",
+        "items": {
+          "type": "object",
+          "properties": {
+            "url": {
```
mikz: Do we need just this? We don't need to control whether it rewrites Host header?

gateway/src/apicast/policy/routing/routing_operation.lua
davidor: Yes, I don't think this has an impact on perf because it's only executed when instantiated as you said. However, in this case, extracting those functions to locals is an easy change and might make things a bit clearer, so I'll do that :+1:

gateway/src/apicast/policy/routing/apicast-policy.json
mikz: I'd suggest `property`, `match`, `key_name` (that would match 1:1 to the key name). Not really sure, naming things is hard :)

gateway/src/apicast/policy/routing/apicast-policy.json
davidor: I'm not sure. Do you think we need a way to prevent rewriting the Host header?
Currently, the host is always rewritten calling this method:
https://github.com/3scale/apicast/blob/75eb7e77d466d5ac6ba0aded9446838c87414e95/gateway/src/apicast/upstream.lua#L153

gateway/src/apicast/policy/routing/routing_operation.lua
mikz: Looks like they depend on the closure, so they'd need to accept additional parameters. And then they become polymorphic interface and can reduce the whole `new_op_with_*` into one constructor and injecting a function :)

davidor: Oh right, didn't see that it uses `header_name`. I'll leave it as it is, then.

gateway/src/apicast/policy/routing/apicast-policy.json
davidor: `match` looks OK to me. I'll change it.

t/apicast-policy-routing.t
davidor: Hmm I think that for each operation (`matches` , `==` ... ) it makes sense to test both the case where it evaluates to true and false. I see what you mean, this test is too verbose, but I think part of it is because sending a request with a JWT requires quite a few lines in the config and the request.

mikz: I think it makes sense to test both cases for each operation, but probably more in a unit test rather than integration one. Integration tests should be focused on the HTTP request itself.
I get that that unit tests might be a bit different because there is no real http request, but that is fine in my opinion. If we find out we have issues with it we try to solve that.

Maybe extracting those common parts into some shared helpers or something.

gateway/src/apicast/policy/routing/apicast-policy.json
mikz: I think we should double check with out requirements and with people using this. 
Host header can be different from the hostname where the traffic should go. 
There might be a need to supply own Host header in some cases.

gateway/src/apicast/policy/routing/apicast-policy.json
davidor: Done: https://github.com/3scale/apicast/pull/976/commits/71e16d98c5378bc3b6e99e8930264541e8799309

gateway/src/apicast/policy/routing/apicast-policy.json
davidor: I think that we can leave this as it is and add the extra parameter later if needed.
Let's check with @vramosp

t/apicast-policy-routing.t
davidor: Fair enough. I think that for this particular case it's enough to test operations when they match. If we try to match a header and it works, it means that the code was able to extract the header specified correctly, which is one of the things we want to do in these integration tests. Adding an extra test with the same information but evaluating to false does not add much value.

All the operations are covered in the unit tests of the `Condition` module.

I deleted the redundant tests following that guideline. They went from 32 to 20: https://github.com/3scale/apicast/pull/976/commits/5fa61d21adc02ca235ac74ac68ad93330c6cd443

gateway/src/apicast/policy/routing/apicast-policy.json
MarkCheshire: Is it feasible to offer the following set of options?

-  Host header overwrite by new upstream host (default)
-  Host header unchanged
-  Host header manual overwrite with a new fixed value

davidor: I created an issue for this #981

PR #982, opened 2019-01-21
Add ability to stop matching mapping rules when there's a match

### LINKED ISSUE #980: Feature: add ability to stop matching rules after first match
The current behavior is that all the mapping rules are matched. If there are several matches, APIcast reports the metrics associated with each rule that matched.

So for example, a request with path `/abc/def`, matches these 2 rules:
```
`/abc` => metric_a + 1
`/abc/def` => metric_b + 1
```

With the current behavior, APIcast reports metric_a + 1 and metric_b + 1.

The request is to be able to specify, for each rule, a new attribute that indicates whether APIcast should stop evaluating more rules if that one matches.

With that new feature, we could specify:
```
`/abc` => metric_a + 1 , last =true
`/abc/def` => metric_b + 1
```

And APIcast would only report metric_a + 1 for a request with path `/abc/def`.

Ref: https://issues.jboss.org/browse/THREESCALE-1344

davidor:
Closes #980


mikz:
Good job! It was well designed in the first place, so this was just a minimal change 👍

PR #983, opened 2019-01-21
Routing policy: allow to modify the Host header

### LINKED ISSUE #981: Feature: add ability to modify Host header in the routing policy
As discussed in https://github.com/3scale/apicast/pull/976#discussion_r247753932

davidor:
Closes #981

### DIFF
--- a/gateway/src/apicast/policy/routing/upstream_selector.lua
+++ b/gateway/src/apicast/policy/routing/upstream_selector.lua
@@ -25,7 +25,13 @@ function _M.select(_, rules, context)
     local cond_is_true = rule.condition:evaluate(context)
 
     if cond_is_true then
-      return Upstream.new(rule.url)
+      local upstream = Upstream.new(rule.url)
+
+      if rule.host_header and rule.host_header ~= '' then
+        upstream:use_host_header(rule.host_header)
+      end
+
+      return upstream
     end
   end
 
--- a/gateway/src/apicast/upstream.lua
+++ b/gateway/src/apicast/upstream.lua
@@ -144,6 +144,10 @@ local function host_header(uri)
     end
 end
 
+function _M:use_host_header(host)
+    self.host = host
+end
+
 --- Rewrite request Host header to what is provided in the argument or in the URL.
 function _M:rewrite_request()
     local uri = self.uri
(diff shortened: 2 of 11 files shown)


gateway/src/apicast/upstream.lua
```diff
@@ -144,6 +144,10 @@ local function host_header(uri)
     end
 end
 
+function _M:set_host_for_host_header(host)
```
mikz: Aren't we just using `host = ...` on other places? I mean without a method, just setting the property.

gateway/src/apicast/policy/routing/upstream_selector.lua
```diff
@@ -25,7 +25,13 @@ function _M.select(_, rules, context)
     local cond_is_true = rule.condition:evaluate(context)
 
     if cond_is_true then
-      return Upstream.new(rule.url)
+      local upstream = Upstream.new(rule.url)
+
+      if rule.host_header then
```
mikz: I think we have to cover case of an empty value.

davidor: Good point :+1:

gateway/src/apicast/upstream.lua
davidor: I was doubting between adding this new method or not.

In the end, I added it because I think that it makes the code easier to understand. Using the method it's pretty clear that I'm only using the host to set the Host header. However, with `.host = ...`, it's not that clear, the host could be used for other things.

mikz: Agreed. Maybe we could tune up the naming a bit? `use_host_header` ?
And use it on other places where `.host = `  is used right now?

davidor: Sounds good :+1:

davidor: Renamed in https://github.com/3scale/apicast/pull/983/commits/4b898cfb13ca98d94697451547e6d3e49f976b3d

Modified calls to the new method from `.host =` in https://github.com/3scale/apicast/pull/983/commits/a399634b5b3f88237f92ffc158cf8fe50518fc01

gateway/src/apicast/policy/routing/upstream_selector.lua
davidor: Done in https://github.com/3scale/apicast/pull/983/commits/50ddc3757d493d7fde93828afafbf58e5a9d2ba0

PR #986, opened 2019-01-30
Add "push_policy" command to the CLI

davidor:
This PR adds a command to push policy schemas to the 3scale admin portal.

I tagged it as WIP because we'll need confirmation from @hallelujah to make sure that the endpoints used and their parameters are correct.

### DIFF
--- a/gateway/src/apicast/policy_pusher.lua
+++ b/gateway/src/apicast/policy_pusher.lua
@@ -1,0 +1,57 @@
+local policy_manifests_loader = require('apicast.policy_manifests_loader')
+local http_ng = require('resty.http_ng')
+
+local setmetatable = setmetatable
+local format = string.format
+
+local _M = {}
+
+local mt = { __index = _M }
+
+function _M.new(http_client, manifests_loader)
+  local self = setmetatable({}, mt)
+  self.http_client = http_client or http_ng.new()
+  self.policy_manifests_loader = manifests_loader or policy_manifests_loader
+  return self
+end
+
+local system_endpoint = '/admin/api/registry/policies'
+
+local function system_url(access_token, admin_portal_domain)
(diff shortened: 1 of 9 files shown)


gateway/src/apicast/policy_pusher.lua
```diff
@@ -0,0 +1,59 @@
...
+
+local function system_url(access_token, admin_portal_domain)
+  return format('https://%s:%s%s', access_token, admin_portal_domain, system_endpoint)
+end
+
+local function push_to_system(name, version, manifest, admin_portal_domain, access_token, http_client)
+  local url = system_url(access_token, admin_portal_domain)
+
+  return http_client.post(
+    url, { name = name, version = version, schema = cjson.encode(manifest) }
```
mikz: You probably want back JSON, so you should send `Accept: application/json` and sending there JSON, so should send `Content-Type: application/json`.

mikz: Or there should be `http_ng.json.post` that does it.

davidor:
@hallelujah In this PR, I assume that there's going to be a `https://access_token:admin_portal_domain/admin/api/registry/policies` endpoint and that the POST request will accept these parameters: `name`, `version`, `schema`.

Can you please confirm that this looks OK to you whenever you have some time? Thanks!

gateway/src/apicast/policy_pusher.lua
davidor: You are right. Fixed :+1:

hallelujah:
@davidor yes, for parameters we will have:

- `name` String which corresponds to the directory name of the policy
- `version` String corresponding to the version in the schema
- `schema` the string representation of the JSON schema. We will validate it upon creation.

PR #987, opened 2019-02-11
added service id and service name headers when X-3scale-debug is used

palmieric:
Added info on the services in the response headers when using X-3scale-debug headers as asked in https://issues.jboss.org/browse/THREESCALE-1849

### DIFF
--- a/gateway/src/apicast/proxy.lua
+++ b/gateway/src/apicast/proxy.lua
@@ -85,6 +85,8 @@ local function output_debug_headers(service, usage, credentials)
     ngx.header["X-3scale-credentials"]   = credentials
     ngx.header["X-3scale-usage"]         = usage
     ngx.header["X-3scale-hostname"]      = ngx.var.hostname
+    ngx.header["X-3scale-svc-id"]        = service.id
+    ngx.header["X-3scale-svc-name"]      = service.serializable.system_name
   end
 end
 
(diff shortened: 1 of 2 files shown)


gateway/src/apicast/proxy.lua
```diff
@@ -85,6 +85,8 @@ local function output_debug_headers(service, usage, credentials)
     ngx.header["X-3scale-credentials"]   = credentials
     ngx.header["X-3scale-usage"]         = usage
     ngx.header["X-3scale-hostname"]      = ngx.var.hostname
+    ngx.header["X-3scale-svc-id"]        = service.id
```
mikz: Why abbreviating to `svc` instead of using `service`?

palmieric: Just for readability X-3scale-service-name looked a really long name.

mikz: @palmieric It is just one character more than `X-3scale-credentials`. I'd prefer `service` instead of `svc`.

palmieric: Ok will change it and resubmit

mikz: You can just force push to this PR.

And you'll need a changelog entry.

palmieric: Done

mikz: You'll still need to add a new entry to CHANGELOG.md file.

PR #991, opened 2019-03-04
clarifies description and removes references to endpoints

kevprice83:
There was some confusing language around **what** was within the scope of the policy, in this case the service object is available in the context and not the metrics/methods. Removed any path specific or endpoint specific references to avoid any confusion.

### DIFF
--- a/gateway/src/apicast/policy/rate_limit/README.md
+++ b/gateway/src/apicast/policy/rate_limit/README.md
@@ -15,8 +15,8 @@ The purpose of this policy is to control the rate of traffic sent to the
 upstream API. There are several kinds of limits that can be defined. Here are
 some examples:
 
-- The client with address 127.0.0.1 can access the specific endpoint
-`/something` up to 100 times per minute.
+- The client with address 127.0.0.1 can access the specific API
+`http://example.com` up to 100 times per minute.
 - Requests that contain a JSON web token (JWT) whose `sub` claim has the value
 `abc` can make up to 10 calls to the API per second.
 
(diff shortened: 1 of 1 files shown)


kevprice83:
@davidor @mikz let me know if the changes made still makes sense. If something is still not clear or if you suggest any further improvements.

mikz:
@kevprice83 it looks like the documentation is missing a section about conditionals, but it is in the UI.

It is possible to configure a rate limit just for one endpoint like `GET /endpoint`. It has no connection to 3scale mapping rules, but it still can be configured.

gateway/src/apicast/policy/rate_limit/README.md
```diff
@@ -15,8 +15,8 @@ The purpose of this policy is to control the rate of traffic sent to the
 upstream API. There are several kinds of limits that can be defined. Here are
 some examples:
 
-- The client with address 127.0.0.1 can access the specific endpoint
-`/something` up to 100 times per minute.
+- The client with address 127.0.0.1 can access the specific API
```
davidor: I think that both things should be possible using this policy. You can limit a whole API/service but you can also limit specific endpoints/paths as it said in the deleted part of the diff.

kevprice83:
Right so I guess we can do something like `{% if uri == '/foo' %} apply some rate limiting {% endif %}`? The conditional policy itself is not available in the UI but maybe this can still be used to perform the same?

mikz:
@kevprice83 no, those conditions are expressed as in the condition policy (not as liquid), but are embedded here:
https://github.com/3scale/apicast/blob/59e19be9b23108ec2b46a6396d8b7855ccfc22f2/gateway/src/apicast/policy/rate_limit/apicast-policy.json#L105-L124
https://github.com/3scale/apicast/blob/59e19be9b23108ec2b46a6396d8b7855ccfc22f2/t/apicast-policy-rate-limit.t#L1215-L1283

kevprice83:
@mikz that's clear now thanks. I'll see if I can update the doc with the relevant section that's missing plus an example.

kevprice83:
This PR has now been superseded by #1040 so closing

PR #1006, opened 2019-04-03
[nginx] allow configuring maximum client certificate chain length

mikz:
some clients need to send the whole chain (certificate, intermediate, root),
and nginx needs to accept that whole chain to pass it to policies

Also fixes a segfault when parsing some certificates (intermediate CA, EC). Can be reproduced with the certificates committed in the repo.

# TODO

- [x] integration test using the full chain


doc/parameters.md
```diff
@@ -155,6 +155,14 @@ Path to a file with passphrases for the SSL cert keys specified with
 - `on`: reuses SSL sessions.
 - `off`: does not reuse SSL sessions.
 
+### `APICAST_PROXY_HTTPS_VERIFY_DEPTH`
+
+**Default:** 1
+**Values:** positive integers
+
+Defines the maximum length of the client certificate chain.
+Length 1 means there can be 1 extra certificate included (eg. intermediate CA).
```
porueesq: Suggestion:
_If this parameter has `1` as its value, it implies that this length might include one additional certificate  (eg. intermediate CA)._

mikz: 👍

gateway/src/resty/openssl/x509.lua
```diff
@@ -55,7 +55,8 @@ local function normalize_pem_cert(str)
   if not str then return end
   if #(str) == 0 then return end
 
-  return re_gsub(str, [[\s(?!CERTIFICATE)]], '\n', 'oj')
+  -- using also jit compiler (j) will result in a segfault with some certificates
```
davidor: Do you know which certificates cause the problem?

t/fixtures/CA/Makefile
```diff
@@ -0,0 +1,29 @@
+all: client.key client.crt server.key server.crt ca-bundle.crt:
```
davidor: This Makefile is going to be super useful :+1:

gateway/src/resty/openssl/x509.lua
mikz: The updated ones in this PR: t/fixtures/CA.  But only when they come from nginx, couldn't reproduce it in busted.

t/fixtures/CA/Makefile
mikz: Yeah, got tired of figuring out how to generate them every time 😄

doc/parameters.md
```diff
@@ -155,6 +155,14 @@ Path to a file with passphrases for the SSL cert keys specified with
 - `on`: reuses SSL sessions.
 - `off`: does not reuse SSL sessions.
 
+### `APICAST_PROXY_HTTPS_VERIFY_DEPTH`
```
mikz: This is wrong. Fixed it in https://github.com/3scale/APIcast/pull/1048

Issue #988, opened 2019-02-17
OpenID/OAuth token validation flow in APICast is incorrect


magnusvage:
I believe that the implemented steps to validate an OpenID/OAuth token are incorrect

Looking at the source code found at https://github.com/3scale/apicast/blob/ab4ce9d08759c657944078ba8dbf1216c533c9da/gateway/src/apicast/oauth/oidc.lua

It explicitly references https://tools.ietf.org/html/rfc7523#section-3, which are the steps of how an AUTHORIZATION server needs to validate a token. However, APICast is not the authorization server, it is the resource server (Or more correctly, acting as a front for several resource servers) if looking at https://tools.ietf.org/html/rfc6749#section-1.1

resource server
      The server hosting the protected resources, capable of accepting
      and responding to protected resource requests using access tokens.

As of now, the aud or azp is taken to be the application client_id (from the source code above):

local app_id = payload.azp or payload.aud

However, APICast should check to opposite: That aud is an identifer of APICast (or even more preferrably and granular, an identifier of the service that is being called).

As described in https://tools.ietf.org/html/rfc7519#section-4.1.3

The "aud" (audience) claim identifies the recipients that the JWT is
   intended for.  Each principal intended to process the JWT MUST
   identify itself with a value in the audience claim.  If the principal
   processing the claim does not identify itself with a value in the
   "aud" claim when this claim is present, then the JWT MUST be
   rejected.

This would require that all services are represented as clients in Keycloak (or any auth server), and that application clients are given access to service clients in the auth server. This is so that when the application requests a token from the auth server (be it a client credentials or authorization code flow), the application can set the scope to the service in the auth request, explicitly requesting access to the service, which is how it was all intended to work.

I am not a super expert with all things OpenID/OAuth, so if I've gotten something terribly wrong and it all works as intended, please feel free to close this.

magnusvage:
To flesh out my reasoning above, I created a scenario.

Two clients exist in Keycloak, representing an application and a service in 3scale. demo-application has no roles/access to demo-service at this point.

<img width="381" alt="screenshot 2019-02-18 at 00 08 44" src="https://user-images.githubusercontent.com/36159342/52920847-72809200-3311-11e9-9f92-986bb164beff.png">

As demo-application, requesting an access token for demo-service does not give me demo-service as aud, since I don't have any access:

Postman token request:
[...]

PR #1007, opened 2019-04-03
[oidc] do not use`aud` JWT claim

mikz:
it might not contain client_id, so better not to use it
https://github.com/3scale/APIcast/issues/988#issuecomment-477972776


gateway/src/apicast/policy/3scale_batcher/reports_batcher.lua
```diff
@@ -91,7 +91,7 @@ function _M:get_all(service_id)
 
     local report = keys_helper.report_from_key_batched_report(key, value)
 
-    if value and value > 0 and report.service_id == service_id then
+    if value and value > 0 and report and report.service_id == service_id then
```
davidor: I think that the failing test was because of this: https://github.com/3scale/APIcast/blob/e825115209d53cf6c564e0aad889412eb5e758d4/t/apicast-policy-3scale-batcher.t#L613 We should change that to "azp".

mikz: Yes, but still `report_from_key_batched_report` can return `nil` and this will crash. That should not happen.

davidor: Yes, that too. I think that apart from that, we should add a warning in `report_from_key_batched_report`, because if it returns nil, it means that the policy received a combination of "auth params" (app_id, app_key, ...) that it's not supported.

mikz: That probably deserves its own PR with a test. Will create one.

mikz: https://github.com/3scale/APIcast/pull/1009

PR #1014, opened 2019-04-09
Rename and clarify get_auth_params() method in the Service module

davidor:
Simple PR that just renames and documents the `get_auth_params()` method in the `Service` module.

The name of the method was confusing. It implied that it had something to do with the auth params (app_id, app_key, etc.), but it does not. This caused some confusion, see https://issues.jboss.org/browse/THREESCALE-1930 and I think it's not the first time that it happens.

### DIFF
--- a/gateway/src/apicast/configuration/service.lua
+++ b/gateway/src/apicast/configuration/service.lua
@@ -162,7 +162,14 @@ function backend_version_credentials.version_oauth(config)
   return setmetatable({ access_token, access_token = access_token }, credentials_oauth_mt)
 end
 
-local function get_auth_params(method)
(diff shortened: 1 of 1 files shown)


mikz:
We should at some point make reading the body optional and even POST should just read the query params.

davidor:
I agree, but that means breaking users who rely on this behavior.
For now, I just wanted to clarify what the method does to avoid confusion.

davidor:
For reference, I ran some benchmarks and summarized the results [here](https://issues.jboss.org/browse/THREESCALE-1930?focusedCommentId=13719594&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13719594)

PR #1023, opened 2019-04-29
Local chain: forward correct context in all the request phases

davidor:
It was done correctly only for `rewrite()`. In the rest of the phases, the context received was not merged with the one coming from `policy_chain:export()`.

This is the same bug that we fixed in https://github.com/3scale/APIcast/pull/673 . However, we only fixed it for the rewrite phase.

I discovered this while working on [this branch](https://github.com/3scale/APIcast/tree/upstream-config-policy). I needed to export some data to the shared context and access it in the balancer phase and I realized that it was not there.

### DIFF
--- a/gateway/src/apicast/policy/local_chain/local_chain.lua
+++ b/gateway/src/apicast/policy/local_chain/local_chain.lua
@@ -41,19 +41,22 @@ for _, phase in policy.phases() do
     local policy_chain = find_policy_chain(context)
 
     if policy_chain then
-      return policy_chain[phase](policy_chain, context, ...)
+      local merged_context = LinkedList.readwrite(context, policy_chain:export())
+      return policy_chain[phase](policy_chain, merged_context, ...)
     end
   end
 end
 
 local rewrite = _M.rewrite
 
+-- Similar to the methods defined just above. The only difference is that this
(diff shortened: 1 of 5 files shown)


gateway/src/apicast/policy/local_chain/local_chain.lua
```diff
@@ -38,22 +38,19 @@ end
 -- forward all policy methods to the policy chain
 for _, phase in policy.phases() do
   _M[phase] = function(_, context, ...)
-    local policy_chain = find_policy_chain(context)
+    local policy_chain
+    if phase == 'rewrite' then
```
mikz: This will break JIT. It will be better to generate different function for the rewrite phase.

davidor: Right. Fixed :+1:

davidor:
@mikz unfortunately, the bug solved in this PR was hiding another one in the 3scale batcher policy.

The batcher policy was not working correctly when combined with the caching one. The caching had no effect because the cache handler was stored in the context and used from the `access` phase. I added a couple of commits that fix the issue. The comment in the commit that touches the integration tests explains why the test did not catch the bug.

PR #1024, opened 2019-04-30
Add service label on metrics

eloycoto:
This commits adds support to have additional information on some metrics[0],
this commit enables users to know the service affected by the metric. 

Due to this can be a problematic for Prometheus[1], to enable this a new ENV
variable(`APICAST_EXTENDED_METRICS`) is defined, so if users that have a few services
they can run the extended metrics without affecting Prometheus stability, if
they thousands of services the recommendation is to have this option disabled. 

In case of the extended_metrics is disabled, all the service labels will use
"all" value and Prometheus performance will not be affected.

The new metrics list looks like this:

```
bash-4.2$ curl http://localhost:9421/metrics -s | grep service
total_response_time_seconds_bucket{service="all",le="00.100"} 91
total_response_time_seconds_bucket{service="all",le="00.200"} 189
total_response_time_seconds_bucket{service="all",le="00.300"} 220
total_response_time_seconds_bucket{service="all",le="00.400"} 220
total_response_time_seconds_bucket{service="all",le="00.500"} 222
total_response_time_seconds_bucket{service="all",le="00.750"} 223
total_response_time_seconds_bucket{service="all",le="01.000"} 223
total_response_time_seconds_bucket{service="all",le="01.500"} 224
total_response_time_seconds_bucket{service="all",le="02.000"} 224
total_response_time_seconds_bucket{service="all",le="03.000"} 224
total_response_time_seconds_bucket{service="all",le="04.000"} 224
total_response_time_seconds_bucket{service="all",le="05.000"} 224
total_response_time_seconds_bucket{service="all",le="10.000"} 224
total_response_time_seconds_bucket{service="all",le="+Inf"} 224
total_response_time_seconds_count{service="all"} 224
total_response_time_seconds_sum{service="all"} 33.616
upstream_response_time_seconds_bucket{service="all",le="00.100"} 93
upstream_response_time_seconds_bucket{service="all",le="00.200"} 190
upstream_response_time_seconds_bucket{service="all",le="00.300"} 220
upstream_response_time_seconds_bucket{service="all",le="00.400"} 220
upstream_response_time_seconds_bucket{service="all",le="00.500"} 223
upstream_response_time_seconds_bucket{service="all",le="00.750"} 224
upstream_response_time_seconds_bucket{service="all",le="01.000"} 224
upstream_response_time_seconds_bucket{service="all",le="01.500"} 224
upstream_response_time_seconds_bucket{service="all",le="02.000"} 224
upstream_response_time_seconds_bucket{service="all",le="03.000"} 224
upstream_response_time_seconds_bucket{service="all",le="04.000"} 224
upstream_response_time_seconds_bucket{service="all",le="05.000"} 224
upstream_response_time_seconds_bucket{service="all",le="10.000"} 224
upstream_response_time_seconds_bucket{service="all",le="+Inf"} 224
upstream_response_time_seconds_count{service="all"} 224
upstream_response_time_seconds_sum{service="all"} 32.226
upstream_status{status="200",service="all"} 224
```

[0] List of affected metrics:
    total_response_time_seconds
    upstream_response_time_seconds
    upstream_status

[1] https://prometheus.io/docs/practices/naming/#labels

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - You can filter services by endpoint name using Regexp [PR #1022](https://github.com/3scale/APIcast/pull/1022) [THREESCALE-1524](https://issues.jboss.org/browse/THREESCALE-1524)
 - "Upstream Connection" policy. It allows to configure several options for the connections to the upstream [PR #1025](https://github.com/3scale/APIcast/pull/1025), [THREESCALE-2166](https://issues.jboss.org/browse/THREESCALE-2166)
 - Enable APICAST_EXTENDED_METRICS env variable to provide more in-depth details [PR #1024](https://github.com/3scale/APIcast/pull/1024) [THREESCALE-2150](https://issues.jboss.org/browse/THREESCALE-2150)
+- Enable APICAST_EXTENDED_METRICS environment variable to provide additional details [PR #1024](https://github.com/3scale/APIcast/pull/1024) [THREESCALE-2150](https://issues.jboss.org/browse/THREESCALE-2150)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - You can filter services by endpoint name using Regexp [PR #1022](https://github.com/3scale/APIcast/pull/1022) [THREESCALE-1524](https://issues.jboss.org/browse/THREESCALE-1524)
 - "Upstream Connection" policy. It allows to configure several options for the connections to the upstream [PR #1025](https://github.com/3scale/APIcast/pull/1025), [THREESCALE-2166](https://issues.jboss.org/browse/THREESCALE-2166)
+- Enable APICAST_EXTENDED_METRICS env variable to provide more in-depth details [PR #1024](https://github.com/3scale/APIcast/pull/1024) [THREESCALE-2150](https://issues.jboss.org/browse/THREESCALE-2150)
 
 ### Fixed
 
--- a/doc/parameters.md
+++ b/doc/parameters.md
@@ -386,6 +386,6 @@ with specific information that will provide more in-depth details about APIcast.
 
 The metrics that will have extended information are:
 
-- total_response_time_seconds: label service
-- upstream_response_time_seconds: label service
-- upstream_status: label service
+- total_response_time_seconds: labels service_id and service_system_name
+- upstream_response_time_seconds: labels service_id and service_system_name
+- upstream_status: labels service_id and service_system_name
--- a/doc/parameters.md
+++ b/doc/parameters.md
@@ -374,3 +374,18 @@ Defines a HTTP proxy to be used for connecting to HTTPS services. Authentication
 **Example:** `foo,bar.com,.extra.dot.com`
 
 Defines a comma-separated list of hostnames and domain names for which the requests should not be proxied. Setting to a single `*` character, which matches all hosts, effectively disables the proxy.
+
+### `APICAST_EXTENDED_METRICS`
+
+**Default:** false
+**Value:** boolean
+**Example:** "true"
+
+Enables additional information on Prometheus metrics; some labels will be used
+with specific information that will provide more in-depth details about APIcast.
+
+The metrics that will have extended information are:
+
+- total_response_time_seconds: label service
+- upstream_response_time_seconds: label service
+- upstream_status: label service
--- a/gateway/src/apicast/metrics/upstream.lua
+++ b/gateway/src/apicast/metrics/upstream.lua
@@ -4,36 +4,40 @@ local prometheus = require('apicast.prometheus')
 
 local _M = {}
 
+local service_label = 'service'
+local status_label = 'status'
+
 local upstream_status_codes = prometheus(
   'counter',
   'upstream_status',
   'HTTP status from upstream servers',
-  { 'status' }
+  { status_label, service_label }
 )
 
 local upstream_resp_times = prometheus(
   'histogram',
   'upstream_response_time_seconds',
-  'Response times from upstream servers'
+  'Response times from upstream servers',
+  { service_label }
 )
 
-local function inc_status_codes_counter(status)
+local function inc_status_codes_counter(status, service)
   if tonumber(status) and upstream_status_codes then
-    upstream_status_codes:inc(1, { status })
+    upstream_status_codes:inc(1, { status, service })
   end
 end
 
-local function add_resp_time(response_time)
+local function add_resp_time(response_time, service)
   local time = tonumber(response_time)
 
   if time and upstream_resp_times then
-    upstream_resp_times:observe(time)
+    upstream_resp_times:observe(time, { service })
   end
 end
 
-function _M.report(status, response_time)
-  inc_status_codes_counter(status)
-  add_resp_time(response_time)
+function _M.report(status, response_time, service)
+  inc_status_codes_counter(status, service)
+  add_resp_time(response_time, service)
 end
 
 return _M
--- a/gateway/src/apicast/metrics/upstream.lua
+++ b/gateway/src/apicast/metrics/upstream.lua
@@ -4,26 +4,31 @@ local prometheus = require('apicast.prometheus')
 
 local _M = {}
 
-local service_label = 'service'
+local service_label = 'service_id'
+local service_system_name_label = 'service_system_name'
 local status_label = 'status'
 
 local upstream_status_codes = prometheus(
   'counter',
   'upstream_status',
   'HTTP status from upstream servers',
-  { status_label, service_label }
+  { status_label, service_label, service_system_name_label }
 )
 
 local upstream_resp_times = prometheus(
   'histogram',
   'upstream_response_time_seconds',
   'Response times from upstream servers',
-  { service_label }
+  { service_label, service_system_name_label }
 )
 
 local function inc_status_codes_counter(status, service)
   if tonumber(status) and upstream_status_codes then
-    upstream_status_codes:inc(1, { status, service })
+    upstream_status_codes:inc(1, {
+      status,
+      service.id or "",
+      service.system_name or ""
+    })
   end
 end
 
@@ -31,11 +36,17 @@ local function add_resp_time(response_time, service)
   local time = tonumber(response_time)
 
   if time and upstream_resp_times then
-    upstream_resp_times:observe(time, { service })
+    upstream_resp_times:observe(time, {
+      service.id or "",
+      service.system_name or ""
+    })
   end
 end
 
 function _M.report(status, response_time, service)
+  if not service then
+    service = {}
+  end
   inc_status_codes_counter(status, service)
   add_resp_time(response_time, service)
 end
--- a/gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
+++ b/gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
@@ -9,6 +9,14 @@ local select = select
 local find = string.find
 local pairs = pairs
 
+-- extended_metrics is a variable used to report multiple labels in some
+-- metrics. This can be useful in small environements, but can be problematic
+-- for large users due this can create a large matrix of metrics.
+-- More info about this can be found in Prometheus doc:
+-- https://prometheus.io/docs/practices/naming/#labels
+
+local extended_metrics = resty_env.enabled('APICAST_EXTENDED_METRICS')
+
 local upstream_metrics = require('apicast.metrics.upstream')
 
 local new = _M.new
@@ -76,7 +84,8 @@ local shdict_free_space_metric = prometheus('gauge', 'openresty_shdict_free_spac
 local response_times = prometheus(
   'histogram',
   'total_response_time_seconds',
-  'Time needed to sent a response to the client (in seconds).'
+  'Time needed to sent a response to the client (in seconds).',
+  { 'service' }
 )
 
 function _M.init()
@@ -118,19 +127,22 @@ function _M:metrics()
   end
 end
 
-local function report_req_response_time()
+local function report_req_response_time(service)
   -- Use ngx.var.original_request_time instead of ngx.var.request_time so
   -- the time spent in the post_action phase is not taken into account.
   local resp_time = tonumber(ngx.var.original_request_time)
-
   if resp_time and response_times then
-    response_times:observe(resp_time)
+    response_times:observe(resp_time, { service })
   end
 end
 
-function _M.log()
-  upstream_metrics.report(ngx.var.upstream_status, ngx.var.upstream_response_time)
-  report_req_response_time()
+function _M.log(_, context)
+  local service = ""
+  if context.service and context.service.id and extended_metrics then
+    service = context.service.id
+  end
+  upstream_metrics.report(ngx.var.upstream_status, ngx.var.upstream_response_time, service)
+  report_req_response_time(service)
 end
 
 return _M
--- a/gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
+++ b/gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
@@ -84,7 +84,7 @@ local shdict_free_space_metric = prometheus('gauge', 'openresty_shdict_free_spac
 local response_times = prometheus(
   'histogram',
   'total_response_time_seconds',
-  'Time needed to sent a response to the client (in seconds).',
+  'Time needed to send a response to the client (in seconds).',
   { 'service' }
 )
 
--- a/gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
+++ b/gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
@@ -85,7 +85,7 @@ local response_times = prometheus(
   'histogram',
   'total_response_time_seconds',
   'Time needed to send a response to the client (in seconds).',
-  { 'service' }
+  { 'service_id', 'service_system_name' }
 )
 
 function _M.init()
@@ -132,14 +132,17 @@ local function report_req_response_time(service)
   -- the time spent in the post_action phase is not taken into account.
   local resp_time = tonumber(ngx.var.original_request_time)
   if resp_time and response_times then
-    response_times:observe(resp_time, { service })
+    response_times:observe(resp_time, {
+      service.id or "",
+      service.system_name or ""
+    })
   end
 end
 
 function _M.log(_, context)
-  local service = ""
-  if context.service and context.service.id and extended_metrics then
-    service = context.service.id
+  local service = { id = "", system_name = "" }
+  if context.service and extended_metrics then
+    service = context.service
   end
   upstream_metrics.report(ngx.var.upstream_status, ngx.var.upstream_response_time, service)
   report_req_response_time(service)
--- a/spec/metrics/upstream_spec.lua
+++ b/spec/metrics/upstream_spec.lua
@@ -32,12 +32,12 @@ describe('upstream metrics', function()
 
     it('increases the counter of status codes', function()
       upstream_metrics.report(200, 0.1)
-      assert.stub(test_counter.inc).was_called_with(test_counter, 1, { 200 })
+      assert.stub(test_counter.inc).was_called_with(test_counter, 1, { 200, "", "" })
     end)
 
     it('adds the latency to the histogram', function()
       upstream_metrics.report(200, 0.1)
-      assert.stub(test_histogram.observe).was_called_with(test_histogram, 0.1, {})
+      assert.stub(test_histogram.observe).was_called_with(test_histogram, 0.1, {"", ""})
     end)
 
     describe('when the status is nil or empty', function()
@@ -58,17 +58,17 @@ describe('upstream metrics', function()
 
 
     describe('With service id', function()
-      local service_metric_id = "42"
+      local service = {id = "123", system_name="foo"}
       it("increases empty service on nil", function()
         upstream_metrics.report(200, 0.1, nil)
-        assert.stub(test_histogram.observe).was_called_with(test_histogram, 0.1, {})
-        assert.stub(test_counter.inc).was_called_with(test_counter, 1, { 200 })
+        assert.stub(test_histogram.observe).was_called_with(test_histogram, 0.1, {"", ""})
+        assert.stub(test_counter.inc).was_called_with(test_counter, 1, { 200, "", "" })
       end)
 
       it("increase a valid service", function()
-        upstream_metrics.report(200, 0.1, service_metric_id)
-        assert.stub(test_histogram.observe).was_called_with(test_histogram, 0.1, { service_metric_id })
-        assert.stub(test_counter.inc).was_called_with(test_counter, 1, { 200, service_metric_id })
+        upstream_metrics.report(200, 0.1, service)
+        assert.stub(test_histogram.observe).was_called_with(test_histogram, 0.1, { service.id, service.system_name })
+        assert.stub(test_counter.inc).was_called_with(test_counter, 1, { 200, service.id, service.system_name })
       end)
     end)
   end)
--- a/spec/metrics/upstream_spec.lua
+++ b/spec/metrics/upstream_spec.lua
@@ -37,7 +37,7 @@ describe('upstream metrics', function()
 
     it('adds the latency to the histogram', function()
       upstream_metrics.report(200, 0.1)
-      assert.stub(test_histogram.observe).was_called_with(test_histogram, 0.1)
+      assert.stub(test_histogram.observe).was_called_with(test_histogram, 0.1, {})
     end)
 
     describe('when the status is nil or empty', function()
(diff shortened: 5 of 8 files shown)


gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
```diff
@@ -118,19 +126,22 @@ function _M:metrics()
...
   end
 end
 
-function _M.log()
-  upstream_metrics.report(ngx.var.upstream_status, ngx.var.upstream_response_time)
-  report_req_response_time()
+function _M.log(_, context)
+  local service = context.service.id
+  if not service or not extended_metrics then
+    service = "all"
```
mikz: Why `all` and not just `nil` ?

gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
```diff
@@ -9,6 +9,13 @@ local select = select
 local find = string.find
 local pairs = pairs
 
+-- extended_metrics is a variable used to report multiple labels in some
+-- metrics. This can be useful in small environements, but can be problematic
+-- for large users due this can create a large matrix of metrics.
+-- More info about this can be found in Prometheus doc:
+-- https://prometheus.io/docs/practices/naming/#labels
+local extended_metrics = resty_env.get('APICAST_EXTENDED_METRICS')
```
mikz: There is `resty_env.enabled` that will do automatic conversion of `true/false/1/0/...`. https://github.com/3scale/lua-resty-env#enabled

gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
```diff
@@ -118,19 +126,22 @@ function _M:metrics()
...
-function _M.log()
-  upstream_metrics.report(ngx.var.upstream_status, ngx.var.upstream_response_time)
-  report_req_response_time()
+function _M.log(_, context)
+  local service = context.service.id
+  if not service or not extended_metrics then
+    service = "all"
+  end
+  upstream_metrics.report(ngx.var.upstream_status, ngx.var.upstream_response_time, service)
+  report_req_response_time(service)
```
mikz: I know this is unrelated to this PR. But we probably should also introduce a counter of how many requests the gateway processed (and have a label for each service). So prometheus can detect requests per minute for example.

t/prometheus-metrics.t
```diff
@@ -275,7 +275,137 @@ qr/upstream_response_time_seconds(.|\n)*upstream_response_time_seconds_bucket\{l
...
+--- request eval
+["GET /", "GET /metrics"]
+--- more_headers eval
+["", "Host: metrics"]
+--- error_code eval
+[ 200, 200 ]
+--- response_body_like eval
+[
+"",
+qr/total_response_time_seconds(.|\n)*total_response_time_seconds_bucket\{service="42",le=".*"\} 1/,
```
mikz: If we do more of these we could refactor this into some helper method.

So we could say just `$prometheus->match_metrics('total_response_time_seconds', 'total_response_time_seconds_bucket')` and that helper method would generate the regexp.

eloycoto: I would like to defer this one for a while, my head is already exploding with Lua :-D

mikz: I can imagine :-D For the record, there are a few helpers already:
https://github.com/3scale/APIcast/blob/master/t/policies.pl
https://github.com/3scale/APIcast/blob/master/t/fixtures/mutual_ssl.pl
https://github.com/3scale/APIcast/blob/master/t/fixtures/tls.pl

gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
eloycoto: If I send nil the following error message appears

```
2019/04/30 07:18:03 [error] 149#149: *3 [lua] prometheus.lua:295: log_error(): Wrong number of labels for upstream_status. Expected 2, got 1 while logging request, client: 127.0.0.1, server: _, request: "GET /some-path?user_key=4e5944f11f575fd50bd6624c8d441f13 HTTP/1.1", upstream: "https://18.213.6.196:443/some-path?user_key=4e5944f11f575fd50bd6624c8d441f13", host: "echo-api.3scale.net"
```

mikz: I see. Might be good to leave there a comment then, so we don't have to ask the question when we revisit that code.

mikz: And an empty string? I'm not sure what is better honestly.

eloycoto: An empty string can be confused, so all have explicit information.

doc/parameters.md
```diff
@@ -351,3 +351,23 @@ Defines a HTTP proxy to be used for connecting to HTTPS services. Authentication
...
+### `APICAST_EXTENDED_METRICS`
+
+**Default:** nil
+**Value:** boolean
+**Example:** "true"
+
+Enables additional information on Prometheus metrics; some labels will be used
+with specific information that will provide more in-depth details about APIcast.
+
+Note: this parameter will create different labels for metrics and can cause
```
davidor: I think we'll need to measure this to be sure about the performance impact, for example when having 10 or 100 different services.

Maybe we should delete this warning until we measure this and know approximately how many services there needs to be to have a noticeable impact on performance.

doc/parameters.md
```diff
@@ -351,3 +351,23 @@ Defines a HTTP proxy to be used for connecting to HTTPS services. Authentication
 **Example:** `foo,bar.com,.extra.dot.com`
 
 Defines a comma-separated list of hostnames and domain names for which the requests should not be proxied. Setting to a single `*` character, which matches all hosts, effectively disables the proxy.
+
+### `APICAST_EXTENDED_METRICS`
+
+**Default:** nil
```
davidor: Shouldn't it be false?

spec/metrics/upstream_spec.lua
```diff
@@ -36,8 +37,8 @@ describe('upstream metrics', function()
     end)
 
     it('adds the latency to the histogram', function()
-      upstream_metrics.report(200, 0.1)
-      assert.stub(test_histogram.observe).was_called_with(test_histogram, 0.1)
+      upstream_metrics.report(200, 0.1, service_metric_name)
```
davidor: We should probably test with and without service. To make sure that both test cases work correctly.

Also, we should do the same for the status codes counter.

spec/metrics/upstream_spec.lua
```diff
@@ -1,6 +1,7 @@
 describe('upstream metrics', function()
   describe('report', function()
     local upstream_metrics
+    local service_metric_name = "42"
```
davidor: This should be an ID instead of a name, right?

gateway/src/apicast/metrics/upstream.lua
```diff
@@ -4,36 +4,39 @@ local prometheus = require('apicast.prometheus')
...
 local _M = {}
 
+local service_label = 'service'
+
 local upstream_status_codes = prometheus(
   'counter',
   'upstream_status',
   'HTTP status from upstream servers',
-  { 'status' }
+  { 'status', service_label }
```
davidor: Should we also extract `status_label` to be consistent?

t/prometheus-metrics.t
```diff
@@ -275,7 +275,137 @@ qr/upstream_response_time_seconds(.|\n)*upstream_response_time_seconds_bucket\{l
...
+  ]
+}
+--- upstream
+  location / {
+     content_by_lua_block {
+       ngx.exit(200);
+     }
+  }
+
+  location /two {
```
davidor: We can delete this one. It's not used in this test.

davidor:
@eloycoto the new metrics should be added to https://github.com/3scale/APIcast/blob/master/doc/prometheus-metrics.md

t/prometheus-metrics.t
```diff
@@ -275,7 +275,137 @@ qr/upstream_response_time_seconds(.|\n)*upstream_response_time_seconds_bucket\{l
...
+["GET /", "GET /metrics"]
+--- more_headers eval
+["", "Host: metrics"]
+--- error_code eval
+[ 200, 200 ]
+--- response_body_like eval
+[
+"",
+qr/total_response_time_seconds(.|\n)*total_response_time_seconds_bucket\{service="42",le=".*"\} 1/,
+qr/upstream_response_time_seconds(.|\n)*upstream_response_time_seconds_bucket\{service="42",le=".*"\} 1(.|\n)*upstream_status\{status="42",service="all"\} 1/,
```
davidor: I see `status="42"`. That's not correct right? Shouldn't it be `2xx`?

t/prometheus-metrics.t
```diff
@@ -275,7 +275,137 @@ qr/upstream_response_time_seconds(.|\n)*upstream_response_time_seconds_bucket\{l
...
+  }
+
+  location /two {
+     content_by_lua_block {
+       ngx.exit(200);
+     }
+  }
+--- request eval
+["GET /", "GET /metrics"]
+--- more_headers eval
```
davidor: I'm not sure about this. I think that you should send the Host header to control which one of the 2 services defined in the configuration will receive the request.

Also, I think this test would be better if we made a request to each service. Then, we could check that each one of them received a request and also verify that the total is 2.

eloycoto:
I think that I found the issue with the test[0]. 

On the executor, where the phase `log` happens, the ngx.ctx.context is nil, so
it creates a new context and the old one is not updated.

https://github.com/3scale/APIcast/blob/66168b7a15655f0ea473d862a41ae9fdbb3fa912/gateway/src/apicast/executor.lua#L40-L51

The main problem on log phase, if the batch policy or cache policy is in place
the context.service is not defined, so the Prometheus metrics raise an error
that the service.id cannot be found.

I debug all the code, and the issue is fixed if we add in this the following
code `require('resty.ctx').apply()`, if we add this the code is working
correctly. 

https://github.com/3scale/APIcast/blob/66168b7a15655f0ea473d862a41ae9fdbb3fa912/gateway/conf.d/apicast.conf#L68-L71

I think that it's safe enough, but I want to make sure that I do not introduce
a new issue. Also IMHO we should trigger a error message on
`executor:shared_build_context` that if the context does not exists maybe we
need to notice. (But I do not have the full overview of the project)

[0] https://github.com/3scale/APIcast/blob/master/t/apicast-policy-3scale-batcher.t#L403-L483

eloycoto:
Added the patch on log phase in apicast.conf, now a different test is failing:

https://github.com/3scale/APIcast/blob/66168b7a15655f0ea473d862a41ae9fdbb3fa912/t/apicast.t#L621-L631

Should a request to invalid service reach `log` phase? Should be denied in the auth phase? @davidor any insight here?

davidor:
@eloycoto The log phase will run even when the `find_service` policy cannot find a service for the request. In that case `context.service` is going to be nil. That explains why the test `TEST 17: invalid service` that you pasted above fails.

We'll need to take that case into account when updating the Prometheus metrics.

eloycoto:
@davidor what about this commit? Is that correct?
https://github.com/3scale/APIcast/pull/1024/commits/190962e6449afa53418e9718cc0a4bd58ea5e05b

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Ability to configure client certificate chain depth [PR #1006](https://github.com/3scale/APIcast/pull/1006)
 - You can filter services by endpoint name using Regexp [PR #1022](https://github.com/3scale/APIcast/pull/1022) [THREESCALE-1524](https://issues.jboss.org/browse/THREESCALE-1524)
+- Enable APICAST_EXTENDED_METRICS env variable to provide more in-depth details [PR #1024](https://github.com/3scale/APIcast/pull/1024) [THREESCALE-2150](https://issues.jboss.org/browse/THREESCALE-2150)
```
```suggestion
- Enable APICAST_EXTENDED_METRICS environment variable to provide additional details [PR #1024](https://github.com/3scale/APIcast/pull/1024) [THREESCALE-2150](https://issues.jboss.org/browse/THREESCALE-2150)
```

gateway/src/apicast/policy/nginx_metrics/nginx_metrics.lua
```diff
@@ -76,7 +83,8 @@ local shdict_free_space_metric = prometheus('gauge', 'openresty_shdict_free_spac
 local response_times = prometheus(
   'histogram',
   'total_response_time_seconds',
-  'Time needed to sent a response to the client (in seconds).'
+  'Time needed to sent a response to the client (in seconds).',
```
porueesq:
```suggestion
  'Time needed to send a response to the client (in seconds).',
```

davidor:
> @davidor what about this commit? Is that correct?
> [190962e](https://github.com/3scale/APIcast/commit/190962e6449afa53418e9718cc0a4bd58ea5e05b)

@eloycoto and I have been investigating this. Adding `require('resty.ctx').apply()` to the log by lua block in `gateway/conf.d/apicast.conf` is not the solution. When starting APIcast with `bin/apicast` we saw that the context is shared correctly in all the phases, including the `log` one. There's only one test that fails and the reason is that it is the only one that defines its own `rewrite_by_lua`: https://github.com/3scale/APIcast/blob/66168b7a15655f0ea473d862a41ae9fdbb3fa912/t/apicast-policy-3scale-batcher.t#L453

Calling `require('resty.ctx').apply()` in that rewrite block fixes the issue.

eloycoto:
Hi, 

PR Updated, all the changes in the same commits, the PR is quite small, issues
fixed:

- For the context issues, now in the `M:log` is checking that the
  context.service exists if not a empty service will be used. So the test that
  had failed last Tuesday are now ok.  
- For the service, no longer the key "all", a empty string will be send. 
- For `t/prometheus-metrics.t` a new helper was added, is like
  response_body_like but with multiple options. I did not add this helper in a
  proper library because I think that it's only needed here, but if you think
  that it's good to move it, I'm happy to move it.   

Please, review it again. 
Eloy

spec/metrics/upstream_spec.lua
```diff
@@ -55,5 +55,19 @@ describe('upstream metrics', function()
         assert.stub(test_histogram.observe).was_not_called()
       end)
     end)
+
+
+    describe('With service id', function()
```
davidor: These test only check the latency histogram. We should also add tests for the status code counter.

t/prometheus-metrics.t
```diff
@@ -275,7 +315,144 @@ qr/upstream_response_time_seconds(.|\n)*upstream_response_time_seconds_bucket\{l
...
+["", "Host: metrics"]
+--- error_code eval
+[ 200, 200 ]
+--- expected_response_body_like_multiple eval
+[
+"",
+[
+    qr/total_response_time_seconds(.|\n)*total_response_time_seconds_bucket\{service="42",le=".*"\} 1/,
```
mikz: Couldn't this be actually two regexps ?

t/prometheus-metrics.t
```diff
@@ -231,7 +271,7 @@ In particular, it shows the status codes and the response times
 --- response_body_like eval
 [
 "",
-qr/upstream_response_time_seconds(.|\n)*upstream_response_time_seconds_bucket\{le=".*"\} 1(.|\n)*upstream_status\{status="200"\} 1/
+qr/upstream_response_time_seconds(.|\n)*upstream_response_time_seconds_bucket\{service="",le=".*"\} 1(.|\n)*upstream_status\{status="200",service=""\} 1/
```
mikz: when expected_response_body_like_multiple is available it might be worth to refactor this?

t/prometheus-metrics.t
```diff
@@ -275,7 +315,144 @@ qr/upstream_response_time_seconds(.|\n)*upstream_response_time_seconds_bucket\{l
...
+--- error_code eval
+[ 200, 200, 200 ]
+--- expected_response_body_like_multiple eval
+[
+"",
+"",
+[
+qr/total_response_time_seconds_bucket\{service="42",le=".*"\} 1/,
+qr/upstream_response_time_seconds_bucket\{service="42",le=".*"\} 1/,
+qr/upstream_status\{status="200",service="42"\}/,
```
davidor: I think it would be good to check the counter as well like in the 2 cases above. (`1/`)
Same for service 21.

davidor:
@eloycoto Sorry, I thought about this after approving the PR. Maybe we should use the service name or  instead of the ID as the label of the metric.

I'm thinking about how this can be used to show data in something like Grafana. The ID will not give much information to the people operating APIcast. It's likely that they'll need to go to the admin portal to check which service that ID corresponds to. Maybe we can save them that step.

eloycoto:
For the record, we discussed offline with @davidor:

If we use the ID, the number will never change. 
There is a name that the users can update, but that means that the historical information will be lost. 
There is a system_name variable that will work, but if you update the Name in the UI the system_name will not be updated, so it's like an old id.

We are going to keep the ID that it's the most reasonable one.

mikz:
I think we discussed this earlier and kind of agreed using system name. The system name is immutable and should be used instead of an ID because it gives more context. It is immediately obvious in alerts and dashboards what service it is. Ids have zero context and it would be impossible for humans to decipher them without some lookup through 3scale API.

eloycoto:
Changes to use system_name added in a new commit. Please review :-)

eloycoto:
Hi, 

Made some perfomance testing around the multiple services in the labels, it's
not a big change, I didn't notice any memory issue, is around 200req/sec slower
and adds a 2ms latency: 

Config: https://gist.githubusercontent.com/eloycoto/37b23b7c74a21645e6a4c37f41197c53/raw/447eaf62434e50cd1e0253b622443f0d6de99d64/3scale-100-nodes.json
wrk setup:

```
function request()
   wrk.headers["Host"] = tostring(math.random(0,99))
   return wrk.format(nil, "/?user_key=foo")
end
```

Result with extended metrics:
```
root@ubuntu-s-4vcpu-8gb-ams3-01:~/APIcast# wrk  --connections 100 --threads 10 --duration 300 'http://172.18.0.3:8080/?user_key=foo' -s wrk-code.lua
Running 5m test @ http://172.18.0.3:8080/?user_key=foo
  10 threads and 100 connections
    Latency    27.49ms    7.93ms 141.59ms   72.14%
    Req/Sec   365.66     40.38     0.95k    75.36%
  1092416 requests in 5.00m, 234.36MB read
Requests/sec:   3640.26
Transfer/sec:    799.69KB
```

Upstream status metrics:
```
bash-4.2# curl http://localhost:9421/metrics -s | grep ^upstream_status
upstream_status{status="200",service="0"} 10772
upstream_status{status="200",service="1"} 10933
upstream_status{status="200",service="10"} 10410
upstream_status{status="200",service="11"} 11157
upstream_status{status="200",service="12"} 10465
upstream_status{status="200",service="13"} 11176
upstream_status{status="200",service="14"} 10485
upstream_status{status="200",service="15"} 10869
upstream_status{status="200",service="16"} 11551
upstream_status{status="200",service="17"} 11190
upstream_status{status="200",service="18"} 11373
upstream_status{status="200",service="19"} 10638
upstream_status{status="200",service="2"} 10568
upstream_status{status="200",service="20"} 10732
upstream_status{status="200",service="21"} 11071
upstream_status{status="200",service="22"} 11077
upstream_status{status="200",service="23"} 10724
upstream_status{status="200",service="24"} 10985
upstream_status{status="200",service="25"} 10688
upstream_status{status="200",service="26"} 10419
upstream_status{status="200",service="27"} 10913
upstream_status{status="200",service="28"} 11312
upstream_status{status="200",service="29"} 11288
upstream_status{status="200",service="3"} 11297
upstream_status{status="200",service="30"} 10395
upstream_status{status="200",service="31"} 10977
upstream_status{status="200",service="32"} 10810
upstream_status{status="200",service="33"} 11141
upstream_status{status="200",service="34"} 10920
upstream_status{status="200",service="35"} 11128
upstream_status{status="200",service="36"} 10456
upstream_status{status="200",service="37"} 10756
upstream_status{status="200",service="38"} 11159
upstream_status{status="200",service="39"} 10872
upstream_status{status="200",service="4"} 10773
upstream_status{status="200",service="40"} 10346
upstream_status{status="200",service="41"} 10553
upstream_status{status="200",service="42"} 11502
upstream_status{status="200",service="43"} 10724
upstream_status{status="200",service="44"} 10953
upstream_status{status="200",service="45"} 11275
upstream_status{status="200",service="46"} 10420
upstream_status{status="200",service="47"} 11296
upstream_status{status="200",service="48"} 11134
upstream_status{status="200",service="49"} 10559
upstream_status{status="200",service="5"} 10817
upstream_status{status="200",service="50"} 10998
upstream_status{status="200",service="51"} 10654
upstream_status{status="200",service="52"} 10794
upstream_status{status="200",service="53"} 10395
upstream_status{status="200",service="54"} 11651
upstream_status{status="200",service="55"} 10933
upstream_status{status="200",service="56"} 11145
upstream_status{status="200",service="57"} 10777
upstream_status{status="200",service="58"} 10043
upstream_status{status="200",service="59"} 11505
upstream_status{status="200",service="6"} 10780
upstream_status{status="200",service="60"} 10272
upstream_status{status="200",service="61"} 10658
upstream_status{status="200",service="62"} 10847
upstream_status{status="200",service="63"} 11367
upstream_status{status="200",service="64"} 11336
upstream_status{status="200",service="65"} 10528
upstream_status{status="200",service="66"} 10825
upstream_status{status="200",service="67"} 11279
upstream_status{status="200",service="68"} 11098
upstream_status{status="200",service="69"} 10708
upstream_status{status="200",service="7"} 11377
upstream_status{status="200",service="70"} 11280
upstream_status{status="200",service="71"} 11224
upstream_status{status="200",service="72"} 10878
upstream_status{status="200",service="73"} 10995
upstream_status{status="200",service="74"} 11010
upstream_status{status="200",service="75"} 10953
upstream_status{status="200",service="76"} 11471
upstream_status{status="200",service="77"} 10683
upstream_status{status="200",service="78"} 10551
upstream_status{status="200",service="79"} 10088
upstream_status{status="200",service="8"} 11209
upstream_status{status="200",service="80"} 11152
upstream_status{status="200",service="81"} 11418
upstream_status{status="200",service="82"} 10981
upstream_status{status="200",service="83"} 10945
upstream_status{status="200",service="84"} 10713
upstream_status{status="200",service="85"} 10293
upstream_status{status="200",service="86"} 11375
upstream_status{status="200",service="87"} 10748
upstream_status{status="200",service="88"} 10630
upstream_status{status="200",service="89"} 11301
upstream_status{status="200",service="9"} 11493
upstream_status{status="200",service="90"} 11578
upstream_status{status="200",service="91"} 10510
upstream_status{status="200",service="92"} 11049
upstream_status{status="200",service="93"} 11309
upstream_status{status="200",service="94"} 10555
upstream_status{status="200",service="95"} 10859
upstream_status{status="200",service="96"} 11179
upstream_status{status="200",service="97"} 10857
upstream_status{status="200",service="98"} 11392
upstream_status{status="200",service="99"} 10790
bash-4.2#
```

Without extended metrics:

wrk output:
```
root@ubuntu-s-4vcpu-8gb-ams3-01:~/APIcast# wrk  --connections 100 --threads 10 --duration 300 'http://172.18.0.3:8080/?user_key=foo' -s wrk-code.lua
Running 5m test @ http://172.18.0.3:8080/?user_key=foo
  10 threads and 100 connections
    Latency    25.81ms    7.61ms 118.59ms   72.33%
    Req/Sec   389.60     46.55     1.07k    79.60%
  1163878 requests in 5.00m, 249.69MB read
Requests/sec:   3878.50
Transfer/sec:    852.02KB
root@ubuntu-s-4vcpu-8gb-ams3-01:~/APIcast#
```

Metrics:
```
bash-4.2# curl http://localhost:9421/metrics -s | grep ^upstream_status
upstream_status{status="200",service=""} 1163966
bash-4.2#
```

mikz:
@eloycoto I realised one shortcoming when using just system_name. It is no longer unique in multi-tenant environments. Would it be a big issue to have both system name and id? The granularity should be the same, so it shouldn't affect prometheus.

eloycoto:
@mikz your comment makes total sense! I do not liked the system_name; a user can't change that and can be out of date quite quickly. 

I would format as ${id}-${name}, with that users can use a custom legend based on ID or names, for legend format, they can use the following promQL query, and we allow users to have the correct names on that: 

https://prometheus.io/docs/prometheus/latest/querying/functions/#label_replace

What do you think about this? Are you ok with that?

Regards

mikz:
@eloycoto I actually thought about using two labels. One for id and one for system_name. So people do not have to bother using the label replace and just see the system_name as a hint for what the service is, but still be unique because of the ID. It is just additional metadata.
And for the ones that want, they can use the label_replace based on the id.

eloycoto:
@mikz the problem if you add two labels, is that Prometheus matrix will be bigger, and the Prometheus performance will be affected, I'm happy with two labels if have different data, but having two labels for the same data, I'm a bit against

mikz:
@eloycoto how it will be bigger? the number of combinations is the same if you make an identifier that contains both or if you have two identifiers and one is unique. Prometheus will not store anything in empty timelines.

eloycoto:
@mikz latest changes now contains the service_id and the service_system_name labels per metric :-) 

Regards.

mikz:
@eloycoto amazing! 👍

davidor:
@eloycoto Good job! :+1:

PR #1025, opened 2019-04-30
[THREESCALE-2166] Add Upstream Connection policy

davidor:
This PR adds a new policy, "Upstream Connection".

This policy allows to configure several options for the connections to the upstream. In particular, 3 timeouts:  connect_timeout, send_timeout, and read_timeout.

Ref: https://issues.jboss.org/browse/THREESCALE-2166

### DIFF
--- a/gateway/src/apicast/balancer.lua
+++ b/gateway/src/apicast/balancer.lua
@@ -1,6 +1,7 @@
 local assert = assert
 local round_robin = require 'resty.balancer.round_robin'
 local resty_url = require 'resty.url'
+local inspect = require 'inspect'
 
 local _M = { default_balancer = round_robin.new() }
 
@@ -52,6 +53,23 @@ local function get_peer(balancer, upstream)
   return peer
 end
 
+local function set_timeouts(balancer, timeouts)
+  if not timeouts then return end
+
+  local _, err = balancer:set_timeouts(
+      timeouts.connect_timeout,
+      timeouts.send_timeout,
+      timeouts.read_timeout
+  )
+
+  if err then
+    ngx.log(ngx.WARN,
+            'Error while setting balancer timeouts: ',
+            inspect(timeouts),
+            ' err: ', err)
+  end
+end
+
 function _M.call(_, context, bal)
   local balancer = assert(bal or _M.default_balancer, 'missing balancer')
   local upstream, peer, err, ok
@@ -71,6 +89,9 @@ function _M.call(_, context, bal)
   ok, err = balancer:set_current_peer(peer[1], peer[2] or upstream.uri.port or resty_url.default_port(upstream.uri.scheme))
 
   if ok then
+    -- context.upstream_connection_opts is set by the "upstream_connection" policy.
+    set_timeouts(balancer, context.upstream_connection_opts)
+
     -- I wish there would be a nicer way, but unfortunately ngx.exit(ngx.OK) does not
     -- terminate the current phase handler and will evaluate all remaining balancer phases.
     context[upstream] = peer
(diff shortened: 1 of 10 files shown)


davidor:
@vramosp I named the policy "Upstream Connection". If you want to rename it, let me know.

vramosp:
@davidor Would something like 'Upstream Timeout Configuration' be too restrictive?

davidor:
@vramosp In my opinion, that's a bit restrictive because in the future we might want to add other configuration parameters to this policy that are not timeouts.

vramosp:
@davidor I can't think of anything better then.

gateway/src/apicast/balancer.lua
```diff
@@ -52,6 +52,20 @@ local function get_peer(balancer, upstream)
...
+  if not timeouts then return end
+
+  local _, err = balancer:set_timeouts(
+      timeouts.connect_timeout,
+      timeouts.send_timeout,
+      timeouts.read_timeout
+  )
+
+  if err then
+    ngx.log(ngx.WARN, 'Error while setting balancer timeouts: ', err)
```
eloycoto: I read the resty balance timeout, and the error message is not too descriptive[0]. I would add the connect_timeout, send_timeout and read_timeout in the warning messages. 

https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/balancer.lua#L166-L206

spec/balancer_spec.lua
```diff
@@ -23,5 +23,37 @@ describe('apicast.balancer', function()
       assert(apicast_balancer:call({ upstream =  upstream }, balancer))
       assert.spy(balancer.set_current_peer).was.called_with(balancer, '127.0.0.2', 443)
     end)
+
+    it('sets the timeouts in the balancer when they are received in the context', function()
+      local timeouts = {
```
eloycoto: I would add a test with timeouts with negative numbers just in case.

gateway/src/apicast/policy/upstream_connection/apicast-policy.json
```diff
@@ -0,0 +1,24 @@
...
+  "configuration": {
+    "type": "object",
+    "properties": {
+      "connect_timeout": {
+        "description": "Timeout for establishing a connection (in seconds).",
+        "type": "integer"
+      },
+      "send_timeout": {
+        "description": "Timeout between two successive write operations (in seconds).",
+        "type": "integer"
```
eloycoto: Just a note here, could we use a type number with a range? Like this for example: 

https://json-schema.org/understanding-json-schema/reference/numeric.html#range 

So validation for negative numbers can be in the schema.

gateway/src/apicast/policy/upstream_connection/apicast-policy.json
```diff
@@ -0,0 +1,24 @@
...
+  "description": "Allows to configure several options for the connections to the upstream",
+  "version": "builtin",
+  "configuration": {
+    "type": "object",
+    "properties": {
+      "connect_timeout": {
+        "description": "Timeout for establishing a connection (in seconds).",
+        "type": "integer"
+      },
+      "send_timeout": {
```
eloycoto: Just a note here, could we use a type number with a range? Like this for example:

https://json-schema.org/understanding-json-schema/reference/numeric.html#range

So validation for negative numbers can be in the schema.

gateway/src/apicast/policy/upstream_connection/apicast-policy.json
davidor: Good idea. Added :+1:

gateway/src/apicast/balancer.lua
davidor: You're right, the error does not show the values. I added an `inspect()` there :+1:

PR #1028, opened 2019-05-03
[standalone] support for defining upstreams

mikz:
route can refer to an upstream(external) and directly send a response
there without going through a policy chain

it snowballed a bit into a bunch of fixes for standalone mode

required for https://github.com/3scale/ostia/pull/57

### Example configuration

```json
{
  "global": { "log_level": "debug", "error_log": "stderr", "access_log": "stdout" },
  "server": { "listen": [{ "port": 8080, "name": "default", "protocol": "http" }] },
  "routes": [
    {
      "name": "hello",
      "match": { "server_port": "default", "uri_path": "/hello" },
      "destination": { "upstream": "https://echo-api.3scale.net" }
    }
  ],
  "internal": [ ],
  "external": [
    { "name": "https://echo-api.3scale.net", "server": "https://echo-api.3scale.net" }
  ]
}
```

### DIFF
--- a/gateway/src/apicast/policy/standalone/configuration.lua
+++ b/gateway/src/apicast/policy/standalone/configuration.lua
@@ -92,7 +92,13 @@ do
         local load = loaders[url.scheme]
         if not load then return nil, 'cannot load scheme' end
 
-        return setmetatable(load(url), config_mt)
+        local t, err = load(url)
+
+        if t then
+            return setmetatable(t, config_mt)
+        else
+            return t, err
+        end
     end
 end
 
--- a/gateway/src/apicast/policy/standalone/configuration.lua
+++ b/gateway/src/apicast/policy/standalone/configuration.lua
@@ -31,6 +31,17 @@ do
     local tostring = tostring
     local path = require('pl.path')
     local file = require('pl.file')
+    local inspect = require('inspect')
+    local tab_clone = require "table.clone"
+    local __tostring = function(self)
+        -- We need to remove metatable of the inspected table, otherwise it points to itself
+        -- and leads to infinite recursion. Cloning the table has the same effect without mutation.
+        -- As this is evaluated only when nginx was compiled with debug flag,
+        -- it means it happens only in development.
+        return inspect(tab_clone(self))
+    end
+
+    local config_mt = { __tostring = __tostring }
 
     local YAML = require('resty.yaml')
     local cjson = require('cjson')
@@ -81,8 +92,7 @@ do
         local load = loaders[url.scheme]
         if not load then return nil, 'cannot load scheme' end
 
-
-        return load(url)
+        return setmetatable(load(url), config_mt)
     end
 end
 
--- a/gateway/src/apicast/policy/standalone/configuration.lua
+++ b/gateway/src/apicast/policy/standalone/configuration.lua
@@ -31,7 +31,7 @@ do
     local path = require('pl.path')
     local file = require('pl.file')
 
-    local YAML = require('lyaml')
+    local YAML = require('resty.yaml')
     local cjson = require('cjson')
 
     local decoders = {
--- a/gateway/src/apicast/policy/standalone/configuration.lua
+++ b/gateway/src/apicast/policy/standalone/configuration.lua
@@ -9,7 +9,7 @@ local mt = { __index = _M }
 
 local allowed_schemes = {
     file = true,
-    -- TODO: support Data URI
+    data = true,
 }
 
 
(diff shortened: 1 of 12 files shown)


gateway/src/apicast/policy/standalone/standalone.lua
```diff
@@ -319,13 +351,27 @@ end
 local function find_service(self, route)
   local destination = route and route.destination
 
-  if self.services and destination then
+  if not destination then return nil, 'no destination' end
+
+  if destination.service then
     return self.services[destination.service]
```
eloycoto: now that you return `service, err` should return here `self.services[destination.service], nil`?

gateway/src/apicast/policy/standalone/standalone.lua
```diff
@@ -319,13 +351,27 @@ end
...
   end
+
+  if destination.upstream then
+    local upstream = self.upstreams[destination.upstream]
+
+    if upstream then
+      return upstream
```
eloycoto: also return here upstream, nil no?

mikz: Return `value` and `value, nil` work exactly the same. So I guess there is not much reason for that.

gateway/src/apicast/policy/standalone/standalone.lua
mikz: The second return value is implicit nil. I don't see why it would be needed.

gateway/src/resty/sandbox.lua
```diff
@@ -14,7 +14,7 @@ local setmetatable = setmetatable
 local concat = table.concat
 
 local _G = _G
-local _M = {}
+local _M = { debug = false }
```
davidor: I agree that this produces a lot of unnecessary logs but leaving it like this is confusing because there's no way to enable this logs apart from changing this line of code. Should we at least add a line here that says that you should change this to true if you ever want to debug this module?

t/apicast-policy-standalone.t
```diff
@@ -32,3 +32,37 @@ GET /t HTTP/1.1
...
+
+
+=== TEST 2: standalone can point to external destination
+--- environment_file: standalone
+--- configuration_format: yaml
+--- configuration
+server:
+  listen:
+  - port: $TEST_NGINX_SERVER_PORT
+    name: test
```
davidor: Everything is named `test`. It's not obvious to know what attributes should have matching names.

gateway/src/apicast/policy/standalone/standalone.lua
```diff
@@ -217,11 +219,31 @@ end
 do
   local External_mt = { __index = External }
 
+  local UpstreamPolicy = Policy.new('Standalone Upstream Policy')
+  local UpstreamPolicy_new = UpstreamPolicy.new
+
+  function UpstreamPolicy.new(upstream)
```
davidor: Would it make sense to use the exiting .new()? Couldn't we pass the config as the exiting constructor expects it? I guess that a rule that matches everything would work.

gateway/src/apicast/policy/standalone/configuration.lua
```diff
@@ -31,6 +31,10 @@ do
     local tostring = tostring
     local path = require('pl.path')
     local file = require('pl.file')
+    local inspect = require('inspect')
+    local tab_clone = require "table.clone"
+    local __tostring = function(self) return inspect(tab_clone(self)) end
```
davidor: Why do we need the clone?

mikz: We need to get rid of the metatable, otherwise, it does infinite recursion. Will mention it in a comment.

gateway/src/apicast/policy/standalone/standalone.lua
mikz: True. Will address it.

t/apicast-policy-standalone.t
mikz: Fair point. Will make better names.

gateway/src/resty/sandbox.lua
mikz: Yes, I'll mention it in a comment.
I don't see any situation other than development when you'd want to debug this. So it being a temporary code change is not that terrible in my opinion.

t/apicast-policy-standalone.t
```diff
@@ -199,3 +199,20 @@ GET /t HTTP/1.1
 --- no_error_log
 [error]
 [warn]
+
+
+
+=== TEST 6: invalid configuration
+--- environment_file: standalone
+--- SKIP: works on OpenResty 1.15.x
```
davidor: I don't understand this.

mikz: I'm testing this on OpenResty 1.15 locally and it works but does not work on the CI (1.13).

We should remove the SKIP once we are on 1.15. They changed passing signals between master/worker and it's shutdown.

davidor: Thanks for the clarification.

PR #1034, opened 2019-05-10
[THREESCALE-2264] Obtain JWT client_id from different claim

### LINKED ISSUE #1008: Make the JWT claim to get client_id configurable [THREESCALE-2264]
Various IDPs store the client_id in different claims of the JWT.
We need to make this configurable, so we can be compatible with different IDPs.

Options include:
* hardcoded claim name
* liquid templating

eloycoto:
This commit adds the option to obtain the client_id from a different claim and
does not use the AZP claim if the identity provider does not need that. This
commit adds a new option `oidc_client_id` that can be used as a liquid template
to retrieve any claim from the JWT payload.

Fixes #1008
Fixes THREESCALE-2264

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - "Upstream Connection" policy. It allows to configure several options for the connections to the upstream [PR #1025](https://github.com/3scale/APIcast/pull/1025), [THREESCALE-2166](https://issues.jboss.org/browse/THREESCALE-2166)
 - Enable APICAST_EXTENDED_METRICS env variable to provide more in-depth details [PR #1024](https://github.com/3scale/APIcast/pull/1024) [THREESCALE-2150](https://issues.jboss.org/browse/THREESCALE-2150)
 - Enable APICAST_EXTENDED_METRICS environment variable to provide additional details [PR #1024](https://github.com/3scale/APIcast/pull/1024) [THREESCALE-2150](https://issues.jboss.org/browse/THREESCALE-2150)
+- Add the option to obtain client_id from any JWT claim [THREESCALE-2264](https://issues.jboss.org/browse/THREESCALE-2264) [PR #1034](https://github.com/3scale/APIcast/pull/1034)
+
 
 ### Fixed
 
--- a/gateway/src/apicast/configuration.lua
+++ b/gateway/src/apicast/configuration.lua
@@ -115,6 +115,7 @@ function _M.parse_service(service)
       backend = backend,
       oidc = {
         issuer_endpoint = value(proxy.oidc_issuer_endpoint),
+        client_id = value(proxy.oidc_client_id),
       },
       credentials = {
         location = proxy.credentials_location or 'query',
--- a/gateway/src/apicast/configuration.lua
+++ b/gateway/src/apicast/configuration.lua
@@ -115,7 +115,8 @@ function _M.parse_service(service)
       backend = backend,
       oidc = {
         issuer_endpoint = value(proxy.oidc_issuer_endpoint),
-        client_id = value(proxy.oidc_client_id),
+        claim_with_client_id = value(proxy.jwt_claim_with_client_id),
+        claim_with_client_id_type = proxy.jwt_claim_with_client_id_type or "plain",
       },
       credentials = {
         location = proxy.credentials_location or 'query',
--- a/gateway/src/apicast/oauth/oidc.lua
+++ b/gateway/src/apicast/oauth/oidc.lua
@@ -16,6 +16,8 @@ local _M = {
   cache_size = 10000,
 }
 
+local default_claim_with_app_id =  "azp"
+
 function _M.reset()
   _M.cache = lrucache.new(_M.cache_size)
 end
@@ -31,19 +33,22 @@ local mt = {
 
 local empty = {}
 
-function _M.new(oidc_config, service)
+function _M.new(oidc_config)
   local oidc = oidc_config or empty
   local issuer = oidc.issuer
   local config = oidc.config or empty
   local alg_values = config.id_token_signing_alg_values_supported or empty
-  local tmpl, tmpl_err
+
+  local client_id_tmpl, client_id_tmpl_err, client_id_tmpl_type
   local err
 
-  if  oidc_config and oidc_config.client_id then
-     tmpl, tmpl_err = TemplateString.new(oidc_config.client_id, "liquid")
-    if tmpl_err then
+  if oidc.claim_with_client_id then
+    client_id_tmpl_type = oidc.claim_with_client_id_type
+    client_id_tmpl, client_id_tmpl_err = TemplateString.new(oidc.claim_with_client_id, client_id_tmpl_type)
+    if client_id_tmpl_err then
       err = 'Invalid client_id template string'
     end
+
   end
 
   if not issuer or #alg_values == 0 then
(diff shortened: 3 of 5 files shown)


gateway/src/apicast/oauth/oidc.lua
```diff
@@ -30,7 +31,7 @@ local mt = {
 
 local empty = {}
 
-function _M.new(oidc_config)
+function _M.new(oidc_config, service)
```
davidor: Do we need to pass the `service`? It's only used to obtain the jwt claim with the client ID. Can that attribute be part of the `oidc_config` param?

gateway/src/apicast/oauth/oidc.lua
```diff
@@ -194,8 +196,17 @@ function _M:transform_credentials(credentials, cache_key)
   end
 
   local payload = jwt_obj.payload
-
   local app_id = payload.azp
+
+  if self.service and self.service.oidc and self.service.oidc.client_id then
+    local tmpl, tmpl_err = TemplateString.new(self.service.oidc.client_id, "liquid")
```
davidor: We could create the `TemplateString` in the constructor to avoid doing this in every request.

davidor: I see that this is always interpreted as liquid. I'm not sure that's a good idea from a UX perspective. That forces the user to write something like `{{ aud }}` in the UI instead of just `aud`.

spec/oauth/oidc_spec.lua
```diff
@@ -227,6 +228,58 @@ describe('OIDC', function()
...
+        local credentials, ttl, _, err = oidc:transform_credentials({ access_token = access_token })
+
+        assert(credentials, err)
+
+        assert.same({ app_id  = "notused" }, credentials)
+        assert.equal(10, ttl)
+      end)
+
+      it('validation of app_id template with functions', function()
+        stub(ngx_variable, 'available_context', function(context) return context end)
```
davidor: This appears in both tests. We could avoid the duplication by extracting this to a `before` function.

spec/oauth/oidc_spec.lua
```diff
@@ -227,6 +228,58 @@ describe('OIDC', function()
       local credentials, _, _, err = oidc:transform_credentials({ access_token = access_token })
       assert(credentials, err)
     end)
+
+    describe('with app_id modification', function()
```
davidor: I think this could be more descriptive. The requirement is not to modify the app_id is to get it from a custom JWT claim.

spec/oauth/oidc_spec.lua
```diff
@@ -227,6 +228,58 @@ describe('OIDC', function()
...
+            sub = 'someone',
+            exp = ngx.now() + 10,
+          },
+        })
+
+        local credentials, ttl, _, err = oidc:transform_credentials({ access_token = access_token })
+
+        assert(credentials, err)
+
+        assert.same({ app_id  = "notused" }, credentials)
```
davidor: I think that setting `notused` as a value, when it's the jwt used for this is a bit confusing :sweat_smile:

t/apicast-oidc.t
```diff
@@ -210,3 +210,55 @@ my $jwt = encode_jwt(payload => {
 "Authorization: Bearer $jwt"
 --- no_error_log
 [error]
+
+=== TEST 4: Use different claim for app_id
```
davidor: This should be `TEST 5`

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - You can filter services by endpoint name using Regexp [PR #1022](https://github.com/3scale/APIcast/pull/1022) [THREESCALE-1524](https://issues.jboss.org/browse/THREESCALE-1524)
 - "Upstream Connection" policy. It allows to configure several options for the connections to the upstream [PR #1025](https://github.com/3scale/APIcast/pull/1025), [THREESCALE-2166](https://issues.jboss.org/browse/THREESCALE-2166)
+- Add the option to obtain client_id from any JWT claim [THREESCALE-2264](https://issues.jboss.org/browse/THREESCALE-2264) [PR #1025](https://github.com/3scale/APIcast/pull/1034)
```
davidor: Should be `[PR #1034]`

gateway/src/apicast/oauth/oidc.lua
eloycoto: I thought about that, but each service can use different `app_id` so I send the service to have a custom value per service, I think that it's the best way.

davidor: I see. Even if that's the case, I don't see any reason to couple this class with service when it only needs the name of a jwt claim from it. This constructor already receives `oidc_config`, a table of options, I think that adding another option to that table is a better solution. The caller will need to add that attribute, but I think that's better than introducing a coupling with Service.

gateway/src/apicast/configuration.lua
```diff
@@ -114,6 +114,8 @@ function _M.parse_service(service)
       backend = backend,
       oidc = {
         issuer_endpoint = value(proxy.oidc_issuer_endpoint),
+        client_id = value(proxy.oidc_client_id),
```
davidor: nitpick: the name of this var suggests that this is the client ID itself, when in reality, it's the jwt claim that contains the client ID. Can we rename this to `jwt_claim_with_client_id` or similar ?

gateway/src/apicast/oauth/oidc.lua
```diff
@@ -30,20 +31,34 @@ local mt = {
 
 local empty = {}
 
-function _M.new(oidc_config)
+function _M.new(oidc_config, service)
```
davidor: `service` is no longer needed, right?

gateway/src/apicast/oauth/oidc.lua
```diff
@@ -30,20 +31,34 @@ local mt = {
...
+
   if not issuer or #alg_values == 0 then
     err = 'missing OIDC configuration'
   end
 
   return setmetatable({
     config = config,
     issuer = issuer,
+    service = service,
+    tmpl = tmpl,
```
davidor: I think that having more descriptive names in these attributes would make the code easier to understand.

gateway/src/apicast/oauth/oidc.lua
```diff
@@ -209,6 +224,17 @@ function _M:transform_credentials(credentials, cache_key)
...
+function _M:get_client_id(jwt_payload)
+  if self.tmpl and self.tmpl_type == "liquid" then
+      return self.tmpl:render(jwt_payload)
+  end
+
+  if self.tmpl and self.tmpl_type == "plain" then
+    return jwt_payload[self.tmpl:render()]
+  end
+
+  return jwt_payload.azp
```
davidor: Maybe extracting `azp` to a `default_claim_with_app_id` constant or similar would help.
`jwt_payload[default_claim_with_app_id]`
What do you think?

spec/oauth/oidc_spec.lua
```diff
@@ -227,6 +228,103 @@ describe('OIDC', function()
...
+      end)
+
+
+      it('use plaintext format to modify app_id', function()
+
+        local oidc = get_oidc({
+          client_id = "aud",
+          client_id_type = "plain"})
+
+        local access_token = jwt:sign(rsa.private, {
```
davidor: Can we use the same token for the 3 tests? If that's the case, we should consider defining it in the `describe('getting client_id from any JWT claim', function()` block to avoid duplication.

gateway/src/apicast/oauth/oidc.lua
eloycoto: 🤦‍♂ sorry, no needed but I didn't amend it.

eloycoto:
@davidor Changes request were not small, so I split that in two new. Ping me if you want that squashed with the previous ones.

davidor:
@eloycoto The changes look good to me :+1: 
Please squash the commits as you see fit and solve the changelog conflicts.

PR #1035, opened 2019-05-13
[THREESCALE-1150] Add option to use only path-based routing

davidor:
Reference: https://issues.jboss.org/browse/THREESCALE-1150

This PR adds a new env, `APICAST_PATH_ROUTING_ONLY`. When set to true, the gateway will use path-based routing and will not fallback to the default host-based routing, which is what happens when using `APICAST_PATH_ROUTING`.

### DIFF
--- a/doc/parameters.md
+++ b/doc/parameters.md
@@ -112,6 +112,16 @@ When configured to authenticate using OAuth, this param specifies the TTL (in se
 
 When this parameter is set to _true_, the gateway will use path-based routing in addition to the default host-based routing. The API request will be routed to the first service that has a matching mapping rule, from the list of services for which the value of the `Host` header of the request matches the _Public Base URL_.
 
+### `APICAST_PATH_ROUTING_ONLY`
+
+**Values:**
+- `true` or `1` for _true_
+- `false`, `0` or empty for _false_
+
+When this parameter is set to _true_, the gateway uses path-based routing and will not fallback to the default host-based routing. The API request will be routed to the first service that has a matching mapping rule, from the list of services for which the value of the `Host` header of the request matches the _Public Base URL_.
+
+This parameter has precedence over `APICAST_PATH_ROUTING`. If `APICAST_PATH_ROUTING_ONLY` is enabled, APIcast will only do path-based routing regardless of the value of `APICAST_PATH_ROUTING`.
+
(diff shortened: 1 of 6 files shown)


doc/parameters.md
```diff
@@ -112,6 +112,16 @@ When configured to authenticate using OAuth, this param specifies the TTL (in se
 
 When this parameter is set to _true_, the gateway will use path-based routing in addition to the default host-based routing. The API request will be routed to the first service that has a matching mapping rule, from the list of services for which the value of the `Host` header of the request matches the _Public Base URL_.
 
+### `APICAST_PATH_ROUTING_ONLY`
+
+**Values:**
+- `true` or `1` for _true_
+- `false`, `0` or empty for _false_
+
+When this parameter is set to _true_, the gateway will use path-based routing and will not fallback to the default host-based routing. The API request will be routed to the first service that has a matching mapping rule, from the list of services for which the value of the `Host` header of the request matches the _Public Base URL_.
```
```suggestion
When this parameter is set to _true_, the gateway uses path-based routing and does not fallback to the default host-based routing. The API request is routed to the first service that has a matching mapping rule, from the list of services for which the value of the `Host` header of the request matches the _Public Base URL_.
```

davidor: Changed :+1: 
Thanks for the review.

spec/policy/find_service/find_service_spec.lua
```diff
@@ -58,11 +61,39 @@ describe('find_service', function()
...
       end)
+
+      describe('and fallback to find by host is disabled', function()
+        before_each(function()
+          ConfigurationStore.path_routing = true
+          ConfigurationStore.path_routing_only = true
+        end)
+
+        it('stores nil in the context even if there is a service for the host', function()
+          local service = { id = '1' }
```
eloycoto: A quick comment, shouldn't need to test with more than one service?

davidor: It's not really needed. This module only cares about what the `HostBasedFinder` and the `PathBasedFinder` modules return and they just return one service. In those 2 modules we have tests with configurations that have multiple services, so I think we are covered.

davidor:
Rebased on top of master to solve conflicts in the changelog.

PR #1039, opened 2019-05-15
[THREESCALE-2236] Add methods option on Keycloak policy.

eloycoto:
This commits add the `methods` option on the keycloack policy. That
allow users to define a new policy from any jwt claim, resource and
method.

To be backwards compatible 'ANY' method is in place, and if methods are
not defined this global method will be used and all will work as normal.

Example policy:

```
"policy_chain": [
  {
    "name": "apicast.policy.keycloak_role_check",
    "configuration": {
      "scopes": [
        {
          "realm_roles": [ { "name": "director" } ],
          "resource": "/confidential",
          "methods": ["POST"]
        }
      ],
      "type": "blacklist"
    }
  },
  { "name": "apicast.policy.apicast" }
]
```

### DIFF
--- a/gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
+++ b/gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
@@ -60,11 +60,13 @@ local default_type = 'plain'
 
 local new = _M.new
 
+local any_method = 'ANY'
+
 local function create_template(value, value_type)
   return TemplateString.new(value, value_type or default_type)
 end
 
-local function build_templates(scopes)
+local function build_scopes(scopes)
   for _, scope in ipairs(scopes) do
 
     if scope.realm_roles then
@@ -86,8 +88,12 @@ local function build_templates(scopes)
 
     scope.resource_template_string = create_template(
       scope.resource, scope.resource_type)
+    if (not scope.methods) or (scope.methods and #scope.methods == 0 ) then
+      scope.methods = { any_method }
+    end
 
   end
+
 end
 
 function _M.new(config)
@@ -95,7 +101,7 @@ function _M.new(config)
   self.type = config.type or "whitelist"
   self.scopes = config.scopes or {}
 
-  build_templates(self.scopes)
+  build_scopes(self.scopes)
 
   return self
 end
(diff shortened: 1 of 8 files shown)


gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
```diff
@@ -87,7 +89,12 @@ local function build_templates(scopes)
     scope.resource_template_string = create_template(
       scope.resource, scope.resource_type)
 
+    if not scope.methods then
```
davidor: Can we change the name of this method? It's not longer about building templates only.

davidor: What if `scope.methods` is an empty array? I think we should also set `scope.methods` to `{ any_method }`.

gateway/src/apicast/policy/keycloak_role_check/keycloak_role_check.lua
```diff
@@ -152,38 +159,54 @@ local function match_client_roles(scope, context)
...
-    return false
-  end
-
-  for _, scope in ipairs(scopes) do
+local function validate_scope_access(scope, context, uri, request_method)
+  for _, method  in ipairs(scope.methods) do
+    -- make a matched method just in case that `ANY` method is defined and
+    -- the mapping rules does not match, the matched_method need to be
+    -- cleared in the next interaction to trigger with the correct method.
+    local matched_method = request_method
```
davidor: It's a bit unfortunate that we need to do this because `MappingRule` does not provide a way to match no matter what the http method is.

If we added that, maybe we could simplify this part of the code. What do you think?

spec/policy/keycloak_role_check/keycloak_role_check_spec.lua
```diff
@@ -7,6 +7,8 @@ describe('Keycloak Role check policy', function()
     ngx.header = {}
     stub(ngx, 'print')
 
+    stub(ngx.req, 'get_method', function() return 'GET' end)
+    stub(ngx_variable, 'available_context', function(context) return context end)
```
davidor: This line is duplicated.

spec/policy/keycloak_role_check/keycloak_role_check_spec.lua
```diff
@@ -572,6 +574,81 @@ describe('Keycloak Role check policy', function()
...
+
+        it("when jwt role matches and one method also match", function()
+          local role_check_policy = KeycloakRoleCheckPolicy.new({
+            scopes = {{
+                client_roles = { { name = "role_of_known_client", client = "known_client" } },
+                resource = "/foo",
+                methods = {"GET", "POST", "PUT"}
+              }}})
+          role_check_policy:access(context)
+          assert.not_same(ngx.status, 403)
```
davidor: I'm not sure I understand this. By default this policy applies "whitelisting" instead of "blacklisting".
If I understood correctly the role matches and also the http method.
Why is the status a 4xx? What am I missing?

t/apicast-policy-keycloak-role-check.t
```diff
@@ -393,3 +393,160 @@ yay, api backend
 --- no_error_log
 [error]
 oauth failed with
+
+
+=== TEST6: Role check with allow methods
```
davidor: Can you add a space between `TEST` and `6` ? It's just because I don't know if that can cause some issue with the script that we have to automatically number tests correctly? https://github.com/3scale/APIcast/blob/master/script/reorder-tests

spec/policy/keycloak_role_check/keycloak_role_check_spec.lua
davidor: On a related note, I think in these tests we should be explicit and mention whether they're whitelisting or blacklisting, and also add the attribute that sets that instead of relying on defaults. Then, we can have another test in a different section, if there is not one already, that checks that when the type is not set, it defaults to whitelisting.

t/apicast-policy-keycloak-role-check.t
eloycoto: I've updated all the test.

davidor:
Good job @eloycoto :+1:

I think that extracting a `Scope` module might simplify some code and also the tests, but there's no need to do that now. We can evaluate that in the future if we need to add more functionality to this policy.

I think the PR could be merged after  #1041

eloycoto:
Code rebased with master branch, should be ready to merge.

PR #1041, opened 2019-05-17
[Test] Fix mocked functions.

eloycoto:
If any function is mocked used stub need to be reverted at some point,
doing something like this

```
before_each(function()
  stub(foo, "validate", function() return false end)
end)

after_each(function()
  foo.validate:revert()
end)

it("Validate stub foo", function()
  assert.same(foo.validate(), false)
end)
```

In APICast all ngx variables are reset on `ngx_helper.lua` but some
methods like `ngx_variables` or `backend_client` are not reset and the
mock function is in there for all other test.

The test in `spec/policy/ngx_variable_spec.lua` validates that
`ngx_variable.available_context` is working correctly, due some other
test can mock this function the test was inestable over the last days.

This commit creates a way to define a cleanup for mocked functions that
are not part of the ngx class.

### DIFF
--- a/spec/ngx_helper.lua
+++ b/spec/ngx_helper.lua
@@ -21,6 +21,13 @@ local ngx_ctx_original = deepcopy(ngx.ctx)
 local ngx_shared_original = deepcopy(ngx.shared)
 local ngx_header_original = deepcopy(ngx.header)
 
+-- Apicast shared variables that are mocked at more than one different test.
+local ngx_variable = require('apicast.policy.ngx_variable')
+local ngx_variable_original = deepcopy(ngx_variable)
+
+local backend_client = require('apicast.backend_client')
+local backend_client_original = deepcopy(backend_client)
+
 local register_getter = misc.register_ngx_magic_key_getter
 local register_setter = misc.register_ngx_magic_key_setter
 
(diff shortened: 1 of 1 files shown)


mikz:
All mocks/stubs should be reverted automatically. If they are applied in some global context (not after/before_each), then they would not be reverted between tests. 

I see this change as a bandaid and yet another way to rollback changes. Why we can't just rely on the rollbacking the stubs?

eloycoto:
@mikz I'm ok to apply the `:revert()` function on the after_each functions, I think that it's a bit clear and It makes things easier to read and debug. 

Last week, I had a chat with @davidor and we agreed that we should use the same way that we use for `ngx.*` methods.

mikz:
@eloycoto but that should not be needed at all. There are global after/before hooks that do that for all registered stubs. There should be no reason to manually revert them because it would be really error-prone.

https://github.com/3scale/APIcast/blob/master/spec/luassert_helper.lua (was significantly refactored in https://github.com/3scale/APIcast/pull/799).

eloycoto:
Will check, thanks!

eloycoto:
Close this one due to is not the correct solution, the correct one is on https://github.com/3scale/APIcast/pull/1042

PR #1044, opened 2019-05-21
[THREESCALE-2540] Fix issues with uppercase upstreams.

eloycoto:
If the proxy API backend is in uppercase it'll not work correctly and
Apicast exists with an execpetion [0]. This commit transform api_backend
always to lowercase string, so the system will not break even if
receives a uppercase string.

[0] Exception
```
2019/05/21 05:59:30 [error] 1037#1037: *3 lua entry thread aborted: runtime error: /home/centos/gateway/src/apicast/policy/apicast/apicast.lua:101: missing upstream
stack traceback:
coroutine 0:
        [C]: in function 'assert'
        /home/centos/gateway/src/apicast/policy/apicast/apicast.lua:101: in function </home/centos/gateway/src/apicast/policy/apicast/apicast.lua:100>
        /home/centos/gateway/src/apicast/policy_chain.lua:190: in function </home/centos/gateway/src/apicast/policy_chain.lua:187>
        /home/centos/gateway/src/apicast/policy_chain.lua:190: in function 'content'
        content_by_lua(lua_JOJems:485):1: in function <content_by_lua(lua_JOJems:485):1> while sending to client, client: 127.0.0.1, server: _, request: "GET /some-path?user_key=4e5944f11f575fd50bd6624c8d441f13 HTTP/1.1", host: "localhost:8080"
```

Fix THREESCALE-2540

### DIFF
--- a/gateway/src/apicast/policy/apicast/apicast.lua
+++ b/gateway/src/apicast/policy/apicast/apicast.lua
@@ -57,8 +57,12 @@ function _M:rewrite(context)
     p:rewrite(service, context)
   end
 
+  local err
   context[self] = context[self] or {}
-  context[self].upstream = p.get_upstream(service)
+  context[self].upstream, err = p.get_upstream(service)
+  if err then
+    ngx.log(ngx.WARN, "upstream api for the service:", service.id, " is invalid, error:", err)
+  end
 
   ngx.ctx.proxy = p
 end
@@ -98,6 +102,11 @@ function _M:access(context)
 end
 
 function _M:content(context)
+  if not context[self].upstream then
+    ngx.log(ngx.WARN, "Upstream server not found for this request")
+    return
+  end
+
   local upstream = assert(context[self].upstream, 'missing upstream')
 
   if upstream then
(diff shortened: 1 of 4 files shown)


gateway/src/apicast/policy/apicast/apicast.lua
```diff
@@ -98,6 +102,11 @@ function _M:access(context)
 end
 
 function _M:content(context)
+  if not context[self].upstream then
+    ngx.log(ngx.WARN, "Upstream server not found for this request")
```
eloycoto: I'm not sure here, but I think that something like this need to be in place, but I'm not sure. 

```
  ngx.status = 504
  ngx.print('')
  return exit()
```

davidor: Why is this `if` needed? Shouldn't we treat this case like we did up until now with the assert below?

eloycoto: The thing is if you have 3 services and one is not correctly set, APICast will die and other services will be affected. 

With this solution, the current service will not work but others will be not affected at all.

davidor: I think that we need a test case showing that apicast no longer crashes in that case.

gateway/src/apicast/policy/apicast/apicast.lua
```diff
@@ -57,8 +57,12 @@ function _M:rewrite(context)
     p:rewrite(service, context)
   end
 
+  local err
   context[self] = context[self] or {}
-  context[self].upstream = p.get_upstream(service)
+  context[self].upstream, err = p.get_upstream(service)
+  if err then
+    ngx.log(ngx.WARN, "Backend api for the service=", service.id, " is invalid, error:", err)
```
mikz: Usually, the logs start with a lower case letter, also it is more common to use `:` rather than `=`

Also, it is not really "backend api", but rather "upstream"

PR #1045, opened 2019-05-21
OpenResty 1.15.8.1 and Perl 5.28.1 compatibility

mikz:
Small changes required for APIcast to work with latest OpenResty and recent Perl.

Both make prove and busted passes locally.

### DIFF
--- a/gateway/src/resty/ctx.lua
+++ b/gateway/src/resty/ctx.lua
@@ -19,11 +19,17 @@ local tonumber = tonumber
 local _M = {
 }
 
+local function __ngx_req()
+  return getfenv(0).__ngx_req
+end
+
+local get_request = base.get_request or __ngx_req
+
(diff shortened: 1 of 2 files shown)


gateway/src/resty/ctx.lua
```diff
@@ -19,11 +19,17 @@ local tonumber = tonumber
 local _M = {
 }
 
+local function __ngx_req()
+  return getfenv(0).__ngx_req
+end
+
+local get_request = base.get_request or __ngx_req
```
davidor: Do we need the `or` in 1.15.8.1 or just to keep it working in 1.13.x ?

mikz: The or is for 1.13. `base.get_request` first appears in `1.15`. It could be cleaned up later.

Issue #1043, opened 2019-05-20
Busted: Fix double stub issue


eloycoto:
At the moment all the stubs are reverted on `spec/luassert_helper.lua` but if a stub happens twice only one is reverted and stub is leaked to other tests. 

That need to be fixed, the options are: 

- Fix the luaasert issue. 
- Check the `spec/luassert_helper.lua` if functions are still spy. 
- Trigger an error if a stub happens if a stub is already in there.

PR #1047, opened 2019-05-23
[Test] Luassert make sure to revert function if it is already mocked.

eloycoto:
This commit overwrites luassert stub function to make sure that before
try to mock any function verifies that is not mocked, and if it is,
revert to the original one. This prevents scenarios where multiple stubs
are in one test and are leaked to other test and that scenario is hard
to debug/notice.

I tried to fix this on `luassert.state`, but looks like that the second stub is not always registered and it was a bit flaky, so this, for now, is the best solution that I found. 

Fix https://github.com/3scale/APIcast/issues/1043

### DIFF
--- a/spec/luassert_helper.lua
+++ b/spec/luassert_helper.lua
@@ -1,7 +1,32 @@
 local busted = require('busted')
 local assert = require('luassert.assert')
+local spy = require('luassert.spy')
+local stub = require('luassert.stub')
 local util = require ('luassert.util')
 local say = require('say')
+
+-- The stub.new and spy.on functions are patched here to make sure that the
+-- target function is not already a mocked function.
+-- The main reason to do that is to avoid some test that a function is mocked
+-- on  the before_each and after in the test also is mocked again, all mocks
+-- are reverted(state.revert in this module), but if the mock is double, it is
+-- only reverted once, so the first mock function will be leaked to other test.
+local original_stub_new = stub.new
+local original_spy_on = spy.on
+
+function stub.new(object, key, ...)
+  if object and spy.is_spy(object[key]) then
+    error("Stub for key='".. key .."' is already a mocked function, revert the parent function")
+  end
+  return original_stub_new(object, key, ...)
+end
+
+function spy.on(object, key)
+  if object and spy.is_spy(object[key]) then
+    error("Spy for key='".. key .."' is already a mocked function, revert the parent function")
+  end
+  return original_spy_on(object, key)
+end
 
 do -- set up reverting stubs
   local state = require("luassert.state")
--- a/spec/policy/3scale_batcher/3scale_batcher_spec.lua
+++ b/spec/policy/3scale_batcher/3scale_batcher_spec.lua
@@ -131,6 +131,7 @@ describe('3scale batcher policy', function()
       describe('and backend is available', function()
         before_each(function()
           local backend_client = require('apicast.backend_client')
+          backend_client.authorize:revert()
           stub(backend_client, 'authorize').returns({ status = 200 })
         end)
 
@@ -144,12 +145,14 @@ describe('3scale batcher policy', function()
       describe('and backend is not available', function()
         before_each(function()
           local backend_client = require('apicast.backend_client')
+          backend_client.authorize:revert()
           stub(backend_client, 'authorize').returns({ status = 500 })
         end)
 
         describe('and the authorization is in the downtime cache', function()
           describe('and it is OK', function()
             before_each(function()
+              batcher_policy.backend_downtime_cache.get:revert()
               stub(batcher_policy.backend_downtime_cache, 'get').returns(200)
             end)
 
@@ -163,6 +166,7 @@ describe('3scale batcher policy', function()
 
           describe('and it is denied', function()
             before_each(function()
+              batcher_policy.backend_downtime_cache.get:revert()
               stub(batcher_policy.backend_downtime_cache, 'get').returns(409)
             end)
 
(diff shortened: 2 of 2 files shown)


spec/luassert_helper.lua
```diff
@@ -1,7 +1,20 @@
...
+local spy = require('luassert.spy')
+local stub = require('luassert.stub')
 local util = require ('luassert.util')
 local say = require('say')
+local tablex = require('pl.tablex')
+local deepcopy = tablex.deepcopy
+
+local original_stub_new = deepcopy(stub.new)
+
+stub.new = function(object, key, ...)
```
mikz: Stub is a spy with extra functionality. So if we probably should fix the spy and that would also fix the stub.

`spy.is_spy(val)` retrurns true for both spies and stubs.

spy.new is called from stub.new, so just overriding spy.new should do the trick

spec/luassert_helper.lua
```diff
@@ -1,7 +1,20 @@
 local busted = require('busted')
 local assert = require('luassert.assert')
+local spy = require('luassert.spy')
+local stub = require('luassert.stub')
 local util = require ('luassert.util')
 local say = require('say')
+local tablex = require('pl.tablex')
+local deepcopy = tablex.deepcopy
+
+local original_stub_new = deepcopy(stub.new)
```
mikz: stub.new is a function, there is no reason to copy it. just a reference will do.

```lua

stub_new = stub.new

function stub.new(...)
  return stub_new(...)
end
```

spec/luassert_helper.lua
```diff
@@ -1,7 +1,20 @@
...
+
+stub.new = function(object, key, ...)
+  if object and spy.is_spy(object[key]) then
+    object[key]:revert()
```
mikz: I think we just could throw an error and let the developer handle it. I think that is more valuable than doing something magical in the background.

spec/luassert_helper.lua
eloycoto: I've just done some test if I use `spy.new()` the callback argument is never spied, so I think that the main problem is here: 

https://github.com/Olivine-Labs/luassert/blob/3b2351c384cf982b953ab6d7964f835acb8cb7db/src/stub.lua#L27-L35

I'm not sure what is the best solution here, any opinion here?

spec/luassert_helper.lua
```diff
@@ -1,8 +1,19 @@
...
+local spy = require('luassert.spy')
+local stub = require('luassert.stub')
 local util = require ('luassert.util')
 local say = require('say')
 
+local original_stub_new = stub.new
+
+function stub.new(object, key, ...)
+  if object and spy.is_spy(object[key]) then
+    assert(false, "Stub for key='".. key .."' is already a mocked function, revert the parent function")
```
mikz: No need for the assert. Just call `error(message)`.

spec/luassert_helper.lua
mikz: I'm not sure if I get it.

I'm just saying overriding `spy.new` instead of `stub.new` will cover both stubs and spies.

eloycoto: This is an example:
 
```
localhost.localdomain ➜  APIcast git:(master) ✗ cat spec/stub_issue.lua
foo = {
    foo = function() return 1 end
}

describe("EloyIssue", function()
    it("Test Stubs", function()
        assert.same(foo.foo(), 1)

        stub(foo, "foo", function() return 2 end)
        assert.same(foo.foo(), 2)

        stub(foo, "foo", function() return 3 end)
        assert.same(foo.foo(), 3)
    end)
end)
localhost.localdomain ➜  APIcast git:(master) ✗ git diff spec/luassert_helper.lua | cat
diff --git a/spec/luassert_helper.lua b/spec/luassert_helper.lua
index 8cd07802..a2e5c8ef 100644
--- a/spec/luassert_helper.lua
+++ b/spec/luassert_helper.lua
@@ -1,8 +1,16 @@
 local assert = require('luassert.assert')
 local util = require ('luassert.util')
+local spy = require ('luassert.spy')
 local say = require('say')

+original_spy_new = spy.new
+
+function spy.new(callback)
+    print("Callback is spy:", spy.is_spy(callback))
+    return original_spy_new(callback)
+end
+
 do -- set up reverting stubs
   local state = require("luassert.state")

```

```
bash-4.2$
bash-4.2$ clear
bash-4.2$ BUSTED_FILES=spec/stub_issue.lua make busted
/usr/local/openresty/luajit/bin/rover install --roverfile=gateway/Roverfile > /dev/null
/usr/local/openresty/luajit/bin/rover exec bin/busted spec/stub_issue.lua
Callback is spy:false
Callback is spy:false
Callback is spy:false
●
1 success / 0 failures / 0 errors / 0 pending : 0.003327 seconds
bash-4.2$
```
For some reason, the spy.new the callback always is never spied, I also tried on spy.on but in the only place that works is on `stub.new()`

mikz: I see. `spy.new` has a different signature from `stub.new`. Then stub `spy.on` ?

It is being called from `stub.new`. https://github.com/Olivine-Labs/luassert/blob/3b2351c384cf982b953ab6d7964f835acb8cb7db/src/stub.lua#L38

spec/luassert_helper.lua
```diff
@@ -1,8 +1,27 @@
 local busted = require('busted')
 local assert = require('luassert.assert')
+local spy = require('luassert.spy')
+local stub = require('luassert.stub')
 local util = require ('luassert.util')
 local say = require('say')
 
+local original_stub_new = stub.new
+local original_spy_on = spy.on
+
```
davidor: The PR looks good. This was tricky :+1:

I think we should add a comment here that explains why overriding these two functions is needed. We should mention that we are reverting stubs and spies on every tests, but if we double-stub/spy, then only the latest is reverted.

PR #1049, opened 2019-05-29
Bump Openresty to 1.15.8.1 version

eloycoto:


gateway/src/apicast/executor.lua
```diff
@@ -38,7 +38,10 @@ local function build_context(executor)
 end
 
 local function shared_build_context(executor)
-    local ctx = ngx.ctx or {}
+    local ctx, err = pcall(function() return ngx.ctx end) or {}
```
mikz: This does not work. `pcall` returns `true/false` as a first value.

>  Its first result is the status code (a boolean), which is true if the call succeeds without errors
http://www.lua.org/manual/5.2/manual.html#pdf-pcall

bin/busted.lua
```diff
@@ -1,3 +1,5 @@
+getmetatable(_G).__newindex = nil
```
davidor: Please add a comment explaining why this is needed :pray:

davidor:
Looks good to me apart from the missing comment :+1:

PR #1050, opened 2019-05-29
 [THREESCALE-2194] Ensure locks are released on configuration_loader. 

### LINKED ISSUE #1010: Ensure lock is released in the ConfigurationLoader module
When downloading the configuration from Porta, we use an `ngx.semaphore` to make sure that there are not several requests at the same time trying to download the same configuration and storing it in memory.

Here's the code function that does that:
https://github.com/3scale/APIcast/blob/6f6aeb034e0c388c9280c6e259f4ff9303be859f/gateway/src/apicast/configuration_loader.lua#L199

The problem is that if there's an error while trying to download the config, the thread dies and the lock is never returned. Subsequent requests that try to download the config for the same service will wait for 15 s and then fail because they will not be able to acquire the lock (it was never returned in the first call). There's a bug reported when this happened:  https://issues.jboss.org/browse/THREESCALE-2194

My first thought was to just wrap that code on an xpcall or pcall. However, I think that we should implement a generic "finally/ensure" clause because we have the same pattern in different parts of the code. For example: https://github.com/3scale/APIcast/blob/e9c8f0a59318fa4f37e9fa41422cce5d046a92ff/gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua#L112

eloycoto:
This commits add a new function `run` to the synchronization module,
that function ensures that a given function is going to be run in a safe
task and make sure that the release of the lock  always happens, so the
lock is not going to be blocked if something goes wrong.

Fix #1010
Fix THREESCALE-2194

### DIFF
--- a/gateway/src/apicast/configuration_loader.lua
+++ b/gateway/src/apicast/configuration_loader.lua
@@ -18,6 +18,8 @@ local assert = assert
 local pcall = pcall
 local tonumber = tonumber
 
+local lazy_load_timeout = 15
+
 local noop = function(...) return ... end
 
 local _M = {
@@ -196,33 +198,30 @@ function lazy.init(configuration)
   configuration.configured = true
 end
 
-function lazy.rewrite(configuration, host)
-  if not host then
-    return nil, 'missing host'
-  end
-
-  local sema = synchronization:acquire(host)
-
-  if ttl() == 0 then
-    configuration = configuration_store.new(configuration.cache_size)
-  end
-
-  local ok, err = sema:wait(15)
-
-  if ok and not _M.configured(configuration, host) then
-    ngx.log(ngx.INFO, 'lazy loading configuration for: ', host)
+local function lazy_load_config(configuration, host)
     local config = _M.load(host)
     if not config then
       ngx.log(ngx.WARN, 'failed to get config for host: ', host)
     end
     _M.configure(configuration, config)
+end
+
+function lazy.rewrite(configuration, host)
+  if not host then
+    return nil, 'missing host'
   end
 
-  if ok then
-    synchronization:release(host)
-    sema:post()
-  else
-    ngx.log(ngx.WARN, 'failed to acquire lock to lazy load: ', host, ' error: ', err)
+  if ttl() == 0 then
+    configuration = configuration_store.new(configuration.cache_size)
+  end
+
+  if _M.configured(configuration, host) then
+    return configuration
+  end
+
+  local ret, result = synchronization:run(host, lazy_load_timeout, lazy_load_config, configuration, host)
+  if not ret then
+    ngx.log(ngx.WARN, 'failed to load config for host: ', host, ' error: ', result)
   end
 
   return configuration
--- a/gateway/src/resty/synchronization.lua
+++ b/gateway/src/resty/synchronization.lua
@@ -3,6 +3,7 @@
 -- @classmod resty.synchronization
 
 local semaphore = require "ngx.semaphore"
+local safe_task = require('resty.concurrent.safe_task_executor')
 
 local rawset = rawset
 local setmetatable = setmetatable
@@ -15,7 +16,7 @@ local mt = {
 }
 --- initialize new synchronization table
 -- @tparam int size how many resources for each key
-function _M.new(_, size)
+function _M.new(size)
   local semaphore_mt = {
     __index = function(t, k)
       local sema = semaphore.new(size or 1)
@@ -24,7 +25,6 @@ function _M.new(_, size)
       return sema
     end
   }
-
 
   local semaphores = setmetatable({}, semaphore_mt)
   return setmetatable({ semaphores = semaphores }, mt)
@@ -51,6 +51,35 @@ function _M:release(key)
     return nil, 'not initialized'
   end
   semaphores[key] = nil
+end
+
+--- run a new function, callback using locks on the given key
+-- @tparam string key: key for the semaphore
+-- @tparam int timeout: timeout for getting the lock before raise an error.
+-- @tparam function callback: function to execute if the lock is acquired correctly
+-- @param ...: the variable number of arguments that are going to be send to the callback function.
+function _M:run(key, timeout, callback, ...)
+  local sema, err = self:acquire(key)
+  if err ~= key then
+    ngx.log(ngx.WARN, 'failed to acquire lock on key: ', key, ' error: ', err)
+    return false
+  end
+
+  local lock_acquired, acquire_err = sema:wait(timeout)
+  if not lock_acquired then
+    ngx.log(ngx.WARN, 'failed to acquire lock on key: ', key, ' error: ', acquire_err)
+    return false
+  end
+
+  local task = safe_task.new(callback)
+  local ret, result, execute_error = task:execute(...)
+
+  if lock_acquired then
+    self.release(key)
+    sema:post()
+  end
+
+  return ret, result, execute_error
 end
 
 return _M
--- a/t/configuration-loading-lazy.t
+++ b/t/configuration-loading-lazy.t
@@ -76,7 +76,7 @@ env APICAST_CONFIGURATION_LOADER=lazy;
 GET /t?user_key=fake
 --- error_code: 200
 --- user_files eval
-[ 
+[
   [ 'config.json', qq|
   {
     "services": [{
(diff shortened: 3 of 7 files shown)


t/configuration-loading-lazy.t
```diff
@@ -75,7 +76,7 @@ env THREESCALE_PORTAL_ENDPOINT=http://127.0.0.1:$TEST_NGINX_SERVER_PORT;
 GET /t?user_key=fake
 --- error_code: 200
 --- user_files eval
-[
+[
```
davidor: :eyes:

gateway/src/resty/synchronization.lua
```diff
@@ -53,4 +53,28 @@ function _M:release(key)
   semaphores[key] = nil
 end
 
+--- run a new function, callback using locks on the given key
+-- @tparam string key: key for the semaphore
+-- @tparam int timeout: timeout for getting the lock before raise an error.
+-- @tparam function callback: function to execute if the lock is adquired correctly
+function _M:run(key, timeout, callback, ...)
+  local sema = self:acquire(key)
```
davidor: `acquire` can return an error.

gateway/src/resty/synchronization.lua
```diff
@@ -53,4 +53,28 @@ function _M:release(key)
   semaphores[key] = nil
 end
 
+--- run a new function, callback using locks on the given key
+-- @tparam string key: key for the semaphore
+-- @tparam int timeout: timeout for getting the lock before raise an error.
+-- @tparam function callback: function to execute if the lock is adquired correctly
```
davidor: typo acquired

davidor: Maybe we should also add `...` to the documentation of the function. I can't remember the syntax now.

gateway/src/resty/synchronization.lua
```diff
@@ -53,4 +53,28 @@ function _M:release(key)
...
+  local ok, err = sema:wait(timeout)
+  if not ok then
+    ngx.log(ngx.WARN, 'failed to acquire lock on key: ', key, ' error: ', err)
+    return false, nil
+  end
+
+  local task = safe_task.new(callback)
+  local ret, result = task:execute(...)
+
+  if ok then
```
davidor: Can we name this `lock_acquired` or similar? I think that would make it perfectly clear what this refers to.

gateway/src/resty/synchronization.lua
```diff
@@ -53,4 +53,28 @@ function _M:release(key)
...
+    self.release(key)
+    sema:post()
+  end
+
+  return ret, result
```
davidor: Are 2 enough? According to the docs of the function, `execute` returns 2 or 3 params depending on whether there was an error or not.

t/configuration-loading-lazy.t
```diff
@@ -95,3 +96,30 @@ GET /t?user_key=fake
   | ]
 ]
 
+=== TEST 4: load invalid json
+To validate that process does not died with invalid config
```
davidor: I'm not sure this test is what we need. I checked that it passes without applying the patch. Also, this description seems to contradict the one in TEST 2.

spec/synchronization.lua
```diff
@@ -0,0 +1,78 @@
...
+    synchronization = require('resty.synchronization').new(1)
+  end)
+
+  it("Valid run execution", function()
+    local callback = spy.new(function() return 1 end)
+
+    local ret, result = synchronization:run(key, 10, callback)
+    assert.same(ret, true)
+    assert.same(result, {1})
+    assert.spy(callback).was.called()
```
davidor: o we need the spies? I think that checking the results of `run` might be enough.

gateway/src/apicast/configuration_loader.lua
```diff
@@ -196,33 +196,30 @@ function lazy.init(configuration)
...
+  if _M.configured(configuration, host) then
+    return configuration
   end
 
-  if ok then
-    synchronization:release(host)
-    sema:post()
-  else
-    ngx.log(ngx.WARN, 'failed to acquire lock to lazy load: ', host, ' error: ', err)
+  local ret, result = synchronization:run(host, 15, lazy.load_config, configuration, host)
```
davidor: Can we define a constant with this 15? I know it's not a problem introduced in this PR, but looks like a good opportunity to fix this.

gateway/src/apicast/configuration_loader.lua
```diff
@@ -196,33 +196,30 @@ function lazy.init(configuration)
   configuration.configured = true
 end
 
+function lazy.load_config(configuration, host)
```
davidor: This method is exposed. It should be private, right?

gateway/src/apicast/configuration_loader.lua
```diff
@@ -196,33 +196,30 @@ function lazy.init(configuration)
...
   end
 
-  if ok then
-    synchronization:release(host)
-    sema:post()
-  else
-    ngx.log(ngx.WARN, 'failed to acquire lock to lazy load: ', host, ' error: ', err)
+  local ret, result = synchronization:run(host, 15, lazy.load_config, configuration, host)
+  if not ret then
+    ngx.log(ngx.WARN, 'failed to acquire lock to lazy load: ', host, ' error: ', result)
```
davidor: this error message is also shown in the synchronization module. Maybe here we should show a more specific error here.

spec/synchronization.lua
```diff
@@ -0,0 +1,78 @@
...
+  it("Validates that run send correctly the arguments", function()
+    local callback = spy.new(function(key, value) return key, value end)
+
+    local ret, result = synchronization:run(key, 4, callback, "bob", "alice")
+    assert.same(ret, true)
+    assert.same(result, {"bob", "alice"})
+    assert.spy(callback).was_called()
+  end)
+
+  it("Validates that run does not breaks if a execption on the callback", function()
```
davidor: typos: `break` `exception`.

spec/synchronization.lua
eloycoto: It's just to make the test completed and easy to understand if someone review the code in future.

t/configuration-loading-lazy.t
eloycoto: just empty whitespace on the end of the line. I think that it's good to have it clean :-)

t/configuration-loading-lazy.t
eloycoto: test 2 validates that the test returns a 404, but the json.decode of an empty string is not going to fail. But json decode of an invalid json will break.

eloycoto: Made this test, looks more legitimate and it's fails: 
```
=== TEST 5: Invalid json structure
To validate that process does not died if the server return an invalid json configuration
--- main_config
env THREESCALE_PORTAL_ENDPOINT=http://127.0.0.1:$TEST_NGINX_SERVER_PORT/;
env APICAST_CONFIGURATION_LOADER=lazy;
--- http_config
  include $TEST_NGINX_UPSTREAM_CONFIG;
  lua_package_path "$TEST_NGINX_LUA_PATH";
--- config
  include $TEST_NGINX_BACKEND_CONFIG;

  location = /admin/api/nginx/spec.json {
    try_files /config.json =404;
  }

  location /api/ {
    echo "all ok";
  }
--- request
GET /t?user_key=fake
--- error_code: 404
--- user_files
>>> config.json
{
    "services": [{
      "id": 1,
      "backend_version": 1,
      "proxy": {
        "api_backend": ["test1", "test2"],
        "proxy_rules": [
          { "pattern": "/t", "http_method": "GET", "metric_system_name": "test" }
        ]
      }
    }]
}
```

spec/synchronization.lua
davidor: :+1:

gateway/src/apicast/configuration_loader.lua
davidor: Just checking. Did you forget to change this or decided not to?

gateway/src/apicast/configuration_loader.lua
```diff
@@ -9,7 +9,7 @@ local oidc_loader = require 'apicast.configuration_loader.oidc'
 local util = require 'apicast.util'
 local env = require('resty.env')
 local resty_url = require('resty.url')
-local synchronization = require('resty.synchronization').new(1)
+local synchronization = require('resty.synchronization').new(nil,1)
```
davidor: Why did this change?

gateway/src/resty/synchronization.lua
```diff
@@ -53,4 +53,33 @@ function _M:release(key)
...
+--- run a new function, callback using locks on the given key
+-- @tparam string key: key for the semaphore
+-- @tparam int timeout: timeout for getting the lock before raise an error.
+-- @tparam function callback: function to execute if the lock is acquired correctly
+-- @param ...: the variable number of arguments that are going to be send to the callback function.
+function _M:run(key, timeout, callback, ...)
+  local sema, err = self:acquire(key)
+  if err ~= key then
+    ngx.log(ngx.WARN, 'failed to acquire lock on key: ', key, ' error: ', err)
+    return false, nil
```
davidor: There's no need for the `, nil`

gateway/src/resty/synchronization.lua
davidor: When there's an error, `execute` returns 3 params and the error msg is the third one, so I think we'll lose it.
https://github.com/3scale/APIcast/blob/6b42baa85c9bde860c4c3bf70e8af78b48a710cf/gateway/src/resty/concurrent/safe_task_executor.lua#L39

spec/synchronization.lua
```diff
@@ -0,0 +1,78 @@
...
+  it("Validates that run send correctly the arguments", function()
+    local callback = spy.new(function(key, value) return key, value end)
+
+    local ret, result = synchronization:run(key, 4, callback, "bob", "alice")
+    assert.same(ret, true)
+    assert.same(result, {"bob", "alice"})
+    assert.spy(callback).was_called()
+  end)
+
+  it("Validates that run does not break if a exception on the callback", function()
```
davidor: but there's no exception in this test, right?

spec/synchronization.lua
```diff
@@ -0,0 +1,78 @@
+describe("Synchronization", function()
```
davidor: this file needs to be named `sychronization_spec.lua`. Without the `_spec` our Makefile will not run the test in this file.

PR #1051, opened 2019-05-29
[THREESCALE-2590] Add matches operation to the Rate Limit policy

davidor:
Fixes [THREESCALE-2590](https://issues.jboss.org/browse/THREESCALE-2590)

Apart from that, I fixed a couple of minor things in the test files that I needed to modify to implement the patch.

### DIFF
--- a/gateway/src/apicast/policy/rate_limit/apicast-policy.json
+++ b/gateway/src/apicast/policy/rate_limit/apicast-policy.json
@@ -63,8 +63,9 @@
             "type": "string"
           },
           "op": {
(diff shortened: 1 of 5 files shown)


gateway/src/apicast/policy/rate_limit/apicast-policy.json
```diff
@@ -64,7 +64,7 @@
           },
           "op": {
             "type": "string",
-            "enum": ["==", "!="]
+            "enum": ["==", "!=", "matches"]
```
mikz: We should describe what is the format of "matches".

davidor: Added :+1:

PR #1058, opened 2019-06-05
[THREESCALE-2340] Make response codes compatible with reporting threads

davidor:
Fixes [THREESCALE-2340](https://issues.jboss.org/browse/THREESCALE-2340)

This PR makes response codes compatible with reporting threads.

The only problem was that the status code (`ngx.var.status`) was checked in a timer and timers do not have access to ngx.var.*.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added the option to manage access based on method on Keycloak Policy. [THREESCALE-2236](https://issues.jboss.org/browse/THREESCALE-2236) [PR #1039](https://github.com/3scale/APIcast/pull/1039)
 - The Rate Limit policy now supports conditions defined with the "matches" operation. [PR #1051](https://github.com/3scale/APIcast/pull/1051), [THREESCALE-2590](https://issues.jboss.org/browse/THREESCALE-2590)
 - Upgrade OpenResty to 1.15.8.1 release.[PR #1049](https://github.com/3scale/APIcast/pull/1049), [THREESCALE-2200](https://issues.jboss.org/browse/THREESCALE-2200)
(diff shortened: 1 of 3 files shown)


andrewdavidmackenzie:
Is the outcome of this change @davidor that enabling response-codes would NOT disable the async reporting, hence avoiding the thruput impact of that?

davidor:
Yes. Just to be clear, the current code allows async reporting with response codes but only with 1 thread.

The env `APICAST_REPORTING_THREADS` is used to report using a pool of several threads. However, the current code ignores it when response codes are enabled. This PR changes that, now it's possible to have both `APICAST_REPORTING_THREADS > 1` and response codes.

andrewdavidmackenzie:
OK. So it sounds like no changes in the docs (on ENV vars and behaviour are needed).

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added the option to manage access based on method on Keycloak Policy. [THREESCALE-2236](https://issues.jboss.org/browse/THREESCALE-2236) [PR #1039](https://github.com/3scale/APIcast/pull/1039)
 - The Rate Limit policy now supports conditions defined with the "matches" operation. [PR #1051](https://github.com/3scale/APIcast/pull/1051), [THREESCALE-2590](https://issues.jboss.org/browse/THREESCALE-2590)
 - Upgrade OpenResty to 1.15.8.1 release.[PR #1049](https://github.com/3scale/APIcast/pull/1049), [THREESCALE-2200](https://issues.jboss.org/browse/THREESCALE-2200)
+- Now it is possible to report status codes when using reporting threads [PR #1058](https://github.com/3scale/APIcast/pull/1058)
```
eloycoto: Quick one, shouldn't be here a JIRA link?

davidor: added :+1:

PR #1062, opened 2019-06-06
Document and test management API only with policies exposed

davidor:
This PR documents how to enable only the `/policies` endpoint in the management API. I also added a test for it. No code changes are required, this was already supported.

### DIFF
--- a/doc/parameters.md
+++ b/doc/parameters.md
@@ -84,7 +84,8 @@ Allows to set the log level for the logs related to OpenID Connect integration
 **Values:**
 
 - `disabled`: completely disabled, just listens on the port
-- `status`: only the `/status/` endpoints enabled for health checks
+- `status`: enables the `/status/` endpoint for health checks, and the `/policies` endpoint that shows the list of available policies.
+- `policies`: enables only the `/policies` endpoint.
 - `debug`: full API is open
 
 The [Management API](./management-api.md) is powerful and can control the APIcast configuration.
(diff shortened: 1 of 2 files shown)


doc/parameters.md
```diff
@@ -84,7 +84,8 @@ Allows to set the log level for the logs related to OpenID Connect integration
 **Values:**
 
 - `disabled`: completely disabled, just listens on the port
-- `status`: only the `/status/` endpoints enabled for health checks
+- `status`: enables the `/status/` endpoints enabled for health checks and `/policies` that shows the list of available policies.
```
```suggestion
- `status`: enables the `/status/` endpoints for health checks and `/policies` that show the list of available policies.
```

davidor: :man_facepalming: 
Thanks!

davidor: Sorry, I didn't notice the second change. It should be "shows" instead of "show" right?

porueesq: Just to clarify: which one(s) show the list of available policies?
a. `/policies`
b. health check and `/policies`

In both cases, the verb is `show`, because of plural. However, the sentence might need to be rephrased to avoid ambiguity.

davidor: It's `a)`, but I think it should be singular because it's the `/policies` endpoint.

porueesq: Suggestion:
status`: enables the `/status/` endpoints for health checks, and the `/policies` endpoint that shows the list of available policies.

davidor: :+1:

Issue #1059, opened 2019-06-05
BACKEND_ENDPOINT_OVERRIDE parameter has an incorrect description


avilatusell:
The description of [BACKEND_ENDPOINT_OVERRIDE](https://github.com/3scale/APIcast/blob/v3.5.0/doc/parameters.md#backend_endpoint_override) states that this environment variable is `Useful when deploying outside OpenShift deployed AMP`. Currently the value is already set to the route of backend-listener so the environment variable is no longer required when deploying APIcast outside the 3scale project. This has also been reported in [THREESCALE-2462](https://issues.jboss.org/browse/THREESCALE-2462).

However, when deploying APIcast in the another project but in the same OpenShift cluster that contains 3scale API Manager, BACKEND_ENDPOINT_OVERRIDE can be used to override the public route for backend-listener and set it to the OpsneShift backend-listener service address.

mikz:
It used to be the case. But then we switched the logic entirely and now it is useful when deploying inside the same cluster, so it can be used to use internal service over external Route.

davidor:
Fixed in #1063

PR #1063, opened 2019-06-06
Backend enpoint override description update

avilatusell:
This PR modifies the description of the BACKEND_ENDPOINT_OVERRIDE environment variable. 
Fixes the issue [#1059](https://github.com/3scale/APIcast/issues/1059).

### DIFF
--- a/doc/parameters.md
+++ b/doc/parameters.md
@@ -240,7 +240,8 @@ This is the value that will be used in the nginx `worker_processes` [directive](
 
 ### `BACKEND_ENDPOINT_OVERRIDE`
 
-URI that overrides the backend endpoint from the configuration, by default it is the external route. Is is useful when deploying APIcast into the same OpenShift cluster than 3scale as to use the internal hostname of the backend listener service instead of the public route. 
+URI that overrides the backend endpoint. By default, it is the external route. 
+This parameter is useful when deploying APIcast into the same OpenShift cluster than 3scale, as when using the internal hostname of the backend listener service instead of the public route. 
 
 **Example**: `http://backend-listener.<AMP_PROJECT>.svc.cluster.local:3000`.
 
--- a/doc/parameters.md
+++ b/doc/parameters.md
@@ -243,7 +243,7 @@ This is the value that will be used in the nginx `worker_processes` [directive](
 URI that overrides the backend endpoint. By default, it is the external route. 
 This parameter is useful when deploying APIcast into the same OpenShift cluster than 3scale, as when using the internal hostname of the backend listener service instead of the public route. 
 
-**Example**: `http://backend-listener.<AMP_PROJECT>.svc.cluster.local:3000`.
+**Example**: `http://backend-listener.<3scale-namespace>.svc.cluster.local:3000`
 
 ### `OPENSSL_VERIFY`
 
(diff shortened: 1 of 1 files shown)


doc/parameters.md
```diff
@@ -240,9 +240,7 @@ This is the value that will be used in the nginx `worker_processes` [directive](
 
 ### `BACKEND_ENDPOINT_OVERRIDE`
 
-URI that overrides backend endpoint from the configuration. Useful when deploying outside OpenShift deployed AMP.
-
-**Example**: `https://backend.example.com`.
+URI that overrides the backend endpoint from the configuration, by default it is the external route. Is is useful when deploying APIcast into the same OpenShift cluster than 3scale as to use the internal hostname of the backend listener service instead of the public route.
```
kevprice83: > ...by default it is the public openshift route. 
>...same OpenShift cluster as 3scale backend and use the internal service hostname of `backend-listener`.

doc/parameters.md
```diff
@@ -240,9 +240,9 @@ This is the value that will be used in the nginx `worker_processes` [directive](
 
 ### `BACKEND_ENDPOINT_OVERRIDE`
 
-URI that overrides backend endpoint from the configuration. Useful when deploying outside OpenShift deployed AMP.
+URI that overrides the backend endpoint from the configuration, by default it is the external route. Is is useful when deploying APIcast into the same OpenShift cluster than 3scale as to use the internal hostname of the backend listener service instead of the public route.
```
```suggestion
URI that overrides the backend endpoint. By default, it is the external route. 
This parameter is useful when deploying APIcast into the same OpenShift cluster than 3scale, as when using the internal hostname of the backend listener service instead of the public route. 
```

doc/parameters.md
```diff
@@ -242,6 +242,8 @@ This is the value that will be used in the nginx `worker_processes` [directive](
 
 URI that overrides the backend endpoint from the configuration, by default it is the external route. Is is useful when deploying APIcast into the same OpenShift cluster than 3scale as to use the internal hostname of the backend listener service instead of the public route. 
 
+**Example**: `http://backend-listener.<AMP_PROJECT>.svc.cluster.local:3000`.
```
kevprice83: `http://backend-listener.<3scale-namespace>.svc.cluster.local:3000`

avilatusell:
```suggestion
**Example**: `http://backend-listener.<AMP_PROJECT>.svc.cluster.local:3000`.
**Example**: `http://backend-listener.<3scale-namespace>.svc.cluster.local:3000`
```

doc/parameters.md
```diff
@@ -240,9 +240,10 @@ This is the value that will be used in the nginx `worker_processes` [directive](
 
 ### `BACKEND_ENDPOINT_OVERRIDE`
 
-URI that overrides backend endpoint from the configuration. Useful when deploying outside OpenShift deployed AMP.
+URI that overrides the backend endpoint. By default, it is the external route.
```
davidor: I'm not sure it's a good idea to focus so much on Openshift given that this is valid for any kind of deployment.

kevprice83: That's the whole point of this PR as I see it. Could you please share what other use cases there are for this parameter then? Also if it's not on OpenShift I don't see when else this would be used.

mikz: Local development for example. Or exposing 3scale backend via a different route. Using geo routing, ...

PR #1064, opened 2019-06-07
Reset _G table to avoid invalid warning message.

eloycoto:
Clean warning on openresty 1.15.8.1, where some global variables are
set, and a warning message is show during startup outside apicast
packages.

### DIFF
--- a/gateway/libexec/boot.lua
+++ b/gateway/libexec/boot.lua
@@ -1,8 +1,12 @@
+-- Clean warning on openresty 1.15.8.1, where some global variables are set,
+-- and a warning message is show during startup outside apicast packages.
+-- Code related: https://github.com/openresty/lua-nginx-module/blob/61e4d0aac8974b8fad1b5b93d0d3d694d257d328/src/ngx_http_lua_util.c#L795-L839
+(getmetatable(_G) or {}).__newindex = nil
+
 package.path = package.path .. ";./src/?.lua;"
 require('apicast.loader')
 
 local configuration = require 'apicast.configuration_loader'
-
 local config = configuration.boot()
 
 ngx.say(config)


mikz:
What global variable is being set?

eloycoto:
```2019/06/10 12:54:54 [warn] 2385#2385: *2 [lua] _G write guard:12: writing a global lua variable ('yaml') which may lead to race conditions between concurrent requests, so prefer the use of 'l
ocal' variables
stack traceback:
        [C]: at 0x7f76d4276d00
        [C]: in function 'require'
        ./lua_modules/share/lua/5.1/lyaml/init.lua:34: in main chunk
        [C]: in function 'require'
        /home/centos/gateway/src/resty/yaml.lua:6: in main chunk
        [C]: in function 'require'
        ...os/gateway/src/apicast/configuration_loader/data_url.lua:9: in main chunk
        [C]: in function 'require'
        /home/centos/gateway/src/apicast/configuration_loader.lua:5: in main chunk
        [C]: in function 'require'
        /home/centos/gateway/libexec/boot.lua:4: in function 'file_gen'
        init_worker_by_lua:51: in function <init_worker_by_lua:49>
        [C]: in function 'xpcall'
        init_worker_by_lua:58: in function <init_worker_by_lua:56>, context: ngx.timer
```

mikz:
Hm. The problem, in general, is `resty-cli` because it evaluates everything in a timer. This code is loaded in also when APIcast is supposed to serve requests, but it should be disabled only when executed from `resty-cli`. Maybe pass some `-e` option to https://github.com/3scale/APIcast/blob/master/gateway/libexec/run or just edit https://github.com/3scale/APIcast/blob/master/gateway/libexec/boot.lua ?

eloycoto:
ok, edit boot also works for me. Thanks

gateway/libexec/boot.lua
```diff
@@ -1,3 +1,8 @@
+-- Clean warning on openresty 1.15.8.1, where some global variables are set,
+-- and a warning message is show during startup outside apicast packages.
+-- Code related: https://github.com/openresty/lua-nginx-module/blob/61e4d0aac8974b8fad1b5b93d0d3d694d257d328/src/ngx_http_lua_util.c#L795-L839
+getmetatable(_G).__newindex = nil
```
mikz: This will crash on OpenResty < 1.15. But I guess that is not that big of a deal.

davidor: It's this is the only incompatibility we have introduced with OpenResty 1.13.x. It might be worth to use `(getmetatable(_G) or {})` instead like in lua-rover.

Issue #957, opened 2018-11-20
Make development failed


y-tabata:
##### Version

CentOS Linux release 7.5.1804 (Core)

docker version:
Client:
 Version:         1.13.1
 API version:     1.26
 Package version: docker-1.13.1-75.git8633870.el7.centos.x86_64
 Go version:      go1.9.4
 Git commit:      8633870/1.13.1
 Built:           Fri Sep 28 19:45:08 2018
 OS/Arch:         linux/amd64

Server:
 Version:         1.13.1
 API version:     1.26 (minimum version 1.12)
 Package version: docker-1.13.1-75.git8633870.el7.centos.x86_64
 Go version:      go1.9.4
 Git commit:      8633870/1.13.1
 Built:           Fri Sep 28 19:45:08 2018
 OS/Arch:         linux/amd64
 Experimental:    false

docker-compose version:
docker-compose version 1.18.0, build 8dd22a9
docker-py version: 2.6.1
CPython version: 3.4.9
OpenSSL version: OpenSSL 1.0.2k-fips  26 Jan 2017

##### Description

The `make development` failed in my environment.
Firstly, the `--detach` option is not supported.

```
[root@app2 apicast]# make development
docker-compose -f docker-compose-devel.yml up --detach
Builds, (re)creates, starts, and attaches to containers for a service.

Unless they are already running, this command also starts any linked services.

The `docker-compose up` command aggregates the output of each container. When
the command exits, all containers are stopped. Running `docker-compose up -d`
starts the containers in the background and leaves them running.

If there are existing containers for a service, and the service's configuration
or image was changed after the container's creation, `docker-compose up` picks
up the changes by stopping and recreating the containers (preserving mounted
volumes). To prevent Compose from picking up changes, use the `--no-recreate`
flag.

If you want to force Compose to stop and recreate all containers, use the
`--force-recreate` flag.

Usage: up [options] [--scale SERVICE=NUM...] [SERVICE...]

Options:
    -d                         Detached mode: Run containers in the background,
                               print new container names. Incompatible with
                               --abort-on-container-exit and --timeout.
    --no-color                 Produce monochrome output.
    --no-deps                  Don't start linked services.
    --force-recreate           Recreate containers even if their configuration
                               and image haven't changed.
                               Incompatible with --no-recreate.
    --no-recreate              If containers already exist, don't recreate them.
                               Incompatible with --force-recreate.
    --no-build                 Don't build an image, even if it's missing.
[...]

PR #1067, opened 2019-06-11
fix for make development failures - github issue #957

sekharvajjhala:
This PR fixes the development failures reported in  [https://github.com/3scale/apicast/issues/957](#957) using separate docker-compose files for volume configuration and -d option instead of --detach command line option for docker-compose.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,19 @@ RUNTIME_IMAGE ?= $(BUILDER_IMAGE)-runtime
 
 DEVEL_IMAGE ?= apicast-development
 DEVEL_DOCKERFILE ?= Dockerfile-development
+
 DEVEL_DOCKER_COMPOSE_FILE ?= docker-compose-devel.yml
+DEVEL_DOCKER_COMPOSE_VOLMOUNT_MAC_FILE ?= docker-compose-devel-volmount-mac.yml
+DEVEL_DOCKER_COMPOSE_VOLMOUNT_DEFAULT_FILE ?= docker-compose-devel-volmount-default.yml
+
+os = "$(shell uname -s)"
+
+# if running on Mac
+ifeq ($(os),"Darwin")
+    DEVEL_DOCKER_COMPOSE_VOLMOUNT_FILE = $(DEVEL_DOCKER_COMPOSE_VOLMOUNT_MAC_FILE)
+else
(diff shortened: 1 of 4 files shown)


kevprice83:
@sekharvajjhala I think you might need to include an entry in the CHANGELOG

Makefile
```diff
@@ -18,7 +18,19 @@ RUNTIME_IMAGE ?= $(BUILDER_IMAGE)-runtime
...
 DEVEL_IMAGE ?= apicast-development
 DEVEL_DOCKERFILE ?= Dockerfile-development
+
 DEVEL_DOCKER_COMPOSE_FILE ?= docker-compose-devel.yml
+DEVEL_DOCKER_COMPOSE_VOLMOUNT_MAC_FILE ?= docker-compose-devel-volmount-mac.yml
+DEVEL_DOCKER_COMPOSE_VOLMOUNT_DEFAULT_FILE ?= docker-compose-devel-volmount-default.yml
+
+os = "$(shell uname -s)"
+
+# if running on Mac
```
davidor: Wouldn't it be more correct to detect if running on a rhel-based distro? As far as I know, those are the ones that have problems with the `:cached ` and `:delegated` modes.

sekharvajjhala: The consistency option (:delegated , :cached) according to this https://docs.docker.com/v17.09/engine/admin/volumes/bind-mounts/#choosing-the--v-or-mount-flag should be ignored on all platforms other than Mac. I did not want to assume that other distros other than are ignoring these options and so may run into this problem.And it is not obvious how to check if a specific non mac platform distro is ignoring these options. Since these options are specific to Mac, I thought it would be more logical to separate them out into a mac specific docker-compose file and use a default default docker-compose files.

davidor: I see. I was wrong :+1:

Makefile
```diff
@@ -190,12 +202,13 @@ development: USER := $(shell id -u $(USER))
 endif
 development: .docker/lua_modules .docker/local .docker/cpanm .docker/vendor/cache
 development: ## Run bash inside the development image
-	- $(DOCKER_COMPOSE) -f $(DEVEL_DOCKER_COMPOSE_FILE) up --detach
+	@echo "Running on $(os)"
```
davidor: Not sure this echo is needed. Does it add any useful info for the user?

sekharvajjhala: Only useful is helping in indicating whether platform specific docker-compose file got loaded (mac or non mac) which in turn would tell whether consistency options are being loaded or not. I agree it is not clear at all since this is being inferred which is not obvious to anyone but the code's author.

PR #1070, opened 2019-06-14
JWT claim policy

eloycoto:
Hi, 

This PR contains the JWT claim policy, not a big change at all, multiple commits all should be descriptive enough. 

Example policy: 

```
{
  "services": [
    {
      "id": 200,
      "endpoint": "http://localhost:8080",
      "backend_version": "1",
      "backend_version": "oauth",
      "proxy": {
        "authentication_method": "oidc",
        "oidc_issuer_endpoint": "http://192.168.15.103:10000/auth/realms/3scale",
        "service_backend_version": "1",
        "hosts": [
          "one"
        ],
        "api_backend": "https://echo-api.3scale.net:443",
        "proxy_rules": [
          {
            "pattern": "/",
            "http_method": "GET",
            "metric_system_name": "hits",
            "delta": 1
          }
        ],
        "policy_chain": [
          {
          },
          {
            "name": "apicast.policy.jwt_claim_check",
            "configuration": {
                "error_message": "Invalid JWT check",
                "operations": [
                    {"op": "==", "match": "acr", "match_type": "plain", "value": "10", "check_type": "whitelist"},
                    {"op": "==", "match": "{{upn}}", "match_type": "liquid", "value": "eloy", "check_type": "whitelist"}
                ]
            }
          }
        ]
      }
    }
  ]
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Upgrade OpenResty to 1.15.8.1 release.[PR #1049](https://github.com/3scale/APIcast/pull/1049), [THREESCALE-2200](https://issues.jboss.org/browse/THREESCALE-2200)
 - Now it is possible to report status codes when using reporting threads [PR #1058](https://github.com/3scale/APIcast/pull/1058), [THREESCALE-2340](https://issues.jboss.org/browse/THREESCALE-2340)
 - New Retry policy. Allows to configure retries for calls to the upstream APIs [PR #1057](https://github.com/3scale/APIcast/pull/1057), [THREESCALE-1517](https://issues.jboss.org/browse/THREESCALE-1517)
+- JWT claim policy. Allows to allow/deny traffic based on JWT claim constraint [PR #1070](https://github.com/3scale/APIcast/pull/1070), [THREESCALE-2265](https://issues.jboss.org/browse/THREESCALE-2265)
 
 ### Fixed
 
--- a/gateway/src/apicast/policy/jwt_claim_check/apicast-policy.json
+++ b/gateway/src/apicast/policy/jwt_claim_check/apicast-policy.json
@@ -1,0 +1,113 @@
+{
+  "$schema": "http://apicast.io/policy-v1/schema#manifest#",
+  "name": "JWT Claim Check",
+  "summary": "Allow or deny traffic based on a JWT claim",
+  "description": [
+    "This Policy allow to block traffic based on a JWT token.",
+    "To verify any JWT claim can be used and can be compared ",
+    "using plain or liquid filters."
+  ],
+  "version": "builtin",
+  "configuration": {
+    "definitions": {
+      "value_type": {
+        "$id": "#/definitions/value_type",
+        "type": "string",
+        "oneOf": [{
+          "enum": ["plain"],
+          "title": "Evaluate as plain text."
+        }, {
+          "enum": ["liquid"],
+          "title": "Evaluate as liquid."
+        }]
+      }
+    },
+    "type": "object",
+    "properties": {
+      "error_message": {
+        "title": "Error message",
+        "description": "Error message to show to user when traffic is blocked",
+        "type": "string"
+      },
+      "rules": {
+        "type": "array",
+        "items": {
+          "properties": {
+            "combine_op": {
+              "type": "string",
+              "enum": ["and", "or"],
+              "default": "and"
+            },
+            "methods": {
+              "description": "Allowed methods",
+              "type": "array",
+              "default": ["ANY"],
+              "items": {
+                "type": "string",
+                "enum": [
+                  "ANY",
+                  "GET",
+                  "HEAD",
+                  "POST",
+                  "PUT",
+                  "DELETE",
+                  "PATCH",
+                  "OPTIONS",
+                  "TRACE",
+                  "CONNECT"
+                ]
+              }
+            },
+            "resource": {
+              "type": "string",
+              "description": "Resource controlled by the rule. This is the same format as Mapping Rules. This matches from the beginning of the string and to make an exact match you need to use '$' at the end."
+            },
+            "resource_type": {
+              "description": "How to evaluate 'resource' field",
+              "$ref": "#/definitions/value_type"
+            },
+            "operations": {
+              "type": "array",
+              "items": {
+                "required": [
+                  "jwt_claim",
+                  "jwt_claim_type",
+                  "op",
+                  "value"
+                ],
+                "type": "object",
+                "properties": {
+                  "jwt_claim": {
+                    "description": "String to get JWT claim", 
+                    "type": "string"
+                  },
+                  "jwt_claim_type": {
+                    "description": "How to evaluate 'jwt_claim' value",
+                    "$ref": "#/definitions/value_type"
+                  },
+                  "op": {
+                    "description": "Match operation to compare JWT claim with the provided value. In case that a not a number is in use in numeric comparison, the value will be transformed to 0.",
+                    "type": "string",
+                    "enum": [
+                      "==",
+                      "!=",
+                      "matches"
+                    ]
+                  },
+                  "value": {
+                    "description": "Value to compare the retrieved JWT claim",
+                    "type": "string"
+                  },
+                  "value_type": {
+                    "description": "How to evaluate 'jwt_claim' value",
+                    "$ref": "#/definitions/value_type"
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}
--- a/gateway/src/apicast/policy/jwt_claim_check/jwt_claim_check.lua
+++ b/gateway/src/apicast/policy/jwt_claim_check/jwt_claim_check.lua
@@ -1,0 +1,96 @@
+local policy = require('apicast.policy')
+local _M = policy.new('JWT check policy')
+
+local Condition = require('apicast.conditions.condition')
+local MappingRule = require('apicast.mapping_rule')
+local Operation = require('apicast.conditions.operation')
+local TemplateString = require('apicast.template_string')
+
+local ipairs = ipairs
+
+local new = _M.new
+
+local default_error_message = "Request blocked due to JWT claim policy"
+local default_template_type = 'plain'
+local default_allowed_methods = { MappingRule.any_method }
+
+local function deny_request(error_msg)
+  ngx.status = ngx.HTTP_FORBIDDEN
+  ngx.say(error_msg)
+  ngx.exit(ngx.status)
+end
+
+function _M.new(config)
+  local self = new(config)
+  self.error_message = config.error_message or default_error_message
+  self.rules = {}
+  for _, rule in ipairs(config.rules) do
+    local conditions = {}
+    for _, condition in ipairs(rule.operations) do
+      if condition.jwt_claim_type == "plain" then
+        -- Due to this need to be fetched from the JWT claim, render  the match
+        -- as liquid to be able to fetch the info from the JWT_claim
+        condition.jwt_claim = "{{"..condition.jwt_claim.."}}"
+      end
+
+      table.insert( conditions,
+        Operation.new(
+          condition.jwt_claim, "liquid",
+          condition.op,
+          condition.value, condition.value_type or default_template_type))
+    end
+    table.insert( self.rules, {
+      condition = Condition.new(conditions, rule.combine_op),
+      methods = rule.methods or default_allowed_methods,
+      resource = TemplateString.new(
+        rule.resource,
+        rule.resource_type or default_template_type),
+    })
+  end
+
+  return self
+end
+
+-- is_rule_denied_request returns true if the request need to be blocked based
+-- on a provided rule with the request context.
+-- This function will only work if the request match on resource and in one of
+-- the methods, the methods that are not  defined in the rule will be allowed.
+local function is_rule_denied_request(rule, context)
+
+  local uri = ngx.var.uri
+  local request_method =  ngx.req.get_method()
+
+  local resource = rule.resource:render(context)
+  local mapping_rule_match = false
+
+  for _, method  in ipairs(rule.methods) do
+    local mapping_rule = MappingRule.from_proxy_rule({
+      http_method = method,
+      pattern = resource,
+      querystring_parameters = {},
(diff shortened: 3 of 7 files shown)


gateway/src/apicast/conditions/operation.lua
```diff
@@ -23,7 +23,13 @@ local evaluate_func = {
   -- false otherwise.
   ['matches'] = function(left, right)
     return (match(tostring(left), tostring(right)) and true) or false
-  end
+  end,
+
+  -- Numerics entries
```
davidor: I'm not sure it's a good idea to convert to 0 any strings that cannot be converted into numbers.

I can see how this can become confusing in policies that expose these operations. Some users might assume that comparisons between strings work and that 'a' < 'b' should evaluate to true, for example.

Should we validate in each policy that these operations are always called with integers and discard the operation if it contains strings?

gateway/src/apicast/policy/jwt_claim_check/jwt_claim_check.lua
```diff
@@ -0,0 +1,55 @@
...
+    table.insert(self.conditions, {
+      op = Operation.new(condition.match, "liquid", condition.op, condition.value, 'plain'),
+      check_type = condition.check_type})
+  end
+  return self
+end
+
+
+function _M:access(context)
+  for _,operation in pairs(self.conditions) do
```
davidor: I think it would be good to support 'conditions' like in other policies like the routing one.
That would be more consistent, and also would allow us to combine operations with `and` and `or`.

gateway/src/apicast/policy/jwt_claim_check/jwt_claim_check.lua
```diff
@@ -0,0 +1,55 @@
...
+
+  return status
+end
+
+function _M.new(config)
+  local self = new(config)
+  self.error_message = config.error_message or default_error_message
+
+  self.conditions = {}
+  for _,condition in ipairs(config.operations) do
```
davidor: Can you declare `ipairs` as a local at the beginning of the file? Same with `pairs` used below.

gateway/src/apicast/policy/jwt_claim_check/apicast-policy.json
```diff
@@ -0,0 +1,86 @@
...
+                "==",
+                "!=",
+                "matches",
+                ">", 
+                ">=",
+                "<",
+                "<="
+              ]
+            },
+            "value": {
```
davidor: Maybe we should support evaluating `value` as liquid as well. We do this in policies like the routing one.
More specifically, one of the options available in the routing policy consists of routing based on a jwt claim and the policy allows users to compare it with a value that can be evaluated as liquid. I think this is basically the same use case as here.

spec/policy/jwt_claim_check/jwt_claim_check_spec.lua
```diff
@@ -0,0 +1,127 @@
...
+      jwt_check:access(context)
+      assert.same(ngx.status, 403)
+    end)
+
+  end)
+
+  describe("Text operations", function()
+    it("equal operation", function()
+      local jwt_check = JWTClaimCheckPolicy.new({operations={
+          {op="==", match="{{foo}}", match_type="liquid", value="foo"},
```
davidor: I think this is a bit confusing. The name of the claim and its value are the same, so it's not immediately obvious what is being tested.

spec/policy/jwt_claim_check/jwt_claim_check_spec.lua
```diff
@@ -0,0 +1,127 @@
...
+        error_message=message})
+      jwt_check:access(context)
+      assert.same(ngx.status, 403)
+      assert.spy(ngx.say).was.called()
+      assert.spy(ngx.say).was.called_with(message)
+  end)
+
+  describe("blacklist mode", function()
+
+    it("Stop processing if match", function()
```
davidor: This would be more complete if we added a second operation that matches, right?

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Upgrade OpenResty to 1.15.8.1 release.[PR #1049](https://github.com/3scale/APIcast/pull/1049), [THREESCALE-2200](https://issues.jboss.org/browse/THREESCALE-2200)
 - Now it is possible to report status codes when using reporting threads [PR #1058](https://github.com/3scale/APIcast/pull/1058), [THREESCALE-2340](https://issues.jboss.org/browse/THREESCALE-2340)
 - New Retry policy. Allows to configure retries for calls to the upstream APIs [PR #1057](https://github.com/3scale/APIcast/pull/1057), [THREESCALE-1517](https://issues.jboss.org/browse/THREESCALE-1517)
+- JWT claim policy. Allows to allow/deny traffic based on JWT claim policy [PR #1070](https://github.com/3scale/APIcast/pull/1070), [THREESCALE-2265](https://issues.jboss.org/browse/THREESCALE-2265)
```
davidor: `policy` at the end of the sentence sounds redundant.

davidor:
@eloycoto I think it would be good to add a brief README like we did for other policies. A brief explanation about what the policy does and some examples that show how to configure it should be enough. I think that can be helpful for other contributors and also for other teams.

eloycoto:
@davidor new changes are now pushed. Things that have been changed. 

- Policy has now a filter by resource and method. Also, OR combine operations.  
- New readme with examples/caveats has been added. 
- A few new integration tests.
- I think that 100% of the cases covered in the unit test. 

If you can have a look at the changes will be great. 

Regards.

gateway/src/apicast/policy/jwt_claim_check/readme.md
```diff
@@ -0,0 +1,134 @@
...
+      "error_message": "Invalid JWT check",
+      "rules": [
+          {
+              "operations": [
+                  {"op": "==", "jwt_claim": "role", "jwt_claim_type": "plain", "value": "admin"}
+              ],
+              "combine_op":"and",
+              "methods": ["GET"],
+              "resource": "/resource", 
+              "resource_type": "liquid"
```
davidor: Should be `plain`.

gateway/src/apicast/policy/jwt_claim_check/readme.md
```diff
@@ -0,0 +1,134 @@
...
+  "configuration": {
+      "error_message": "Invalid JWT check",
+      "rules": [
+          {
+              "operations": [
+                  {"op": "==", "jwt_claim": "role", "jwt_claim_type": "plain", "value": "role1"}
+              ],
+              "combine_op":"and",
+              "methods": ["ANY"],
+              "resource": "/resource",
```
davidor: Should be `/resource1`

gateway/src/apicast/policy/jwt_claim_check/readme.md
```diff
@@ -0,0 +1,134 @@
...
+  "configuration": {
+      "error_message": "Invalid JWT check",
+      "rules": [
+          {
+              "operations": [
+                  {"op": "==", "jwt_claim": "role_{{aud}}", "jwt_claim_type": "liquid", "value": "client1"},
+              ],
+              "combine_op":"and",
+              "methods": ["ANY"],
+              "resource": "/resource",
```
davidor: Should be `/resource1`

gateway/src/apicast/policy/jwt_claim_check/readme.md
```diff
@@ -0,0 +1,134 @@
...
+      "error_message": "Invalid JWT check",
+      "rules": [
+          {
+              "operations": [
+                  {"op": "==", "jwt_claim": "role", "jwt_claim_type": "plain", "value": "role1"}
+              ],
+              "combine_op":"and",
+              "methods": ["ANY"],
+              "resource": "/resource", 
+              "resource_type": "liquid"
```
davidor: "plain" ?

gateway/src/apicast/policy/jwt_claim_check/readme.md
```diff
@@ -0,0 +1,134 @@
...
+      "error_message": "Invalid JWT check",
+      "rules": [
+          {
+              "operations": [
+                  {"op": "==", "jwt_claim": "role_{{aud}}", "jwt_claim_type": "liquid", "value": "client1"},
+              ],
+              "combine_op":"and",
+              "methods": ["ANY"],
+              "resource": "/resource", 
+              "resource_type": "liquid"
```
davidor: "plain" ?

gateway/src/apicast/policy/jwt_claim_check/readme.md
```diff
@@ -0,0 +1,134 @@
...
+          }
+      ]
+  }
+}
+```
+
+- When you want to allow those who have the client role including the client ID
+  of the application client (the recipient of the access token) to access
+  `/resource1`. Use the `"liquid"` to specify the JWT information to the
```
davidor: ... `liquid` resource_type to ... ?

gateway/src/apicast/policy/jwt_claim_check/readme.md
```diff
@@ -0,0 +1,134 @@
...
+              ],
+              "combine_op": "and",
+              "methods": ["ANY"],
+              "resource": "/resource_{{jwt.aud}}", 
+              "resource_type": "liquid"
+          }
+      ]
+  }
+}
+```
```
davidor: Would be good to show an example with more than one rule to show that it's supported.

gateway/src/apicast/policy/jwt_claim_check/apicast-policy.json
```diff
@@ -0,0 +1,113 @@
...
+  ],
+  "version": "builtin",
+  "configuration": {
+    "definitions": {
+      "value_type": {
+        "$id": "#/definitions/value_type",
+        "type": "string",
+        "oneOf": [{
+          "enum": ["plain"],
+          "title": "Evaluate 'value' as plain text."
```
davidor: This assumes that the field using this is called `value` which is not true for all the cases in this schema. There's one called `resource`.  We should probably change the description to "Evaluate as plain text."

eloycoto: good point, thanks!

gateway/src/apicast/policy/jwt_claim_check/readme.md
```diff
@@ -0,0 +1,134 @@
+# JWT claim Check Policy
+
```
davidor: I think we need to add a note somewhere in this document like the one we added in the routing policy when using JWT: `In order to be able to route based on the value of a jwt claim, there needs to be a policy in the chain that validates the jwt and stores it in the context that the policies share.`: https://github.com/3scale/APIcast/tree/master/gateway/src/apicast/policy/routing#rule-that-uses-a-jwt-claim

spec/policy/jwt_claim_check/jwt_claim_check_spec.lua
```diff
@@ -0,0 +1,286 @@
...
+            combine_op="and",
+            methods  = {"POST", "GET"},
+            resource = ngx.var.uri
+          }
+        },
+        error_message=message})
+      jwt_check:access(context)
+      assert.same(ngx.status, 403)
+      assert.spy(ngx.say).was.called()
+      assert.spy(ngx.say).was.called_with(message)
```
davidor: This one makes the previous line redundant, right?

gateway/src/apicast/policy/jwt_claim_check/jwt_claim_check.lua
```diff
@@ -0,0 +1,100 @@
...
+local Condition = require('apicast.conditions.condition')
+local MappingRule = require('apicast.mapping_rule')
+local Operation = require('apicast.conditions.operation')
+local TemplateString = require('apicast.template_string')
+
+local ipairs = ipairs
+
+local new = _M.new
+
+local default_error_message = "Request block due JWT claim policy"
```
davidor: `blocked`

gateway/src/apicast/policy/jwt_claim_check/jwt_claim_check.lua
```diff
@@ -0,0 +1,100 @@
...
+
+      table.insert( conditions,
+        Operation.new(
+          condition.jwt_claim, "liquid",
+          condition.op,
+          condition.value, condition.value_type or default_template_type))
+    end
+    table.insert( self.rules, {
+      condition = Condition.new(conditions, rule.combine_op),
+      methods = rule.methods,
```
davidor: Should we add { `ANY` } as a default here?

gateway/src/apicast/policy/jwt_claim_check/jwt_claim_check.lua
```diff
@@ -0,0 +1,100 @@
...
+      method_scope_match = true
+      break
+    end
+  end
+
+  if not method_scope_match then
+    return false
+  end
+
+  if rule.condition:evaluate(context.jwt) then
```
davidor: The end of the method could be simplified to
```
return not rule.condition:evaluate(context.jwt)
```

right?

gateway/src/apicast/policy/jwt_claim_check/jwt_claim_check.lua
```diff
@@ -0,0 +1,100 @@
...
+-- on a provided rule with the request context.
+-- This function will only work if the request match on resource and in one of
+-- the methods, the methods that are not  defined in the rule will be allowed. 
+local function is_rule_denied_request(rule, context)
+
+  local uri = ngx.var.uri
+  local request_method =  ngx.req.get_method()
+
+  local resource = rule.resource:render(context)
+  local method_scope_match = false
```
davidor: I think the name of this var is confusing because it does not just refer to the method. It also refers to the path of the request, doesn't it?

gateway/src/apicast/conditions/operation.lua
davidor: I'm still concerned about this. I think at the very least we should add an explanation in the README.

spec/policy/jwt_claim_check/jwt_claim_check_spec.lua
eloycoto: I really liked this one, because can be called without arguments, so if that happens we know at least that was called.

davidor: :+1:

PR #1071, opened 2019-06-17
Test: Bump test apicast to 0.24

eloycoto:

### DIFF
--- a/gateway/cpanfile.snapshot
+++ b/gateway/cpanfile.snapshot
@@ -445,10 +445,10 @@ DISTRIBUTIONS
       Test::More 0
       Time::HiRes 0
       XML::Simple 0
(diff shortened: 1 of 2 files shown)


davidor:
`cpanfile.snapshot` should be updated too to lock the version.

eloycoto:
My love&hate relation with perl/cpan/releases. It's fixed now, sorry!

gateway/cpanfile.snapshot
```diff
@@ -529,6 +960,60 @@ DISTRIBUTIONS
       List::Util 1.09
       Scalar::Util 1.09
       Test::Builder 0
+  Test-Harness-3.42
```
davidor: Something does not look right. There are dependencies that were not here before, like this one.

PR #1080, opened 2019-06-28
Policy: HTTP proxy

eloycoto:
This policy adds the option to set a service to use a http proxy (As is
in use for HTTP_PROXY parameter)

This service will allow Apicast users to send the traffic over a proxy
and do request modifications in 3-party proxies.

Traffic flow:

```
	   ,-.
	   `-'
	   /|\
		|            ,-------.          ,---------.          ,----------.
	   / \           |Apicast|          |HTTPPROXY|          |APIBackend|
	  User           `---+---'          `----+----'          `----------'
	   |  GET /resource  |                   |                    |
	   | --------------->|                   |                    |
	   |                 |                   |                    |
	   |                 |  Get /resource    |                    |
	   |                 |------------------>|                    |
	   |                 |                   |                    |
	   |                 |                   |  Get /resource/    |
	   |                 |                   | - - - - - - - - - >|
	   |                 |                   |                    |
	   |                 |                   |     response       |
	   |                 |                   |<- - - - - - - - - -|
	   |                 |                   |                    |
	   |                 |     response      |                    |
	   |                 |<------------------|                    |
	   |                 |                   |                    |
	   |                 |                   |                    |
	   | <---------------|                   |                    |
	  User           ,---+---.          ,----+----.          ,----------.
	   ,-.           |Apicast|          |HTTPPROXY|          |APIBackend|
	   `-'           `-------'          `---------'          `----------'
	   /|\
		|
	   / \

```

Configuration:
```
{
  "services": [
	{
	  "id": 200,
	  "backend_version":  1,
	  "backend_authentication_type": "service_token",
	  "backend_authentication_value": "token-value",
	  "proxy": {
		"hosts": [
		  "one"
		],
		"api_backend": "https://echo-api.3scale.net/",
		"proxy_rules": [
		  {
			"pattern": "/",
			"http_method": "GET",
			"metric_system_name": "hits",
			"delta": 1
		  }
		],
		"policy_chain": [
		  {
		  },
		  {
			"configuration": {
				"all_proxy": "http://192.168.15.103:8888/",
				"https_proxy": "https://192.168.15.103:8888/",
				"http_proxy": "https://192.168.15.103:8888/"
			}
		  }
		]
	  }
	}
  ]
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Introduce possibility of specifying policy order restrictions in their schemas. APIcast now shows a warning when those restrictions are not respected [#1088](https://github.com/3scale/APIcast/pull/1088), [THREESCALE-2896](https://issues.jboss.org/browse/THREESCALE-2896)
 - Added new parameters to logging policy to allow custom access log [PR #1081](https://github.com/3scale/APIcast/pull/1089) [THREESCALE-1234](https://issues.jboss.org/browse/THREESCALE-1234)[THREESCALE-2876](https://issues.jboss.org/browse/THREESCALE-2876)
+- Added http_proxy policy to use an HTTP proxy in api_backed calls. [THREESCALE-2696](https://issues.jboss.org/browse/THREESCALE-2696) [PR #1080](https://github.com/3scale/APIcast/pull/1080)
 
 ## [3.6.0-rc1] - 2019-07-04
 
--- a/gateway/src/apicast/policy/http_proxy/Readme.md
+++ b/gateway/src/apicast/policy/http_proxy/Readme.md
@@ -1,0 +1,81 @@
+# HTTP proxy policy
+
+This policy allows users to define a HTTP proxy where the traffic will be send
+over the defined proxy, the example traffic flow is the following:
+
+```
+   ,-.
+   `-'
+   /|\
+    |            ,-------.          ,---------.          ,----------.
+   / \           |Apicast|          |HTTPPROXY|          |APIBackend|
+  User           `---+---'          `----+----'          `----------'
+   |  GET /resource  |                   |                    |
+   | --------------->|                   |                    |
+   |                 |                   |                    |
+   |                 |  Get /resource    |                    |
+   |                 |------------------>|                    |
+   |                 |                   |                    |
+   |                 |                   |  Get /resource/    |
+   |                 |                   | - - - - - - - - - >|
+   |                 |                   |                    |
+   |                 |                   |     response       |
+   |                 |                   |<- - - - - - - - - -|
+   |                 |                   |                    |
+   |                 |     response      |                    |
+   |                 |<------------------|                    |
+   |                 |                   |                    |
+   |                 |                   |                    |
+   | <---------------|                   |                    |
+  User           ,---+---.          ,----+----.          ,----------.
+   ,-.           |Apicast|          |HTTPPROXY|          |APIBackend|
+   `-'           `-------'          `---------'          `----------'
+   /|\
+    |
+   / \
+```
+
+All APIcast traffic to 3scale backend will not use the proxy, due to this only
+applies for the service and the communication between Apicast and API backend.
+
+If you want to use all traffic through a Proxy, an HTTP_PROXY env var need to be
+used.
+
+## Configuration
+
+```
+"policy_chain": [
+    {
+      "name": "apicast.policy.apicast"
+    },
+    {
+      "name": "apicast.policy.http_proxy",
+      "configuration": {
+          "all_proxy": "http://192.168.15.103:8888/",
+          "https_proxy": "https://192.168.15.103:8888/",
+          "http_proxy": "https://192.168.15.103:8888/"
+      }
+    }
+]
+```
+
+- If http_proxy or https_proxy is not defined the all_proxy will be taken. 
+
+## Caveats
+
+- This policy will disable all load-balancing policies and traffic will be
+  always send to the proxy. 
+- In case of HTTP_PROXY, HTTPS_PROXY or ALL_PROXY parameters are defined, this
+  policy will overwrite those values. 
+- Proxy connection does not support authentication.
+
+
+## Example Use case
+
+This policy was designed to be able to apply more fined grained policies and
+transformation using Apache Camel.
+
+An example project can be found
+[here](https://github.com/zregvart/camel-netty-proxy). This project is an HTTP
+Proxy that transforms to uppercase all the response body given by the API
+backend.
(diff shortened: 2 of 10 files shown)


davidor:
@eloycoto  The commit with title `Policy: Add http_proxy integration test` appears as empty here, and I can't see any integration tests in other commits, so I guess something went wrong.

mikz:
Also, the feature + tests should be in the same commit anyway :)

gateway/src/apicast/policy/http_proxy/proxy.lua
```diff
@@ -0,0 +1,47 @@
...
+
+  if config.all_proxy then
+    local err
+    self.all_proxy, err =  resty_url.parse(config.all_proxy)
+    if err then
+      ngx.log(ngx.WARN, "All proxy '", config.all_proxy, "' is not correctly defined, err:", err)
+    end
+  end
+
+  for _, proto in ipairs(proxies) do
```
davidor: `local ipairs = ipairs` ?

gateway/src/resty/http/proxy.lua
```diff
@@ -108,6 +108,12 @@ local function find_proxy_url(request)
     local uri = parse_request_uri(request)
     if not uri then return end
 
+    -- request can have a local proxy defined and env variables have lower
+    -- priority, if the proxy is defined in the request that will be used.
+    if request.proxy_uri then
```
davidor: This could be simplified:

`return request.proxy_uri or _M.find(uri)`

spec/policy/http_proxy/http_proxy_spec.lua
```diff
@@ -0,0 +1,53 @@
...
+describe('HTTP proxy  policy', function()
+  local all_proxy_val = "http://all.com"
+  local http_proxy_val = "http://plain.com"
+  local https_proxy_val = "http://secure.com"
+
+  it("http[s] proxies are defined if all_proxy is in there", function()
+    local proxy = proxy_policy.new({
+      all_proxy = all_proxy_val
+    })
+    assert.same(proxy:find_proxy("http"), resty_url.parse(all_proxy_val))
```
davidor: In all these tests, I think we should check what `export` returns instead of `find_proxy`, because `export` is what should be exposed to other modules.

t/apicast-policy-http-proxy.t
```diff
@@ -0,0 +1,172 @@
...
+--- request
+GET /?user_key=value
+--- response_body
+yay, api backend
+--- error_code: 200
+--- error_log env
+using proxy: $TEST_NGINX_HTTP_PROXY
+
+=== TEST 3: using HTTPS proxy for backend
+--- ONLY
```
davidor: :eyes:

t/apicast-policy-http-proxy.t
```diff
@@ -0,0 +1,172 @@
...
+--- backend env
+listen $TEST_NGINX_RANDOM_PORT ssl;
+
+ssl_certificate $TEST_NGINX_SERVER_ROOT/html/server.crt;
+ssl_certificate_key $TEST_NGINX_SERVER_ROOT/html/server.key;
+
+location /transactions/authrep.xml {
+  access_by_lua_block {
+    assert = require('luassert')
+    assert.equal('https', ngx.var.scheme)
```
davidor: Shouldn't all this be in `upstream` instead of `backend` ?

gateway/src/apicast/policy/http_proxy/apicast-policy.json
```diff
@@ -0,0 +1,27 @@
...
+  "summary": "Adds an HTTP proxy to the service.",
+  "description": [
+    "With this policy all the traffic for this service will be routed accross ",
+    "the defined proxy"
+  ],
+  "version": "builtin",
+  "configuration": {
+      "type": "object",
+      "properties": {
+        "all_proxy": {
```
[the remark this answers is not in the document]
eloycoto: I took care in the code, also added a unittest. 
```
it("all_proxy does not overwrite http/https proxies", function()
    local proxy = proxy_policy.new({
      all_proxy = all_proxy_val,
      http_proxy = http_proxy_val,
      https_proxy = https_proxy_val
    })
    local callback = proxy:export()

    assert.same(callback.has_http_proxy(_, http_uri), resty_url.parse(http_proxy_val))
    assert.same(callback.has_http_proxy(_, https_uri), resty_url.parse(https_proxy_val))
  end)
```

gateway/src/apicast/policy/http_proxy/Readme.md
```diff
@@ -0,0 +1,80 @@
...
+
+
+## Example Use case
+
+This policy was designed to be able to apply more fined grained policies and
+transformation using Apache Camel.
+
+An example HTTP proxy transformation can be the following project; this project
+transforms the response given by and endpoint and converts the body to
+uppercaint and convert the body to uppercase.
```
davidor: This sentence does not sound right.

gateway/src/apicast/policy/http_proxy/apicast-policy.json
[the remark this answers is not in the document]
davidor: :+1:
I think the description of `all_proxy` or the description above should mention this. Otherwise, the users will not be aware of this when looking at the form to configure the policy.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Introduce possibility of specifying policy order restrictions in their schemas. APIcast now shows a warning when those restrictions are not respected [#1088](https://github.com/3scale/APIcast/pull/1088), [THREESCALE-2896](https://issues.jboss.org/browse/THREESCALE-2896)
 - Added new parameters to logging policy to allow custom access log [PR #1081](https://github.com/3scale/APIcast/pull/1089) [THREESCALE-1234](https://issues.jboss.org/browse/THREESCALE-1234)[THREESCALE-2876](https://issues.jboss.org/browse/THREESCALE-2876)
+- Added http_proxy policy to use an HTTP proxy in api_backed calls. [THREESCALE-2696](https://issues.jboss.org/browse/THREESCALE-2696)
```
davidor: Would be good to add the link to this PR as well.

gateway/src/apicast/policy/http_proxy/proxy.lua
```diff
@@ -0,0 +1,48 @@
+local policy = require('apicast.policy')
+local _M = policy.new('http_proxy')
```
davidor: Can you please add the version? It's needed in order to make the feature of checking the order of policies work.
I needed to add the version in all the policies here: https://github.com/3scale/APIcast/pull/1088/files
In the future we might want to see if there's a better way to do this.

gateway/src/apicast/policy/http_proxy/proxy.lua
```diff
@@ -0,0 +1,48 @@
...
+  return self
+end
+
+local function find_proxy(self, scheme)
+  return self.proxies[scheme]
+end
+
+function _M:export()
+  return  {
+    get_http_proxy = function(_, uri)
```
davidor: I don't see why this function needs the first param.
Can't we get rid of it and then just call `context.get_http_proxy(self.uri)` from the upstream module instead of `context:get_http_proxy(self.uri)` ?

PR #1089, opened 2019-07-05
Policy: logging add custom access log options (JSON, Service vars, NGX vars)

### LINKED ISSUE #1082: Feature: more logging variables available for the access log
At the moment, the log format of the access log in APICast is the following:

https://github.com/3scale/APIcast/blob/bfa3ac9be38c99da5c01cfba4becd0688513b0fe/gateway/http.d/apicast.conf.liquid#L1

There is a variable there called $post_action_impact, which is a variable created by the APICast Lua extensions. It would be great if more variables are set, so that I can set up a log format most suitable for me. Right now, the use case would be to see which service was called, and which method of that service. With that data, working with the logs in ELK or similar would become much easier (filtering on service  and method names instead of paths matching a particular regex etc.)

The change could be completely backwards compatible by just making the variables available but leaving the default logging format as it is.

eloycoto:
Some users requested different ways to log the access log with more metadata,
different formats or conditional logging based on multiple request values. 

This policy address this, two new variables are now set, where to allow or disallow
to print a custom log message, and another one `extened_access_log` just store
all the information to print that. 

The policy has multiple options, here a few examples:

Custom log format
```
{
  "name": "apicast.policy.logging",
  "configuration": {
    "enable_access_logs": false
    "custom_logging": "\"{{request}}\" to service {{service.id}} and {{service.name}}",
  }
}
```

Only log the entry if the status is 200

```
{
  "name": "apicast.policy.logging",
  "configuration": {
    "enable_access_logs": false
    "custom_logging": "\"{{request}}\" to service {{service.id}} and {{service.name}}",
    "condition": {
      "operations": [
        {"op": "==", "match": "{{status}}", "match_type": "liquid", "value": "200"}
      ],
      "combine_op": "and"
    }
  }
}
```

This pr fixed #1082 and THREESCALE-1234 and THREESCALE-2876

### DIFF
--- a/gateway/http.d/apicast.conf.liquid
+++ b/gateway/http.d/apicast.conf.liquid
@@ -1,5 +1,11 @@
 log_format time '[$time_local] $host:$server_port $remote_addr:$remote_port "$request" $status $body_bytes_sent ($request_time) $post_action_impact';
 
+map $status $extended_access_log {
+    default '';
+}
+
+log_format extended escape=none '$extended_access_log';
+
 server {
   listen {{ port.management | default: 8090 }};
   server_name {{ server_name.management | default: 'management _' }};
@@ -45,6 +51,8 @@ server {
 
   set $access_logs_enabled '1';
   access_log {{ access_log_file | default: "/dev/stdout" }} time if=$access_logs_enabled;
+  set $extended_access_logs_enabled '0';
+  access_log {{ access_log_file | default: "/dev/stdout" }} extended if=$extended_access_logs_enabled;
 
   {%- assign http_port = port.apicast | default: 8080 %}
   {%- assign https_port = env.APICAST_HTTPS_PORT %}
--- a/gateway/src/apicast/policy/logging/logging.lua
+++ b/gateway/src/apicast/policy/logging/logging.lua
@@ -1,13 +1,22 @@
 --- Logging policy
 
 local _M  = require('apicast.policy').new('Logging Policy', 'builtin')
-
 local new = _M.new
 
-local default_enable_access_logs = true
+local Condition = require('apicast.conditions.condition')
+local LinkedList = require('apicast.linked_list')
+local Operation = require('apicast.conditions.operation')
+local TemplateString = require('apicast.template_string')
+local cjson = require('cjson')
 
 -- Defined in ngx.conf.liquid and used in the 'access_logs' directive.
 local ngx_var_access_logs_enabled = 'access_logs_enabled'
+local ngx_var_extended_access_logs_enabled = 'extended_access_logs_enabled'
+local ngx_var_extended_access_log = 'extended_access_log'
+
+local default_enable_access_logs = true
+local default_template_type = 'plain'
+local default_combine_op = "and"
 
 -- Returns the value for the ngx var above from a boolean that indicates
 -- whether access logs are enabled or not.
@@ -29,12 +38,115 @@ function _M.new(config)
   end
 
   self.enable_access_logs_val = val_for_ngx_var[enable_access_logs]
+  self.custom_logging = config.custom_logging
+  self.enable_json_logs = config.enable_json_logs
+  self.json_object_config = config.json_object_config or {}
+
+  self:load_condition(config)
 
   return self
 end
 
-function _M:log()
+function _M:load_condition(config)
+  if not config.condition then
+    return
+  end
+
+  ngx.log(ngx.DEBUG, 'Enabling extended log with conditions')
+  local operations = {}
+  for _, operation in ipairs(config.condition.operations) do
+    table.insert( operations,
+      Operation.new(
+        operation.match, operation.match_type,
+        operation.op,
+        operation.value, operation.value_type or default_template_type))
+  end
+  self.condition = Condition.new( operations, config.condition.combine_op or default_combine_op)
+end
+
+local function get_request_context(context)
+  local ctx = { }
+  ctx.req = {
+    headers=ngx.req.get_headers(),
+  }
+
+  ctx.resp = {
+    headers=ngx.resp.get_headers(),
+  }
+
+  ctx.service = context.service or {}
+  return LinkedList.readonly(ctx, ngx.var)
+end
+
+local function enable_extended_access_log()
+  ngx.var[ngx_var_extended_access_logs_enabled] = 1
+end
+
+local function disable_extended_access_log()
+  ngx.var[ngx_var_extended_access_logs_enabled] = 0
+end
+
+local function disable_default_access_logs()
+  ngx.var[ngx_var_access_logs_enabled] = 0
+end
+
+--- log_dump_json: returns an string with the json output.
+local function log_dump_json(self, extended_context)
+  local result = {}
+  for _, value in ipairs(self.json_object_config) do
+    result[value.key] = TemplateString.new(value.value, value.value_type or default_template_type):render(extended_context)
+  end
+
+  local status, data = pcall(cjson.encode, result)
+  if not status then
+    ngx.log(ngx.WARN, "cannot serialize json on logging, err:", data)
+    -- Disable access log due to no valid information can be returned
+    disable_extended_access_log()
+    return ""
+  end
+
+  return data
+end
+
+-- log_dump_line: render the liquid custom_logging value and return it.
+local function log_dump_line(self, extended_context)
+  local tmpl = TemplateString.new(self.custom_logging, "liquid")
+  return tmpl:render(extended_context)
+end
+
+-- get_log_line return the log line based on the kind of log defined in the
+-- service, if Json is enabled will dump a json object, if not will render the
+-- simple log line.
+function _M:get_log_line(extended_context)
+  if self.enable_json_logs then
+    return log_dump_json(self, extended_context)
+  end
+  return log_dump_line(self, extended_context)
+end
+
+
+function _M:use_default_access_logs()
+  return not (self.custom_logging or self.enable_json_logs)
+end
+
+function _M:log(context)
   ngx.var[ngx_var_access_logs_enabled] = self.enable_access_logs_val
+  if self:use_default_access_logs() then
+    return
+  end
+  -- Extended log is now enaled, disable the default access_log
+  disable_default_access_logs()
+
+  local extended_context = get_request_context(context or {})
+  if self.condition and not self.condition:evaluate(extended_context) then
+    -- Access log is disabled here, request does not match, so log is disabled
+    -- for this request
+    disable_extended_access_log()
+    return
+  end
+
+  enable_extended_access_log()
+  ngx.var[ngx_var_extended_access_log] = self:get_log_line(extended_context)
 end
 
 return _M
--- a/gateway/src/apicast/policy/logging/logging.lua
+++ b/gateway/src/apicast/policy/logging/logging.lua
@@ -9,6 +9,8 @@ local Operation = require('apicast.conditions.operation')
 local TemplateString = require('apicast.template_string')
 local cjson = require('cjson')
 
+local ipairs = ipairs
+
 -- Defined in ngx.conf.liquid and used in the 'access_logs' directive.
 local ngx_var_access_logs_enabled = 'access_logs_enabled'
 local ngx_var_extended_access_logs_enabled = 'extended_access_logs_enabled'
--- a/gateway/src/apicast/policy/logging/readme.md
+++ b/gateway/src/apicast/policy/logging/readme.md
@@ -1,8 +1,8 @@
 # Logging policy
 
 This policy has two purposes: one is to enable and disable access log output,
-and and the other is to create a custom access log format for each service and
-be able to set conditions to write custom access log.
+and the other is to create a custom access log format for each service and be
+able to set conditions to write custom access log.
 
 
 ## Exported variables
@@ -20,7 +20,7 @@ response.
 
 ## Examples
 
-### Disabling access log:
+### Disabling access log
 
 ```json
 {
@@ -31,7 +31,7 @@ response.
 }
 ```
 
-### Enabling custom access log:
+### Enabling custom access log
 
 ```json
 {
@@ -43,7 +43,8 @@ response.
 }
 ```
 
-### Enabling custom access log with the service ID:
+### Enabling custom access log with the service identifier
+
 ```json
 {
   "name": "apicast.policy.logging",
--- a/gateway/src/apicast/policy/logging/readme.md
+++ b/gateway/src/apicast/policy/logging/readme.md
@@ -120,6 +120,7 @@ response.
 }
 ```
 
+
 ## Caveats
 
 - If `custom_logging` or `enable_json_logs` property is enabled, default access
--- a/gateway/src/apicast/policy/logging/readme.md
+++ b/gateway/src/apicast/policy/logging/readme.md
@@ -1,0 +1,172 @@
+# Logging policy
+
+This policy has two primary purposes: one is to enable and disable access log
+output, and the second one is to be able to create a custom access log format
+for each service and be able to set conditions to write custom access log. 
+
+
+## Exported variables
+
+Liquid templating can be used on custom logging, the exported variables are:
+
+- NGINX default log_format directive variables, as example: `{{remote_addr}}`.
+[Log format documentation](http://nginx.org/en/docs/http/ngx_http_log_module.html). 
+
+- Response and request headers using `{{req.headers.FOO}}` for getting FOO
+header in the request, or `{{res.headers.FOO}}` to retrieve FOO header on
+response.
+- Service information, as `{{service.id}}` and all service properties as the
+  `THREESCALE_CONFIG_FILE` or `THREESCALE_PORTAL_ENDPOINT` parameters provide.
+
+## Examples
+
+### Disable access log:
+
+```json
+{
+  "name": "apicast.policy.logging",
+  "configuration": {
+    "enable_access_logs": false
+  }
+}
+```
+
+### Enable custom access log:
+
+```json
+{
+  "name": "apicast.policy.logging",
+  "configuration": {
+    "enable_access_logs": false,
+    "custom_logging": "[{{time_local}}] {{host}}:{{server_port}} {{remote_addr}}:{{remote_port}} \"{{request}}\" {{status}} {{body_bytes_sent}} ({{request_time}}) {{post_action_impact}}",
+  }
+}
+```
+
+### Enable custom access log with the service ID:
+```json
+{
+  "name": "apicast.policy.logging",
+  "configuration": {
+    "enable_access_logs": false,
+    "custom_logging": "\"{{request}}\" to service {{service.id}} and {{service.name}}",
+  }
+}
+```
+
+### Write access log in JSON format:
+
+```json
+{
+  "name": "apicast.policy.logging",
+  "configuration": {
+    "enable_access_logs": false,
+    "enable_json_logs": true,
+    "json_object_config": [
+      {
+        "key": "host",
+        "value": "{{host}}",
+        "value_type": "liquid"
+      },
+      {
+        "key": "time",
+        "value": "{{time_local}}",
+        "value_type": "liquid"
+      },
+      {
+        "key": "custom",
+        "value": "custom_method",
+        "value_type": "plain"
+      }
+    ]
+  }
+}
+```
+
+### Write a custom access log only for a successful request.
+
+```json
+{
+  "name": "apicast.policy.logging",
+  "configuration": {
+    "enable_access_logs": false,
+    "custom_logging": "\"{{request}}\" to service {{service.id}} and {{service.name}}",
+    "condition": {
+      "operations": [
+        {"op": "==", "match": "{{status}}", "match_type": "liquid", "value": "200"}
+      ],
+      "combine_op": "and"
+    }
+  }
+}
+```
+
+### Write a custom access log where reponse status match 200 or 500.
+
+```json
+{
+  "name": "apicast.policy.logging",
+  "configuration": {
+    "enable_access_logs": false,
+    "custom_logging": "\"{{request}}\" to service {{service.id}} and {{service.name}}",
+    "condition": {
+      "operations": [
+        {"op": "==", "match": "{{status}}", "match_type": "liquid", "value": "200"},
+        {"op": "==", "match": "{{status}}", "match_type": "liquid", "value": "500"}
+      ],
+      "combine_op": "or"
+    }
+  }
+}
+```
+
+## Caveats
+
+- If `custom_logging` or `enable_json_logs` property is enabled, default access
+  log will be dissabled.
+- If `enable_json_logs` is enabled, `custom_logging` field will be omitted
+
+## Global configuration for all services. 
+
+Logging options can be useful in all services, to avoid having issues with logs
+that are not correctly formated in other services, a custom Apicast environment
+can be set and all services will implement a specifc policy, in this case
+logging.
+
+Here is an example of a policy that is loaded in all services: 
+
+custom_env.lua
+```
+local cjson = require('cjson')
+local PolicyChain = require('apicast.policy_chain')
+local policy_chain = context.policy_chain
+
+local logging_policy_config = cjson.decode([[
+{
+  "enable_access_logs": false,
+  "custom_logging": "\"{{request}}\" to service {{service.id}} and {{service.name}}"
+}
+]])
+
+policy_chain:insert( PolicyChain.load_policy('logging', 'builtin', logging_policy_config), 1) 
+
+return {
+  policy_chain = policy_chain,
+  port = { metrics = 9421 },
+}
+```
+
+To run Apicast with this specific environment: 
+
+```
+docker run --name apicast --rm -p 8080:8080 \
+    -v $(pwd):/config \
+    -e APICAST_ENVIRONMENT=/config/custom_env.lua \
+    -e THREESCALE_PORTAL_ENDPOINT=https://ACCESS_TOKEN@ADMIN_PORTAL_DOMAIN \
+    quay.io/3scale/apicast:master
+```
+
+Key concepts of the docker command: 
+  - Current lua file need to be shared to the container `-v $(pwd):/config`
+  - `APICAST_ENVIRONMENT` variable need to be set to the lua file that is
+    stored on `/config` directory.
--- a/gateway/src/apicast/policy/logging/readme.md
+++ b/gateway/src/apicast/policy/logging/readme.md
@@ -1,13 +1,13 @@
 # Logging policy
 
-This policy has two primary purposes: one is to enable and disable access log
-output, and the second one is to be able to create a custom access log format
-for each service and be able to set conditions to write custom access log. 
+This policy has two purposes: one is to enable and disable access log output,
+and and the other is to create a custom access log format for each service and
+be able to set conditions to write custom access log.
 
 
 ## Exported variables
 
-Liquid templating can be used on custom logging, the exported variables are:
+Liquid templating can be used on custom logging. The exported variables include:
 
 - NGINX default log_format directive variables, as example: `{{remote_addr}}`.
 [Log format documentation](http://nginx.org/en/docs/http/ngx_http_log_module.html). 
@@ -20,7 +20,7 @@ response.
 
 ## Examples
 
-### Disable access log:
+### Disabling access log:
 
 ```json
 {
@@ -31,7 +31,7 @@ response.
 }
 ```
 
-### Enable custom access log:
+### Enabling custom access log:
 
 ```json
 {
@@ -43,7 +43,7 @@ response.
 }
 ```
 
-### Enable custom access log with the service ID:
+### Enabling custom access log with the service ID:
 ```json
 {
   "name": "apicast.policy.logging",
@@ -54,7 +54,8 @@ response.
 }
 ```
 
-### Write access log in JSON format:
+
+### Configuring access logs in JSON format
 
 ```json
 {
@@ -83,7 +84,7 @@ response.
 }
 ```
 
-### Write a custom access log only for a successful request.
+### Configuring a custom access log only for a successful request
 
 ```json
 {
@@ -101,7 +102,9 @@ response.
 }
 ```
 
-### Write a custom access log where reponse status match 200 or 500.
+
+
+### Customizing access logs where reponse status match 200 or 500
 
 ```json
 {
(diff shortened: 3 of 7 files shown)


gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+### Write access log in JSON format:
+
+```json
+{
+  "name": "apicast.policy.logging",
+  "configuration": {
+    "enable_access_logs": false,
+    "custom_logging": "{\"time_local\": \"{{time_local}}\", \"host\" : \"{{host}}\"}",
```
mikz: I don't think this would work. It is exploitable by sending values like `\"` in headers or the host header. It really needs to be properly serialized.

eloycoto:
Hi, 

I had a chat with Michal today about this policy, there are a couple of issues
about the default logging all described in the issue #1082

## My proposal for default logging:##

Two new ENV variables: 
```
CUSTOM_ACCESS_LOG=true
CUSTOM_ACCESS_LOG_FORMAT='{{ host }}'
```

So, if the CUSTOM_ACCESS_LOG is enabled, a new policy like
[DEFAULT_POLICIES](https://github.com/3scale/APIcast/blob/master/gateway/src/apicast/policy_chain.lua#L69)
will be enabled, the  the custom_logging parameter will be  the
`CUSTOM_ACCESS_LOG_FORMAT` env variable.

This provides a few good things:
- All keeps in a policy. 
- Can render all the variables, context, service, response/request headers. 

As discussed, this behaviour is not good enough, policies can't be loaded by
ENV, and that rule can't be break. 

## ENV variable to modify the access_log:

This variable will set the access log by default using ENV variable, if not
default one, the example will be like this: 

```
Line of Code: 
https://github.com/3scale/APIcast/blob/master/gateway/http.d/apicast.conf.liquid#L1

Current code: 

log_format time '[$time_local] $host:$server_port $remote_addr:$remote_port "$request" $status $body_bytes_sent ($request_time) $post_action_impact';

Updated code:
log_format {{LOG_FORMAT | '`'[$time_local] $host:$server_port $remote_addr:$remote_port "$request" $status $body_bytes_sent ($request_time) $post_action_impact\''}};

```

This has the following strengths:

- No liquid rendering, all happens in Nginx. 
- Default logging and no policies are loaded.

This has the following weaknesses:

- Users in this case need to write the ENV variable in nginx format `$host` and
  in the policy with the liquid format `{{host}}`.
- All context information, services, headers, etc.. are lost here. 
- Json format with fields will be lost

## Json render in logs:

Michal also commented that the current JSON render is not a valid option, we
need to deal with JSON using a Lua table and render one by one, and create the
JSON using cjson:

Example policy configuration:
```
{ 
    log_format: [
        "host": "{{host}}", 
        "time_local": "{{time_local}}",
        "method": "{{method}}",
    ]
}
```

This has the following weaknesses:

- Need to call Liquid by each key, instead of a single call.
- For default access with the change in liquid is almost impossible, is
  generated by NGINX, and no Json render by default in Nginx.(There is a
  extension) 
- For ENV variables for the access log you need to some kind of format like
  `export DEFAULT_LOG_FORMAT=host={{host}},time={{time_local}},method={{method}}`
- We need to create the form to enable or not JSON, I guess that some users need
  to have a standard log format and not all users need JSON logs.  

Workaround: 

Use escape in Liquid log, example: 
```
{
  "name": "apicast.policy.logging",
  "configuration": {
    "enable_access_logs": false,
    "custom_logging": "{\"time_local\": \"{{time_local | escape}}\", \"host\" : \"{{host|escape}}\"}",
  }
}
```

Please, share your thoughts here, 

Thanks 
Eloy!

mikz:
Customers already can enforce policies for all services by using the Account Management API and just add them to the policy chain. All that can be automated without changing APIcast at all.

The default logging is just a subset of a feature that would enable configuring any policy by env. I think we should not make a specific case for logging. Either go the full way and have a general solution for all policies or don't do it at all. This is already possible by using a Lua script (environment file). See our APIcast Cloud Hosted: https://github.com/3scale/apicast-cloud-hosted/blob/master/apicast/config/cloud_hosted.lua
Just make own image by using BuildConfig and configure policies globally as you want. I don't see the need to do it in a limited fashion through ENV variables.

Regarding the log_format customization, I don't get the drawbacks. Not all the Lua variables are accessible, but some are exposed (like service_id, https://github.com/3scale/APIcast/blob/4a71c1d762cc4a3e57f05b6813daec4294e24a0d/gateway/conf.d/apicast.conf#L137-L148). Also all the headers and the request/response/upstream information is available through nginx variables. JSON formatting can be done using `escape=json` and just making the json format like https://stackoverflow.com/questions/25049667/how-to-generate-a-json-log-from-nginx for example.

Regarding the JSON serialization. I think that proper JSON serialization is the only way how to do JSON formatting from Lua. We don't know how big performance drawback it would be yet, so I'd not say it is an issue to serialize them in a loop. Also, we just have to bite the bullet at some point and optimize the liquid library to be JITed and the benefits will be everywhere.


To summarize it:

To have a default policy somewhere in the global chain, just use the environment file like the APIcast Cloud Hosted. That can be done just by mounting it via a ConfigMap or building it into the image.

But for a simpler, per deployment configuration, just allowing to set the custom log format and it's escaping by ENV variables is enough for most cases. I think all the request information that makes sense in logs is available by nginx variables.

JSON serialization in Lua has to be done by `cjson`. We should measure the performance impact of liquid processing in a loop, but this is done on other places already (like adding headers).

eloycoto:
@mikz sorry, 

I still have two questions regarding this: 

1) escape=json will work ok, but I have a question around this, what happens if the log line is not json? do you want to have extendedLogJson=true|false and extendedLog=true for normal log and define 3 kind of access log?

2) If we use buildConfig we can use logging policy and users can use that, I'm happy with that. 

Regards.

mikz:
A logging policy should use custom format or JSON (by cjson), but in both cases should be escape=none. 
If someone would want to change the default logging to a different format (like JSON), they should also be able to set the escaping. Then it is entirely up to the user what they want to do.

Not only BuildConfig, but even just mounting the file from ConfigMap works.

eloycoto:
Ok, 

To summarize: 

1) I'll document in the policy how to set up a `buildconfig` with a default
logging and load policy in all services. Also the integration test.  
2)  I'll change the logging configuration to something like this: 

Custom logging with single log line data:

```
{
  "name": "apicast.policy.logging",
  "configuration": {
    "enable_access_logs": false,
    "custom_logging": "[{{time_local}}] {{host}}:{{server_port}}"
  }
}
```

Custom logging with json log line data:

```
{
  "name": "apicast.policy.logging",
  "configuration": {
    "enable_access_logs": false,
    "enable_json_logs": true,
    "json_object_config": [
      {
        "key": "host",
        "value": "{{host}}",
        "value_type": "liquid"
      },
      {
        "key": "time",
        "value": "{{time_local}}",
        "value_type": "liquid"
      },
      {
        "key": "custom",
        "value": "custom_method",
        "value_type": "plain"
      }
    ]
  }
}
```

[Form example](https://mozilla-services.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6eyJuYW1lIjoiIiwiY29uZmlndXJhdGlvbiI6eyJlbmFibGVfYWNjZXNzX2xvZ3MiOmZhbHNlLCJlbmFibGVfanNvbl9sb2dzIjpmYWxzZSwianNvbl9vYmplY3RfY29uZmlnIjpbeyJrZXkiOiIiLCJ2YWx1ZSI6IiIsInZhbHVlX3R5cGUiOiIifSx7ImtleSI6IiIsInZhbHVlIjoiIiwidmFsdWVfdHlwZSI6IiJ9XX0sImZpcnN0TmFtZSI6IkNodWNrIn0sInNjaGVtYSI6eyJkZWZpbml0aW9ucyI6e30sIiRzY2hlbWEiOiJodHRwOi8vanNvbi1zY2hlbWEub3JnL2RyYWZ0LTA3L3NjaGVtYSMiLCIkaWQiOiJodHRwOi8vZXhhbXBsZS5jb20vcm9vdC5qc29uIiwidHlwZSI6Im9iamVjdCIsInRpdGxlIjoiVGhlIFJvb3QgU2NoZW1hIiwicmVxdWlyZWQiOlsibmFtZSIsImNvbmZpZ3VyYXRpb24iXSwicHJvcGVydGllcyI6eyJuYW1lIjp7IiRpZCI6IiMvcHJvcGVydGllcy9uYW1lIiwidHlwZSI6InN0cmluZyIsInRpdGxlIjoiVGhlIE5hbWUgU2NoZW1hIiwiZGVmYXVsdCI6IiIsImV4YW1wbGVzIjpbImFwaWNhc3QucG9saWN5LmxvZ2dpbmciXSwicGF0dGVybiI6Il4oLiopJCJ9LCJjb25maWd1cmF0aW9uIjp7IiRpZCI6IiMvcHJvcGVydGllcy9jb25maWd1cmF0aW9uIiwidHlwZSI6Im9iamVjdCIsInRpdGxlIjoiVGhlIENvbmZpZ3VyYXRpb24gU2NoZW1hIiwicmVxdWlyZWQiOlsiZW5hYmxlX2FjY2Vzc19sb2dzIiwiZW5hYmxlX2pzb25fbG9ncyIsImpzb25fb2JqZWN0X2NvbmZpZyJdLCJwcm9wZXJ0aWVzIjp7ImVuYWJsZV9hY2Nlc3NfbG9ncyI6eyIkaWQiOiIjL3Byb3BlcnRpZXMvY29uZmlndXJhdGlvbi9wcm9wZXJ0aWVzL2VuYWJsZV9hY2Nlc3NfbG9ncyIsInR5cGUiOiJib29sZWFuIiwidGl0bGUiOiJUaGUgRW5hYmxlX2FjY2Vzc19sb2dzIFNjaGVtYSIsImRlZmF1bHQiOmZhbHNlLCJleGFtcGxlcyI6W2ZhbHNlXX0sImVuYWJsZV9qc29uX2xvZ3MiOnsiJGlkIjoiIy9wcm9wZXJ0aWVzL2NvbmZpZ3VyYXRpb24vcHJvcGVydGllcy9lbmFibGVfanNvbl9sb2dzIiwidHlwZSI6ImJvb2xlYW4iLCJ0aXRsZSI6IlRoZSBFbmFibGVfanNvbl9sb2dzIFNjaGVtYSIsImRlZmF1bHQiOmZhbHNlLCJleGFtcGxlcyI6W3RydWVdfSwianNvbl9vYmplY3RfY29uZmlnIjp7IiRpZCI6IiMvcHJvcGVydGllcy9jb25maWd1cmF0aW9uL3Byb3BlcnRpZXMvanNvbl9vYmplY3RfY29uZmlnIiwidHlwZSI6ImFycmF5IiwidGl0bGUiOiJUaGUgSnNvbl9vYmplY3RfY29uZmlnIFNjaGVtYSIsIml0ZW1zIjp7IiRpZCI6IiMvcHJvcGVydGllcy9jb25maWd1cmF0aW9uL3Byb3BlcnRpZXMvanNvbl9vYmplY3RfY29uZmlnL2l0ZW1zIiwidHlwZSI6Im9iamVjdCIsInRpdGxlIjoiVGhlIEl0ZW1zIFNjaGVtYSIsInJlcXVpcmVkIjpbImtleSIsInZhbHVlIiwidmFsdWVfdHlwZSJdLCJwcm9wZXJ0aWVzIjp7ImtleSI6eyIkaWQiOiIjL3Byb3BlcnRpZXMvY29uZmlndXJhdGlvbi9wcm9wZXJ0aWVzL2pzb25fb2JqZWN0X2NvbmZpZy9pdGVtcy9wcm9wZXJ0aWVzL2tleSIsInR5cGUiOiJzdHJpbmciLCJ0aXRsZSI6IlRoZSBLZXkgU2NoZW1hIiwiZGVmYXVsdCI6IiIsImV4YW1wbGVzIjpbImhvc3QiXSwicGF0dGVybiI6Il4oLiopJCJ9LCJ2YWx1ZSI6eyIkaWQiOiIjL3Byb3BlcnRpZXMvY29uZmlndXJhdGlvbi9wcm9wZXJ0aWVzL2pzb25fb2JqZWN0X2NvbmZpZy9pdGVtcy9wcm9wZXJ0aWVzL3ZhbHVlIiwidHlwZSI6InN0cmluZyIsInRpdGxlIjoiVGhlIFZhbHVlIFNjaGVtYSIsImRlZmF1bHQiOiIiLCJleGFtcGxlcyI6WyJ7e2hvc3R9fSJdLCJwYXR0ZXJuIjoiXiguKikkIn0sInZhbHVlX3R5cGUiOnsiJGlkIjoiIy9wcm9wZXJ0aWVzL2NvbmZpZ3VyYXRpb24vcHJvcGVydGllcy9qc29uX29iamVjdF9jb25maWcvaXRlbXMvcHJvcGVydGllcy92YWx1ZV90eXBlIiwidHlwZSI6InN0cmluZyIsInRpdGxlIjoiVGhlIFZhbHVlX3R5cGUgU2NoZW1hIiwiZGVmYXVsdCI6IiIsImV4YW1wbGVzIjpbImxpcXVpZCJdLCJwYXR0ZXJuIjoiXiguKikkIn19fX19fX19LCJ1aVNjaGVtYSI6eyJmaXJzdE5hbWUiOnsidWk6YXV0b2ZvY3VzIjp0cnVlLCJ1aTplbXB0eVZhbHVlIjoiIn0sImFnZSI6eyJ1aTp3aWRnZXQiOiJ1cGRvd24iLCJ1aTp0aXRsZSI6IkFnZSBvZiBwZXJzb24iLCJ1aTpkZXNjcmlwdGlvbiI6IihlYXJ0aGlhbiB5ZWFyKSJ9LCJiaW8iOnsidWk6d2lkZ2V0IjoidGV4dGFyZWEifSwicGFzc3dvcmQiOnsidWk6d2lkZ2V0IjoicGFzc3dvcmQiLCJ1aTpoZWxwIjoiSGludDogTWFrZSBpdCBzdHJvbmchIn0sImRhdGUiOnsidWk6d2lkZ2V0IjoiYWx0LWRhdGV0aW1lIn0sInRlbGVwaG9uZSI6eyJ1aTpvcHRpb25zIjp7ImlucHV0VHlwZSI6InRlbCJ9fX19)

In case of cjson error, log line will be disabled and nothing will be logged.

magnusvage:
Hi!

@eloycoto pointed me this way. I am not very worried about the technical solution. I'm sure you'll come up with a way that gets the job done and is workable for both you and us. However, @mikz you mention that "not all Lua variables are exposed, but some". Is there an exhaustive list somewhere over which variables are available? We have pretty clear requirements regarding what we need to put in the access log. Would it be possible to get service system name and method/metric system name into the access log row with the suggested solution above?

mikz:
@magnusvage https://github.com/3scale/APIcast/blob/5df6144630ece85b2c7ee8e606ebf6a9f1f2251f/gateway/conf.d/apicast.conf#L137-L153 those should be available. It does not contain service name, as it is not multi tenant, but rather just service id. The `usage` is the query encoded usage reported to 3scale backend.

davidor:
I didn't know that tool @mikz . It would be good to add those checks to circleCI.

gateway/src/apicast/policy/logging/logging.lua
```diff
@@ -29,12 +40,101 @@ function _M.new(config)
...
+          operation.value, operation.value_type or default_template_type))
+    end
+    self.condition = Condition.new( operations, config.condition.combine_op or default_combine_op)
+  end
 
   return self
 end
 
-function _M:log()
+local function get_request_context(context)
```
davidor: Is there a reason to define these headers here instead of reusing `ngx_variable` ?
I think everything on `ngx_variable` should be available. We should define there the response headers, though.

spec/policy/logging/logging_spec.lua
```diff
@@ -36,4 +40,187 @@ describe('Logging policy', function()
...
+              { key = "bar", value="{{req.headers.bar}}", value_type="liquid"},
+            }
+          })
+          spy.on(logging, "log_dump_json")
+          logging:log(ctx)
+
+          assert.equals(0, ngx.var.access_logs_enabled)
+          assert.equals(1, ngx.var.extended_access_logs_enabled)
+          assert.equals(ngx.var.extended_access_log, expected_json)
+          assert.spy(logging.log_dump_json).was.called()
```
davidor: Why is this needed? The previous assert already verifies that a JSON was generated, no?

gateway/src/apicast/policy/logging/logging.lua
```diff
@@ -29,12 +40,101 @@ function _M.new(config)
   end
 
   self.enable_access_logs_val = val_for_ngx_var[enable_access_logs]
+  self.custom_logging = config.custom_logging
+  self.enable_json_logs = config.enable_json_logs
+	self.json_object_config = config.json_object_config or {}
+
+  if config.condition then
+    ngx.log(ngx.DEBUG, 'Enabling extended log with conditions')
```
davidor: Looks like a good candidate to be extracted into its own function.

gateway/src/apicast/policy/logging/logging.lua
eloycoto: just because it is easy to provide to users information about the request and response in an easy way.

spec/policy/logging/logging_spec.lua
eloycoto: just to make sure that the call to the correct function for future changes.

t/apicast-policy-logging.t
```diff
@@ -126,8 +126,333 @@ qr/"GET \W+ HTTP\/1.1" 200/
 GET /
 --- error_code: 200
 --- grep_error_log eval
-qr/"GET \W+ HTTP\/1.1" 200/
+[qr/"GET \W+ HTTP\/1.1" 200/]
 --- grep_error_log_out
 "GET / HTTP/1.1" 200
 --- no_error_log
 [error]
+
```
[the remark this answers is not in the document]
eloycoto: discussed offiline, fixed this one.

t/apicast-policy-logging.t
```diff
@@ -126,8 +126,333 @@ qr/"GET \W+ HTTP\/1.1" 200/
...
+  }
+--- request
+GET /
+--- error_code: 200
+--- error_log eval
+[qr/^\{\}/]
+--- no_error_log eval
+[qr/\[error/, qr/GET \/ HTTP\/1.1\" 200/]
+
+=== TEST 10: json log with valid data return a valid json
```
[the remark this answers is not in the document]
eloycoto: It's on the caveats

gateway/http.d/apicast.conf.liquid
```diff
@@ -1,5 +1,11 @@
 log_format time '[$time_local] $host:$server_port $remote_addr:$remote_port "$request" $status $body_bytes_sent ($request_time) $post_action_impact';
 
+map $status $extended_access_log {
```
davidor: Why is this map needed? Sorry, I forgot :sweat_smile:

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,175 @@
+# Logging policy
```
davidor: @porueesq  Can you review this file please? I guess you'll use it as the base for your docs.

spec/policy/logging/logging_spec.lua
```diff
@@ -36,4 +40,187 @@ describe('Logging policy', function()
...
+        end)
+
+
+        it("render correctly liquid and plain entries", function()
+          local logging = LoggingPolicy.new({
+            enable_json_logs = true,
+            json_object_config = {
+              { key = "foo", value="{{foo}}", value_type="liquid"},
+              { key = "bar", value="barValue", value_type="liquid"},
```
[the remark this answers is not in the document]
davidor: I think you forgot about this one.

gateway/src/apicast/policy/logging/logging.lua
davidor: But this limits us to the information in the service plus the headers.
I guess it's fine for now but in the future, users might ask for other info that we have in `ngx_variable` like it happened for other policies.

gateway/http.d/apicast.conf.liquid
eloycoto: if not nignx will not start due to extended_access_log is not defined.

gateway/src/apicast/policy/logging/readme.md
porueesq: @davidor Yes, I'll review the file.
Just for the record, have you updated the functionality of the Logging policy? We currently have this documented:
https://access.redhat.com/documentation/en-us/red_hat_3scale_api_management/2.5/html/administering_the_api_gateway/apicast_policies#logging_policy

davidor: The functionality introduced in this PR will be available in 3scale 2.7 @porueesq

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,175 @@
...
+        {"op": "==", "match": "{{status}}", "match_type": "liquid", "value": "200"},
+        {"op": "==", "match": "{{status}}", "match_type": "liquid", "value": "500"}
+      ],
+      "combine_op": "or"
+    }
+  }
+}
+```
+
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
eloycoto: :man_facepalming:

gateway/http.d/apicast.conf.liquid
davidor: And a simple `set` does not work?

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
+# Logging policy
+
+This policy has two primary purposes: one is to enable and disable access log
```
porueesq: `This policy has two purposes: one is to enable and disable access log`

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
...
+output, and the second one is to be able to create a custom access log format
```
porueesq: `output, and the other is to create a custom access log format`

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
...
+for each service and be able to set conditions to write custom access log. 
+
+
+## Exported variables
+
+Liquid templating can be used on custom logging, the exported variables are:
```
porueesq: `Liquid templating can be used on custom logging. The exported variables include:`

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
...
+
+- Response and request headers using `{{req.headers.FOO}}` for getting FOO
+header in the request, or `{{res.headers.FOO}}` to retrieve FOO header on
+response.
+- Service information, as `{{service.id}}` and all service properties as the
+  `THREESCALE_CONFIG_FILE` or `THREESCALE_PORTAL_ENDPOINT` parameters provide.
+
+## Examples
+
+### Disable access log:
```
porueesq: `### Disabling access log:`

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
...
+```json
+{
+  "name": "apicast.policy.logging",
+  "configuration": {
+    "enable_access_logs": false
+  }
+}
+```
+
+### Enable custom access log:
```
porueesq: `### Enabling custom access log:`

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
...
+{
+  "name": "apicast.policy.logging",
+  "configuration": {
+    "enable_access_logs": false,
+    "custom_logging": "[{{time_local}}] {{host}}:{{server_port}} {{remote_addr}}:{{remote_port}} \"{{request}}\" {{status}} {{body_bytes_sent}} ({{request_time}}) {{post_action_impact}}",
+  }
+}
+```
+
+### Enable custom access log with the service ID:
```
porueesq: `### Enabling custom access log with the service ID:`

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
...
+{
+  "name": "apicast.policy.logging",
+  "configuration": {
+    "enable_access_logs": false,
+    "custom_logging": "\"{{request}}\" to service {{service.id}} and {{service.name}}",
+  }
+}
+```
+
+### Write access log in JSON format:
```
porueesq: `### Writing access logs in JSON format:`

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
...
+        "key": "custom",
+        "value": "custom_method",
+        "value_type": "plain"
+      }
+    ]
+  }
+}
+```
+
+### Write a custom access log only for a successful request.
```
porueesq: `### Configuring a custom access log only for a successful request.`

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
...
+      "operations": [
+        {"op": "==", "match": "{{status}}", "match_type": "liquid", "value": "200"}
+      ],
+      "combine_op": "and"
+    }
+  }
+}
+```
+
+### Write a custom access log where reponse status match 200 or 500.
```
porueesq: `### Customizing access logs where reponse status match 200 or 500`

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
...
+      "combine_op": "or"
+    }
+  }
+}
+```
+
+## Caveats
+
+- If `custom_logging` or `enable_json_logs` property is enabled, default access
+  log will be dissabled.
```
porueesq: `log will be disabled.`

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
...
+- If `enable_json_logs` is enabled, `custom_logging` field will be omitted
```
porueesq: `If `enable_json_logs` is enabled, `custom_logging` field will be omitted.`

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Global configuration for all services.
```
porueesq: `## Global configuration for all services`

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Caveats
+
+- If `custom_logging` or `enable_json_logs` property is enabled, default access
+  log will be dissabled.
+- If `enable_json_logs` is enabled, `custom_logging` field will be omitted
+
+## Global configuration for all services. 
+
+Logging options can be useful in all services, to avoid having issues with logs
```
porueesq: `In all services, logging options help to avoid having issues with logs`

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
...
+that are not correctly formated in other services, a custom Apicast environment
```
porueesq: `that are not correctly formated in other services, a custom APIcast environment`

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
...
+
+policy_chain:insert( PolicyChain.load_policy('logging', 'builtin', logging_policy_config), 1) 
+
+return {
+  policy_chain = policy_chain,
+  port = { metrics = 9421 },
+}
+```
+
+To run Apicast with this specific environment:
```
porueesq: APIcast

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
...
+```
+docker run --name apicast --rm -p 8080:8080 \
+    -v $(pwd):/config \
+    -e APICAST_ENVIRONMENT=/config/custom_env.lua \
+    -e THREESCALE_PORTAL_ENDPOINT=https://ACCESS_TOKEN@ADMIN_PORTAL_DOMAIN \
+    quay.io/3scale/apicast:master
+```
+
+Key concepts of the docker command: 
+  - Current lua file need to be shared to the container `-v $(pwd):/config`
```
porueesq: `Current Lua file must be shared to the container `-v $(pwd):/config``

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -0,0 +1,172 @@
...
+  - `APICAST_ENVIRONMENT` variable need to be set to the lua file that is
```
porueesq: ``APICAST_ENVIRONMENT` variable must be set to the Lua file that is`

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -1,13 +1,13 @@
 # Logging policy
 
-This policy has two primary purposes: one is to enable and disable access log
-output, and the second one is to be able to create a custom access log format
-for each service and be able to set conditions to write custom access log. 
+This policy has two purposes: one is to enable and disable access log output,
+and and the other is to create a custom access log format for each service and
```
porueesq: `and the other is to create a custom access log format for each service, as well as to` 

Remove the first `and`.

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -20,7 +20,7 @@ response.
 
 ## Examples
 
-### Disable access log:
+### Disabling access log:
```
porueesq: Remove colon `:` from all headings.

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -31,7 +31,7 @@ response.
 }
 ```
 
-### Enable custom access log:
+### Enabling custom access log:
```
porueesq: Remove colon.

gateway/src/apicast/policy/logging/readme.md
```diff
@@ -43,7 +43,7 @@ response.
 }
 ```
 
-### Enable custom access log with the service ID:
+### Enabling custom access log with the service ID:
```
porueesq: `### Enabling custom access logs with service identifier`

magnusvage:
@davidor when can we expect these changes to be available for us users to try out?

eloycoto:
Hi @magnusvage 

This policy is currently merged and working on master and will be released in the next version 3.7, that release should be out in a couple of months, but some extensive testing needs to be done, and maybe it's delayed, it's a bit early to say a date.

Regards

PR #1092, opened 2019-07-12
Simplify tostring() in resty resolver

davidor:
A small code cleanup.
I noticed this while debugging some problems in the circleCI profile task.


davidor:
Rebased on top of master to pick the fix added in #1093 
`circle:ci profile` fails, but it's not related to this PR.

PR #1093, opened 2019-07-15
Fix verify ssl test in async_resty

davidor:
The error message changed because of this commit:
https://github.com/chromium/badssl.com/commit/6806596afaca81090df1965e59f99eae33674d56


davidor:
`circle: profile` fails for a reason not related to this PR. We're investigating it.

PR #1094, opened 2019-07-15
CHANGELOG: move policy order to correct version

davidor:


davidor:
`circleci:profile` fails for reasons not related to this PR.

PR #1099, opened 2019-07-22
Load service configs one by one lazily

davidor:
This PR adds the option to add service configurations lazily.

This might be useful for users with many services configured. By default, APIcast loads all the services each time it downloads its configuration from the 3scale admin portal. With a large number of services, this could become problematic.

When this option is enabled, the configurations are loaded lazily. APIcast will only load the ones configured for the host specified in the host header of the request.

The option is not enabled by default. It is enabled using the `APICAST_LOAD_SERVICES_WHEN_NEEDED` env.

The PR also includes a couple of commits that clean up a bit the `remote_v2` module used in this PR and also adds some comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Introduce possibility of specifying policy order restrictions in their schemas. APIcast now shows a warning when those restrictions are not respected [#1088](https://github.com/3scale/APIcast/pull/1088), [THREESCALE-2896](https://issues.jboss.org/browse/THREESCALE-2896)
 - Added new parameters to logging policy to allow custom access log [PR #1081](https://github.com/3scale/APIcast/pull/1089) [THREESCALE-1234](https://issues.jboss.org/browse/THREESCALE-1234)[THREESCALE-2876](https://issues.jboss.org/browse/THREESCALE-2876)
 - Added http_proxy policy to use an HTTP proxy in api_backed calls. [THREESCALE-2696](https://issues.jboss.org/browse/THREESCALE-2696) [PR #1080](https://github.com/3scale/APIcast/pull/1080)
+- Option to load service configurations one by one lazily [PR #1099](https://github.com/3scale/APIcast/pull/1099)
 
 ## [3.6.0-rc1] - 2019-07-04
 
--- a/doc/parameters.md
+++ b/doc/parameters.md
@@ -51,6 +51,25 @@ Double colon (`:`) separated list of environments (or paths) APIcast should load
 It can be used instead of `-e` or `---environment` parameter on the CLI and for example
 stored in the container image as default environment. Any value passed on the CLI overrides this variable.
 
+### `APICAST_LOAD_SERVICES_WHEN_NEEDED`
+**Values:**
+- `true` or `1` for _true_
+- `false`, `0` or empty for _false_
+
+**Default:** _false_
+
+This option might be useful for users with many services configured. By default,
+APIcast loads all the services each time it downloads its configuration from the
+admin portal. With a large number of services, this could become problematic.
+When this option is enabled, the configurations are loaded lazily. APIcast will
+only load the ones configured for the host specified in the host header of the
+request.
+
+Notes:
+- The caching defined by `APICAST_CONFIGURATION_CACHE` applies.
+- This option will be disabled when `APICAST_CONFIGURATION_LOADER` is `boot`.
+- Not compatible with `APICAST_PATH_ROUTING`.
+
 ### `APICAST_LOG_FILE`
 
 **Default:** _stderr_
--- a/doc/parameters.md
+++ b/doc/parameters.md
@@ -58,12 +58,16 @@ stored in the container image as default environment. Any value passed on the CL
 
 **Default:** _false_
 
-This option might be useful for users with many services configured. By default,
-APIcast loads all the services each time it downloads its configuration from the
-admin portal. With a large number of services, this could become problematic.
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Introduce possibility of specifying policy order restrictions in their schemas. APIcast now shows a warning when those restrictions are not respected [#1088](https://github.com/3scale/APIcast/pull/1088), [THREESCALE-2896](https://issues.jboss.org/browse/THREESCALE-2896)
+- Option to load service configurations one by one lazily [PR #1099](https://github.com/3scale/APIcast/pull/1099)
```
```suggestion
- Option to implement lazy loading of service configurations one by one [PR #1099](https://github.com/3scale/APIcast/pull/1099)
```

doc/parameters.md
```diff
@@ -51,6 +51,25 @@ Double colon (`:`) separated list of environments (or paths) APIcast should load
...
 stored in the container image as default environment. Any value passed on the CLI overrides this variable.
 
+### `APICAST_LOAD_SERVICES_WHEN_NEEDED`
+**Values:**
+- `true` or `1` for _true_
+- `false`, `0` or empty for _false_
+
+**Default:** _false_
+
+This option might be useful for users with many services configured. By default,
```
```suggestion
This option is useful when there are many services configured. By default,
```

doc/parameters.md
```diff
@@ -51,6 +51,25 @@ Double colon (`:`) separated list of environments (or paths) APIcast should load
...
+APIcast loads all the services each time it downloads its configuration from the
+admin portal. With a large number of services, this could become problematic.
```
```suggestion
Admin Portal. This can cause performance issues with large number of services.
```

doc/parameters.md
```diff
@@ -51,6 +51,25 @@ Double colon (`:`) separated list of environments (or paths) APIcast should load
 It can be used instead of `-e` or `---environment` parameter on the CLI and for example
 stored in the container image as default environment. Any value passed on the CLI overrides this variable.
 
+### `APICAST_LOAD_SERVICES_WHEN_NEEDED`
```
davidor: I wanted to include "lazy" or "lazily" in the name of the ENV. However, I thought that would be a bit confusing because `APICAST_CONFIGURATION_LOADER` already accepts `lazy` as an option. I am open to suggestions for better names :smile:

doc/parameters.md
davidor: I wrote "might" because it depends a lot on factors such as the number of services, the latency between apicast and the 3scale portal admin, the TTL of the config, etc. We cannot guarantee that this option will always work better than the default.

doc/parameters.md
porueesq: Some ideas:

1. APICAST_LAZY_LOADING
2. APICAST_LAZY_LOAD_SERVICES
3. APICAST_LAZY_LOAD_CONFIG

@davidor Do you have guidelines or any restrictions when creating env var names?

doc/parameters.md
porueesq: Got you. This is important.
Suggestion:
`This option can be used when there are many services configured. However, its performance depends on additional factors such as the number of services, the latency between APIcast and the 3scale Admin Portal, the Time To Live (TTL) of the configuration, etc.` <better if you can include a line break before starting the next line with _By default_>

gateway/src/apicast/configuration_loader/remote_v2.lua
```diff
@@ -107,44 +176,39 @@ function _M:index(host)
...
+  local load_just_for_host = load_just_the_services_needed()
+
   if self == _M  or not self then
     local host = environment
     local m = _M.new()
     local ret, err = m:index(host)
 
     if not ret then
-      return m:call()
+      if load_just_for_host then
```
eloycoto: Does not make sense to me to keep the negative condition here, I think that can be something like this: 

```
if ret then
return ret, err
end

if load_just_for_host then
...
```

I think that makes things more clear.

davidor:
@porueesq I added your suggestions here https://github.com/3scale/APIcast/pull/1099/commits/40f97968e84252fb834436b95d33ba2c9c584d38

gateway/src/apicast/configuration_loader/remote_v2.lua
```diff
@@ -77,6 +77,75 @@ local function array()
...
+  return format(
+      "%s/admin/api/services/proxy/configs/%s.json?%s",
+      portal_endpoint,
+      env,
+      query_args
+  )
+end
+
+local function parse_resp_body(self, resp_body)
+  local json = cjson.decode(resp_body)
```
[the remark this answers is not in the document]
davidor: Right. Good idea. I also added a test for it.

gateway/src/apicast/configuration_loader/remote_v2.lua
```diff
@@ -77,6 +77,75 @@ local function array()
...
+
+  local config = { services = array(), oidc = array() }
+
+  local proxy_configs = json.proxy_configs or {}
+
+  for i=1, #proxy_configs do
+    local proxy_config = proxy_configs[i].proxy_config
```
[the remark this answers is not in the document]
davidor: It was like this before, but I don't see why not use ipairs.
Changed :+1:

gateway/src/apicast/configuration_loader/remote_v2.lua
davidor: You're absolutely right. I've simplified this :+1:

davidor:
Fixed conflicts in the changelog.

PR #1103, opened 2019-08-01
Routing Policy  enhancements

eloycoto:
Hi, 

Multiple commits here: 

1) Added original request information in the executor in the ngx.ctx. 
2) Clean a little bit the ngx_variable, I think that nothing wrong around this one, but please have a look. 
Other commits are to allow liquid match in the routing policy to accomplish @ppatierno feedback. 

Regards

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Option to load service configurations one by one lazily [PR #1099](https://github.com/3scale/APIcast/pull/1099), [THREESCALE-3168](https://issues.jboss.org/browse/THREESCALE-3168)
 - New maintenance mode policy, useful for maintenance periods. [PR #1105](https://github.com/3scale/APIcast/pull/1105), [THREESCALE-3189](https://issues.jboss.org/browse/THREESCALE-3189)
 - Remove dnsmasq process for APIcast [PR #1090](https://github.com/3scale/APIcast/pull/1090), [THREESCALE-1555](https://issues.jboss.org/browse/THREESCALE-1555)
+- Enable liquid operations and original request variable on routing policy [PR #1103](https://github.com/3scale/APIcast/pull/1103) [THREESCALE-3239](https://issues.jboss.org/browse/THREESCALE-3239)
 - Allow to use capture function in liquid templates. [PR #1107](https://github.com/3scale/APIcast/pull/1107), [THREESCALE-1911](https://issues.jboss.org/browse/THREESCALE-1911)
 - OAuth 2.0 MTLS policy [PR #1101](https://github.com/3scale/APIcast/pull/1101) [Issue #1003](https://github.com/3scale/APIcast/issues/1003)
 - Add an option to enable keepalive_timeout on gateway [THREESCALE-2886](https://issues.jboss.org/browse/THREESCALE-2886) [PR #1106](https://github.com/3scale/APIcast/pull/1106)
@@ -23,6 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Fix issues when OPENTRACING_FORWARD_HEADER was set [PR #1109](https://github.com/3scale/APIcast/pull/1109), [THREESCALE-1660](https://issues.jboss.org/browse/THREESCALE-1660)
 - New TLS termination policy [PR #1108](https://github.com/3scale/APIcast/pull/1108), [THREESCALE-2898](https://issues.jboss.org/browse/THREESCALE-2898)
+
 
 ## [3.6.0-rc1] - 2019-07-04
 
--- a/gateway/src/apicast/executor.lua
+++ b/gateway/src/apicast/executor.lua
@@ -37,6 +37,28 @@ local function build_context(executor)
     return linked_list.readwrite({}, config)
 end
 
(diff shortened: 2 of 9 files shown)


gateway/src/apicast/policy/routing/README.md
```diff
@@ -340,6 +340,35 @@ the path is not `/accounts`:
     }
   }
 ```
+## Liquid templating on match
+
+It is possible to use liquid templating in the match section. This allows to
+define rules with dynamic information and provide a way to match with any
+request info.
```
davidor: I think we should be more specific in this description. I think we should mention what information of the `original_request` is available, and also, the variables that can be accessed because they're in `ngx.var`. Should we link https://nginx.org/en/docs/varindex.html (if we expose all of them) ?

gateway/src/apicast/policy/ngx_variable.lua
```diff
@@ -3,17 +3,11 @@ local LinkedList = require('apicast.linked_list')
...
-    remote_addr = ngx.var.remote_addr,
-    remote_port = ngx.var.remote_port,
-    scheme = ngx.var.scheme,
-    server_addr = ngx.var.server_addr,
-    server_port = ngx.var.server_port,
+  local context = {
     headers = ngx.req.get_headers(),
     http_method = ngx.req.get_method(),
   }
+  return LinkedList.readonly(context, ngx.var)
```
davidor: I think this is OK, but it concerns me a bit. We were exposing just a few vars and now all of them are.
Can we be sure that nothing confidential is exposed? Maybe we should be more conservative here and select only the few vars that we know are safe? There are many vars: https://nginx.org/en/docs/varindex.html

gateway/src/apicast/executor.lua
```diff
@@ -37,6 +37,26 @@ local function build_context(executor)
...
 
+local function store_original_request(context)
+  -- There are a few phases[0] that req and var are not set and API is disabled.
+  -- The reason to call this using a pcall function is to avoid to define the
+  -- phases manually that are error prone. Also in openresty is not a good method
+  -- to check this. [1]
+  -- [0] invalid phases: init_worker, init, timer and ssl_cer
+  -- [1] https://github.com/openresty/lua-resty-core/blob/9937f5d83367e388da4fcc1d7de2141c9e38d7e2/lib/resty/core/request.lua#L96
+
+  local result, _ =  pcall(function()
```
davidor: Should we have an `if` here that sets `context.original_request` only when it's nil?
Otherwise, we might overwrite this var when some policy has already modified the headers, or the path. The `shared_build_context` function might guarantee that this never happens, but I'm not 100% sure.

Also, I have doubts about this. You mention that this is going to fail in the `init_worker` phase. However, the `shared_build_context` function will build the context in that phase and store it in `ctx.context`. Subsequent calls to the method will use what's stored in `ctx.context` instead of building again the context, so when is `original_request` initialized?

eloycoto:
Hi, 

Finally, I remove the linked list to ngx.var due to the security concerts; some critical data can get from "ngx.var" variable (number of connections, client certs, etc..) 

I was a bit optimistic when I set that, so it's as it was the only thing that I added is the request path to the ngx variable. 

Thanks, David, for the catch!

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Option to load service configurations one by one lazily [PR #1099](https://github.com/3scale/APIcast/pull/1099)
 - New maintenance mode policy, useful for maintenance periods. [PR #1105](https://github.com/3scale/APIcast/pull/1105), [THREESCALE-3189](https://issues.jboss.org/browse/THREESCALE-3189)
 - Remove dnsmasq process for APIcast [PR #1090](https://github.com/3scale/APIcast/pull/1090), [THREESCALE-1555](https://issues.jboss.org/browse/THREESCALE-1555)
+- Enable liquid operations and original request variable on routing policy [PR #1103](https://github.com/3scale/APIcast/pull/1103)
```
davidor: Remember to include the JIRA as well.

gateway/src/apicast/executor.lua
```diff
@@ -47,8 +69,8 @@ local function shared_build_context(executor)
 
     if not context then
         context = build_context(executor)
-
         ctx.context = context
+        store_original_request(ctx)
```
davidor: Does this guarantee that the original request will be initialized on every request?

In other words, `store_original_request` only runs when the context has not been initialized, but `store_original_request` fails in some phases, so I wonder if the following case is possible:
- The body of this `if` runs and initializes `ctx.context`
- `store_original_request` is called but fails to assign the original request data
- The body of this `if` is not executed again and the original request data is never initialized

eloycoto: due to error happens in the same function call, I do not think that it's the case.

davidor: @eloycoto I've reviewed the latest changes of the PR and everything looks good to me except this. Maybe I'm missing something.

I see 2 scenarios:
- `context` is nil. `store_original_request` is called and it fails (there's a `pcall` so there are cases where it can fail). In that case, the original request would never be initialized in the context.
- `context` is nil, but when `store_original_request is called it never fails. In that case, the `pcall` would be unnecessary.

What am I missing?

PR #1105, opened 2019-08-06
[THREESCALE-3189] Add maintenance policy

davidor:
This PR adds a new simple policy that allows users to reject incoming requests with a specified status code and message. It's useful for maintenance periods or to temporarily block an API.

This continues the work started by @alanmoran in https://github.com/3scale/APIcast/pull/561/

I used his commits, converted the policy to a builtin one (it was an example), and polished a few things and added some tests.


gateway/src/apicast/policy/maintenance_mode/maintenance_mode.lua
```diff
@@ -0,0 +1,34 @@
...
+
+function _M.new(configuration)
+  local policy = new(configuration)
+
+  policy.status_code = default_status_code
+  policy.message = default_message
+
+  if configuration then
+    policy.status_code = tonumber(configuration.status) or policy.status_code
+    policy.message = configuration.message or policy.message
```
mikz: It might be useful to also set some header in case this is active. So you can have the payload as JSON for example.

gateway/src/apicast/policy/maintenance_mode/maintenance_mode.lua
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  return policy
+end
+
+function _M:rewrite()
+  ngx.status = self.status_code
+  ngx.say(self.message)
+
+  return ngx.exit(ngx.status)
```
mikz: This should stop evaluating the policy chain. Is that expected?

mikz: I see the test. It is :)

gateway/src/apicast/policy/maintenance_mode/maintenance_mode.lua
davidor: Not sure what you mean by setting a header here.
Do you mean returning JSON or plain text according to the value of the `Accept` header of the request?

mikz: I mean setting also Content-Type header when the policy is active. JSON payload is incorrect without proper Content-Type. And using two policies to activate mantenance mode would be weird.

davidor: :+1:

https://github.com/3scale/APIcast/pull/1105/commits/bcee0a9400b1604fc90ceb96a05df61f59165b0f

PR #1107, opened 2019-08-09
Liquid: Available context move to readwrite linked list.

eloycoto:
This commit allows users to use capture in their liquid expressions and
assign variables if they needed.

Due to the available_context return a readonly list, capture function
cannot assign the variable and things start to mess up[0]. Due to allow
users to define variables in liquid is quite bad, a new read-write
linked list is created with an empty table to allow write to a temporal
table.

[0] Exception:
```
2019/08/09 13:51:23 [error] 15620#15620: *9 lua entry thread aborted: runtime error: /home/centos/gateway/src/apicast/linked_list.lua:19: readonly list
stack traceback:
coroutine 0:
        [C]: in function 'error'
        /home/centos/gateway/src/apicast/linked_list.lua:19: in function '__newindex'
        ./lua_modules/share/lua/5.1/liquid.lua:2512: in function 'define_var'
        ./lua_modules/share/lua/5.1/liquid.lua:2459: in function 'visit'
        ./lua_modules/share/lua/5.1/liquid.lua:1975: in function 'render'
        /home/centos/gateway/src/apicast/policy/headers/headers.lua:91: in function 'run_commands'
        /home/centos/gateway/src/apicast/policy/headers/headers.lua:151: in function </home/centos/gateway/src/apicast/policy/headers/headers.lua:147>
        /home/centos/gateway/src/apicast/policy_chain.lua:200: in function 'rewrite'
        ...s/gateway/src/apicast/policy/local_chain/local_chain.lua:59: in function <...s/gateway/src/apicast/policy/local_chain/local_chain.lua:54>
        /home/centos/gateway/src/apicast/policy_chain.lua:200: in function 'rewrite'
        rewrite_by_lua(lua_Iy6Gdl:498):3: in main chunk, client: 172.18.0.1, server: _, request: "GET /bridge-1/?user_key=132 HTTP/1.1", host: "one"

```

Fix THREESCALE-1911

### DIFF
--- a/t/apicast-policy-headers.t
+++ b/t/apicast-policy-headers.t
@@ -723,3 +723,62 @@ yay, api backend
 --- no_error_log
 [error]
 oauth failed with
+
+=== TEST 12: Liquid validate capture works
+Validate that capture function works correctly THREESCALE-1911
+--- backend
+  location /transactions/authrep.xml {
+    content_by_lua_block {
+      ngx.exit(200)
+    }
+  }
+--- configuration
+{
+  "services": [
+    {
+      "id": 42,
+      "backend_version":  1,
+      "backend_authentication_type": "service_token",
+      "backend_authentication_value": "token-value",
+      "proxy": {
+        "api_backend": "http://test:$TEST_NGINX_SERVER_PORT/",
+        "proxy_rules": [
+          { "pattern": "/", "http_method": "GET", "metric_system_name": "hits", "delta": 2 }
+        ],
+        "policy_chain": [
+          { "name": "apicast.policy.apicast" },
+          {
+            "name": "apicast.policy.headers",
+            "configuration":
+              {
+                "request":
+                  [
+                    {
+                      "op": "set",
+                      "header": "New-Header-1",
+                      "value": "{% capture foo%}bar{% endcapture%}{{foo}}",
+                      "value_type": "liquid"
+                    }
+                  ]
+              }
+          }
+        ]
+      }
+    }
+  ]
+}
+--- upstream
+  location / {
+     content_by_lua_block {
+       local assert = require('luassert')
+       assert.same("bar", ngx.req.get_headers()['New-Header-1'])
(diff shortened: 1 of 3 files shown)


t/apicast-policy-headers.t
```diff
@@ -723,3 +723,62 @@ yay, api backend
...
+          {
+            "name": "apicast.policy.headers",
+            "configuration":
+              {
+                "request":
+                  [
+                    {
+                      "op": "set",
+                      "header": "New-Header-1",
+                      "value": "{% capture foo%}bar{% endcapture%}{{foo}}",
```
davidor: Can we limit the number of allocations using the liquid lib? Otherwise, this might not be safe.
Ref: https://github.com/3scale/APIcast/blob/f2c4168c5be272bc3e1a7f12c63736ffd94dad6d/gateway/src/apicast/template_string.lua#L70

gateway/src/apicast/policy/ngx_variable.lua
```diff
@@ -17,7 +17,8 @@ local function context_values()
 end
 
 function _M.available_context(policies_context)
-  return LinkedList.readonly(context_values(), policies_context)
+  return LinkedList.readwrite({}, LinkedList.readonly(
```
davidor: Should we make it read-write in the `LiquidTemplateString:render` function instead?
That way other callers of this function will get a read-only list which I think is less surprising.

t/apicast-policy-headers.t
eloycoto: Does not look like that there is a way to limit it: 
https://github.com/chenxianyu2015/liquid-lua/blob/58e0a197445300538adab021d574f9375ebb6eba/lib/liquid.lua#L2610-L2620

And this one: 

https://github.com/chenxianyu2015/liquid-lua/blob/58e0a197445300538adab021d574f9375ebb6eba/lib/liquid.lua#L901

PR #1053, opened 2019-05-29
[WIP] TLS termination policy

mikz:
WIP of https://github.com/3scale/APIcast/issues/999

/cc @didierofrivia the UI depends on https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v1.5.0

PR #1108, opened 2019-08-15
TLS termination policy

eloycoto:
This PR got some work made by @mikz  around TLS termination in PR https://github.com/3scale/APIcast/pull/1053

### DIFF
--- a/gateway/src/apicast/policy/tls/tls.lua
+++ b/gateway/src/apicast/policy/tls/tls.lua
@@ -1,7 +1,7 @@
 -- This is a tls description.
 
 local policy = require('apicast.policy')
-local _M = policy.new('tls')
+local _M = policy.new('tls', "builtin")
 local tab_new = require('table.new')
 local ssl = require('ngx.ssl')
 local cjson = require('cjson')
@@ -77,7 +77,6 @@ end
 
 local function parse_certificates(self, certificate, private_key)
   local err
-
   self.certificate, err = ssl.parse_pem_cert(certificate)
   if err then return nil, err end
 
@@ -139,7 +138,6 @@ local function parse_config_certificates(configs)
 
   for _, config in ipairs(configs)do
     local cert, err = parse_config(config)
-
     if err then
       ngx.log(ngx.WARN, 'could not parse certificate ', cjson.encode(config))
     else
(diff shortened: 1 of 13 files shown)


gateway/src/apicast/policy/tls/tls.lua
```diff
@@ -154,10 +152,45 @@ end
...
+  local ok, err = ssl.set_priv_key(cert.certificate_key)
+  if not ok then
+    ngx.log(ngx.ERR, "failed to set DER private key: ", err)
+    return err
+  end
+end
+
+function _M:ssl_certificate()
+  if #self.certificates == 0 then
+    ngx.log(ngx.WARN, "No valid certificates loaded")
```
mikz: This can be printed just once when the policy is loaded, right?

gateway/src/apicast/policy/tls/tls.lua
```diff
@@ -154,10 +152,45 @@ end
...
+
+  local ok, err = ssl.clear_certs()
+  if not ok then
+      ngx.log(ngx.ERR, "failed to clear existing (fallback) certificates")
+      return ngx.exit(ngx.ERROR)
+  end
+
+  for _, cert in ipairs(self.certificates) do
+    if not set_certificate(cert) then
+      -- Certificate is set correctly, use this one and end the loop.
```
mikz: Shouldn't this terminate the ssl_certificate phase then?

gateway/src/apicast/policy/tls/tls.lua
```diff
@@ -154,10 +152,45 @@ end
...
+  end
+
+  local ok, err = ssl.clear_certs()
+  if not ok then
+      ngx.log(ngx.ERR, "failed to clear existing (fallback) certificates")
+      return ngx.exit(ngx.ERROR)
+  end
+
+  for _, cert in ipairs(self.certificates) do
+    if not set_certificate(cert) then
```
mikz: Why it returns false on success?

gateway/src/apicast/policy/tls/tls.lua
eloycoto: I'm not sure If I get this one, just abort the rest of ssl_certificate policies?

mikz: What is the desired outcome if you have this policy twice? Or some that tries to do the same. How they would know it is already set? You can terminate the phase by ngx.exit(ngx.OK)

gateway/src/resty/data_url.lua
```diff
@@ -1,9 +1,9 @@
+local _M = {}
```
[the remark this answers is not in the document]
eloycoto: If the comment is first if fails to load the module. This is the only way that I get it working.

spec/policy/tls/tls_spec.lua
```diff
@@ -0,0 +1,91 @@
+local _M = require('apicast.policy.tls')
+local ssl = require('ngx.ssl')
+
+local function assert_certificate_set()
+    assert.spy(ssl.clear_certs).was_called()
```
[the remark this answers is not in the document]
eloycoto: the parameters are pointers, so I do not think at all. Also are tested on the integration test with real data.

gateway/src/resty/data_url.lua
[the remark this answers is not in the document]
davidor: Would be good to add a comment that says that this is because there's another module with the same name, just so we do not forget.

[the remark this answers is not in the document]
eloycoto: Ok, spoke yesterday that I added that description on the commit message, I'll change now.

PR #1109, opened 2019-08-21
Tracing: Fix exception with OPENTRACING_FORWARD_HEADER parameter.

eloycoto:
Due to the value of a ENV variable in the apicast.conf is a table,
replace can't be used directly on liquid variables.

This change add a new tostring method in the template to be able to use
string filters on nginx templates.

Added a new method because table need to be used to be able to use
empty conditions.

FIX THREESCALE-1660

### DIFF
--- a/t/opentracing.t
+++ b/t/opentracing.t
@@ -39,3 +39,53 @@ qr/uber-trace-id: /
 --- udp_query eval
 qr/jaeger.version/
 --- wait: 10
+
+=== TEST 2: OpenTracing forward header
+Opentracing forward header is send to the upstream.
+--- env eval
+(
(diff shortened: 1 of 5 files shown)


t/opentracing.t
```diff
@@ -39,3 +39,53 @@ qr/uber-trace-id: /
...
+    }
+  }
+--- request
+GET /a_path?
+--- more_headers eval
+"foobar: value"
+--- error_code: 200
+--- no_error_log
+[error]
+--- udp_listen: 6831
```
davidor: Why do we need these ones?

eloycoto: Just to make sure that Opentracing is also enabled and no issues with the config.

PR #1115, opened 2019-09-13
RELEASE 3.7-beta1

eloycoto:


gateway/src/apicast/version.lua
```diff
@@ -1 +1 @@
-return "3.6.0"
+return "3.7.0"
```
davidor: We should use `3.7.0-beta1`.
We switch to `3.7.0` in the CR because CRs can become final.

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## [Unreleased]
+## [3.7.0-beta1]- 2019-09-13
```
davidor: There are a couple of lines at the end of the changelog that need to be modified too. Like in this commit of a previous beta version: https://github.com/3scale/APIcast/commit/49e639d7f1e1cadf4e4e30f913bf78d0eafad7e0

PR #1116, opened 2019-09-13
Logging: Added target host on logging policy.

eloycoto:
This commit add the way to get the target host on logging policy without
getting request headers.

Related to #1082

### DIFF
--- a/gateway/src/apicast/policy/logging/logging.lua
+++ b/gateway/src/apicast/policy/logging/logging.lua
@@ -77,6 +77,7 @@ local function get_request_context(context)
   }
 
   ctx.service = context.service or {}
+  ctx.original_request = context.original_request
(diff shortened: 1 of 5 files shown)


gateway/src/apicast/policy/logging/logging.lua
```diff
@@ -77,6 +77,7 @@ local function get_request_context(context)
   }
 
   ctx.service = context.service or {}
+  ctx.service.host = context.host
```
davidor: I think this modifies the `host` attr for the service meaning that every request for that service will be affected.

eloycoto: This happens on log_by_lua block, so it only modifies when the request finishes, so nothing is affected no?

PR #1125, opened 2019-10-14
Routing: Fix issues with invalid metric update on routing policy.

eloycoto:
Due to the changes in APIAP in 3Scale, a service can have multiple
backends(using routing policy) and multiple mapping rules can be hit and
it creates a bad behaviour around metrics reported to Apisonator.

This commit adds ownership on routing and mapping rules, where a
ownership can be used to filter the mapping rules that you want to
report. Here an example:

Routing policy:

```
path      backend           owner
----------------------------------
/foo/bar  http://1.1.1.1/   t1
/foo      http://2.2.2.2/   t2
```

Mapping rules:

```
path      metric        owner
-----------------------------------
/foo      hits          t2
/foo/bar  barMetric     t1
```

In this case, due to the request is going to /foo/bar the metric that
will be updated will be `barMetric` and it'll use `2.2.2.2` backend, the
mapping rules to match will filter by the owner, in this case, t2.

### DIFF
--- a/gateway/src/apicast/mapping_rule.lua
+++ b/gateway/src/apicast/mapping_rule.lua
@@ -13,6 +13,11 @@ local re_match = ngx.re.match
 local insert = table.insert
 local re_gsub = ngx.re.gsub
 
+-- This is introduced by API as a product feature. There are two kinds of
+-- mapping rules owner: `BackendAPI` that means that is used by API as a
+-- product and `Proxy` that means that it's a normal mapping rule.
+local BackendAPIconst = "BackendApi"
+
 local _M = {
   any_method = "ANY"
 }
@@ -79,7 +84,7 @@ local function matches_uri(rule_pattern, uri)
   return re_match(uri, format("^%s", rule_pattern), 'oj')
 end
 
-local function new(http_method, pattern, params, querystring_params, metric, delta, last)
+local function new(http_method, pattern, params, querystring_params, metric, delta, last, owner_id, owner_type)
   local self = setmetatable({}, mt)
 
   local querystring_parameters = hash_to_array(querystring_params)
@@ -91,6 +96,11 @@ local function new(http_method, pattern, params, querystring_params, metric, del
   self.system_name = metric or error('missing metric name of rule')
   self.delta = delta
   self.last = last or false
+
+  if owner_type == BackendAPIconst then
+    self.owner_id = owner_id
+  end
+
 
   self.querystring_params = function(args)
     return matches_querystring_params(querystring_parameters, args)
@@ -123,7 +133,9 @@ function _M.from_proxy_rule(proxy_rule)
     proxy_rule.querystring_parameters,
     proxy_rule.metric_system_name,
     proxy_rule.delta,
-    proxy_rule.last
+    proxy_rule.last,
+    tonumber(proxy_rule.owner_id),
+    proxy_rule.owner_type
   )
 end
 
--- a/gateway/src/apicast/mapping_rules_matcher.lua
+++ b/gateway/src/apicast/mapping_rules_matcher.lua
@@ -52,4 +52,21 @@ function _M.matches(method, uri, args, rules)
   return false
 end
 
+function _M.clean_usage_by_owner_id(rules, owner_id)
+  local usage = Usage.new()
+  if not owner_id then
+      return usage
+  end
+
+  for _, rule in ipairs(rules) do
+    -- Checking that the rule has owner_id, if it does not have owner_id means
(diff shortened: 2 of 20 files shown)


gateway/src/apicast/mapping_rule.lua
```diff
@@ -123,7 +124,8 @@ function _M.from_proxy_rule(proxy_rule)
     proxy_rule.querystring_parameters,
     proxy_rule.metric_system_name,
     proxy_rule.delta,
-    proxy_rule.last
+    proxy_rule.last,
+    tonumber(proxy_rule.owner_id)
```
guicassolato: If you really want to use `owner_id` here instead of the proposed `backend_id`, then you will also need `owner_type`. That's because the proxy rule can belong to a product (`owner_type == "Proxy"`) or to a backend (`owner_type == "BackendApi"`). In the first case, it should be considered for the match despite `owner_id`, while in the latter case, only if `owner_id` matches the one of the routing.

gateway/src/apicast/mapping_rules_matcher.lua
```diff
@@ -52,4 +52,21 @@ function _M.matches(method, uri, args, rules)
...
+  local usage = Usage.new()
+  if not owner_id then
+      return usage
+  end
+
+  for _, rule in ipairs(rules) do
+    -- Checking that the rule has owner_id, if it does not have owner_id means
+    -- that does not belongs to a backendAPI.
+    if rule.owner_id and owner_id ~= rule.owner_id then
+      usage:add(rule.system_name, 0 - (rule.delta or 0))
```
davidor: Nitpick: I think the first 0 is redundant.

gateway/src/apicast/proxy.lua
```diff
@@ -300,10 +302,19 @@ function _M:rewrite(service, context)
   context.credentials = ctx.credentials
 end
 
-function _M:access(service, usage, credentials, ttl)
+function _M:access(service)
```
davidor: Why did you remove the params? The call from `apicast.lua` still sends them: https://github.com/3scale/APIcast/blob/f02d7d89c521db910c7bf885b13097cbf227a3ae/gateway/src/apicast/policy/apicast/apicast.lua#L100

I'm not sure how this affects the return below where these params are ignored.

gateway/src/apicast/proxy.lua
```diff
@@ -300,10 +302,19 @@ function _M:rewrite(service, context)
...
   local ctx = ngx.ctx
+  local usage = ctx.usage
+
+  -- If routing policy changes the upstream and it only belongs to a specified
+  -- owner, we need to filter out the usage for APIs that are not used at all.
+  if ctx.context.route_upstream then
```
davidor: I think that we need to do this in the batching policy as well.

The batching policy acts in the access phase and grabs the usage in the context. When the batching policy is configured, we make sure that the apicast policy does not act in the access phase too by setting a flag. This is to avoid authorizing in both of them https://github.com/3scale/APIcast/blob/f02d7d89c521db910c7bf885b13097cbf227a3ae/gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua#L187

gateway/src/apicast/mapping_rules_matcher.lua
eloycoto: it's clear that we are substracting that, I would keep to make it clear.

spec/policy/routing/routing_spec.lua
```diff
@@ -17,12 +17,13 @@ describe('Routing policy', function()
 
       it('calls call() on the upstream passing the context as param', function()
         local routing = RoutingPolicy.new()
-        routing.upstream_selector = upstream_selector
 
+        routing.upstream_selector = upstream_selector
+        routing:access(context)
```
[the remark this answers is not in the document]
eloycoto: Yes, because select happens on access.

gateway/src/apicast/mapping_rule.lua
```diff
@@ -123,7 +130,9 @@ function _M.from_proxy_rule(proxy_rule)
     proxy_rule.querystring_parameters,
     proxy_rule.metric_system_name,
     proxy_rule.delta,
-    proxy_rule.last
+    proxy_rule.last,
+    tonumber(proxy_rule.owner_id),
```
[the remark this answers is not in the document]
eloycoto: It's always an integer.

eloycoto:
@davidor I added the change here: 
https://github.com/3scale/APIcast/pull/1125/files#diff-31313c92616b54028dac1ba183e6f79aR354-R356

PR #1129, opened 2019-10-31
Return 404 to client if no upstream defined.

eloycoto:
Due to the large use of routing policy, nowadays and upstream maybe is
not defined at all, APIcast returns 200 back and logs a new message
that a upstream is not defined.

Due to the work in THREESCALE-3775, the team decided to return a 404
back to the client to be aware that the upstream is not correct. Is not
a 502 status code due to the gateway is not defined at all and can be
confusing.

Related-to THREESCALE-3775
Related-to https://github.com/3scale/porta/pull/1385

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Return 404 back if the upstream is not defined [THREESCALE-3775](https://issues.jboss.org/browse/THREESCALE-3775) [PR #1129](https://github.com/3scale/APIcast/pull/1129)
+
+## [Unreleased]
+
+### Added
+
 - Introduce possibility of specifying policy order restrictions in their schemas. APIcast now shows a warning when those restrictions are not respected [#1088](https://github.com/3scale/APIcast/pull/1088), [THREESCALE-2896](https://issues.jboss.org/browse/THREESCALE-2896)
 - Added new parameters to logging policy to allow custom access log [PR #1089](https://github.com/3scale/APIcast/pull/1089), [THREESCALE-1234](https://issues.jboss.org/browse/THREESCALE-1234)[THREESCALE-2876](https://issues.jboss.org/browse/THREESCALE-2876), [PR #1116] (https://github.com/3scale/APIcast/pull/1116)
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Return 404 back if the upstream is not defined [THREESCALE-3775](https://issues.jboss.org/browse/THREESCALE-3775) [PR #1129](https://github.com/3scale/APIcast/pull/1129)
+
+## [Unreleased]
```
davidor: This looks wrong. There are 2 "unreleased" sections.

gateway/src/apicast/policy/apicast/apicast.lua
```diff
@@ -104,7 +105,7 @@ end
 function _M:content(context)
   if not context[self].upstream then
     ngx.log(ngx.WARN, "Upstream server not found for this request")
-    return
+    return errors.upstream_not_found(context.service)
```
davidor: What if the policy chain has a routing policy and an apicast policy with the `api_backend` defined?
This would ignore the `api_backend` in the apicast policy, right?

CHANGELOG.md
eloycoto: yes, this is to avoid merge conflicts with 3.7 release.

gateway/src/apicast/policy/apicast/apicast.lua
eloycoto: Routing policy made the call() to the upstream in the content phase, so never reaches this. 

Regards.

davidor: Sorry, when I looked at this I thought this was part of the Routing policy.

If I understood correctly, before this change, having a chain with apicast (no api_backend) -> routing worked, but now it's going to return a 404. Is that right?

eloycoto: yes!

davidor: Is this a breaking change? I think that before this, Porta always sent an api_backend. If this can be null only from this version, then I think it should be fine :+1:

eloycoto: No, it's not a breaking change. 

@guicassolato the api_backend as far as I saw in the config is not nil, could you point to the right thing here?

PR #1134, opened 2019-11-04
Release 3.7-CR1

eloycoto:


davidor:
`version.lua` should be updated as well.

PR #1141, opened 2019-11-19
OIDC: Using caching to retrieve issuer configs.

eloycoto:
This commit fix issues with the flipflop connections to the issuer.
Multiple users have hundred of services with the same issuer, and
multiple requests happen at the same time.

This commit has two pourposes:
  - Make the issuer config retrieve a bit faster in the APIcast and do
  not make multiple requests to retrieve the same info.
  - In case that the issuer has internment issues, avoid that one service
  loads correctly, and others do not load at all.

Fix [THREESCALE-3809](https://issues.redhat.com/projects/THREESCALE/issues/THREESCALE-3809)

- [ ] Should this caching be enabled with an Env variable?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
--  Avoid OIDC flip-flops issues when OIDC connectivity fails, now issuer configuration is cached. [THREESCALE-3809](https://issues.jboss.org/browse/THREESCALE-3809) [PR #1141](https://github.com/3scale/APIcast/pull/1141)
+-  Now the configuration of the issuer is cached to avoid flip-flop issues when OIDC connectivity fails.  [THREESCALE-3809](https://issues.jboss.org/browse/THREESCALE-3809) [PR #1141](https://github.com/3scale/APIcast/pull/1141)
 
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Added
+
+-  Avoid OIDC flip-flops issues when OIDC connectivity fails, now issuer configuration is cached. [THREESCALE-3809](https://issues.jboss.org/browse/THREESCALE-3809) [PR #1141](https://github.com/3scale/APIcast/pull/1141)
+
+
+### Fixed
+
+
 ## [3.7.0] - 2019-11-27
 
 `3.7.0-rc2` was considered final and became `3.7.0`.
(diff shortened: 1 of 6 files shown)


doc/parameters.md
```diff
@@ -26,6 +26,8 @@ Specifies the period (in seconds) that the configuration will be stored in the c
 - a positive number ( > 0 ): specifies the interval in seconds between configuration reload. For example, when APIcast is started with `APICAST_CONFIGURATION_CACHE=300` and `APICAST_CONFIGURATION_CACHE=boot`, it will load the configuration on boot, and will reload it every 5 minutes (300 seconds).
 - a negative number ( < 0 ): disables reloading. The cache entries will never be removed from the cache once stored, and the configuration will never be reloaded.
 
+This parameter is also used to store OpenID discovery configuration in the local cache, as the same behaviour as described above.
```
```suggestion
This parameter is also used to store OpenID discovery configuration in the local cache, as the same behavior as described above.
```

CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Added
+
+-  Avoid OIDC flip-flops issues when OIDC connectivity fails, now issuer configuration is cached. [THREESCALE-3809](https://issues.jboss.org/browse/THREESCALE-3809) [PR #1141](https://github.com/3scale/APIcast/pull/1141)
```
```suggestion
-  Now the configuration of the issuer is cached to avoid flip-flop issues when OIDC connectivity fails.  [THREESCALE-3809](https://issues.jboss.org/browse/THREESCALE-3809) [PR #1141](https://github.com/3scale/APIcast/pull/1141)
```

t/configuration-loading-with-oidc.t
```diff
@@ -220,3 +219,102 @@ GET /?user_key=uk
...
+location /transactions/authrep.xml {
+  content_by_lua_block {
+    ngx.exit(200)
+  }
+}
+
+location /api/ {
+  echo 'yay, api backend';
+}
+--- request
```
davidor: Question. Maybe I'm missing something. The configuration reload mode is set to "lazy". Shouldn't there be one request to each of the services? And then verify somehow that in one of the request there's the "retrieving OIDC" message and in the other it's not there because it should be cached?

gateway/src/resty/oidc/discovery.lua
```diff
@@ -40,10 +46,41 @@ function _M.new(http_backend)
             ssl = { verify = resty_env.enabled('OPENSSL_VERIFY') }
         }
     }
+    return _M.new_with_http_client(http_client)
+end
+
+function _M.new_with_http_client(http_client)
+    local self = setmetatable({ http_client = http_client }, mt)
+    self:init_cache()
```
davidor: Nitpick: I think that it's a bit unexpected that the method assigns `self.something`.
I'd change this to `self.cache = init_cache()` or just inline if it's simple enough.

gateway/src/resty/oidc/discovery.lua
```diff
@@ -40,10 +46,41 @@ function _M.new(http_backend)
...
+
+  return self.cache:get(issuer)
+end
+
+--- Insert the OIDC configuration for the issuer in the local cache with a
+--limit. If no ttl set will be 0
+function _M:save_issuer_in_cache(issuer, config, ttl)
+  local expires = tonumber(ttl) or 0
+
+  if not self.cache then
```
davidor: Nitpick : I think it's better to move this `if` to be the first line of the function.
I think it clearly says that if there's no cache there's nothing else to do.

t/configuration-loading-with-oidc.t
eloycoto: there is no host, so retrieve the configuration for both.

PR #1145, opened 2019-12-03
Openresty: Update base executor to custom Openresty package

eloycoto:
Due to the work to add custom modules in APIcast this commit updates
configuration to enable a new images, that uses a custom Openresty build
that are now build by us with custom modules.

WIP PR


t/http-proxy.t
```diff
@@ -1,7 +1,8 @@
 use lib 't';
 use Test::APIcast::Blackbox 'no_plan';
 
-require("t/http_proxy.pl");
+require("http_proxy.pl");
+# require("t/http_proxy.pl");
```
davidor: Not sure why this changed, but you can delete the commented one if not needed.

.circleci/config.yml
```diff
@@ -109,12 +109,12 @@ executors:
...
       S2I_VERSION: "1.1.12-2a783420"
       DOCKER_COMPOSE_VERSION: "1.16.1"
-      OPENRESTY_VERSION: "1.15.8.1-0"
+      OPENRESTY_VERSION: "1.17.4.1-0-centos8"
 
   openresty:
     working_directory: /opt/app-root/apicast
     docker:
-    - image: quay.io/3scale/s2i-openresty-centos7:1.15.8.1-0
+    - image: quay.io/3scale/s2i-openresty-centos7:1.17.4.1-0-centos8
```
davidor: Just a note so we don't forget: we should clean this. Having both centos7 and centos8 in the same image is confusing.
After we do that, we might need to change the docker-compose files used for development. Not sure if affects something else.

eloycoto: Something that I'm aware of, I'll update in the following days.

PR #1150, opened 2019-12-17
PATH-routing: url was not escaped correctly

eloycoto:
If PATH routing was defined, the uri was not escaped correctly so the
mapping rules will not work as expected.

Reported by: Carlo Palmiari <cpalmier@redhat.com>
Fix: THREESCALE-3468
Related-to:112fdfa502483983b462fb82a5e789003ab9f750


kevprice83:
This actually fixes [THREESCALE-4152](https://issues.redhat.com/browse/THREESCALE-4152)

PR #1152, opened 2020-01-03
Websocket support.

eloycoto:
This commit adds the option to enable Websocket pass trough to the
upstream API if the original connection is a websocket connection.

### DIFF
--- a/gateway/src/apicast/policy/websocket/websocket.lua
+++ b/gateway/src/apicast/policy/websocket/websocket.lua
@@ -1,0 +1,23 @@
+-- This policy enables Web socket pass through on APIcast
+
+local _M = require('apicast.policy').new('Websocket', 'builtin')
+
+local new = _M.new
+
+function _M.new(configuration)
+  local policy = new(configuration)
+  return policy
+end
+
+local function is_websocket_connection()
+  local headers = ngx.req.get_headers()
+  return headers["Upgrade"] ~= nil and headers["Sec-WebSocket-Key"] ~= nil
+end
+
+function _M:rewrite()
+  if is_websocket_connection() then
+    ngx.var.upstream_connection_header = "Upgrade"
+  end
+end
+
+return _M
--- a/t/apicast-policy-websocket.t
+++ b/t/apicast-policy-websocket.t
@@ -1,0 +1,164 @@
+use lib 't';
+use Test::APIcast::Blackbox 'no_plan';
+
+repeat_each(2);
+
+run_tests();
+
+__DATA__
+
+=== TEST 1: Simple Websocket connection pass through
+--- configuration
+{
+  "services": [
+    {
+      "id": 42,
+      "backend_version":  1,
+      "backend_authentication_type": "service_token",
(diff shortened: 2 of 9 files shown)


t/apicast-policy-websocket.t
```diff
@@ -0,0 +1,321 @@
+use lib 't';
+use Test::APIcast::Blackbox 'no_plan';
+
+repeat_each(1);
```
davidor: Any reason not to run them twice?

t/apicast-policy-websocket.t
```diff
@@ -0,0 +1,321 @@
...
+  }
+}
+--- request
+GET /?user_key=value
+--- error_code: 200
+--- no_error_log
+[error]
+
+
+=== TEST 2: Simple Websocket connection pass through
```
davidor: The tests numbers are wrong starting from this one.

davidor: Also, isn't this test checking the same as TEST 1?

t/apicast-policy-websocket.t
```diff
@@ -0,0 +1,321 @@
...
+}
+--- response_body env
+received: Data: Sending from client Type: text
+--- error_code: 200
+--- timeout: 10
+--- no_error_log
+[error]
+
+
+=== TEST 2: No websocket connection with policy does not change Upgrade header
```
davidor: Looks like this test is duplicated.

gateway/src/apicast/policy/websocket/websocket.lua
```diff
@@ -0,0 +1,23 @@
...
+end
+
+local function is_websocket_connection()
+  local headers = ngx.req.get_headers()
+  return headers["Upgrade"] ~= nil and headers["Sec-WebSocket-Key"] ~= nil
+end
+
+function _M:rewrite()
+  if is_websocket_connection() then
+    ngx.var.upstream_connection_header = "Upgrade"
```
davidor: Isn't the caller responsible for setting this header?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism#Upgrading_to_a_WebSocket_connection

eloycoto: It's reset on apicast.conf and this is to make sure that we set to Upgrade.

t/apicast-policy-websocket.t
eloycoto: it's usingg `ws` url.

gateway/src/apicast/policy/websocket/websocket.lua
davidor: but why is this header not set by the application that calls apicast?

eloycoto: Is set, but is overwritten by default here: 

https://github.com/3scale/APIcast/blob/master/gateway/conf.d/apicast.conf#L88

So, the changes allow only to change that, just because is a websocket server.

PR #1159, opened 2020-01-20
Mapping rules: Avoid issues with double slash.

eloycoto:
Due to the work make in APIAP, the double slash can happen quite easily,
as described in THREESCALE-3950. With this commit, when a new
mapping_rule is defined, double slash will be simplified to one to be
able to match with ngx.var.uri.

FIX THREESCALE-3950

### DIFF
--- a/gateway/src/apicast/policy/soap/soap.lua
+++ b/gateway/src/apicast/policy/soap/soap.lua
@@ -17,6 +17,7 @@ local MappingRule = require('apicast.mapping_rule')
 local Usage = require('apicast.usage')
 local mapping_rules_matcher = require('apicast.mapping_rules_matcher')
 local MimeType = require('resty.mime')
+local resty_url = require 'resty.url'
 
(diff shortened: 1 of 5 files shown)


gateway/src/apicast/policy/soap/soap.lua
```diff
@@ -90,6 +91,13 @@ end
 function _M:rewrite(context)
   local soap_action_uri = extract_soap_uri()
 
+  -- If the soap_action_uri is a complete url, the mapping rule is only the
```
davidor: I'm missing something. How is this change in the SOAP policy related to the original issue?

eloycoto: The action in the content-type can be a full url, so we need to pass it correctly: 

https://github.com/3scale/APIcast/blob/1266bc8dda609093958ef2b1c8592cd539f5a129/spec/policy/soap/soap_spec.lua#L112-L115

PR #1170, opened 2020-03-02
**POC** : Upstream Load Balancing Policy 

samu91gn:
This pull request is the result of a conversation about [THREESCALE-4579](https://issues.redhat.com/browse/THREESCALE-4579) and it is only intended as a Proof Of Concept to describe that RFE.


eloycoto:
We spoke offline, close this one, was only a poc.

PR #1179, opened 2020-03-11
GRPC policy: Fix routing upstream issue.

eloycoto:
This commit makes things super simpler in GRPC, and avoid the
miss-configuration that was made previously around upstream&balancers.

Before on GRPC we set the balancer, and all the balancer logic happened
on the policy itself, instead of using the apicast.upstream and
apicast.balancer.

With this change, the location is changed on the upstream.lua, so it
does not matter if was set in the apicast.policy, or in another policy,
in this case the routing policy.

This finally fix THREESCALE-4684


eloycoto:
This issue has test coverage on QE framework. @jsmadis

jsmadis:
> This issue has test coverage on QE framework.
@eloycoto Right now I am testing it manually. I will add a regression test asap I can.

PR #1184, opened 2020-04-02
Nginx variable request_id available in liquid templates

vvidovic:
Added possibility to use all Nginx variables in liquid templating for
setting headers. Copied from logging policy.
The use case is that I would like to be able to use Nginx request_id as
a tracking id value and propagate it to backend services so we can
have one unique identifier from API gateway to backends.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
-The format is based on [Keep a Changelog](http://keepachangelog.com/) 
+The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Added Rate-limit headers policy [THREESCALE-3795](https://issues.jboss.org/browse/THREESCALE-3795) [PR #1166](https://github.com/3scale/APIcast/pull/1166)
+
+- Added Nginx request_id variable to context [PR #1184](https://github.com/3scale/APIcast/pull/1184)
 
 ## [3.8.0] - 2020-03-24
 
@@ -209,7 +211,7 @@ Apart from the changes mentioned in this section, this version also includes the
 
 ## [3.4.0-rc2] - 2018-11-16
 
-### Fixed 
+### Fixed
 
 - Fix bug in the Default credentials policy. It was using the default credentials in some cases where it should not [PR #954](https://github.com/3scale/apicast/pull/954), [THREESCALE-1547](https://issues.jboss.org/browse/THREESCALE-1547)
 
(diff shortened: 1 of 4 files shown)


eloycoto:
Hi, 

You don't really need to do this. The information that you add in there is already in the context, so that information, in this case, is duplicated. 

See for example the following test: 
https://github.com/3scale/APIcast/blob/827aca2f2808160a4ce2dcd7452dd653df3e6829/t/apicast-policy-headers.t#L681-L694

If you have doubts about the information that it's stored in the context, you can use the liquid context debug policy: 

https://github.com/3scale/APIcast/blob/827aca2f2808160a4ce2dcd7452dd653df3e6829/t/apicast-policy-liquid-context-debug.t#L42-L72

vvidovic:
> Hi,
> 
> You don't really need to do this. The information that you add in there is already in the context, so that information, in this case, is duplicated.
> 
> See for example the following test:
> https://github.com/3scale/APIcast/blob/827aca2f2808160a4ce2dcd7452dd653df3e6829/t/apicast-policy-headers.t#L681-L694
> 
> If you have doubts about the information that it's stored in the context, you can use the liquid context debug policy:
> 
> https://github.com/3scale/APIcast/blob/827aca2f2808160a4ce2dcd7452dd653df3e6829/t/apicast-policy-liquid-context-debug.t#L42-L72

Hi,
are you sure that I can get Nginx request_id variable?

I used context debug policy but couldn't find information about request_id there.
For example:
* if I use {{ request_id }} in logging policy it shows Nginx request_id variable (like: 54fd9ce12a9af3822b22ffbfbaa28e48)
* if I use {{ request_id }} in headers policy it evaluates to blank
I tested this in local 3scale apicast installation (by changing policy lua code and restarting apicast) and after changes done here headers policy can use {{ request_id }} in same way as logging policy.

eloycoto:
Ahh ok, 

Got it now, If you need request_id in headers you don't need to do that, you can do it on the following function: 

https://github.com/3scale/APIcast/blob/827aca2f2808160a4ce2dcd7452dd653df3e6829/gateway/src/apicast/policy/ngx_variable.lua#L5-L18

Regards

vvidovic:
> Ahh ok,
> 
> Got it now, If you need request_id in headers you don't need to do that, you can do it on the following function:
> 
> https://github.com/3scale/APIcast/blob/827aca2f2808160a4ce2dcd7452dd653df3e6829/gateway/src/apicast/policy/ngx_variable.lua#L5-L18
> 
> Regards

Hi,
Am I understanding you correctly that I should build a new AIPcast for our 3scale if I want to change ngx_variable.lua (or even just add simple post hook to OpenShift template to change/replace this file)?

I added `request_id = ngx.var.request_id,` to return value of context_values() function and after restarting apicast variable is available in context.

Thanks a lot for clearing up this issue :)

eloycoto:
> 
I added request_id = ngx.var.request_id, to return value of context_values() function and after restarting apicast variable is available in context.

Yes, If you update the PR with this single change, It'll be merged, I'm ok with that. 

The main reason why all variables are exposed is to take care that no sensitive information is leaked due to invalid policy configuration. 

Regards.

PR #1186, opened 2020-04-22
Adds APICAST_CONFIGURATION_CACHE_STALE env variable

eloycoto:
This commit adds a way to store more time the configuration in the
cache if the config reload fails, it'll use the stale configuration for
the time that is specified.

FIX THREESCALE-4136

### DIFF
--- a/doc/parameters.md
+++ b/doc/parameters.md
@@ -8,8 +8,8 @@ Note that when deploying APIcast v2 with OpenShift, some of these parameters can
 
 ### `APICAST_BACKEND_CACHE_HANDLER`
 
-**Values:** strict | resilient  
-**Default:** strict  
+**Values:** strict | resilient
+**Default:** strict
 **Deprecated:** Use [Caching](../gateway/src/apicast/policy/caching/apicast-policy.json) policy instead.
 
 Defines how the authorization cache behaves when backend is unavailable.
@@ -18,7 +18,7 @@ Resilient will do so only on getting authorization denied from backend.
 
 ### `APICAST_CONFIGURATION_CACHE`
 
-**Values:** _a number_  
+**Values:** _a number_
 **Default:** 0
 
 Specifies the period (in seconds) that the configuration will be stored in the cache for. Can take the following values:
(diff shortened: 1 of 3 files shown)


doc/parameters.md
```diff
@@ -8,8 +8,8 @@ Note that when deploying APIcast v2 with OpenShift, some of these parameters can
 
 ### `APICAST_BACKEND_CACHE_HANDLER`
 
-**Values:** strict | resilient  
-**Default:** strict  
+**Values:** strict | resilient
+**Default:** strict
 **Deprecated:** Use [Caching](../gateway/src/apicast/policy/caching/apicast-policy.json) policy instead.
 
 Defines how the authorization cache behaves when backend is unavailable.
```
porueesq, line 15: It would be good to add the difference between strict and resilient.

doc/parameters.md
```diff
@@ -18,7 +18,7 @@ Resilient will do so only on getting authorization denied from backend.
 
 ### `APICAST_CONFIGURATION_CACHE`
 
-**Values:** _a number_  
+**Values:** _a number_
```
porueesq, line 21:
```suggestion
**Values:** _integer_
```

doc/parameters.md
```diff
@@ -18,7 +18,7 @@ Resilient will do so only on getting authorization denied from backend.
...
 **Default:** 0
 
 Specifies the period (in seconds) that the configuration will be stored in the cache for. Can take the following values:
```
porueesq, line 24:
```suggestion
Specifies the period in seconds that the configuration will be stored in the cache for. It can take the following values:
```

doc/parameters.md
```diff
@@ -28,6 +28,25 @@ Specifies the period (in seconds) that the configuration will be stored in the c
 
 This parameter is also used to store OpenID discovery configuration in the local cache, as the same behavior as described above.
 
+### `APICAST_CONFIGURATION_CACHE_STALE`
+
+**Values:** _a positive number_
```
porueesq, line 33:
```suggestion
**Values:** _integer_
```

doc/parameters.md
```diff
@@ -28,6 +28,25 @@ Specifies the period (in seconds) that the configuration will be stored in the c
...
+**Default:** 0
+
+When configuration cannot be gather correctly due to network errors, or any
```
porueesq, line 36: Suggestion:
_You can use this variable for the specified number of times to gather configuration details, so requests will not fail due to not found services. This variables is useful when the configuration cannot be correctly obtained due to network errors, or any other issues such as an old configuration stored in the cache._

doc/parameters.md
```diff
@@ -28,6 +28,29 @@ Specifies the period (in seconds) that the configuration will be stored in the c
...
+You can use this variable to specify the number of times that gather
+configuration can consequently fail. This is useful when any external issue on
+reload Apicast configuration, so APICast will use stale information, and it'll
+not delete the caching configuration. 
+
+Note, if a service is deleted, and this value is set, the service will be
+present in APICast configuration for the number of times that this parameter
+set.
+
+The number means, that `APICAST_CONFIGURATION_CACHE` time will be multiplied by
```
davidor, line 45: I'm not sure this is the best option from a UX perspective. Wouldn't it be clearer if this ENV just indicated the number of seconds directly instead of being a multiplier of another ENV?

eloycoto: I'm not sure, because time can be lower than the APICAST_CACHE, if someone forgot to change that the values will be a bit mess. 

With this way, the value is always CACHE time plus one stale configuration.

kevprice83:
After discussing with @eloycoto I'm not sure this is a great solution and in fact it seems like the stale data should be returned from the cache according to: https://github.com/openresty/lua-resty-lrucache#get

Looks like the problem might be [here](https://github.com/3scale/APIcast/blob/f64716c9a8420cdc8ea95265812147d434ffac89/gateway/src/apicast/configuration_store.lua#L96-L98) as it doesn't return the stale data and just exits hence we get `nil`. I reckon only customers who have very large configurations (not necessarily a large number of services) will expose this bug and maybe if there is a slow network speed at that moment it causes the problems we have been seeing.

PR #1187, opened 2020-04-27
Rewrite by verb 1130

### LINKED ISSUE #1130: Re-write rules determined by HTTP Verb
As far as I can tell it isn't possible to create a URL re-write rule based on whether the HTTP verb is a GET, POST, etc. If I have a single endpoint for all verbs like _/api_ then I can't apply re-write rules if I want GET to have a different target to POST. Instead I have to create _/api/retrieve_ for GET and _/api/create_ for POST. The reason this is an issue is that the REST API for JBoss BPMS 6.4 & RHPAM 7.x has different endpoints for POST and DELETE (I know that this will change with the Kogito release, where domain-specific APIs are possible). There are doubtless other use cases where being able to apply re-write rules based on the verb - e.g. in CQRS a GET might go to Data Grid and a POST might go to a DBMS.

pimg:
This resolves #1130 by taking an array of HTTP methods in the config for either the commands or the query_args_command. When the one or HTTP methods are applied in the config the rewrite rule only executes if the request method matches one in the config. If no methods are provided in the config the rewrite command will execute on all request methods.

### DIFF
--- a/gateway/src/apicast/policy/url_rewriting/apicast-policy.json
+++ b/gateway/src/apicast/policy/url_rewriting/apicast-policy.json
@@ -14,6 +14,59 @@
      "chain, then the mapping rules will apply to the original path."],
   "version": "builtin",
   "configuration": {
+    "definitions": {
+      "methods": {
+        "$id": "#/definitions/methods",
+        "type": "array",
+        "items": {
+          "type": "string",
+          "oneOf": [
+            {
+              "enum": [
+                "GET"
+              ],
+              "title": "GET"
+            },
+            {
+              "enum": [
+                "POST"
+              ],
+              "title": "POST"
+            },
+            {
+              "enum": [
+                "PUT"
+              ],
+              "title": "PUT"
+            },
+            {
+              "enum": [
+                "PATCH"
+              ],
+              "title": "PATCH"
+            },
+            {
+              "enum": [
+                "DELETE"
+              ],
+              "title": "DELETE"
+            },
+            {
+              "enum": [
+                "HEAD"
+              ],
+              "title": "HEAD"
+            },
+            {
+              "enum": [
+                "OPTIONS"
+              ],
+              "title": "OPTIONS"
+            }
+          ]
+        }
+      }
+    },
     "type": "object",
     "properties": {
       "commands": {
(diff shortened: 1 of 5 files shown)


eloycoto:
Hi @pimg 

Thanks for your contribution. Have been super busy days here, I'll review this policy asap, but I do not forget it. 

Regards

gateway/src/apicast/policy/url_rewriting/apicast-policy.json
```diff
@@ -107,6 +144,57 @@
               ],
               "default": "plain"
             }
+          },
+          "methods": {
```
eloycoto, line 148: It'll be nice if a new definition for methods can be done, we're trying to implement that because make things easier to follow and less verbose. As a example: 
https://github.com/3scale/APIcast/blob/master/gateway/src/apicast/policy/jwt_claim_check/apicast-policy.json#L13

Reference
https://json-schema.org/understanding-json-schema/structuring.html

gateway/src/apicast/policy/url_rewriting/url_rewriting.lua
```diff
@@ -28,6 +28,30 @@ local function substitute(func, subject, regex, replace, options)
   return new_uri, num_changes > 0
 end
 
+-- Returns true if the Method of the request is in the methods of the command meaning the rewrite rule should be applied
+-- Returns true if no Method is provided in the config for backwardscompatibility
+local function should_apply(methods)
+  local request_method = ngx.req.get_method()
+  local next = next
```
eloycoto, line 35: I would move this next to the top of the file.

gateway/src/apicast/policy/url_rewriting/url_rewriting.lua
```diff
@@ -28,6 +28,30 @@ local function substitute(func, subject, regex, replace, options)
...
+
+  local should_apply_command = true
+
+  if methods ~= nil and next(methods) ~= nil  then
```
eloycoto, line 39: This code can be a bit simpler, what about: 

```
local function should_apply(methods)
  local request_method = ngx.req.get_method()

  if methods == nil and next(methods) == nil  then
    return true
  end

  for _,v in pairs(methods) do
    if v == request_method then
      return true
    end
  end
  return false
end

```

gateway/src/apicast/policy/url_rewriting/url_rewriting.lua
```diff
@@ -28,6 +28,30 @@ local function substitute(func, subject, regex, replace, options)
   return new_uri, num_changes > 0
 end
 
+-- Returns true if the Method of the request is in the methods of the command meaning the rewrite rule should be applied
+-- Returns true if no Method is provided in the config for backwardscompatibility
+local function should_apply(methods)
```
eloycoto, line 33: What about if we change the function name to `is_match_method(methods)`

gateway/src/apicast/policy/url_rewriting/url_rewriting.lua
pimg: Hi Eloy thanks for the review comments, I'm working to resolve them.
One question: should't _if methods == nil **and** next(methods) == nil  then_
be: _if methods == nil **or** next(methods) == nil  then_
(So or instead of and) because in the case of and when methods == nil next(methods) will throw an error? Or am I missing something here?

gateway/src/apicast/policy/url_rewriting/apicast-policy.json
pimg: fixed

gateway/src/apicast/policy/url_rewriting/url_rewriting.lua
pimg: fixed

gateway/src/apicast/policy/url_rewriting/url_rewriting.lua
pimg: rest is changed

gateway/src/apicast/policy/url_rewriting/url_rewriting.lua
pimg: fixed

PR #1188, opened 2020-04-28
Policy: Add custom metrics

eloycoto:
This policy adds the avalability to add a new metric on the post_action,
so user can report back things like metric upstream headers, status code
or any other information.

FIX THREESCALE-5098

### DIFF
--- a/gateway/src/apicast/policy/custom_metrics/Readme.md
+++ b/gateway/src/apicast/policy/custom_metrics/Readme.md
@@ -1,0 +1,73 @@
+# Custom metrics policies
+
+This policy adds the availability to add metrics after the Upstream API
+response.  The main use case for this policy is to add metrics based on response
+code status, headers or different Nginx variables.
+
+## Caveats:
+
+Due to this needs some information from the upstream API, if the authrep with
+backend happens before the upstream call (Is not cached) the metric will not be
+incremented. This policy only increments metrics when the authrep is cached.
+
+This policy does not work with batching policy.
+
+## Configuration examples
+
+This policy increments the metric error, by the header increment, if the
+Upstream API returns a 400 status:
+
+```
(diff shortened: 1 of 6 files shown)


gateway/src/apicast/policy/custom_metrics/Readme.md
```diff
@@ -0,0 +1,73 @@
+# Custom metrics policies
+
+This policy adds the availability to add metrics after the Upstream API
+response.  The main use case for this policy is to add metrics based on response
+code status, headers or different Nginx variables.
+
+## Caveats:
+
+Due to this needs some information from the upstream API, if the authrep with
```
davidor, line 9: I think it would be better to use the authorize and report endpoints instead of authrep.
That way we could authorize => call upstream => report. The problem with the current approach is that it can lose many reports.

gateway/src/apicast/policy/custom_metrics/Readme.md
```diff
@@ -0,0 +1,73 @@
+# Custom metrics policies
+
+This policy adds the availability to add metrics after the Upstream API
```
davidor, line 3: I'd add somewhere that every metric used in this policy needs to be created manually in the admin portal.
Some users might expect APIcast to automatically create them when needed.

t/apicast-policy-custom-metrics.t
```diff
@@ -0,0 +1,429 @@
...
+     }
+  }
+--- request eval
+["GET /?user_key=value", "GET /?user_key=value"]
+--- error_code eval
+[200, 200]
+--- no_error_log
+[error]
+
+=== TEST 4: Rule does not left, metric is not added
```
davidor, line 254: I'm not sure what you mean

PR #1190, opened 2020-05-06
Pathrouting fix

palmieric:
This PR fix THREESCALE-5149

### DIFF
--- a/gateway/src/apicast/policy/find_service/path_based_finder.lua
+++ b/gateway/src/apicast/policy/find_service/path_based_finder.lua
@@ -1,6 +1,5 @@
 local mapping_rules_matcher = require 'apicast.mapping_rules_matcher'
 local escape = require("resty.http.uri_escape")
-local pretty = require("pl.pretty")
 
 local _M = {}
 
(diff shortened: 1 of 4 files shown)


kevprice83:
Would be good to update the unit tests. For example [here](https://github.com/3scale/APIcast/blob/master/spec/policy/find_service/path_based_finder_spec.lua#L15) we are passing an empty table. There are a few instances of this in the unit test, would be nice to add a test for both the happy and unhappy paths.

gateway/src/apicast/policy/find_service/path_based_finder.lua
```diff
@@ -17,7 +17,7 @@ function _M.find_service(config_store, host)
       if hosts[h] == host then
         local name = service.system_name or service.id
         ngx.log(ngx.DEBUG, 'service ', name, ' matched host ', hosts[h])
-        local matches = mapping_rules_matcher.matches(method, uri, {}, service.rules)
+        local matches = mapping_rules_matcher.matches(method, uri, ngx.req.get_uri_args(), service.rules)
```
eloycoto, line 20: This is ok, but I would move this to the top. Calling to a function is expensive, and we should be worried about.

Do the same as method, uri, etc..

palmieric: Ok sure.

PR #1191, opened 2020-05-06
HTTPS Proxy: Fix large body request.

eloycoto:
When request need to be stored in a temp file, the
ngx.req.get_body_data() will return nil. On some uses cases, users
report that large body is not correctly set. So things are failing on
their end.

This commit will check if the file has been created, read the data from
there and set the body correctly.

Fix THREESCALE-3863

### DIFF
--- a/gateway/src/apicast/http_proxy.lua
+++ b/gateway/src/apicast/http_proxy.lua
@@ -4,6 +4,7 @@ local resty_url = require "resty.url"
 local resty_resolver = require 'resty.resolver'
 local round_robin = require 'resty.balancer.round_robin'
 local http_proxy = require 'resty.http.proxy'
+local file_reader = require("resty.file").file_reader
 
 local _M = { }
 
@@ -96,9 +97,29 @@ local function forward_https_request(proxy_uri, uri)
         -- We cannot use resty.http's .get_client_body_reader().
         -- In POST requests with HTTPS, the result of that call is nil, and it
         -- results in a time-out.
+        --
+        --
+        -- If ngx.req.get_body_data is nil, can be that the body is too big to
+        -- read and need to be cached in a local file. This request will return
+        -- nil, so after this we need to read the temp file.
+        -- https://github.com/openresty/lua-nginx-module#ngxreqget_body_data
         body = ngx.req.get_body_data(),
         proxy_uri = proxy_uri
     }
+
+    if not request.body then
+        local temp_file_path = ngx.req.get_body_file()
+        ngx.log(ngx.INFO, "HTTPS Proxy: Request body is bigger than client_body_buffer_size, read the content from path='", temp_file_path, "'")
+
+        if temp_file_path then
+          local body, err = file_reader(temp_file_path)
+          if err then
+            ngx.log(ngx.ERR, "HTTPS proxy: Failed to read temp body file, err: ", err)
+            ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
+          end
+          request.body = body
+        end
+    end
 
     local httpc, err = http_proxy.new(request)
 
--- a/gateway/src/resty/file.lua
+++ b/gateway/src/resty/file.lua
@@ -1,0 +1,31 @@
+local co_yield = coroutine._yield
+local open = io.open
+
+local co_wrap_iter = require("resty.coroutines").co_wrap_iter
+
+local chunk_size = 2^13 -- 8kb
+
+local _M = {}
+
+-- returns an iterator if the file is correct where can be read in chuinks of
+-- 8kb.
+-- If file cannot be open, it'll return nil function and the error message.
+function _M.file_reader(filename)
+    local handle, err = open(filename)
+    if err then
+      return nil, err
+    end
+
+    return co_wrap_iter(function()
+      while true do
+          local chunk = handle:read(chunk_size)
+          if not chunk then
+            break
+          end
+          co_yield(chunk)
+      end
+      handle:close()
+    end)
+end
+
+return _M
--- a/t/http-proxy.t
+++ b/t/http-proxy.t
@@ -3,6 +3,16 @@ use Test::APIcast::Blackbox 'no_plan';
 
 require("http_proxy.pl");
 
+sub large_body {
+  my $res = "";
+  for (my $i=0; $i <= 1024; $i++) {
+    $res = $res . "1111111 1111111 1111111 1111111\n";
+  }
+  return $res;
+}
+
+$ENV{'LARGE_BODY'} = large_body();
+
 repeat_each(3);
 
 run_tests();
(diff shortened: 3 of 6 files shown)


gateway/src/apicast/http_proxy.lua
```diff
@@ -96,10 +97,24 @@ local function forward_https_request(proxy_uri, uri)
...
+        -- https://github.com/openresty/lua-nginx-module#ngxreqget_body_data
         body = ngx.req.get_body_data(),
         proxy_uri = proxy_uri
     }
 
+    if not request.body then
+        local temp_file_path = ngx.req.get_body_file()
+        ngx.log(ngx.INFO, "HTTPS Proxy: Request body is too big, read the content from path='", temp_file_path, "'")
+        if temp_file_path then
+          request.body = util.read_file(temp_file_path)
```
mikz, line 114: This is not safe for big files, because it loads everything in memory. It needs to be an iterator that reads the body chunk by chunk.

eloycoto: Hi,

I'm not sure about this, but Lua already make this on read_all option:

Call:
https://github.com/lua/lua/blob/61a4e64a6667bedaa882571c48a173ef5a4ba73b/liolib.c#L583

Read_all
https://github.com/lua/lua/blob/61a4e64a6667bedaa882571c48a173ef5a4ba73b/liolib.c#L526-L538

So, I'm not sure what do you want me to do here.

mikz: But `apicast.util.read_file` is https://github.com/3scale/APIcast/blob/14df1a809fd422809fc1c3de9e44f82e6a2a5ada/gateway/src/apicast/util.lua#L25-L37

Right?
It reads the whole file into memory.

It should not read the whole file in memory, but rather return an iterator that reads each chunk and returns it. 

Like https://github.com/ledgetech/lua-resty-http#get_client_body_reader. Actually, that looks exactly like that.

eloycoto: Hey,

I was looking at this, but at some point, the information need to be saved in
memmory, maybe I'm missing something, but let explain the body reader that you sent to me:


There are two functions involved:

- **_body_reader**: This is using coroutines, and this is correct, all here
  looks good, it returns a iterator and all is ok

https://github.com/ledgetech/lua-resty-http/blob/a6bd2e0eb1390e330e4fb10a48cced5a1f21fb66/lib/resty/http.lua#L444-L505


- **_read_body**: In this case, when reader() is called, all the information is
  saved in a table, and after that is all in a single variable. This, in my
  opinion, is a bit worse, because you're allocate a table instead of append of
  the string (But I need to do a performance check here ;-), not a strong word)

https://github.com/ledgetech/lua-resty-http/blob/a6bd2e0eb1390e330e4fb10a48cced5a1f21fb66/lib/resty/http.lua#L512-L536


So, I was thinking to have a metatable with some kind of magic method, but
AFAIK, always the content should be stored in memory. I was thinking something
like Filereader in Golang, but I didn't found a way to do that.

mikz: Nginx buffers large bodies (more than a few kb) into files, so they don't need to be stored in memory. 
Reading the *whole* file into memory can lead to DoS as it can have megabytes or gigabytes.
So a reading iterator will read only a chunk of the file at the time and iterating over it.

I have no idea if the iterator I linked would actually work when the body is buffered into a file. It was a mere suggestion of what should happen when reading a file from a filesystem - read it in chunks as an iterator. Maybe penlight has something?

gateway/src/apicast/http_proxy.lua
```diff
@@ -96,10 +97,30 @@ local function forward_https_request(proxy_uri, uri)
...
+        -- read and need to be cached in a local file. This request will return
+        -- nil, so after this we need to read the temp file.
+        -- https://github.com/openresty/lua-nginx-module#ngxreqget_body_data
         body = ngx.req.get_body_data(),
         proxy_uri = proxy_uri
     }
 
+    if not request.body then
+        local temp_file_path = ngx.req.get_body_file()
+        ngx.log(ngx.INFO, "HTTPS Proxy: Request body is too big, read the content from path='", temp_file_path, "'")
```
mikz, line 112: I think this could be a mere debug log level. Not fitting few kb does not really mean "too big".

gateway/src/apicast/util.lua
```diff
@@ -90,4 +90,6 @@ function _M.to_hash(table)
   return t
 end
 
+_M.read_file = read
```
mikz, line 93: Still needed?

gateway/src/resty/file.lua
```diff
@@ -0,0 +1,32 @@
+local co_yield = coroutine.yield
```
mikz, line 1: OpenResty has own version of `yield` https://github.com/openresty/lua-resty-core/commit/97b5825929ee4860b4a54b3905e2e191fc61f41e
Would be good to investigate and see what is the one we should use.

The API is injected in https://github.com/openresty/lua-nginx-module/blob/2a190736a58674086c3a27bf71a7993383fffb55/src/ngx_http_lua_coroutine.c#L288

Also since https://github.com/openresty/lua-nginx-module/pull/1239 is in it might be worth looking at how it affects us.

Also https://github.com/Tieske/Penlight/issues/265#issuecomment-423576100 is quite true :)

gateway/src/resty/file.lua
```diff
@@ -0,0 +1,32 @@
...
+      return nil, err
+    end
+
+    return co_wrap_iter(function()
+      while true do
+          local chunk = handle:read(chunk_size)
+          if not chunk then
+            break
+          end
+          chunk_size = chunk_size + chunk_size
```
mikz, line 25: This is increasing the chunk size with every iteration, right?
Can you test this manually calling the iterator?

t/http-proxy.t
```diff
@@ -952,3 +962,52 @@ proxy request: CONNECT 127.0.0.1:$TEST_NGINX_RANDOM_PORT HTTP/1.1
...
+    echo_foreach_split '\r\n' $echo_client_request_headers;
+    echo $echo_it;
+    echo_end;
+
+    access_by_lua_block {
+       assert = require('luassert')
+       assert.equal('https', ngx.var.scheme)
+       assert.equal('$TEST_NGINX_RANDOM_PORT', ngx.var.server_port)
+       assert.equal('test', ngx.var.ssl_server_name)
+    }
```
mikz, line 1003: Maybe you want to test the received body? At least its size?

eloycoto:
The same problem as the code, If I call `ngx.req.get_body_data` in the upstream API, it'll return nil, and I need to read the temp file to validate that it's working correctly, so you're testing with the same tools, no?

Also, http_proxy timeout if body is nil, so test it's working as expected. 

@jsmadis  could you take care of this in the testsuite?

Regards

mikz:
You can read it as one chunk using penlight for example. Or use the echo module from nginx to print the body?

PR #1192, opened 2020-05-25
Policy: IPCheck validate that port is not used.

eloycoto:
This commit fix the issue when port is added into the forwarder-for
header and IPCheck will not match correctly.

Fix THREESCALE-5258

### DIFF
--- a/gateway/src/apicast/policy/ip_check/client_ip.lua
+++ b/gateway/src/apicast/policy/ip_check/client_ip.lua
@@ -1,5 +1,6 @@
 local ipairs = ipairs
 local re = require('ngx.re')
+local resty_url = require 'resty.url'
 
 local _M = {}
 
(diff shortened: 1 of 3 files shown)


palmieric:
Looks great to me!

gateway/src/apicast/policy/ip_check/client_ip.lua
```diff
@@ -23,7 +24,13 @@ local function ip_from_x_forwarded_for_header()
     return nil
   end
 
-  return re.split(forwarded_for, ',', 'oj')[1]
+  -- THREESCALE-5258 forwarded_for can contain port. If port is in there IP
+  -- value will not be parsed correctly, parsing as url will get the correct
+  -- host
+  -- `:` split is not valid in case that it's IPv6 notation.
+  local host = re.split(forwarded_for, ',', 'oj')[1]
+  local uri = resty_url.parse("my://".. host)
```
davidor, line 32: Why is `my://` needed here?

eloycoto: if no scheme, the resty_url will fail, so it's the safest way to detect if IP/PORT is in there, also with IPv4/IPv6 combitation.

davidor: :+1:
It'd be useful to add a comment that clarifies that.

eloycoto: It's already in there :\

davidor: ah sorry, I missed that.

PR #1193, opened 2020-06-05
Policy: Camel proxy

eloycoto:
Camel netty proxy was used via http-proxy policy, but on TLS
connections, the CONNECT method[0] is not allowed, so the request needs
to be terminated by camel-proxy and send to the upstream API.

With this change, we duplicate the http-proxy and we set the proxy to
terminate the connection so that will work correctly on camel TLS
connections.

[0] https://www.ietf.org/rfc/rfc2817.txt
Fix THREESCALE-4867

### DIFF
--- a/gateway/src/apicast/policy/camel/Readme.md
+++ b/gateway/src/apicast/policy/camel/Readme.md
@@ -1,0 +1,77 @@
+#  Camel proxy policy
+
+This policy allows users to define a camel proxy where the traffic will be send
+over the defined proxy, the example traffic flow is the following:
+
+```
+   ,-.
+   `-'
+   /|\
+    |            ,-------.          ,---------.          ,----------.
+   / \           |Apicast|          |  CAMEL  |          |APIBackend|
+  User           `---+---'          `----+----'          `----------'
+   |  GET /resource  |                   |                    |
+   | --------------->|                   |                    |
+   |                 |                   |                    |
+   |                 |  Get /resource    |                    |
+   |                 |------------------>|                    |
+   |                 |                   |                    |
+   |                 |                   |  Get /resource/    |
+   |                 |                   | - - - - - - - - - >|
+   |                 |                   |                    |
+   |                 |                   |     response       |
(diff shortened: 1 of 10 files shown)


t/apicast-policy-camel.t
```diff
@@ -0,0 +1,208 @@
...
+
+require("http_proxy.pl");
+
+repeat_each(3);
+
+run_tests();
+
+__DATA__
+
+=== TEST 1: API backend connection uses proxy
```
davidor, line 12: Should this say "http proxy" to differentiate it from the other 2 tests?

gateway/src/apicast/policy/camel/Readme.md
```diff
@@ -0,0 +1,77 @@
+#  Camel proxy policy
```
davidor, line 1: Not sure if this is the best name for the policy. There's nothing camel-specific, it's valid for any proxy, right?

t/apicast-policy-camel.t
```diff
@@ -0,0 +1,208 @@
+use lib 't';
+use Test::APIcast::Blackbox 'no_plan';
+
+require("http_proxy.pl");
+
+repeat_each(3);
```
davidor, line 6: Why 3 times?

gateway/src/apicast/policy/camel/camel.lua
```diff
@@ -0,0 +1,57 @@
...
+  if not upstream then
+    return
+  end
+
+  upstream:set_skip_https_connect_on_proxy()
+end
+
+function _M:export()
+  return  {
+    get_http_proxy = function(uri)
```
davidor, line 48: I think it'd be good to document here who expects this function to be in the context.

gateway/src/apicast/policy/camel/Readme.md
eloycoto: The TLS is only for camel (With termination) and the policy name is suggested to be like that: 
https://issues.redhat.com/browse/THREESCALE-4867

PR #1197, opened 2020-06-15
Policy: rate_limit_headers fix cache issues.

eloycoto:
When testing rate_limit_headers the number of APICast workers are always set to
1. When the number of workers increased, the lru_cache is not longer effective
due to multiple threads.

Because we only have information about usage statistics on Authrep, and this is
using ngx.shared info, this policy also need to use shared information to be
able to reply without doing the authrep in all sessions.

Fix THREESCALE-3795

### DIFF
--- a/gateway/src/apicast/policy/rate_limit_headers/cache_entry.lua
+++ b/gateway/src/apicast/policy/rate_limit_headers/cache_entry.lua
@@ -2,6 +2,7 @@ local now = ngx.now
 
 local counter = require "resty.counter"
 local countdown_counter =  require("apicast.policy.rate_limit_headers.countdown_counter")
+local stringx = require 'pl.stringx'
 
 local _M = {}
 local mt = { __index = _M }
@@ -40,5 +41,26 @@ function _M:reset(max, remaining, reset)
   self.remaining = counter.new(remaining)
   self.reset = countdown_counter.new(reset, now())
 end
+
+function _M:export()
+  return string.format("%s#%s#%s",
+    self.limit:__tostring(),
+    self.remaining:__tostring(),
+    self.reset:remaining_secs_positive(now()))
+end
+
+function _M.import(usage, exported_data)
+  if not usage then
+    return nil
+  end
+
+  if not exported_data then
+    return _M.Init_empty(usage)
+  end
+
+  local data = stringx.split(exported_data, "#")
+  return _M.new(usage, data[1] or 0, data[2] or 0, data[3] or 0)
+end
+
 
 return _M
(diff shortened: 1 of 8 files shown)


gateway/src/apicast/policy/rate_limit_headers/cache_entry.lua
```diff
@@ -41,4 +42,25 @@ function _M:reset(max, remaining, reset)
...
+function _M.import(usage, exported_data)
+  if not usage then
+    return nil
+  end
+
+  if not exported_data then
+    return _M.Init_empty(usage)
+  end
+
+  local data = stringx.split(exported_data, "#")
```
davidor, line 61: Nitpick: the "#" could be extracted to a constant that could be used in this method and in `export()`.

spec/policy/rate_limit_headers/cache_entry_spec.lua
```diff
@@ -26,5 +26,35 @@ describe("Cache key", function()
...
+
+    it("Works as expected", function()
+
+      local entry = cache_entry.new(usage, 1, 10, 3)
+      assert.same(entry:export(), "1#10#3")
+
+      local data = cache_entry.import(usage, entry:export())
+      assert.same(data.limit:__tostring(), "1")
+      assert.same(data.remaining:__tostring(), "10")
+    end)
```
davidor, line 39: You forgot one assert for `reset` (3).

davidor: Same for the tests below.

spec/policy/custom_metrics/custom_metrics_spec.lua
```diff
@@ -0,0 +1,79 @@
...
+    stub(ngx.resp, 'get_headers', function() return {} end)
+
+    context = { usage = {} }
+    stub(context.usage, 'merge', function() return end)
+
+    usage = Usage.new()
+    usage:add("foo", 1)
+  end)
+
+  it('if Auth cache is disabled, report the usage', function()
```
davidor, line 26: Not too important, but in cases like this, I like to put the condition outside the `it` clause, so the `it` clause reads like natural English. Something like this:
```lua
describe('when the auth cache is disabled', function()
  it('reports the usage', function()
     ...
  )
)
```

gateway/src/apicast/policy/rate_limit_headers/cache_entry.lua
```diff
@@ -41,4 +42,25 @@ function _M:reset(max, remaining, reset)
   self.reset = countdown_counter.new(reset, now())
 end
 
+function _M:export()
```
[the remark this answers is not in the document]
eloycoto: I do not think so, export/import is more declarative when you read the code. tostring was my first idea :sweat_smile:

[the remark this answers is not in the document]
davidor: :+1:

PR #1205, opened 2020-06-26
Release 3.9 CR1

eloycoto:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 
-## [Unreleased]
+## [3.9.0-alpha1] 2020-06-26
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


gateway/src/apicast/version.lua
```diff
@@ -1 +1 @@
-return "3.6.0"
+return "3.9.0"
```
davidor, line 1: this should have an alpha or beta suffix, right?

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 
-## [Unreleased]
+## [3.9.0-cr1] 2020-06-26
```
davidor, line 8: Should be `alpha1` because it's an ER release, right?

gateway/src/apicast/version.lua
eloycoto: I think that we always keep it as 0. but let me change.

PR #1209, opened 2020-06-30
Policy: rate limit headers does not send invalid value.

eloycoto:
When the backend does not reply with limit headers, APIcast should send
no limit headers at all. This change make sure that the value is not
send at all.

Also, make sure that the value of limit and reset is always positive or
0.

THREESCALE-3795

### DIFF
--- a/gateway/src/apicast/policy/rate_limit_headers/rate_limit_headers.lua
+++ b/gateway/src/apicast/policy/rate_limit_headers/rate_limit_headers.lua
@@ -2,6 +2,7 @@ local policy = require('apicast.policy')
 local _M = policy.new('Rate Limit Headers', 'builtin')
 
 local usage_cache = require "cache"
+local math_max = math.max
 
 local new = _M.new
 local get_phase = ngx.get_phase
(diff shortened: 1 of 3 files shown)


gateway/src/apicast/policy/rate_limit_headers/rate_limit_headers.lua
```diff
@@ -55,9 +55,22 @@ local function decrement(self, usage)
     return self.cache:decrement_usage_metric(usage)
 end
 
+-- return 0 if the number is negative
+local function positive_number(number)
+  if tonumber(number) <= 0 then
```
davidor, line 60:
```lua
return math.max(tonumber(number), 0)
```
?

gateway/src/apicast/policy/rate_limit_headers/rate_limit_headers.lua
```diff
@@ -55,9 +55,22 @@ local function decrement(self, usage)
...
+    return 0
+  end
+  return tonumber(number)
+end
+
 local function add_headers(info)
-    ngx.header[limit_header] = info.limit
-    ngx.header[remaining_header] = info.remaining
+    ngx.header["eloy"] = "test"
```
davidor, line 67: :eyes:

eloycoto: sorry, my head :-(

PR #1195, opened 2020-06-11
Proxy Protocol support

y-tabata:
Closes https://github.com/3scale/APIcast/issues/1194.

I have no idea for the prove test of the proxy protocol. I would be very happy if you give me any kind of comment.

PR #1211, opened 2020-07-09
Proxy Protocol support

### LINKED ISSUE #1194: Proxy Protocol support
There is a big demand for Proxy Protocol, for example, the case NLB is used as an OpenShift router, and the IP check policy is used to access control.
So, we'd like to add proxy protocol support.

# Spec
To support the proxy protocol, we only need to add the "proxy_protocol" parameter to the listen directive.
https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol/

- Prepare `APICAST_HTTP_PROXY_PROTOCOL` and `APICAST_HTTPS_PROXY_PROTOCOL` envs for both HTTP and HTTPS listers to add the "proxy_protocol" parameter.
- Add `proxy_protocol_addr` to the client_ip_sources of the IP check policy, get the client IP from `$proxy_protocol_addr` nginx variable when using proxy protocol.

eloycoto:
Closes #1194.

This is the same as PR https://github.com/3scale/APIcast/pull/1195 but with integration test.

### DIFF
--- a/t/proxy-protocol.t
+++ b/t/proxy-protocol.t
@@ -1,0 +1,189 @@
+use lib 't';
+use Test::APIcast::Blackbox 'no_plan';
+
+# These test are a bit difficult to understand. Due to the Nginx is set to listen
+# proxy_protocol, all backends upstream need to use another port, if not the
+# backend client and API will fail, this is why a new server is added on
+# --backend test section
+#
+# Also, this is the reason why ip check policy is tested here.
+
+
+require("policies.pl");
+repeat_each(3);
+$ENV{TEST_NGINX_HTML_DIR} ||= "$Test::Nginx::Util::ServRoot/html";
+run_tests();
+
+__DATA__
+
+=== TEST 1: Simple proxy-procol connection
+--- init eval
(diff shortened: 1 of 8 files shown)


t/proxy-protocol.t
```diff
@@ -0,0 +1,189 @@
+use lib 't';
+use Test::APIcast::Blackbox 'no_plan';
+
+# These test are a bit difficult to understand. Due to the Nginx is set to listen
```
porueesq, line 4:
```suggestion
# Due to the Nginx is set to listen
```

t/proxy-protocol.t
```diff
@@ -0,0 +1,189 @@
...
+# proxy_protocol, all backends upstream need to use another port, if not the
+# backend client and API will fail, this is why a new server is added on
+# --backend test section
+#
+# Also, this is the reason why ip check policy is tested here.
```
porueesq, line 9:
```suggestion
# Additionally, this is the reason why the IP Check policy is tested here.
```
You can add this sentence to the previous paragraph.

PR #1216, opened 2020-08-06
Policy:3scale_batcher: Fix issues when caching policy is active.

eloycoto:
A user reported that when the caching mode is enabled, in allow mode, and 3scale-batcher is active, the handler for allowing all calls is no working as expected, and users got 403 response.

This commit allows users to get a correct authz when allow mode is enabled.

Fix THREESCALE-5753


gateway/src/apicast/policy/3scale_batcher/3scale_batcher.lua
```diff
@@ -143,6 +143,14 @@ local function handle_backend_denied(self, service, transaction, status, headers
   return error(service, rejection_reason)
 end
 
+local function handle_backend_response_using_cache_handler(cache, transaction, backend_status, cache_handler)
```
davidor, line 146: This looks almost the same as the existing `update_downtime_cache` function.

davidor (approved):

PR #1217, opened 2020-08-06
Logging: Add credentials on the obje

eloycoto:
Some users want to debug credentials on logs, so this adds credentials
object to the liquid object

Fix THREESCALE-5723


davidor:
I'm not sure this is a good idea. It could lead to security issues.
I think that at least, this should be optional and off by default.

eloycoto:
Well, context.credentials is shared to any other liquid policy no? So user can send credentials in any liquid policy, like headers :\

davidor:
I think that case is a bit different because the user needs to say explicitly in the configuration of the policy that something needs to be done with the credentials. In this case, the credentials will be logged just by enabling the policy.

eloycoto:
No, in this case, users need to enable the credentials on logs, will not be logged as default in any case.

davidor:
Ah I see, I misunderstood :+1:

PR #1222, opened 2020-08-24
Config: increase cache limits on demand.

eloycoto:
By default, 1000 services were enough, currently a lot of new users are hitting
thousands of services in a single APICast instance. This commit adds a option
to have a bigger cache size if the user desired that.

Fix THREESCALE-5308


davidor (approved):

dfennessy (commented):
One suggestion.

PR #1227, opened 2020-09-04
Policy: Add request/response content limit

eloycoto:
This policy check the content-caching header value and rejects if the
content is larger than  the value.


t/apicast-policy-limits.t
```diff
@@ -0,0 +1,301 @@
...
+          response = response .. "XXXXXXXXXX"
+      end
+      ngx.header['Content-Length'] = #response
+      ngx.say(response)
+     }
+  }
+
+--- request
+POST /test
+--- error_code: 413
```
davidor, line 239: I think this status code is just for the request size, no the response one.

gateway/src/apicast/policy/limits/limits.lua
```diff
@@ -0,0 +1,59 @@
+local _M  = require('apicast.policy').new('Logging Policy', 'builtin')
```
davidor, line 1: Incorrect name.

gateway/src/apicast/policy/limits/limits.lua
```diff
@@ -0,0 +1,59 @@
...
+      context.request_limited = true;
+      ngx.status = 413
+      return ngx_exit(413)
+  end
+end
+
+function _M:body_filter(context)
+  if context.request_limited then
+    ngx.arg[1] = "Payload Too Large"
+    ngx.arg[2] = true
```
davidor, line 54: Why is this one needed?

t/apicast-policy-limits.t
eloycoto: can't follow this one.

gateway/src/apicast/policy/limits/limits.lua
eloycoto: Because if not response body cannot be written at all.

t/apicast-policy-payload_limits.t
```diff
@@ -0,0 +1,301 @@
...
+"POST /test \n" . $ENV{LARGE_BODY}
+--- response_body
+Payload Too Large
+--- error_code: 413
+--- no_error_log
+[error]
+
+
+=== TEST 3: Request body size smaller than the limit
+-- ONLY
```
davidor, line 131: :eyes:

t/apicast-policy-limits.t
davidor: What I mean is that I think that the response code 413 is used only when the request payload is higher than the limit. However, I'm not sure what the code should be when the _response_ payload returned by the API upstream is higher than the limit.

davidor (approved):

PR #1229, opened 2020-09-24
SERVICES_FILTER: Fix issues with oidc.

eloycoto:
The oidc-service mapping is not based on a key, so each time that a
service is deleted, the mapping to the given service is broken.

With this change, the OIDC config is stored with the service id, so in
case that some services are removed, it'll keep it correctly mapped.

FIX THREESCALE-6042

Missing unit test

### DIFF
--- a/t/apicast-subset-of-services.t
+++ b/t/apicast-subset-of-services.t
@@ -1,5 +1,8 @@
 use Test::APIcast::Blackbox 'no_plan';
 
+our $private_key = `cat t/fixtures/rsa.pem`;
+our $public_key = `cat t/fixtures/rsa.pub`;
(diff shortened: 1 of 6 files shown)


t/apicast-subset-of-services.t
```diff
@@ -184,3 +187,83 @@ __DATA__
 ["yay, api backend\n", "yay, api backend two\n", ""]
 --- error_code eval
 [200, 200, 404]
+
+
+=== TEST 4: Verify OIDC issue
```
davidor, line 192: Can you add the link to the issue here?

davidor (approved):

PR #1232, opened 2020-10-13
Release 3.10.0-alpha1

eloycoto:


gateway/src/apicast/version.lua
```diff
@@ -1 +1 @@
-return "3.6.0"
+return "3.10.0-ER1"
```
davidor, line 1: This should be `alpha1` right?

davidor (approved):

PR #1235, opened 2020-10-27
Resolver: using Fully qualified names.

eloycoto:
When adding a dot to the end of the domain, the search string from
resolv.conf should not be added, because the domain is absolute.

Adding this, will save a few DNS request to the DNS server because no
cluster.local, etc.. will be added to the Qname.

More info in relation to this:
https://superuser.com/a/1468139

Fix THREESCALE-4752

### DIFF
--- a/gateway/src/resty/resolver.lua
+++ b/gateway/src/resty/resolver.lua
@@ -3,6 +3,7 @@ local next = next
 local open = io.open
 local gmatch = string.gmatch
 local match = string.match
+local sub = string.sub
 local format = string.format
 local find = string.find
 local insert = table.insert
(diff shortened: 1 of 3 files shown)


gateway/src/resty/resolver.lua
```diff
@@ -280,24 +280,40 @@ local function valid_answers(answers)
...
     answers, err = dns:query(query, options)
     if valid_answers(answers) then
       cache:save(answers)
-      break
+      return answers, nil
+    end
+    return nil, err
+  end
+
+  if string.sub(qname, -1) == "." then
```
mikz, line 304: This is not optimal. Would be better to cache the `string.sub` to a local variable. The same for `qname:sub`.

https://www.youtube.com/watch?v=FfhEdF40nhQ

gateway/src/resty/resolver.lua
```diff
@@ -280,24 +280,41 @@ local function valid_answers(answers)
...
     answers, err = dns:query(query, options)
     if valid_answers(answers) then
       cache:save(answers)
-      break
+      return answers, nil
+    end
+    return nil, err
+  end
+
+  local qname_last_char = string.sub(qname, -1)
```
mikz, line 304: This is not optimal. Would be better to cache the `string.sub` to a local variable. The same for goes for `qname:sub`. Not the output of the call, but the function being called `local sub = string.sub`.

https://www.youtube.com/watch?v=FfhEdF40nhQ

mikz: As explained in the video, using `string.sub` needs to lookup `string`, which is in `_G` and going that far up in stack breaks JIT. 

This might not be on a hot path, but better make it everywhere, so you don't have to think about it :)

eloycoto: sleepy morning :-( Fixed it now :sweat_smile:

mikz (approved):

PR #1240, opened 2020-11-11
CLI: increase body size on template render.

eloycoto:
A user reported that when a lot of services in K8s environment started
with apicast name, APICast cannot started because of the body size of
the template.

Because the body limit is set to nil, but liquid template set to 100000
[0] and started to fail in there.

The main reason to print all ENV variables on the nginx.conf, is because
the following code [1]

[0] https://github.com/3scale/liquid-lua/blob/a7f4119e96054f08df1fb8dce2be3469eb7ffb5f/lib/liquid.lua#L2675-L2686
[1] https://github.com/3scale/APIcast/blob/fba67b02200a4df231546145296927dea816c33f/gateway/http.d/init.conf#L46-L58

Fix THREESCALE-6315

### DIFF
--- a/gateway/src/apicast/cli/template.lua
+++ b/gateway/src/apicast/cli/template.lua
@@ -94,7 +94,10 @@ function _M:interpret(str)
     local context = self.context
     local filesystem = self.filesystem
     local filter_set = FilterSet:new()
-    local resource_limit = ResourceLimit:new(nil, 1000, nil)
+    -- The 1000000 limit, is to increase the body limits. This was triggered by
+    -- a issue THREESCALE-6315 where a lot of ENV variables are render in the
+    -- nginx.conf file.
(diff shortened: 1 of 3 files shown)


gateway/src/apicast/cli/template.lua
```diff
@@ -94,7 +94,7 @@ function _M:interpret(str)
     local context = self.context
     local filesystem = self.filesystem
     local filter_set = FilterSet:new()
-    local resource_limit = ResourceLimit:new(nil, 1000, nil)
+    local resource_limit = ResourceLimit:new(1000000, 1000, nil)
```
davidor, line 97: Would be good to have a link to the issue here so we remember why this value is not the default one.

davidor (approved):

PR #1248, opened 2021-01-08
Add APICAST_ACCESS_LOG_BUFFER option

eloycoto:
This commit add a new way to write access log, based on chunks, so
access log writes only happens when a buffer is full and this improves the
performance by reducing the number of system calls.

### DIFF
--- a/doc/parameters.md
+++ b/doc/parameters.md
@@ -105,6 +105,15 @@ Specifies the log level for the OpenResty logs.
 Defines the file that will store the access logs.
 
 
+### APICAST_ACCESS_LOG_BUFFER
+
+**Values:** integer
+**Default**: nil
+
+Allows access log writes to be writes in chunks on bytes, so less system calls
+that improves the performance of the whole gateway.
+
+
 ### `APICAST_OIDC_LOG_LEVEL`
 
 **Values:** debug | info | notice | warn | error | crit | alert | emerg  
--- a/doc/parameters.md
+++ b/doc/parameters.md
@@ -110,8 +110,8 @@ Defines the file that will store the access logs.
 **Values:** integer
 **Default**: nil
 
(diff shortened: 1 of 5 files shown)


davidor (approved):
Looks good. Do you have any numbers? Just to get an idea about the perf gains that a log buffer of 1K or 4K for example gives.

doc/parameters.md
```diff
@@ -105,6 +105,15 @@ Specifies the log level for the OpenResty logs.
 Defines the file that will store the access logs.
 
 
+### APICAST_ACCESS_LOG_BUFFER
+
+**Values:** integer
+**Default**: nil
+
+Allows access log writes to be writes in chunks on bytes, so less system calls
```
porueesq, line 113:
```suggestion
Allows access log writes to be included in chunks of bytes, resulting on fewer system calls
```

doc/parameters.md
```diff
@@ -105,6 +105,15 @@ Specifies the log level for the OpenResty logs.
...
+that improves the performance of the whole gateway.
```
porueesq, line 114:
```suggestion
that improve the performance of the whole gateway.
```

t/apicast-policy-logging.t
```diff
@@ -585,3 +585,61 @@ OK
 [qr/^Status\:\:200 USER_KEY\:\:123 42/]
 --- no_error_log
 [error]
+
+=== TEST 14: APICAST_ACCESS_LOG_BUFFER env parameter
+When buffer is enabled, log will be bump in chunks
```
porueesq, line 590:
```suggestion
When buffer is enabled, log will be bumped in chunks
```

porueesq (changes requested):
Some suggestions.

porueesq (approved):

PR #1250, opened 2021-01-12
Performance improvements

eloycoto:
Fix: https://issues.redhat.com/browse/THREESCALE-6570

### DIFF
--- a/gateway/conf/nginx.conf.liquid
+++ b/gateway/conf/nginx.conf.liquid
@@ -34,6 +34,7 @@ events {
 
 http {
   sendfile           on;
+  sendfile_max_chunk 512k;
   tcp_nopush         on;
   tcp_nodelay        on;
   server_tokens off;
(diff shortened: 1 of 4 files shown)


gateway/conf/nginx.conf.liquid
```diff
@@ -34,6 +34,7 @@ events {
 
 http {
   sendfile           on;
+  sendfile_max_chunk 512k;
```
davidor, line 37: I've read the description in https://nginx.org/en/docs/http/ngx_http_core_module.html#sendfile_max_chunk but it's not clear to me in which cases this could be worse than the default. What scenario did you test and what perf numbers did you get?

eloycoto: Mainly, when large bodies are send. We have an Hyperfoil testcase where a lot of large bodies are in place. With this change, we got around 5% more rps.

davidor: :+1:
Do you know if there's a performance penalty for use cases with small or no request bodies?

eloycoto: no penalty, because is write in chunks  instead of fill the chunk, so not issues at all.

davidor (approved):

PR #1251, opened 2021-01-13
CORS: Enable multiple Origins value based on regexp.

eloycoto:
This will check if the allow_origins is a valid regexp, and if it is,
will match with the Origin header value, it'll set the
Access-Control-Allow-Origin header to the Origin value.

Examples:
```
Config             | Origin Request Header Value  | Access-Control-Allow-Origin header
(api|web).test.com | http://web.test.com          | http://web.test.com
(api|web).test.com | http://api.test.com          | http://api.test.com
(api|web).test.com | http://staging.test.com      | Not header set
http://test.com    | http://api.test.com          | http://test.com
*                  | http://api.test.com          | *
blank              | http://api.test.com          | http://api.test.com
```

Fix: THREESCALE-6569

### DIFF
--- a/gateway/src/apicast/policy/cors/apicast-policy.json
+++ b/gateway/src/apicast/policy/cors/apicast-policy.json
@@ -46,7 +46,7 @@
         }
       },
       "allow_origin": {
-        "description": "Origin allowed for CORS requests. The field expects only one origin (e.g. https://example.com) or '*'. If left blank, the value of the 'Origin' request header will be used.",
+        "description": "Origin allowed for CORS requests. The field expects only one origin (e.g. 'https://example.com') or '*'. If left blank, the value of the 'Origin' request header will be used. In order to allow more than one origin it is possible to use a regular expression, if it matches with Origin header value, the value will be set to the Origin Value. In case it does not match, the header will not set at all. Example: '(api|web).test.com' wil match both 'api.test.com' and 'web.test.com'.",
         "type": "string"
       },
       "allow_credentials": {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Buffer access log on chunks [PR #1248](https://github.com/3scale/APIcast/pull/1248) [THREESCALE-6563](https://issues.redhat.com/browse/THREESCALE-6563)
 - Added sendfile_max_chunk to the worker [PR #1250](https://github.com/3scale/APIcast/pull/1250) [THREESCALE-6570](https://issues.redhat.com/browse/THREESCALE-6570)
 - Increased api-keys shared memory size [PR #1250](https://github.com/3scale/APIcast/pull/1250) [THREESCALE-6570](https://issues.redhat.com/browse/THREESCALE-6570)
+- Add support to multiple Origin based on regexp [PR #1251](https://github.com/3scale/APIcast/pull/1251) [THREESCALE-6569](https://issues.redhat.com/browse/THREESCALE-6569)
+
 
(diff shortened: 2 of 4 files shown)


gateway/src/apicast/policy/cors/apicast-policy.json
```diff
@@ -46,7 +46,7 @@
         }
       },
       "allow_origin": {
-        "description": "Origin allowed for CORS requests. The field expects only one origin (e.g. https://example.com) or '*'. If left blank, the value of the 'Origin' request header will be used.",
+        "description": "Origin allowed for CORS requests. The field expects only one origin (e.g. https://example.com) or '*'. If left blank, the value of the 'Origin' request header will be used. In case of valid regexp, and if it matches with Origin header value, the value will be set to the Origin Value. In case of does not match, the header will not set at all.",
```
damianpm, line 49: I think this messages could be improved if we add some examples and a bit of markup (I don't know if html is being escaped):

```suggestion
        "description": "Origin allowed for CORS requests. The field expects only one origin (e.g. <code>https://example.com</code>) or '*'. If left blank, the value of the 'Origin' request header will be used. <br> In order to allow more than one origin it is possible to use a regular expression, if it matches with Origin header value, the value will be set to the Origin Value. In case it does not match, the header will not set at all. Example: <code>(api|web).test.com</code> wil match both <code>api.test.com</code> and <code>web.test.com</code>",
```
It will look like:
Screenshot 2021-01-14 at 11 08 41

damianpm: WDYT @thomasmaas ?

thomasmaas: Yes but I think it has to be plain text, not HTML.

damianpm: In that case, we can simply use quotes instead of <code> tags:
```suggestion
        "description": "Origin allowed for CORS requests. The field expects only one origin (e.g. "https://example.com") or '*'. If left blank, the value of the 'Origin' request header will be used. In order to allow more than one origin it is possible to use a regular expression, if it matches with Origin header value, the value will be set to the Origin Value. In case it does not match, the header will not set at all. Example: "(api|web).test.com" wil match both "api.test.com" and "web.test.com".",
```
It will look like this:

> Screenshot 2021-01-14 at 11 48 00

gateway/src/apicast/policy/cors/apicast-policy.json
```diff
@@ -46,7 +46,7 @@
         }
       },
       "allow_origin": {
-        "description": "Origin allowed for CORS requests. The field expects only one origin (e.g. https://example.com) or '*'. If left blank, the value of the 'Origin' request header will be used. In case of valid regexp, and if it matches with Origin header value, the value will be set to the Origin Value. In case of does not match, the header will not set at all.",
+        "description": "Origin allowed for CORS requests. The field expects only one origin (e.g. <code>https://example.com</code>) or '*'. If left blank, the value of the 'Origin' request header will be used. <br> In order to allow more than one origin it is possible to use a regular expression, if it matches with Origin header value, the value will be set to the Origin Value. In case it does not match, the header will not set at all. Example: <code>(api|web).test.com</code> wil match both <code>api.test.com</code> and <code>web.test.com</code>",
```
damianpm, line 49: It seems like GH picked the first suggestion instead of the second (corrected without html markup) :(

eloycoto: Fixed now! Thanks!

damianpm (approved):
LGTM

davidor (approved):

PR #1253, opened 2021-01-18
adds method matching check to rewrite url captures policy

kevprice83:
Adds a list of methods to the transformation. Checks the request method against the method configured in the policy just as it is done in the URL rewrite policy.

Fixes: THREESCALE-6270

### DIFF
--- a/gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
+++ b/gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
@@ -114,11 +114,11 @@ end
 -- Returns true if no Method is provided in the config for backwardscompatibility
 local function is_match_methods(methods)
 
-  local request_method = ngx.req.get_method()
-
   if methods == nil or next(methods) == nil  then
     return true
   end
+
+  local request_method = ngx.req.get_method()
 
   for _,v in pairs(methods) do
     if v == request_method then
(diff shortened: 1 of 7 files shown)


gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
```diff
@@ -99,16 +99,35 @@ end
...
+  self.methods = methods
 
   return self
 end
 
+-- Returns true if the Method of the request is in the methods of the command meaning the rewrite rule should be applied
+-- Returns true if no Method is provided in the config for backwardscompatibility
+local function is_match_methods(methods)
+
+  local request_method = ngx.req.get_method()
```
eloycoto, line 117: I would move this line after the condition, but it's ok if you want to keep it.

t/apicast-policy-rewrite-url-captures.t
```diff
@@ -233,3 +233,51 @@ yay, api backend
 --- error_code: 200
 --- no_error_log
 [error]
+
+=== TEST 5: one transformation with method that doesn't match
```
eloycoto, line 237: We should add two test that match at least one method:

Test6: One single method works. 
Test7: Array with methods [POST, UPDATE, GET] will match if GET method is in there. (Want to check the next method of the array is working)

eloycoto (changes requested):

gateway/src/apicast/policy/rewrite_url_captures/named_args_matcher.lua
kevprice83: ah yes makes sense, not using it in the conditional statement anyway, thanks

eloycoto (approved):

PR #1254, opened 2021-01-26
THREESCALE-6552 (RFE)

samugi:
Adds the ability to select a list of upstrams (backends) to apply maintenance mode to + key/value liquid enabled operations.

Updates a variable in the liquid context when the Upstream Policy changes the Upstream.


samugi:
included too many commits :P

PR #1255, opened 2021-01-26
THREESCALE-6552 - RFE

samugi:
- Adds the ability to select a list of upstrams (backends) to apply maintenance mode to + key/value liquid enabled operations.
- Updates a variable in the liquid context when the Upstream Policy changes the Upstream.

### DIFF
--- a/gateway/src/apicast/policy/maintenance_mode/maintenance_mode.lua
+++ b/gateway/src/apicast/policy/maintenance_mode/maintenance_mode.lua
@@ -1,4 +1,4 @@
--- This is a simple policy. It allows you reject incoming requests with a
+-- This policy allows to reject incoming requests with a
 -- specified status code and message.
 -- It's useful for maintenance periods or to temporarily block an API
 
@@ -10,6 +10,9 @@ local new = _M.new
 local default_status_code = 503
 local default_message = "Service Unavailable - Maintenance"
 local default_message_content_type = "text/plain; charset=utf-8"
+
+local Condition = require('apicast.conditions.condition')
+local Operation = require('apicast.conditions.operation')
 
 function _M.new(configuration)
   local policy = new(configuration)
@@ -24,15 +27,47 @@ function _M.new(configuration)
     policy.message_content_type = configuration.message_content_type or policy.message_content_type
   end
 
+  policy:load_condition(configuration)
   return policy
 end
 
-function _M:rewrite()
+function _M:load_condition(config)
+  if not config or not config.condition then
+    return
+  end
+
+  local operations = {}
+  for _, operation in ipairs(config.condition.operations or {}) do
+    table.insert( operations,
+      Operation.new(
+        operation.left,
+        operation.left_type or default_template_type,
+        operation.op,
+        operation.right,
+        operation.right_type or default_template_type))
+  end
+  self.condition = Condition.new( operations, config.condition.combine_op or default_combine_op)
+end
+
+function set_maintenance_mode(self)
   ngx.header['Content-Type'] = self.message_content_type
   ngx.status = self.status_code
   ngx.say(self.message)
 
   return ngx.exit(ngx.status)
+end
+
+function _M:access(context)
+  -- If no condition was configured, maintenance mode will be enabled by default
+  if self.condition == nil then
+    return set_maintenance_mode(self)
+  end
+
+  local upstream = context.route_upstream or context:get_upstream() or {}
+  context.upstream = upstream.uri or upstream
+  if self.condition:evaluate(context) then
+    return set_maintenance_mode(self)
+  end
 end
 
 return _M
--- a/gateway/src/apicast/policy/upstream/upstream.lua
+++ b/gateway/src/apicast/policy/upstream/upstream.lua
@@ -58,6 +58,10 @@ function _M:rewrite(context)
   end
 end
 
+function _M:access(context)
+  context.route_upstream = context[self] or context.route_upstream
+end
+
 function _M:content(context)
   local upstream = context[self]
 
--- a/spec/policy/maintenance_mode/maintenance_mode_spec.lua
+++ b/spec/policy/maintenance_mode/maintenance_mode_spec.lua
@@ -1,30 +1,41 @@
 local MaintenancePolicy = require('apicast.policy.maintenance_mode')
+local ngx_variable = require('apicast.policy.ngx_variable')
+   
 
 describe('Maintenance mode policy', function()
-  describe('.rewrite', function()
+
+  local test_host = 'backend.example.org'
+  local test_host2 = 'backend2.example.org'
+  local test_path = '/foo/bar/'
+  local test_path2 = '/bar/foo/'
+
+  describe('.access', function()
     before_each(function()
+      local test_upstream = { uri = { scheme = 'https', host = test_host, port = '443', path = test_path } }
+      ctx = { get_upstream = function() return test_upstream  end }
       stub(ngx, 'say')
       stub(ngx, 'exit')
       ngx.header = {}
+      stub(ngx_variable, 'available_context', function(context) return context end)
     end)
 
     context('when using the defaults', function()
       local maintenance_policy = MaintenancePolicy.new()
 
       it('returns 503', function()
-        maintenance_policy:rewrite()
+        maintenance_policy:access(ctx)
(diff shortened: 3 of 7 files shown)


gateway/src/apicast/policy/upstream/upstream.lua
```diff
@@ -58,6 +58,10 @@ function _M:rewrite(context)
   end
 end
 
+function _M:access(context)
+  context.route_upstream = context[self] or context.route_upstream
```
samugi, line 62: is this ok?

gateway/src/apicast/policy/maintenance_mode/maintenance_mode.lua
```diff
@@ -19,20 +26,100 @@ function _M.new(configuration)
...
+  if not port2 then
+    port2 = (scheme2 == "https" and 443 or 80)
+  end
+
+  local path1 = uri1.path or ""
+  path1 = string.gsub(path1, "/$", "")
+  local path2 = uri2.path or ""
+  path2 = string.gsub(path2, "/$", "")
+
+  return scheme1 == scheme2 and uri1.host == uri2.host and port1 == port2 and path1 == path2
```
palmieric, line 80: Why are you checking the path too?

samugi: @palmieric fixed it, I was supposed to "match" the paths instead of using ==. Thanks!

gateway/src/apicast/policy/upstream/upstream.lua
kevprice83: so basically you're exposing the upstream URL defined in the the proxy_config during the access phase here right? Do we do this anywhere else? One quick test to check this doesn't break anything is to run the entire test suite locally.

spec/policy/maintenance_mode/maintenance_mode_spec.lua
```diff
@@ -1,8 +1,15 @@
 local MaintenancePolicy = require('apicast.policy.maintenance_mode')
+local ngx_variable = require('apicast.policy.ngx_variable')
 
 describe('Maintenance mode policy', function()
-  describe('.rewrite', function()
+
+  describe('.access', function()
     before_each(function()
+      ctx = { service = { api_backend = 'http://backend2.example.org:80' } }
+      ctx2 = {
```
kevprice83, line 9: does this mean we have 2 different instances of the context object during runtime? that could be problematic no as from anywhere we only ever reference it as `context` or `ctx`

samugi: my intention here was to use a different context (ctx2) just for some of the tests. I guess I can also achieve this by modifying the same ctx object from within the test where I need a different value.

samugi: fixed this as per our discussion, I also added a couple more tests.

gateway/src/apicast/policy/upstream/upstream.lua
samugi: yeah we are doing it [here in the routing policy](https://github.com/3scale/APIcast/blob/2ae88d8581288c9782d6124d072a7bbd1e75e55c/gateway/src/apicast/policy/routing/routing.lua#L56). I thought it would be nice to have a default variable in the context where, should a policy modify the upstream, it would store the new value, so I used the same variable here. This way we would know where to look for an updated upstream, regardless of which policy changed it. The reason I ask is I am not sure this was intended to be used this way and maybe I should use a dedicated variable for the Upstream policy instead.
According to the tests it doesn't break anything at the moment :P

gateway/src/apicast/policy/maintenance_mode/maintenance_mode.lua
```diff
@@ -19,20 +22,54 @@ function _M.new(configuration)
   policy.message_content_type = default_message_content_type
 
   if configuration then
+    policy.maintenance_upstreams = configuration.upstreams
```
eloycoto, line 25: This is no longer needed no?

gateway/src/apicast/policy/maintenance_mode/maintenance_mode.lua
```diff
@@ -19,20 +22,54 @@ function _M.new(configuration)
...
+function set_maintenance_mode(self)
   ngx.header['Content-Type'] = self.message_content_type
   ngx.status = self.status_code
   ngx.say(self.message)
 
   return ngx.exit(ngx.status)
 end
 
+function _M:access(context)
+  -- Before the condition was not set, so the maintenance mode when condition is
```
eloycoto, line 62: In this past, this policy does not have condition. So by default, if condition is not set it set the maintenance mode.

gateway/src/apicast/policy/maintenance_mode/maintenance_mode.lua
```diff
@@ -19,20 +22,54 @@ function _M.new(configuration)
...
+  -- nil should work.
+  if self.condition == nil then
+    return set_maintenance_mode(self)
+  end
+
+  context.upstream = context.route_upstream or context:get_upstream() or {}
+  context.upstream = context.upstream.uri or context.upstream
```
eloycoto, line 69: This is incorrect, no?

spec/policy/maintenance_mode/maintenance_mode_spec.lua
```diff
@@ -67,10 +78,109 @@ describe('Maintenance mode policy', function()
...
+        maintenance_policy:access(ctx)
 
         assert.equals('application/json', ngx.header['Content-Type'])
       end)
     end)
+
+    context('when Maintenance Mode is configured for a specific upstream', function()  
+      local maintenance_policy = MaintenancePolicy.new({
+        condition = {
+          operations={{op="==", left="{{ upstream.scheme }}://{{ upstream.host }}{{ upstream.port }}{{ upstream.path }}", left_type="liquid", right="https://"..test_host.."443"..test_path, right_type="plain"}},
```
eloycoto, line 90: I would indent this a bit, in all testsuite TBH.

spec/policy/maintenance_mode/maintenance_mode_spec.lua
```diff
@@ -67,10 +78,109 @@ describe('Maintenance mode policy', function()
...
+          combine_op="and"
+        }
+      })
+      it('returns 503', function()
+        maintenance_policy:access(ctx)
+        assert.stub(ngx.exit).was_called_with(503)
+      end)
+    end)
+
+    context('when Maintenance Mode is configured for a specific upstream (different host)', function()
```
eloycoto, line 100: This can be all in the it expression TBH, and the same for all others

t/apicast-policy-maintenance-mode.t
```diff
@@ -173,11 +177,235 @@ Service Unavailable - Maintenance
...
 --- response_body
 { "msg": "Be back soon" }
 --- response_headers
 Content-Type: application/json
 --- error_code: 503
 --- no_error_log
 [error]
+
+
+=== TEST 5: Maintenance mode is applied with routing policy + matching upstream condition
```
eloycoto, line 190: I'm missing here the following test:

-> Works with api-backend, I think that it's not longer used by system, but just in case. 
-> A OR condition, where the first is not matching, but matches the second one. It's covered by unit, but I prefer to have here too.

Regards

eloycoto (changes requested):
Now only nitpicks! I'm happy to see this PR :-)

gateway/src/apicast/policy/maintenance_mode/maintenance_mode.lua
eloycoto:
```
local upstream = context.route_upstream or context:get_upstream() or {}
context.upstream = upstream.uri or upstream
```

This is a bit better

gateway/src/apicast/policy/maintenance_mode/maintenance_mode.lua
samugi: yup, forgot it :)

eloycoto:
Add a Changelog and it'll be merged :-)

samugi:
> Add a Changelog and it'll be merged :-)

all done!

eloycoto (approved):

PR #1258, opened 2021-02-04
Add /etc/hosts support (e.g. for k8s hostAliases)

DracoBlue:
WHY:

If you are trying to set alternative resolutions for ip's via /etc/hosts, they are ignored because the resolver.lua does not respect the /etc/hosts file. 

Thus if you are using k8s hostAliases annotation to rewrite an external domain to an internal ip, it will not resolve those.

WHAT:

This patch adds a hosts_map generated from /etc/hosts at startup and will answer the dns lookup's to those directly without querying the nameservers at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added sendfile_max_chunk to the worker [PR #1250](https://github.com/3scale/APIcast/pull/1250) [THREESCALE-6570](https://issues.redhat.com/browse/THREESCALE-6570)
 - Increased api-keys shared memory size [PR #1250](https://github.com/3scale/APIcast/pull/1250) [THREESCALE-6570](https://issues.redhat.com/browse/THREESCALE-6570)
 - Add support to multiple Origin based on regexp [PR #1251](https://github.com/3scale/APIcast/pull/1251) [THREESCALE-6569](https://issues.redhat.com/browse/THREESCALE-6569)
+- Added Support for /etc/hosts [PR #1258](https://github.com/3scale/APIcast/pull/1258)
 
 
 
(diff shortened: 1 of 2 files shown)


eloycoto:
Hi, 

I've just do a quick review on this, and I'm a bit worried about: 

1)  /etc/hosts can have multiple address and things can be complicated, test are going to be needed.  
2) Security concerns, I do not want to read that just in case. 

What are you trying to do? I mean, why don't use a service in Kubernetes and use the FQDN for that service? Could you share a bit your use case?

Regards

DracoBlue:
> /etc/hosts can have multiple address and things can be complicated

I will read up on how this must be handled

> test are going to be needed.

I am happy to provide them for those cases.

> Security concerns, I do not want to read that just in case.

Do you want to disable it. So we can put it behind an environment variable flag?

> What happens when IPv6 entries, etc..

As far as I understand the source code, you won't use ipv6 ip's if they are given and always fall back to dns search. I would thus skip ipv6 from etc hosts, if this is requested behavior,.

DracoBlue:
> What are you trying to do?

* The api cast for this project can run either on the same k8s cluster or externally.
* As of my knowdledge I can only set one api backend in 3scale, so I choose the external one (e.g. api-production.example.org)
* So api casts outside of the k8s cluster work
* But api casts inside of the k8s cluster won't work as they will go to the internet and back to answer the requests
* With hostAliases (so /etc/hosts) I can route the traffic from the api cast directly to it's clusterIP internally)

I hope this gives some context. If we really don't want to make /etc/hosts and hostAliases possible - it IMHO must be part of the documentation of the Dockerimage that they will not be respected. it took me some time to figure that this did not have an effect :).

eloycoto:
Hi, 

After speaking offline, we're not going to merge this PR. HostAliases is something that in APICast we're not sure about.

The workaround for this can be the following:

1) Setting a custom DNS server with an alias to where it should be pointed to using the following:
https://kubernetes.io/docs/reference/kubernetes-api/workloads-resources/pod-v1/#hostname-and-name-resolution 
2) Setting a service with the hostname that you want to use: 
https://kubernetes.io/docs/concepts/services-networking/service/#externalname

These are the workaround that should be used. 
Regards!

DracoBlue:
Thanks for the chat and clarification! Have a nice day!

PR #1259, opened 2021-02-12
Caching: Fix race

eloycoto:
When caching is disabled with none, the cache key can be in there, and a
race condition can be happend with this.

To test this:

1) Strict caching mode
2) Made a request, 200 OK cache.set("$KEY", 200)
3) Update config to none
4) apicast/proxy:authorize -> cache.get($KEY) is 200
5) request go to the API
6) Authrep, cache for $KEY is deleted.

Next request will return 403.

Fixed behaviour:

1) Strict caching mode
2) Made a request, 200 OK cache.set("$KEY", 200)
3) Update config to none
4) apicast/proxy:authorize -> cache.get($KEY) is 200
5) cache_is_disabled = true -> execute backend call.
5) request go to the API, and fails
6) Authrep, cache for $KEY is deleted.

Fix THREESCALE-4464

### DIFF
--- a/gateway/src/apicast/policy/caching/caching.lua
+++ b/gateway/src/apicast/policy/caching/caching.lua
@@ -101,17 +101,26 @@ local function handler(config)
   return res
 end
 
+local function is_disabled(config)
+  return config.caching_type and config.caching_type == "none"
+end
+
 --- Initialize a Caching policy.
 -- @tparam[opt] table config
 -- @field caching_type Caching type (strict, resilient, allow, none)
 function _M.new(config)
   local self = new(config)
   self.cache_handler = handler(config or {})
+  self.is_disabled = is_disabled(config or {})
   return self
 end
 
+
 function _M:export()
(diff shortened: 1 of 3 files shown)


gateway/src/apicast/policy/caching/caching.lua
```diff
@@ -101,17 +101,29 @@ local function handler(config)
   return res
 end
 
+local function is_disabled(config)
+  if config.caching_type and config.caching_type == "none" then
```
davidor, line 105: This can be simplified:
```lua
return config.caching_type and config.caching_type == "none"
```

unleashed (approved):
This looks ok if the expectation is that caching will be enabled by default and only disabled when explicitly set to none. Is this assumption correct? Other than that, LGTM.

eloycoto:
> This looks ok if the expectation is that caching will be enabled by default and only disabled when explicitly set to none. Is this assumption correct? Other than that, LGTM.

YUP

PR #1262, opened 2021-02-15
fix camel proxy policy config example, minor copy edits

smccarthy-ie:
`https_proxy` and `http_proxy` config examples should use `http` and not `https`.


eloycoto (approved):

eloycoto:
Thanks, will merge!

PR #1263, opened 2021-02-17
THREESCALE-6723 fix hostname_rewrite incompatible with routing policy

samugi:
This makes hostname_rewrite work even when the routing policy is in the policy chain.


eloycoto (commented):
The nitpick and also the Changelog. Awesome work!

samugi:
> The nitpick and also the Changelog. Awesome work!

all done

eloycoto (approved):

PR #1264, opened 2021-03-04
http-ng: fix HOST header

eloycoto:
This strips the default port and use the correct host header value when
using http-ng

```
URL                          Host Header
--------------------------------------------------------------
http://foo.com               foo.com
http://foo.com:80            foo.com
http://foo.com:80/test       foo.com
http://foo.com:8080/test     foo.com:8080
https://foo.com/             foo.com
https://foo.com:8043/        foo.com:8043
https://foo.com:8043/test    foo.com:8043
```

Fix THREESCALE-2235

### DIFF
--- a/gateway/src/resty/http_ng/request.lua
+++ b/gateway/src/resty/http_ng/request.lua
@@ -2,6 +2,9 @@ local find = string.find
 local sub = string.sub
 local assert = assert
 local setmetatable = setmetatable
+local resty_url = require 'resty.url'
+local format = string.format
+
 
 ------------
 -- @module middleware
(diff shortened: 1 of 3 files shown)


gateway/src/resty/http_ng/request.lua
```diff
@@ -20,13 +23,20 @@ local request = { }
...
 
-local function extract_host(url)
-  local _, last = find(url, '://', 0, true)
-  local len = find(url, '/', last + 1, true)
+local function host_header(uri)
+    local port = uri.port
+    local default_port = resty_url.default_port(uri.scheme)
 
-  if len then len = len - 1 end
+    if port and port ~= default_port then
```
davidor, line 30: I've seen in the JIRA issue that this strips the port only if it's the default one to fix some issue when using OIDC. However, this is a generic client, doesn't it affect other use cases?

eloycoto: No, not a issue at all.

davidor (approved):

PR #1265, opened 2021-03-10
BPFTrace: Add initial trace scripts

eloycoto:
BPFtrace is a great tool to debug how APICast behave, this commit add
a few scripts that will help debugging weird behaviours.

### DIFF
--- a/docker-compose-devel-volmount-default.yml
+++ b/docker-compose-devel-volmount-default.yml
@@ -1,7 +1,10 @@
 version: '2.2'
 services:
   development:
+    privileged: true
+    pid: "host"
     volumes:
       - .:/home/centos/
       - .docker/cpanm:/home/centos/.cpanm
-
+      - /sys/kernel/debug:/sys/kernel/debug
+      - /lib/modules:/lib/module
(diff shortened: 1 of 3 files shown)


docker-compose-devel-volmount-default.yml
```diff
@@ -1,7 +1,10 @@
 version: '2.2'
 services:
   development:
+    privileged: true
+    pid: "host"
     volumes:
       - .:/home/centos/
       - .docker/cpanm:/home/centos/.cpanm
-
+      - /sys/kernel/debug:/sys/kernel/debug
```
davidor, line 9: Why are these 2 needed?

eloycoto: just to attach to the kernel, need to read that

davidor: I guess this is fine, but I'm not sure I understand all the implications of this. Should they be mounted as read only?

eloycoto: If it's ro, the main issue is that bpftrace cannot be installed because no space left :\  and I do not want to have bpftrace installed by default. 

Regards

davidor (approved):

PR #1267, opened 2021-04-15
Update APICAST_HTTPS_VERIFY_DEPTH description

pehala:
Depth 1 should mean `Root CA -> Certificate`, Depth 2 should be `Root CA -> Intermediate CA -> Certificate` and so on.

### DIFF
--- a/doc/parameters.md
+++ b/doc/parameters.md
@@ -408,7 +408,7 @@ Path to a file with the X.509 certificate secret key in the PEM format.
 **Values:** positive integers
 
 Defines the maximum length of the client certificate chain.
-If this parameter has 1 as its value, it implies that this length might include one additional certificate (eg. intermediate CA).
+If this parameter has `1` as its value, it is possible to include an additional certificate in the client certificate chain. For example, root certificate authority.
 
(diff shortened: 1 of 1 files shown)


doc/parameters.md
```diff
@@ -408,7 +408,7 @@ Path to a file with the X.509 certificate secret key in the PEM format.
 **Values:** positive integers
 
 Defines the maximum length of the client certificate chain.
-If this parameter has 1 as its value, it implies that this length might include one additional certificate (eg. intermediate CA).
+If this parameter has 1 as its value, it implies that this length might include one additional certificate (eg. root CA).
```
porueesq, line 411:
```suggestion
If this parameter has `1` as its value, it implies that this parameter includes an additional certificate. For example,  root CA.
```

porueesq (changes requested):
A suggestion

doc/parameters.md
pehala: Thanks, I changed it based on your suggestion.

porueesq (approved):
image

PR #1269, opened 2021-04-28
Policy: Add GC policy for shared dicts.

eloycoto:
This policy allows APICast to remove data that it's staled from shared
dictionaries.  By default, Openresty only deletes this when trying to
write a new entry in the shared dict.  When no new requests in this
instance, will keep allocated the memory forever.

The main reason for this change was a user using blue-green deployment,
where the proxy is not receiving more traffic. However, the shared
dictionaries are full, so the data is still allocated and cannot be used
by other processes.


eguzki:
Is this PR still valid? needs any further progress?

kevprice83:
@samugi @eguzki it looks like this PR is a _"nice to have"_ related to fixing [THREESCALE-6718](https://issues.redhat.com/browse/THREESCALE-6718), that issue has been fixed in 2.11 but we should definitely look into what further benefits this policy can have and whether we consider that useful or not to eventually merge this. Right now, it seems the problem is fixed according to the JIRA so we could probably close this PR and come back to it later.

PR #1270, opened 2021-05-06
Test: Changes to expected_json

eloycoto:
Due to the new changes that will be arrive in APICast, the integration
test cannot wait to the same order of the Json,so we need to use
expected_json handler to validate this.


eloycoto:
Merged on https://github.com/3scale/APIcast/pull/1272/files

PR #1272, opened 2021-05-18
Openresty Update

eloycoto:
Hi, 

I bumped the Openresty Version to 1.9.3, the latest stable, prior we had almost this version, but with a few commits on top of that. 

The most important change is related to how Luajit iterate over metatables, so the order is no longer the creation one; it's arbitrary, so the `ngx.decode`, `cjon.decode`, pairs can a different result each time. Saying that almost all test failed because we tested the string. 

Example: 

```
bash-4.4$ cat eloy/encode_args.lua
local oo = {
  a = 1,
  b= 1,
  c = 1,
  d= 1,
}
for i=1,5 do
  print(ngx.encode_args(oo))
end
bash-4.4$ resty eloy/encode_args.lua
a=1&b=1&c=1&d=1
a=1&b=1&c=1&d=1
a=1&b=1&c=1&d=1
a=1&b=1&c=1&d=1
a=1&b=1&c=1&d=1
bash-4.4$ resty eloy/encode_args.lua
c=1&a=1&d=1&b=1
c=1&a=1&d=1&b=1
c=1&a=1&d=1&b=1
c=1&a=1&d=1&b=1
c=1&a=1&d=1&b=1
bash-4.4$
```

Regarding other changes:
- APICAst Module: Is removed, no longer needed, and we don't want to keep in there. Doc team you should remove from docs. 
- Oath2 support: Also removed, Doc team you should remove it. 
- Small FFI calls need to be fixed.
- Some Integration and unit tests need to be fixed
- Now the context is shared on ssl_service, so the PATH routing when SSL was failing, because the path cannot be retrieved, we should return instead of setting the service. 
- Rate limit with Redis was failing because it used semaphores when it was not allowed; fixed that in the best way possible. 

Long PR, each commit is self-documented, so should be easy to get the context.


doc/parameters.md
```diff
@@ -134,13 +134,6 @@ Allows to set the log level for the logs related to OpenID Connect integration
 The [Management API](./management-api.md) is powerful and can control the APIcast configuration.
 You should enable the debug level only for debugging.
 
-### `APICAST_MODULE`
-
-**Default:** apicast  
-**Deprecated:** Use [policies](./policies.md) instead.
-
-Specifies the name of the main Lua module that implements the API gateway logic. Custom modules can override the functionality of the default `apicast.lua` module. See [an example](../examples/custom-module) of how to use modules.
-
```
dfennessy, line 143: @eloycoto Will there be further text to review? If there is only the removal of text, then it LGTM :smile:

eloycoto: My job is to make you happy, so no changes in doc :-)

gateway/src/resty/http_ng/headers.lua
```diff
@@ -54,7 +56,7 @@ local header_mt = {
   __tostring = function(t)
     local tmp = {}
 
-    for k,v in pairs(t) do
+    for k,v in tablex_sort(t) do
```
davidor, line 59: Why do headers need to be sorted? Is it to pass some tests or is there some part of the code that expects them sorted?

t/apicast-async-reporting.t
```diff
@@ -129,8 +129,13 @@ GET /test?user_key=foo
 --- response_body
 api response
 --- error_code: 200
---- error_log env
-backend client uri: https://127.0.0.1:$TEST_NGINX_RANDOM_PORT/transactions/authrep.xml?service_token=token-value&service_id=42&usage%5Bhits%5D=1&user_key=foo ok: true status: 200
+--- error_log env eval
+[
+  qr/backend client uri\: https\:\/\/127\.0\.0\.1:$TEST_NGINX_RANDOM_PORT\/transactions\/authrep.xml\?.*?(service_id=42).*? ok\: true status\: 200/,
```
davidor, line 134: Would it be better to check the params in a `location`? Similar to what we have in other tests modified in this PR: https://github.com/3scale/APIcast/pull/1272/commits/4779efe9f29544aa58ad858d38d0d93c418cbde8#diff-14b2ba18db23d9d85a7017bca7bf61564525518b3f6305633788656396c1757fR614

spec/backend_client_spec.lua
```diff
@@ -5,13 +5,35 @@ local http_ng_resty = require 'resty.http_ng.backend.resty'
...
+  local result = ""
+  local first = true
+
+  for k,v in tablex.sort(data) do
+    local prefix = "&"
+    if first then
+      prefix = ""
+      first = false
+    end
+    result = string.format("%s%s%s=%s",result, prefix, k, v)
```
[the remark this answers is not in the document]
eloycoto: concat has issues with sort.

gateway/src/resty/http_ng/headers.lua
eloycoto: Just to pass the test, is the only way to have consistent data.

gateway/src/resty/ctx.lua
```diff
@@ -37,7 +37,7 @@ function _M.ref()
 
   local _ = ngx.ctx -- load context
 
-  local ctx_ref = C.ngx_http_lua_ffi_get_ctx_ref(r)
+  local ctx_ref = C.ngx_http_lua_ffi_get_ctx_ref(r, ffi.new("int[1]"), ffi.new("int[1]"))
```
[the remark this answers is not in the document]
eloycoto: The C call definition changed, and now it need to pass a integer.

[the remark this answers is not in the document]
davidor: Why 1? Does it matter? Or we can just send anything?
Would be good to clarify with a comment.

spec/policy/find_service/find_service_spec.lua
```diff
@@ -130,20 +130,10 @@ describe('find_service', function()
...
-            end
-          end)
-          stub(PathBasedFinder, 'find_service')
           local find_service = FindService.new()
-
           find_service:ssl_certificate(context)
-
-          assert.stub(PathBasedFinder.find_service).was_not_called()
-          assert.equals(service, context.service)
+          assert.falsy(context.service)
```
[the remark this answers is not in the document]
eloycoto: Fixed.

spec/configuration_store_spec.lua
```diff
@@ -255,10 +255,18 @@ describe('Configuration Store', function()
 
       local service1 = { id = '42' }
       local service2 = { id = '43' }
+
       store:add(service1)
       store:add(service2)
 
-      assert.same({ service1, service2 }, store:all())
+      -- I'm crying, but I need to do this hack, table-sort return a iterator
```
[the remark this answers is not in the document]
eloycoto: Fixed.

t/apicast-async-reporting.t
```diff
@@ -56,6 +56,7 @@ yay, api backend: 127.0.0.1:$TEST_NGINX_SERVER_PORT
 
 === TEST 2: https api backend works
 with async background reporting
+--- ONLY
```
davidor, line 59: You forgot this

davidor (approved):

PR #1276, opened 2021-05-25
Keepalive: pool based on variable.

eloycoto:
When using keepalive connections, can be problematic when the
microservice is behind a LB.

When the upstream is shared accross services, if a connection is like:
APICast->LB->serviceA, the pool will be based on the IP. For the
serviceB, the request will land to the serviceA, because no SNI check at
all.

When using plaintext HTTP, most of the LB check the HOST header to
verify the upstream Service, but Host header is not in there when the LB
only do passthrough.

Fix THREESCALE-6849


gateway/http.d/upstream.conf
```diff
@@ -4,6 +4,7 @@ upstream upstream {
   balancer_by_lua_block { require('apicast.executor'):balancer() }
 
   keepalive 1024;
+  keepalive_pool $upstream_keepalive_key;
```
davidor, line 7: Where is this directive defined?

gateway/src/apicast/upstream.lua
```diff
@@ -196,6 +196,17 @@ end
 function _M:set_skip_https_connect_on_proxy()
   self.skip_https_connect = true
 end
+
+
+function _M:set_keepalive_key_if_https(context)
```
davidor, line 201: I think it's a big strange to find an "if_something" in a function name. I find it more natural to have the if outside the function.

gateway/src/apicast/upstream.lua
```diff
@@ -196,6 +196,17 @@ end
...
+  if self.uri.scheme == "https" or self.uri.scheme == "wss" then
+    local key = self.uri.host
+    local service_id = ((context or {}).service or {}).id or "" or ""
```
davidor, line 204: Redundant `or ""`.

davidor (approved):

PR #1277, opened 2021-05-26
Usage: fix issues on encoding 

eloycoto:
Usage keys are based on ngx.encode_args, BUT due to the Luajit update
the data is not always the same, so the following usage metatable:

```
{
  metric_a=1,
  metric_b=1,
}
```

can be `metric_a=1&metric_b=1` or `metric_b=1&metric_a=1`, so the cached
key is not the same, and auth call will happen a few more times when it
shouldn't.

The way that I followed is the same as Openresty, and the escape_uri
should always be a string due to the usage method.

Fix THREESCALE-7122

### DIFF
--- a/spec/usage_spec.lua
+++ b/spec/usage_spec.lua
@@ -169,10 +169,39 @@ describe('usage', function()
     end)
   end)
 
-  it("encode_format", function()
-    local usage = Usage.new()
-    usage:add('hits', 0)
-    assert.are.same(usage:encoded_format(), "usage%5Bhits%5D=0")
+  describe("encode format", function()
+    it("works with a single metric", function()
+      local usage = Usage.new()
+      usage:add('hits', 0)
+      assert.are.same(usage:encoded_format(), "usage%5Bhits%5D=0")
+    end)
+
+    it("with multiple metrics return the same", function()
+      local usage = Usage.new()
+      usage:add('a', 1)
+      usage:add('c', 3)
+      usage:add('b', 2)
+      assert.are.same(usage:encoded_format(), "usage%5Ba%5D=1&usage%5Bb%5D=2&usage%5Bc%5D=3")
+      -- Multiple cases just in case
+      local usage = Usage.new()
+      usage:add('c', 3)
+      usage:add('a', 1)
+      usage:add('b', 2)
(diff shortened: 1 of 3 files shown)


spec/usage_spec.lua
```diff
@@ -169,10 +169,38 @@ describe('usage', function()
...
+  describe("encode format", function()
+    it("works with a single metric", function()
+      local usage = Usage.new()
+      usage:add('hits', 0)
+      assert.are.same(usage:encoded_format(), "usage%5Bhits%5D=0")
+    end)
+
+    it("with multiple metrics return the same", function()
+      local usage = Usage.new()
+      usage:add('a', 0)
```
davidor, line 181: It'd be better to give different values to the metrics to test that they are assigned properly in the resulting string.

spec/usage_spec.lua
```diff
@@ -169,10 +169,38 @@ describe('usage', function()
...
+      assert.are.same(usage:encoded_format(), "usage%5Ba%5D=0&usage%5Bb%5D=0&usage%5Bc%5D=0")
+      -- Multiple cases just in case
+      local usage = Usage.new()
+      usage:add('c', 0)
+      usage:add('a', 0)
+      usage:add('b', 0)
+      assert.are.same(usage:encoded_format(), "usage%5Ba%5D=0&usage%5Bb%5D=0&usage%5Bc%5D=0")
+    end)
+
+    it("integer metrics", function()
```
davidor, line 193: Do we need to test this? I don't think this is supported :thinking:

eloycoto: I was thinking the same, but didn't harm to have just in case.

spec/usage_spec.lua
```diff
@@ -169,10 +169,39 @@ describe('usage', function()
...
+      usage:add('hits', 0)
+      assert.are.same(usage:encoded_format(), "usage%5Bhits%5D=0")
+    end)
+
+    it("with multiple metrics return the same", function()
+      local usage = Usage.new()
+      usage:add('a', 0)
+      usage:add('c', 3)
+      usage:add('b', 2)
+      assert.are.same(usage:encoded_format(), "usage%5Ba%5D=1&usage%5Bb%5D=2&usage%5Bc%5D=3")
```
davidor, line 184: Tests are failing because of this. It expects a = 1, but above it's set to 0.

davidor (approved):

PR #1278, opened 2021-06-04
Policy: Content Caching disable totally

eloycoto:
When using content_cache and proxy_no_cache, some directives are still
in place, and was causing issues to a few users regarding HEAD request,
etc..

This commit disabled the proxy_cache globally based on cache_zone
variable.

FIX THREESCALE-7016

### DIFF
--- a/t/apicast-policy-content-caching.t
+++ b/t/apicast-policy-content-caching.t
@@ -429,3 +429,107 @@ __DATA__
 ["X-Cache-Status: MISS", "X-Cache-Status: MISS"]
 --- no_error_log
 [error]
+
+
+
+=== TEST 8: HEAD request is enabled if no content-caching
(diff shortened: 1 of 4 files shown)


t/apicast-policy-content-caching.t
```diff
@@ -429,3 +429,103 @@ __DATA__
 ["X-Cache-Status: MISS", "X-Cache-Status: MISS"]
 --- no_error_log
 [error]
+
+
+
+=== TEST 8: HEAD request is enabled if no content-caching
```
davidor, line 435: Would be good to add a brief explanation here or in `TEST 9` that clarifies that this is tested because of https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_convert_head

davidor (approved):

PR #1279, opened 2021-06-09
Policy: Nginx header filter

eloycoto:
Nginx, by default, checks/validates some request headers. This policy
allows the user to skips these checks and sends them to the upstream
servers.

FIX THREESCALE-6704

### DIFF
--- a/gateway/src/apicast/policy/nginx_filters/Readme.md
+++ b/gateway/src/apicast/policy/nginx_filters/Readme.md
@@ -1,0 +1,45 @@
+# Nginx Filters policy
+
+## Description
+
+Nginx, by default, checks/validates some request headers. This policy allows the
+user to skips these checks and sends them to the upstream servers. 
+
+The primary use case is like If-Match headers, where Nginx filters by default,
+but some users want these headers on the upstream server because Nginx is not
+responsible for this header. 
+
+There is an option to delete it or to append it to the upstream server. 
+
+## Warning
+
+If one header is also managed by the multiple headers policy, this can cause
+conflicts.
+
+
+## Examples
+
+Send If-Match header to the upstream server, and avoid Nginx 412 response:
+
+```
+{ "name": "apicast.policy.nginx_filters",
+  "configuration": {
+    "headers": [
+      {"name": "If-Match", "append": true}
+    ]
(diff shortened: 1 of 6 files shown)


gateway/src/apicast/policy/nginx_filters/nginx_filters.lua
```diff
@@ -0,0 +1,46 @@
...
+  for key,val in pairs(self.headers) do
+      local original_header = headers[key]
+      if original_header then
+        ngx.req.clear_header(key)
+        if val then
+          context.nginx_filters_headers[key] = original_header
+        end
+      end
+  end
+  return
```
davidor, line 33: Is this redundant?

gateway/src/apicast/policy/nginx_filters/nginx_filters.lua
```diff
@@ -0,0 +1,46 @@
...
+
+function _M:rewrite(context)
+  if not self.validate then
+    return
+  end
+
+  local headers = get_headers()
+  context.nginx_filters_headers = {}
+
+  for key,val in pairs(self.headers) do
```
davidor, line 24: I think this would look better with something like `header_name, append`. Then in the `if` below you would had `if append` which would make this easier to understand, in my opinion.

gateway/src/apicast/policy/nginx_filters/nginx_filters.lua
```diff
@@ -0,0 +1,46 @@
+local _M  = require('apicast.policy').new('Logging Policy', 'builtin')
```
davidor, line 1: Wrong name

gateway/src/apicast/policy/nginx_filters/Readme.md
```diff
@@ -0,0 +1,40 @@
+# Nginx Filters policy
```
davidor, line 1: This policy could create conflicts with the headers one.
Should we specify exactly which headers should be configured with this policy (link to nginx docs) or can we list them?

davidor (approved):

eloycoto:
Adding skip-changelog because it's failing the check :facepalm:

PR #1280, opened 2021-06-16
Policy: MTLS with trusted store and verify

eloycoto:
When using upstream_mtls policy, the policy only sends the client certificates
    but does not verify the connection. Therefore, the policy name is not honouring
    what it did.

    A new x509_store will be created with this change, and it'll be used to verify
    the SSL_context. This change will use some FFI functions that are part of
    apicast-nginx-module[0]

    Fix THREESCALE-7099
    [0] https://github.com/3scale/apicast-nginx-module


gateway/src/apicast/policy/upstream_mtls/upstream_mtls.lua
```diff
@@ -82,6 +89,27 @@ local function read_certificate_key(value, value_type)
...
+local function read_ca_certificates(ca_certificates)
+  local valid = false
+  local store = X509_STORE.new()
+  for _,certificate in pairs(ca_certificates) do
+    local cert, err = X509.parse_pem_cert(certificate)
+    if cert then
+      valid = true
+      store:add_cert(cert)
+    else
+        ngx.log(ngx.INFO, "cannot load certificate, err: ", err)
```
davidor, line 101: Different indentation.

gateway/src/apicast/policy/upstream_mtls/upstream_mtls.lua
```diff
@@ -82,6 +89,27 @@ local function read_certificate_key(value, value_type)
...
+    end
+  end
+
+  if valid then
+    return store.store
+  end
+
+  store = nil
+  return
```
davidor, line 110: Unnecessary return

spec/policy/upstream_mtls/upstream_mtls_spec.lua
```diff
@@ -119,5 +129,82 @@ describe('Upstream MTLS policy', function()
...
+  describe("CA_Certificates", function()
+
+
+    before_each(function()
+      stub.new(upstream_mtls, 'set_ca_cert', function() return true end)
+    end)
+
+    it("ca_store is nil if no certificates", function()
+      local config = {
+        certificate = "XXXX",
```
davidor, line 141: You could extract this string to a variable accessible from all the tests and reuse it everywhere.

davidor (commented):
I just added minor comments. I'm not really familiar with this part of the codebase and can't help as much as I'd like, but it looks good to me :+1:

davidor (approved):

PR #1284, opened 2021-06-22
Logs: Fix host on access logs

eloycoto:
When using ngx.var.host the access_logs are not logging the target host.
With this change, by default we log the host that was hit in the
upstream.

Fix THREESCALE-6193


eloycoto:
@samugi you'll be happy with this one.

davidor (approved):

samugi:
> @samugi you'll be happy with this one.

yassss

PR #1289, opened 2021-07-19
CIRCLECI: push containers with latest tag when build master git branch

eloycoto:
When using master branch, the quay.io images are reference as master.
The operator team wants to have the name as latest, so all components
can have the same naming.

FIX https://issues.redhat.com/browse/THREESCALE-7251

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -146,7 +146,12 @@ jobs:
       - run: make test-runtime-image gateway-logs --keep-going
       - login-docker:
           command: |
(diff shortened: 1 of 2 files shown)


unleashed (commented):
The change looks good, but it is replacing the master tag with latest, so there will no longer be a master tag. Is that what we want? We could still create both tags and push them at once.

unleashed (approved):

eloycoto:
Yep, latest is the preferred option now :-)

unleashed:
Well, I was asking because IIUIC now we'll lack a `master` tag, so if anyone was following those they'll be stuck.

PR #1295, opened 2021-08-11
TEST: Fix Uptream MTLs policy integration test

eloycoto:
When using invalid host, the ngx_ssl_host was not used, so the verify
was working when it shouldn't.

This PR is part of THREESCALE-768

### DIFF
--- a/t/apicast-policy-upstream_mtls.t
+++ b/t/apicast-policy-upstream_mtls.t
@@ -385,7 +385,7 @@ $Test::Nginx::Util::ENDPOINT_SSL_PORT = Test::APIcast::get_random_port();
       "backend_authentication_type": "service_token",
       "backend_authentication_value": "token-value",
       "proxy": {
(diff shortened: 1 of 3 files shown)


t/apicast-policy-upstream_mtls.t
```diff
@@ -575,3 +575,72 @@ ssl_client_i_dn: CN=localhost,OU=APIcast,O=3scale
 --- error_code: 200
 --- no_error_log
 [error]
+
+
+=== TEST 8: MTLS policy with correct CA certificate, but invalid host
```
unleashed, line 580: Where are you specifying the invalid host here?

unleashed: Ok, can we add a comment to the test adding this information, ie. the certificates are set up for `localhost` but here we will use a different hostname?

unleashed (approved):
LGTM - but first let's look into documenting why the test is expected to fail, ie. because the TLS certs are set up to apply to a different hostname `localhost`.

PR #1296, opened 2021-08-12
FIX: routing policy catch-all condition

eloycoto:
When operations is empty was working as expected, but when it was nil,
raises an excpetion and routing policy was not working as expected.

FIX https://issues.redhat.com/browse/THREESCALE-6415

### DIFF
--- a/t/apicast-policy-routing.t
+++ b/t/apicast-policy-routing.t
@@ -2305,3 +2305,114 @@ yay, api backend
 --- error_code: 200
 --- no_error_log
 [error]
+
+
+=== TEST 34: Match all conditions working as expected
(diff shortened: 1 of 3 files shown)


t/apicast-policy-routing.t
```diff
@@ -2304,4 +2304,58 @@ GET /a_path
...
+                {
+                  "url": "http://test:$TEST_NGINX_SERVER_PORT/",
+                  "condition": {
+                    "operations": [
+                      {
+                        "match": "path",
+                        "op": "==",
+                        "value": "/a_path"
+                      }
+                    ]
```
unleashed, line 2332: shouldn't we add a test with `conditions.operations` equal to `nil`?

unleashed (approved):

PR #1300, opened 2021-08-20
Doc: Add liquid examples

eloycoto:

### DIFF
--- a/doc/liquid_conditions.md
+++ b/doc/liquid_conditions.md
@@ -1,0 +1,67 @@
+# Liquid conditions examples
+
+When using liquid, a lot of things can be achieved, and here are some examples
+of conditions that can be used across multiple policies
+
+
+## Headers
+
+### Header is not present
+
+Condition pass if header 'Backend' is sent and it's staging
+
+```
+"condition":{
+   "operations":[
+      {
+         "left":" {% if headers['Backend'] %}{{ headers['Backend'] }}{%else%}Notpresent{%endif%}",
+         "left_type":"liquid",
+         "op":"==",
+         "right":"staging"
+      }
+   ]
+}
+```
+
+If want to test when the header is not present, the condition will be like this:
+
+```
+"condition":{
+   "operations":[
+      {
+         "left":" {% if headers['Backend'] %}IsPresent{%else%}Notpresent{%endif%}",
+         "left_type":"liquid",
+         "op":"==",
+         "right":"Notpresent"
+      }
+   ]
+}
+```
+
+## Numeric operations
+
+These are harder, by default, all the values on liquid are strings, so we need
+to get any integer before do any operation. We can use `minus` or `plus`
+filters using 0, and do something like this:
+
+
+```
+"condition":{
+  "operations":[
+    {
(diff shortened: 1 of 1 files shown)


doc/liquid_conditions.md
```diff
@@ -0,0 +1,67 @@
+# Liquid conditions examples
+
+When using liquid, a lot of things can be achieved, and here are some examples
```
porueesq, line 3: You can use Liquid to configure conditions applicable across multiple policies to <insert_purpose>

Some questions to help you:
* What's the benefit of using Liquid to configure conditions across multiple policies?
* Are there any prerequisites to use these examples?
* Are there any limitations with these examples? For instance, can these examples be used in a production environment?

doc/liquid_conditions.md
```diff
@@ -0,0 +1,67 @@
...
+of conditions that can be used across multiple policies
+
+
+## Headers
+
+### Header is not present
+
+Condition pass if header 'Backend' is sent and it's staging
```
porueesq, line 11:
```suggestion
Condition passes if the  'Backend' header is sent and it's staging
```

Questions:
a. Who/What sends the  'Backend' header ?
b. What does `staging` refer to?

doc/liquid_conditions.md
```diff
@@ -0,0 +1,67 @@
...
+         "left":" {% if headers['Backend'] %}{{ headers['Backend'] }}{%else%}Notpresent{%endif%}",
+         "left_type":"liquid",
+         "op":"==",
+         "right":"staging"
+      }
+   ]
+}
+```
+
+If want to test when the header is not present, the condition will be like this:
```
porueesq, line 26:
```suggestion
If you want to test when the `Backend` header is not present, the condition is as follows:
```

doc/liquid_conditions.md
```diff
@@ -0,0 +1,67 @@
...
+         "op":"==",
+         "right":"Notpresent"
+      }
+   ]
+}
+```
+
+## Numeric operations
+
+These are harder, by default, all the values on liquid are strings, so we need
```
porueesq, line 43:
```suggestion
Numeric operations require additional steps. By default, all the values on liquid are strings, so you need
```

doc/liquid_conditions.md
```diff
@@ -0,0 +1,67 @@
...
+to get any integer before do any operation. We can use `minus` or `plus`
```
porueesq, line 44:
```suggestion
to get any integer before do any operation. Use `minus` or `plus`
```

doc/liquid_conditions.md
```diff
@@ -0,0 +1,67 @@
...
+filters using 0, and do something like this:
```
porueesq, line 45:
```suggestion
filters along with `0` (zero), and apply the following configuration:
```

doc/liquid_conditions.md
```diff
@@ -0,0 +1,67 @@
...
+    }
+ ]
+}
+```
+
+@TODO: something like a int filter should be added.
+@TODO: default filter should be added too.
+
+
+## Another examples:
```
porueesq, line 65:
```suggestion
## Additional examples:
```

porueesq (changes requested):
Some questions and suggestions

eguzki (approved):

PR #1306, opened 2021-09-01
OIDC: fix issue on remote v2.

eloycoto:
This is a second patch, and it's related to PR #1304.

When using remotev2 loader, the config.oidc is created in there, and
there is also a caching mechanism for OIDC[0], so we need to append the
service_id and send a valid table in case of non OIDC setup.

Because this oidc is cached, we also do a deepcopy, to avoid to use
always the same reference, so the config.oidc[i].service_id is always
different and we didn't get a invalid filtering.

This is only failing when using APICAST_SERVICE_LIST

[0] Commit: 720cd99f5dcbd2e11e3b3f07d5ee5888865740d7
Related to: THREESCALE-6042

### DIFF
--- a/spec/configuration_loader/remote_v2_spec.lua
+++ b/spec/configuration_loader/remote_v2_spec.lua
@@ -549,7 +549,10 @@ UwIDAQAB
       assert.truthy(config)
       assert.equals('string', type(config))
 
-      assert.equals(1, #(cjson.decode(config).services))
+      result_config = cjson.decode(config)
+      assert.equals(1, #result_config.services)
+      assert.equals(1, #result_config.oidc)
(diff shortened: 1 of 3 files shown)


spec/configuration_loader/remote_v2_spec.lua
```diff
@@ -562,5 +565,62 @@ UwIDAQAB
       assert.is_nil(config)
       assert.equals('Expected object key string but found invalid token at character 3', err)
     end)
+
+    it('treturns configuration with oidc config complete', function()
```
unleashed, line 569: typo `t`returns

unleashed (approved):

PR #1314, opened 2021-09-30
Release 3scale 2.11 stable 

eloycoto:


gsaslis:
I see this is just fast-forwarding stable to candidate, so I'll go ahead and 👍 this one, so we can handover to QE for CR1

gsaslis (approved):

PR #1319, opened 2021-12-21
THREESCALE-7906

djmarinb:
THREESCALE-7906
Under some error circumstances (like 414 Request-URI Too Large) the  access log were generated in a phase previous to having defined the access_log_enabled and the extended_access_logs_enabled.

This causes a warning to arise, but most importanty, prevented that acces to be logged.

We use the workarround of declaring the variables as a map with a default value because maps can be used in the http context and therefore will be available and initialized even on the given error case.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -146,7 +146,8 @@ jobs:
       - run: make test-runtime-image gateway-logs --keep-going
       - login-docker:
           command: |
-            IMAGE_TAG="${CIRCLE_TAG:-${CIRCLE_BRANCH}}"
+            CIRCLE_BRANCH_LOWER=`echo $CIRCLE_BRANCH | tr '[:upper:]' '[:lower:]'`
+            IMAGE_TAG="${CIRCLE_TAG:-${CIRCLE_BRANCH_LOWER}}"
             if [ "${IMAGE_TAG}" == "master" ]
             then
               IMAGE_TAG="latest"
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -146,7 +146,8 @@ jobs:
       - run: make test-runtime-image gateway-logs --keep-going
       - login-docker:
           command: |
-            IMAGE_TAG="${CIRCLE_TAG:-${CIRCLE_BRANCH}}"
+            CIRCLE_BRANCH_LOWER=`echo $CIRCLE_BRANCH | tr '[:upper:]' '[:lower:]'`
```
eloycoto, line 149: I do not think that these changes should be part of this PR, maybe a separate commit or separate PR. Why is needed?

eloycoto (commented):
Nginx part LGTM, I cannot understand the circleCI changes.

.circleci/config.yml
djmarinb: The s2i-runtime failed because it used the branch name as a tag for the docker image, but it fails with uppercase characters. This change is to transform them to lowercase

eloycoto: :+1:

djmarinb: I separated the circleci changes on a different commit

gateway/http.d/apicast.conf.liquid
```diff
@@ -4,6 +4,15 @@ map $status $extended_access_log {
     default '';
 }
 
+map $status $access_logs_enabled {
```
guicassolato, line 7: Maps works like switch..case selectors to set values to conditional variables in nginx, if I am not mistaken. If I understand correctly then, when `default` is provided alone, the value of the variable will be the default value (when even `default` is omitted, it will be the empty string, by the way).

So why using `$status`? Out of curiosity, would something like the following work instead?

```suggestion
map "" $access_logs_enabled {
```

guicassolato (approved):

unleashed:
Does this PR need any further work or review?

unleashed (approved):
Looks good to me!

PR #1320, opened 2022-01-12
Threescale 6594

djmarinb:
Enables the date filter on liquid templated


unleashed (approved):

unleashed:
We need to review the required tests in CI to discard those that have issues (ie. frequent timeouts) or those that make little sense to have in the required list.

PR #1322, opened 2022-01-26
Threescale 79063 scale 2.12 candidate

djmarinb:


gateway/src/apicast/version.lua
```diff
@@ -1 +1 @@
-return "3.6.0"
+return "3.11.0"
```
eloycoto, line 1: Should be 12, no?

PR #1323, opened 2022-02-04
[WIP] - Threescale 7967 http proxy patch

kevprice83:
I will leave this as marked WIP until we are able to safely merge to master again in the future. This PR is an example for the new, temporary patch process discussed with the 3scale program.

Fixes THREESCALE-7967

### DIFF
--- a/gateway/src/resty/http/proxy.lua
+++ b/gateway/src/resty/http/proxy.lua
@@ -72,8 +72,9 @@ local function _connect_proxy_https(httpc, request, host, port)
 end
 
 local function connect_proxy(httpc, request, skip_https_connect)
+    -- target server requires hostname not IP and DNS resolution is left to the proxy itself as specified in the RFC #7231
+    -- https://httpwg.org/specs/rfc7231.html#CONNECT for the CONNECT method
     local uri = request.uri
-    --local host, port = httpc:resolve(uri.host, uri.port, uri)
     local proxy_uri = request.proxy
 
     if proxy_uri.scheme ~= 'http' then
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,10 +1,10 @@
 # Change Log
+
```
samugi, line 2: were all these newlines and formatting changes in the README meant to be committed?

gateway/src/resty/http/proxy.lua
```diff
@@ -94,15 +95,17 @@ local function connect_proxy(httpc, request, skip_https_connect)
...
+
     if uri.scheme == 'http' then
         -- http proxy needs absolute URL as the request path
-        request.path = format('%s://%s:%s%s', uri.scheme, host, port, uri.path or '/')
+        request.path = format('%s://%s:%s%s', uri.scheme, uri.host, uri.port, uri.path or '/')
         return httpc
     elseif uri.scheme == 'https' and skip_https_connect then
-        request.path = format('%s://%s:%s%s', uri.scheme, host, port, request.path or '/')
-        return _connect_tls_direct(httpc, request, host, port)
+        request.path = format('%s://%s:%s%s', uri.scheme, uri.host, uri.port, request.path or '/')
```
eloycoto, line 105: I think that this will work, but please test this with routing policy, I remember that we had issues with that. Also, with Camel policy

kevprice83: I will check with the routing policy yep! However, the camel policy I might need to get some help with because I don't understand how that works and the tests for that specific policy are not passing even without my changes. It seems like something is broken there or maybe it's expected and we need to test in another way?

CHANGELOG.md
kevprice83: no I didn't make those changes, not sure where they came from to be honest. I only added a CHANGELOG entry, I assume you are referring to the CHANGELOG and not README though right?

samugi: yes I was :) - fixed it

kevprice83:
@samugi I've pushed the first iteration of a fix for the upstream http problem but there is a definite issue to take care of with args and so tests 13, 15 & 16 are partially failing as a result. Let's try to fix that last item on Tuesday.

PR #1324, opened 2022-02-10
1316: support resilient redis mode for cache util

### LINKED ISSUE #1316: Threescale Util fails the current request if redis connection is not initiated successfully
[Redis utility here](https://github.com/3scale/APIcast/blob/master/gateway/src/apicast/threescale_utils.lua#L153-L156) seems to not be designed with resiliency in mind, in case redis fails to connect for any reason, subsequent call to `_M.error()` will invoke `ngx.say` and `ngx.exit` and terminate the current request.

##### Version

V2.9

##### Steps To Reproduce

1. Import `redis` from `threescale_utils` module.
2. call `connect_redis`, e.g. local red = ts:connect_redis()`.
3. run APICast instance not connected to redis (like staging) or simulate redis disconnect.
4. Entire request fails.

##### Current Result
Entire request terminates.

##### Expected Result
Provide at least an key of `option` argument to be like `secondary` or `resilient`, and if passed `_M.error()` is not invoke if redis is not available and request gracefully continues.

abarrak:
This resolves #1316 and make it useful to use the cache without fail in case of network disconnect, etc.

Tests passing:
_(results included in commit message)_
Screen Shot 2022-02-10 at 12 07 17 PM

### DIFF
--- a/gateway/src/apicast/threescale_utils.lua
+++ b/gateway/src/apicast/threescale_utils.lua
@@ -151,29 +151,36 @@ function _M.connect_redis(options)
   red:set_timeout(opts.timeout)
 
   local ok, err = red:connect(_M.resolve(host, port))
-  if not ok and not resilient then
-    return nil, _M.error("failed to connect to redis on ", host, ":", port, ": ", err)
-  elseif not ok and resilient then
-    return nil, _M.error_gracefully("failed to connect to redis on ", host, ":", port, ": ", err)
+
+  if not ok then
+    if not resilient then
+      return nil, _M.error("failed to connect to redis on ", host, ":", port, ": ", err)
+    else
+      return nil, _M.error_gracefully("failed to connect to redis on ", host, ":", port, ": ", err)
+    end
   end
 
   if opts.password then
     ok = red:auth(opts.password)
 
-    if not ok and not resilient then
-      return nil, _M.error("failed to auth on redis ", host, ":", port)
-    elseif not ok and resilient then
-      return nil, _M.error_gracefully("failed to auth on redis ", host, ":", port)
+    if not ok then
+      if not resilient then
+        return nil, _M.error("failed to auth on redis ", host, ":", port)
+      else
+        return nil, _M.error_gracefully("failed to auth on redis ", host, ":", port)
+      end
     end
   end
 
   if opts.db then
     ok = red:select(opts.db)
 
-    if not ok and not resilient then
-      return nil, _M.error("failed to select db ", opts.db, " on redis ", host, ":", port)
-    elseif not ok and resilient then
-      return nil, _M.error_gracefully("failed to select db ", opts.db, " on redis ", host, ":", port)
+    if not ok then
+      if not resilient then
+        return nil, _M.error("failed to select db ", opts.db, " on redis ", host, ":", port)
+      else
+        return nil, _M.error_gracefully("failed to select db ", opts.db, " on redis ", host, ":", port)
+      end
     end
   end
 
--- a/gateway/src/apicast/threescale_utils.lua
+++ b/gateway/src/apicast/threescale_utils.lua
@@ -151,35 +151,29 @@ function _M.connect_redis(options)
   red:set_timeout(opts.timeout)
 
   local ok, err = red:connect(_M.resolve(host, port))
-  if not ok then
-    if not resilient then
-      return nil, _M.error("failed to connect to redis on ", host, ":", port, ": ", err)
-    else
-      return nil, _M.error_silently("failed to connect to redis on ", host, ":", port, ": ", err)
-    end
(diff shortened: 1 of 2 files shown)


unleashed (dismissed):

abarrak:
Thanks @eloycoto and @unleashed ..
Just covered both scenarios and made sure to keep backward compatibility as the default.

Tests run is passing:
```shell
$ make busted-util
EXTRA_CFLAGS="-DHAVE_EVP_KDF_CTX=1" /usr/local/openresty/luajit/bin/rover install --roverfile=gateway/Roverfile > /dev/null
/usr/local/openresty/luajit/bin/rover exec bin/busted "spec/threescale_utils_spec.lua" 
●●●
3 successes / 0 failures / 0 errors / 0 pending : 0.022514 seconds
```

abarrak:
Guys I thought again about current implementation and think it's better to keep `_M.error()` signature intact and introduce a new function (e.g. `_M.silent_error()`) using same signature `...` and without nginx exit call.

This way will make the redis connect use either based on the `resilient` flag and without impacting original function usage.
Shared chunk in the function will be reused without duplication.

What do you think @unleashed @eloycoto @kevprice83 ?

gateway/src/apicast/threescale_utils.lua
```diff
@@ -152,22 +152,34 @@ function _M.connect_redis(options)
 
   local ok, err = red:connect(_M.resolve(host, port))
   if not ok then
-    return nil, _M.error("failed to connect to redis on ", host, ":", port, ": ", err)
+    if not resilient then
```
kevprice83, line 155: @abarrak I think it's better to add this as another condition rather than a nested if statement. `if not ok and if not resilient then` because we only want to execute something when both return false or both return true. It's not super important and maybe @eloycoto might be able to comment whether the nested if statement could introduce any additional unwanted processing.

gateway/src/apicast/threescale_utils.lua
```diff
@@ -187,17 +199,25 @@ function _M.match_xml_element(xml, element, value)
   return string.find(xml, pattern, xml_header_len, xml_header_len, true)
 end
 
--- error and exit
-function _M.error(...)
+-- error without exit
+function _M.error_silently(...)
```
kevprice83, line 203: I would maybe rename this to `error_gracefully()` because it's not a silent error as we are going to log the failed connection to redis.

abarrak: Noted 📝

gateway/src/apicast/threescale_utils.lua
abarrak: And what would you suggest for the other case?
Does `elseif not ok then …` sound good enough?

kevprice83: I'd just suggest something like:

```
if not ok and not resilient then
  return nil, _M.error("failed to connect to redis on ", host, ":", port, ": ", err)
else
  return nil, _M.error_silently("failed to connect to redis on ", host, ":", port, ": ", err)
end
```

This might just be about style though so let¡s wait for a comment from someone more knowledgeable in Lua than myself because if it doesn't introduce unnecessary processing then we can leave it as you have already written it. I assume this is executed on every single request though right? That's why I was a little concerned, I don't see us opening  a connection with Redis and reusing that in this code, we open a new connection for every inbound request.

abarrak: Alright will wait for @eloycoto feedback.
I considered it in the first, but changed to the current structure, due the fact that original condition was just `if not ok` only, and using `else` alone here would catch the cases where response is `ok`.

Many thanks for the review comments so far 🌹 will make sure to implement it in upcoming commit.

eguzki:
@kevprice83 please review this PR and we decide to merge or close it

gateway/src/apicast/threescale_utils.lua
```diff
@@ -124,7 +124,7 @@ function _M.connect_redis(options)
   local opts = {}
 
   local url = options and options.url or env.get('REDIS_URL')
-
+  local resilient = options.resilient
```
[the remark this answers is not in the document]
abarrak: It should be set by calling sites to this utility as part of `option` table argument.
I pass it in a custom policy that rely on redis for some caching requirement.
This, defaults to null when it's nothing is set, hence preserving main usage and backward compatibility.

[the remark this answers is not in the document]
kevprice83: @abarrak could you document an example of how to pass this option? I think that would be really helpful for others who would also benefit from this feature.

abarrak:
Hi @eguzki @kevprice83, could you decide on this PR clousre?
It was originally approved but got new changes after review.

PR #1329, opened 2022-02-25
adds jwk alg matcher

kevprice83:
This PR adds a function to match the algorithm of the public key retrieved from the `jwks_uri` against the `jwt.header.alg` field. Currently we only match the JWT's `alg` against a whitelist which is populated based on the response from the discovery endpoint but this further check is required to ensure we do the verification correctly.

Fixes [THREESCALE-8249](https://issues.redhat.com/browse/THREESCALE-8249)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed issues with Upstream MTLS certs [PR #1307](https://github.com/3scale/APIcast/pull/1307) [THREESCALE-7508](https://issues.redhat.com/browse/THREESCALE-7508)
 - Fixed warning messages [PR #1318](https://github.com/3scale/APIcast/pull/1318) [THREESCALE-7906](https://issues.redhat.com/browse/THREESCALE-7906)
 - Fixed dirty context [PR #1328](https://github.com/3scale/APIcast/pull/1328) [THREESCALE-8000](https://issues.redhat.com/browse/THREESCALE-8000) [THREESCALE-8007](https://issues.redhat.com/browse/THREESCALE-8007)
+- Fixed jwk alg confusion [PR #1329](https://github.com/3scale/APIcast/pull/1329) [THREESCALE-8249](https://issues.redhat.com/browse/THREESCALE-8249)
 
 ### Added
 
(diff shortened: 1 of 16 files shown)


eloycoto (approved):
woah, good one.

gsaslis:
@eloycoto I am wondering if we can have this PR merged, but I see a couple of broken CI checks. 

On the one hand, I see the `prove` job has been failing ... almost consistently in - at least - a few other PRs I checked. As such, I am inclined to ignore it. 

On the other hand, I am not sure about the `busted` job failure? Should that failure block the merging of this PR, do you think?

Thanks in advance!

eloycoto:
Both should work, someone should have a look at that, both are important.

kevprice83:
@gsaslis before merging I need to write the integration & unit tests. At the moment it's not urgent as we are planning to included this in the 2.12 release but I have it as a top priority for next week!

gateway/.s2i/bin/assemble
```diff
@@ -3,7 +3,9 @@ set -euo pipefail
 
 mkdir -p /tmp/.s2i
 
-if [[ -n "${GIT_BRANCH:-}" && "${GIT_BRANCH:-}" != master && -z "${GIT_TAG:-}" ]]; then
+PATCH_BRANCH=THREESCALE
+
+if [[ -n "${GIT_BRANCH:-}" && "${GIT_BRANCH:-}" != master && "${GIT_BRANCH:-}" != *"${PATCH_BRANCH}"* ]]; then
```
samugi, line 8: I believe this was maybe included by mistake in this PR?

samugi:
assuming the circleci failures are all due to external factors: lgtm

samugi (changes requested):
assuming the circleci failures are all due to external factors: lgtm

eguzki:
LGTM

I share the [comment](https://github.com/3scale/APIcast/pull/1329#discussion_r838632920) for the change in `gateway/.s2i/bin/assemble`

Other than that, is should be rebased on top of master. 

Great work

gateway/.s2i/bin/assemble
kevprice83: well it's not a big deal but yeah cleaner to remove it I guess

kevprice83: Done!

samugi (approved):
lgtm

PR #1335, opened 2022-04-27
Development image upgraded to ubi8

eguzki:
### what

* Remove dependency from openresty base image from https://github.com/3scale/s2i-openresty
* Added `Dockerfile.devel` with the devel image content
* Fixed  `spec/resty/openssl/x509_spec.lua` unittest. Reason:
  * invalid certificate error message changed due to upgrade on `openssl-libs` dep:
  * from `openssl-libs-1.1.1g-15.el8_3.x86_64` in `centos8` -> `openssl-libs-1.1.1k-6.el8_5.x86_64` in `ubi8`
* Development container had overlapping volume mounts, causing (in linux) to create files with the root owner. The overlapping volume mounts have been removed from docker compose config.

### verification steps

run development image

```
make development
```
Inside the container, download dependencies
```
make dependencies
```
Inside the container, run apicast
```
$ cat <<EOF >config.json
{
   "services": [
      {
         "proxy": {
             "hosts": ["one"],
             "proxy_rules": [],
             "api_backend": "https://echo-api.3scale.net",
             "policy_chain": []
         }
      }
   ]
}
EOF

$ APICAST_LOG_LEVEL=debug APICAST_WORKER=1 THREESCALE_CONFIG_FILE=config.json BACKEND_ENDPOINT_OVERRIDE=http://localhost:8081 ./bin/apicast
```
Send request from externally from the host
```
$ APICAST_IP=$(docker inspect apicast_build_0_development_1 | yq e -P '.[0].NetworkSettings.Networks.apicast_build_0_default.IPAddress' -)

$ curl -v http://${APICAST_IP}:8080
*   Trying 192.168.32.3:8080...
* TCP_NODELAY set
* Connected to 192.168.32.3 (192.168.32.3) port 8080 (#0)
> GET / HTTP/1.1
> Host: 192.168.32.3:8080
> User-Agent: curl/7.68.0
> Accept: */*
> 
< HTTP/1.1 404 Not Found
< Server: openresty
< Date: Wed, 27 Apr 2022 16:27:12 GMT
< Content-Type: text/plain
< Transfer-Encoding: chunked
< Connection: keep-alive
< 
* Connection #0 to host 192.168.32.3 left intact
```

### DIFF
--- a/Dockerfile.devel
+++ b/Dockerfile.devel
@@ -44,7 +44,7 @@ ENV PATH="./lua_modules/bin:/usr/local/openresty/luajit/bin/:${PATH}" \
     LUA_CPATH="./lua_modules/lib/lua/5.1/?.so;;" \
     LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/opt/app-root/lib"
 
-RUN yum install -y luarocks && \
+RUN yum install -y luarocks-${LUAROCKS_VERSION} && \
     luarocks install --server=http://luarocks.org/dev lua-rover && \
     rover -v && \
     yum -y remove luarocks && \
--- a/Dockerfile.devel
+++ b/Dockerfile.devel
@@ -1,0 +1,68 @@
+FROM registry.access.redhat.com/ubi8:8.5
+
+ARG OPENRESTY_RPM_VERSION="1.19.3"
+ARG LUAROCKS_VERSION="2.3.0"
+
+WORKDIR /tmp
+
+ENV APP_ROOT=/opt/app-root \
+    HOME=/opt/app-root/src \
+    PATH=/opt/app-root/src/bin:/opt/app-root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
+    PLATFORM="el8"
+
+RUN sed -i s/enabled=./enabled=0/g /etc/yum/pluginconf.d/subscription-manager.conf
+
+RUN yum upgrade -y 
+
+RUN dnf install -y 'dnf-command(config-manager)'
+
+RUN yum install -y \
+        gcc make git which curl expat-devel kernel-headers openssl-devel m4 \
+        libyaml libyaml-devel perl-local-lib perl-App-cpanminus
+
+# perl-Test-Nginx
+RUN cpanm https://cpan.metacpan.org/authors/id/A/AG/AGENT/Test-Nginx-0.29.tar.gz
+
+RUN yum config-manager --add-repo http://packages.dev.3sca.net/dev_packages_3sca_net.repo
+
+RUN yum install -y \
+        openresty-${OPENRESTY_RPM_VERSION} \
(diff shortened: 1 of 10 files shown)


Dockerfile.devel
```diff
@@ -0,0 +1,68 @@
...
+
+COPY site_config.lua /usr/share/lua/5.1/luarocks/site_config.lua
+COPY config-*.lua /usr/local/openresty/config-5.1.lua
+
+ENV PATH="./lua_modules/bin:/usr/local/openresty/luajit/bin/:${PATH}" \
+    LUA_PATH="./lua_modules/share/lua/5.1/?.lua;./lua_modules/share/lua/5.1/?/init.lua;/usr/lib64/lua/5.1/?.lua;/usr/share/lua/5.1/?.lua" \
+    LUA_CPATH="./lua_modules/lib/lua/5.1/?.so;;" \
+    LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/opt/app-root/lib"
+
+RUN yum install -y luarocks && \
```
kevprice83, line 47: Shouldn't this be `luarocks-${LUAROCKS_VERSION}` ?

docker-compose-devel.yml
```diff
@@ -1,28 +1,22 @@
...
       - redis
-    working_dir: /home/centos/
-    volumes:
-      - .docker/lua_modules:/home/centos/lua_modules
-      - .docker/local:/home/centos/local
-      - .docker/vendor/cache:/home/centos/vendor/cache
-      # no need to access those from docker
-      - /home/centos/.docker
-      - /home/centos/.git
+    working_dir: /opt/app-root/src/
```
kevprice83, line 8: This breaks existing envrionments because the permissions are not the same, either need to take care of this ourselves OR document that the project should just be recreated.

kevprice83 (changes requested):
A couple of changes I've mentioned but also another question I have is that if this PR is intended to replace the s2i process with a Dockerfile instead shoudln't we be removing all the s2i resources? Or are we still going to rely on those for downstream image builds and use Dockerfile only for upstream?

docker-compose-devel.yml
eguzki: I  think I get your point. 

If you run "make development", the image being loaded will be `quay.io/3scale/apicast-ci:openresty-1.19.3` and there should not be any permission issue. 

Only  if you run "make development IMAGE=mycustom_image", then you need to make sure that "my_custom_image" is ubi8 based and has been built with "make dev-build" target. Usually the devel image is not custom image. Only when we neeed to upgrade some dep existing in the devel image, we need to test the new devel image with a custom image before pushing a new release

kevprice83 (approved):
So after testing again and reproducing the permissions issues it seems like the best solution is to notify the developers (community?) that the new changes require either those dirs to be removed or simply to delete the project and reclone it.

Otherwise nothing else left to review here and the changes look good to me.

PR #1340, opened 2022-05-04
THREESCALE-7906 - initialization of $post_action_impact and $target_host variables as tables

samugi:
As a follow up to https://github.com/3scale/APIcast/pull/1319, 2 more variables require to be initialized with a default value to avoid generating "uninitialized variables" logs in case a large URI is passed in the request (that causes 414 Request-URI Too Large): $target_host and $post_action_impact

### DIFF
--- a/gateway/http.d/apicast.conf.liquid
+++ b/gateway/http.d/apicast.conf.liquid
@@ -14,6 +14,13 @@ map "" $extended_access_logs_enabled {
   default '0';
 }
 
+map "" $post_action_impact {
+  default '';
+}
+
+map "" $target_host {
+  default '$host';
+}
 
 log_format extended escape=none '$extended_access_log';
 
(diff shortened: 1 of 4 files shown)


gateway/http.d/apicast.conf.liquid
```diff
@@ -14,6 +14,13 @@ map "" $extended_access_logs_enabled {
   default '0';
 }
 
+map "" $post_action_impact {
+  default 0;
```
kevprice83, line 18: I think the value is defined as a string `'0'`

kevprice83 (approved):
Summary of fix as I understand this:

[THREESCALE-7906](https://issues.redhat.com/browse/THREESCALE-7906) reported 2 uninitialised variables being referenced in the access_logs so #1319 was fixing that **BUT** in fact once those were fixed another 2 variables were causing the same problem but were not caught because of a lack of testing for this log warning. Adding a test to catch this `warn` statement will make future occurrences much easier to detect.

Fix and proper tests look good to me. Nice work @samugi 💪

gateway/http.d/apicast.conf.liquid
samugi: I've set it to empty string to reflect [this](https://github.com/3scale/APIcast/blob/e17e421944bdffe97366e306123a83c2874720dd/gateway/conf.d/apicast.conf#L159)

gateway/http.d/apicast.conf.liquid
```diff
@@ -14,6 +14,13 @@ map "" $extended_access_logs_enabled {
   default '0';
 }
 
+map "" $post_action_impact {
```
eguzki, line 17: Shouldn't be removed from https://github.com/3scale/APIcast/blob/e17e421944bdffe97366e306123a83c2874720dd/gateway/conf.d/apicast.conf#L159 ?

eguzki (changes requested):

eguzki (approved):

gateway/http.d/apicast.conf.liquid
samugi: currently investigating this. Removing this line would make sense but it seems to require reviewing our non-blackbox tests (that don't use apicast.conf.liquid).

samugi: non blackbox tests conversion to Blackbox tracked here: https://github.com/3scale/APIcast/pull/1343

eguzki (approved):

PR #1342, opened 2022-05-16
THREESCALE-8426 - request path is stripped for proxied https requests

samugi:
This is a fix for THREESCALE-8426 where in some situations the client's request path is stripped away from the proxied upstream request.

### Steps to reproduce the issue:
- Backend with public path = /
- Upstream with https:// scheme and a path in its private URL, like: https://echo-api.3scale.net/some/path
- Https proxy (configured on APIcast for upstream connections)
- Send a request to APIcast with a certain path appended http://public-base-url/foo/bar
- Verify the request was actually sent to https://echo-api.3scale.net/some/path instead of the expected https://echo-api.3scale.net/some/path/foo/bar

### Problem
The problem is that we set the upstream request's path using `uri.path` (when available) [here](https://github.com/3scale/APIcast/blob/132a91cb1869a4a607db69bfd3bb5ad5f2a03053/gateway/src/apicast/http_proxy.lua#L85).

`uri.path`, is [upstream.uri](https://github.com/3scale/APIcast/blob/132a91cb1869a4a607db69bfd3bb5ad5f2a03053/gateway/src/apicast/http_proxy.lua#L161).path, where `upstream.uri` is [self.uri](https://github.com/3scale/APIcast/blob/132a91cb1869a4a607db69bfd3bb5ad5f2a03053/gateway/src/apicast/upstream.lua#L164) i.e. only the prefix (aka the upstream path that is hardcoded in the Private URL when the Backend is defined), which explains the problem as described above.

### Fix
Currently we rewrite `ngx.var.uri` to contain the exact path for the upstream request [here](https://github.com/3scale/APIcast/blob/132a91cb1869a4a607db69bfd3bb5ad5f2a03053/gateway/src/apicast/upstream.lua#L166) by appending the value of `ngx.var.uri` to the Backend prefix.
This means we should use `ngx.var.uri` as the path for the proxied request [here](https://github.com/3scale/APIcast/blob/132a91cb1869a4a607db69bfd3bb5ad5f2a03053/gateway/src/apicast/http_proxy.lua#L85), instead of `uri.path`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed dirty context (part 2 of PR #1328) when tls termination policy is in the policy chain [PR #1333](https://github.com/3scale/APIcast/pull/1333)
 - Fixed NGINX filters policy error [PR #1339](https://github.com/3scale/APIcast/pull/1339) [THREESCALE-7349](https://issues.redhat.com/browse/THREESCALE-7349)
 - Fix to avoid uninitialized variables when request URI is too large [PR #1340](https://github.com/3scale/APIcast/pull/1340) [THREESCALE-7906](https://issues.redhat.com/browse/THREESCALE-7906)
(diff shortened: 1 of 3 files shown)


eguzki (approved):

kevprice83 (approved):
Looks good, added a new test as well

PR #1344, opened 2022-05-19
Remove s2i

eguzki:
### what

Fixes: https://issues.redhat.com/browse/THREESCALE-8132

### Verification steps

Make sure s2i is not installed in the system

* build runtime image
```
make runtime-image
....
Successfully tagged apicast-runtime-image:latest
```

* test runtime image just built
```
make test-runtime-image IMAGE_NAME=apicast-runtime-image:latest
```

Depends on #1335


Dockerfile
```diff
@@ -0,0 +1,92 @@
+FROM registry.access.redhat.com/ubi8:8.5
+
+ARG OPENRESTY_RPM_VERSION="1.19.3"
+
+LABEL summary="3scale's API gateway (APIcast) is an OpenResty application which consists of two parts: Nginx configuration and Lua files." \
```
dfennessy, line 5:
```suggestion
LABEL summary="The 3scale API gateway (APIcast) is an OpenResty application, which consists of two parts: NGINX configuration and Lua files." \
```

Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+      description="APIcast is not a standalone API gateway therefore it needs connection to the 3scale API management platform. The container includes OpenResty and uses LuaRocks to install dependencies (rocks are installed in the application folder)." \
```
dfennessy, line 6:
```suggestion
      description="APIcast is not a standalone API gateway. It needs connection to the 3scale API management platform. The container includes OpenResty and uses LuaRocks to install dependencies. Install LuaRocks in the application folder." \
```

Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+      io.k8s.description="APIcast is not a standalone API gateway therefore it needs connection to the 3scale API management platform. The container includes OpenResty and uses LuaRocks to install dependencies (rocks are installed in the application folder)." \
```
dfennessy, line 7:
```suggestion
      io.k8s.description="APIcast is not a standalone API gateway. It needs connection to the 3scale API management platform. The container includes OpenResty and uses LuaRocks to install dependencies. Install LuaRocks in the application folder)." \
```

doc/build.md
```diff
@@ -20,20 +20,23 @@ Lua Dependencies are defined in [`apicast-VERSION.rockspec`](https://github.com/
 
 ## Components
 
-APIcast is using [source-to-image](https://github.com/openshift/source-to-image) to build the final Docker image.
-You'll need to have source-to-image installed and available on your system.
+APIcast is using [docker](https://www.docker.com/) to build the APIcast runtime image.
+You'll need to have docker installed and available on your system.
```
dfennessy, line 24:
```suggestion
You must have docker installed and available on your system.
```

doc/build.md
```diff
@@ -20,20 +20,23 @@ Lua Dependencies are defined in [`apicast-VERSION.rockspec`](https://github.com/
...
 
-The builder image used is [s2i-openresty](https://github.com/3scale/s2i-openresty).
+The development image used is based on [UBI](https://developers.redhat.com/products/rhel/ubi).
 It builds on heavy OpenShift base images.
-In the future, we would like to utilize s2i extended build and use a very minimal runtime image.
+In the future, we would like to leverage multistage dockerfiles and use a very minimal runtime image.
```
dfennessy, line 28:
```suggestion
In future releases, we will leverage multistage docker files and use a minimal runtime image.
```

doc/build.md
```diff
@@ -20,20 +20,23 @@ Lua Dependencies are defined in [`apicast-VERSION.rockspec`](https://github.com/
...
 
 ## Build process
 
-The build is defined in `Makefile`. The [`make build`](https://github.com/3scale/apicast/blob/bc8631fcf91fcab25cae84152e16536ce01d22be/Makefile#L31-L32) is meant for development and uses s2i incremental build.
-The [`make release`](https://github.com/3scale/apicast/blob/bc8631fcf91fcab25cae84152e16536ce01d22be/Makefile#L34-L35) is for the release build. 
+**runtime image**
 
-Both use the [s2i-openresty](https://github.com/3scale/s2i-openresty) builder image pushed to [`quay.io/3scale/s2i-openresty-centos`](https://quay.io/repository/3scale/s2i-openresty-centos7?tag=latest).
+The runtime image build is defined in the `Makefile`. The `make runtime-image` is meant for the upstream release build.
```
dfennessy, line 34:
```suggestion
The runtime image build is defined in the `Makefile`. The `make runtime-image` is for the upstream release build.
```

Dockerfile
eguzki: this is not a user guide, it is a description of the image.

Dockerfile
eguzki: this is not a user guide, it is a description of the image.

dfennessy (approved):
lgtm!

.circleci/config.yml
```diff
@@ -158,38 +143,17 @@ jobs:
       - cleanup-docker
 
   # Try to build s2i builder image and run smoke tests
```
kevprice83, line 145: Comment should read "Try to build development image and run smoke tests"

.circleci/config.yml
```diff
@@ -158,38 +143,17 @@ jobs:
...
-            if [ "${IMAGE_TAG}" == "master-builder" ]
-            then
-              IMAGE_TAG="latest-builder"
-            fi
-            IMAGE_NAME="apicast:${IMAGE_TAG}"
-            docker tag apicast-builder-image ${IMAGE_NAME}
-            make push
-
+            docker rmi --no-prune ${DEVEL_IMAGE} 2>/dev/null || :
+            make dev-build
```
kevprice83, line 156: I would expect us to be executing the make target which then runs those smoke tests no? Probably rename the make target `test-runtime-image` to `test-dev-image` and execute that. Would need to pass in the correct image which looks to be `quay.io/3scale/apicast-ci:openresty-1.19.3`. However, given that is a fixed tag rather than floating we might want to think about a better way to reference that otherwise we would need to update this circleci config every time the image updates??

Makefile
```diff
@@ -139,50 +132,41 @@ PROVE_PATTERN = "{t,examples}/**/*.t"
...
-dev: USER = root
-dev: builder-image apicast-source ## Run APIcast inside the container mounted to local volume
-	$(DOCKER_COMPOSE) run --user=$(USER) --service-ports --rm --entrypoint=bash $(SERVICE) -i
+gateway-logs: export IMAGE_NAME = does-not-matter
+gateway-logs:
+	$(DOCKER_COMPOSE) logs gateway
 
-test-builder-image: export IMAGE_NAME ?= apicast-test
-test-builder-image: clean-containers ## Smoke test the builder image. Pass any docker image in IMAGE_NAME parameter.
+test-runtime-image: export IMAGE_NAME ?= $(RUNTIME_IMAGE)
```
kevprice83, line 161: Do we only want to test the runtime image or also the development image? I guess only dev image makes sense but just wanted to check this with you first.

kevprice83 (changes requested):

.circleci/config.yml
eguzki: The thing is that the development image is a ubi8 + openresty. There is no apicast in there at all. So what to tests?

I have checked the former devel image repo, [s2i-openresty](https://github.com/3scale/s2i-openresty). There is a makefile target for tests. What are those test about? The summary is on the header of [test/run](https://github.com/3scale/s2i-openresty/blob/master/test/run) file:
```
# The 'run' performs a simple test that verifies the S2I image.
# The main focus here is to exercise the S2I scripts.
#
# For more information see the documentation:
# https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md
```
The devel image of s2i-openresty has scripts that are run in the s2i workflow at different stages. The tests are meant to test the scripts work as expected.

However, the current base image is nothing else than a set of packages installed, some env variables set and so on. So, even though we could come up with some tests, I do not think it is worth doing. wdyt?

Makefile
eguzki: Related to the previous [comment](https://github.com/3scale/APIcast/pull/1344#discussion_r883490073), I think that only the runtime image should be tests. It does not make sense to test the devel image, or it has low value doing it. wdyt?

.circleci/config.yml
eguzki: The development image does not have any apicast code. 

The development container (running the development image) has mounted your `$PWD` in `/opt/app-root/src`. Configured [here](https://github.com/3scale/APIcast/blob/remove-s2i/docker-compose-devel-volmount-default.yml). With this approach, you change the apicast lua code or whatever using your favorite IDE and the changes are immediately available in the development container.

samugi: alright that makes a lot of sense now thanks for clarifying @eguzki. Then I guess I would agree with your statement above: probably it's not necessary to test this image.

Makefile
```diff
@@ -2,8 +2,7 @@ MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 PROJECT_PATH := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
 .DEFAULT_GOAL := help
 DOCKER_COMPOSE = docker-compose
-DOCKER ?= $(shell which podman 2> /dev/null || which docker 2> /dev/null || echo "docker")
-S2I = script/s2i
+DOCKER ?= $(shell which docker 2> /dev/null || echo "docker")
```
samugi, line 5: This was done because Podman requires to have a running container in order to mount the volume from it (see https://github.com/3scale/APIcast/pull/1344/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R56). Should we consider leaving the `-source` container running (instead of [stopping it](https://github.com/3scale/APIcast/pull/1344/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R89)) to keep Podman support or would that potentially cause other side-effects?

eguzki: not sure that podman works, even leaving the container running.  

That said, the root issue comes from the fact that in the `docker-compose.yaml` service `prove` has `volumes_from` which requires a running(podman) or stopped (docker) container. With just a volume should be enough. But a container is required.

The service `prove`  in the   `docker-compose.yaml` is only used in makefile's target `prove-docker`. Which is an alternative way, not the main one in the development workflow, to run integration tests.  I plan to fix this and have `prove-docker` use a volume, instead of a volume in a (running or stopped) container. But this PR is already big enough and that deserves a new PR.

samugi: understood, thanks for clarifying

samugi (approved):
based on the conversations above, it looks good to me.

PR #1346, opened 2022-05-27
t/configuration-loading-lazy.t: convert to APIcast::Blackbox

eguzki:


t/configuration-loading-lazy.t
```diff
@@ -1,178 +1,130 @@
...
 $ENV{APICAST_CONFIGURATION_LOADER} = 'lazy';
 
-env_to_nginx(
-    'APICAST_CONFIGURATION_LOADER',
-    'TEST_NGINX_APICAST_PATH',
-    'THREESCALE_CONFIG_FILE',
-    'THREESCALE_PORTAL_ENDPOINT'
-);
-
+repeat_each(2);
```
eguzki, line 8: without `repeat_each` directive the test was doing two request. I do not know why, as I expected only one according to https://metacpan.org/pod/Test::Nginx::Socket#repeat_each. So I defined explicitly.

samugi: I think it comes from [here](https://github.com/3scale/Test-APIcast/blob/c3e059369795f76156cbc7fdd6076ad2fcc1a5c7/lib/Test/APIcast.pm#L91) and then [inherited by Blackbox](https://github.com/3scale/Test-APIcast/blob/c3e059369795f76156cbc7fdd6076ad2fcc1a5c7/lib/Test/APIcast/Blackbox.pm#L7).

t/configuration-loading-lazy.t
```diff
@@ -1,178 +1,130 @@
...
-    echo "";
-  }
---- request
-GET /t
+--- env eval
+(
+  'APICAST_CONFIGURATION_LOADER' => 'lazy',
+  'THREESCALE_PORTAL_ENDPOINT' => "http://test:$ENV{TEST_NGINX_SERVER_PORT}"
+)
+--- upstream env
```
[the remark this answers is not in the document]
eguzki: done

t/configuration-loading-lazy.t
eguzki: Then, for blackbox tests, we should be leveraging `$ENV{TEST_NGINX_REPEAT_EACH}` env var and the framework should call `repeat_each(M)` according to that.

eguzki: does not work :/
only `repeat_each`

samugi: I see we use repeat_each in multiple tests, no sure why the env var is not picked up but I'd say we could safely continue using the function

eguzki:
ready for review @samugi

PR #1343, opened 2022-05-17
THREESCALE-8453 tests conversion to APIcast::Blackbox

samugi:
This PR tracks the conversion of some test files from `Test::APIcast` to the `Test::APIcast::Blackbox` framework.

Below the list of test files involved:

- apicast-request-logs.t
- resty-ctx.t
- apicast-policy-rate-limit.t
- apicast.t
- configuration-loading-lazy.t
- configuration-loading-boot-remote.t
- apicast-caching.t
- configuration-loading-boot-without-config.t
- custom-config.t
- apicast-async-reporting.t
- apicast-policy-3scale-batcher.t
- backend.t
- resolver.t
- balancer.t
- apicast-upstream-balancer.t
- management.t
- apicast-policy-chains.t
- apicast-mapping-rules.t
- backend-cache-handler.t

Of the above, the following are blocking https://github.com/3scale/APIcast/pull/1340:

- [x] apicast-policy-rate-limit.t (sam)
- [x] apicast.t (sam) https://github.com/3scale/APIcast/pull/1347
- [x] configuration-loading-lazy.t  (eguzki) https://github.com/3scale/APIcast/pull/1346
- [x] apicast-caching.t (eguzki) https://github.com/3scale/APIcast/pull/1351
- [x] apicast-async-reporting.t (sam) https://github.com/3scale/APIcast/pull/1348
- [x] apicast-policy-3scale-batcher.t (sam) https://github.com/3scale/APIcast/pull/1349
- [x] apicast-policy-chains.t (sam) https://github.com/3scale/APIcast/pull/1350

PR #1347, opened 2022-05-30
t/apicast.t: convert to APIcast::Blackbox

samugi:
apicast tests converted to Blackbox framework as part of https://github.com/3scale/APIcast/pull/1343

### DIFF
--- a/t/apicast.t
+++ b/t/apicast.t
@@ -1,176 +1,154 @@
 use lib 't';
-use Test::APIcast 'no_plan';
+use Test::APIcast::Blackbox 'no_plan';
 
-BEGIN {
-    $ENV{APICAST_POLICY_LOAD_PATH} = 't/fixtures/policies';
-}
+$ENV{APICAST_ACCESS_LOG_FILE} = "$Test::Nginx::Util::ErrLogFile";
+$ENV{TEST_NGINX_HTML_DIR} ||= "$Test::Nginx::Util::ServRoot/html";
+$ENV{APICAST_POLICY_LOAD_PATH} = 't/fixtures/policies';
 
-env_to_nginx(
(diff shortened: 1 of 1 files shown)


t/apicast.t
```diff
@@ -727,92 +634,75 @@ limits exceeded!
...
       -- Notice that the user_key sent in the body does not appear here.
       local expected = "service_token=token-value&service_id=42&usage%5Bhits%5D=2"
       require('luassert').same(ngx.decode_args(expected), ngx.req.get_uri_args(0))
     }
   }
-
 --- request eval
 "POST /
-user_key=value-".( "1" x 1024)
+user_key=value-".( "1" x 16 x 1024)
```
samugi, line 666: It appears there is no way to set `client_body_buffer_size 1;` with Blackbox so I increased the body size to exceed the default buffer's size (16K).

t/apicast.t
```diff
@@ -453,152 +401,119 @@ MHcCAQEEIFCV3VwLEFKz9+yTR5vzonmLPYO/fUvZiMVU1Hb11nN8oAoGCCqGSM49
...
---- no_error_log
-[error]
 --- response_headers
 X-3scale-matched-rules: /
 X-3scale-usage: usage%5Bhits%5D=2
+--- no_error_log
+[error]
+
 
 === TEST 16: uses endpoint host as Host header
```
eguzki, line 490: this test is kind of duplicated from https://github.com/3scale/APIcast/pull/1348

If it is duplicated, makes sense to remove one of them?

eguzki (approved):
good job

t/apicast.t
samugi: I removed this one because the other test (async report) covers this one too (and then sends an additional request) so I think it makes more to only keep that one.

PR #1349, opened 2022-06-01
t/apicast-policy-3scale-batcher.t: convert to APIcast::Blackbox

samugi:


t/apicast-policy-3scale-batcher.t
```diff
@@ -407,176 +356,148 @@ The caching policy will cache the first result and return it while backend is
...
+          if k == 'transactions[0][app_key]' then
+            ngx.exit(500)
+          elseif k == 'transactions[0][usage][hits]' then
+            usage_match = v == '3'
+          elseif k == 'transactions[0][app_id]' then
+            app_id_match = v == 'appid'
+          end
+        end
+        ngx.shared.result = usage_match and app_id_match
+      }
```
samugi, line 482: this is executed when `/force_report_to_backend` is called, when the backend_client is instantiated, so it includes the usage (hits) from `3` calls now instead of 2, because the `/force_report_to_backend` endpoint is an upstream too so it increases the hits as well. It used to be `2` in the old test because that endpoint was accessed without going through APIcast.

samugi, line 482: this is executed when `/force_report_to_backend` is called, when the backend_client is instantiated, so it includes the usage (hits) from `3` calls now instead of 2, because the `/force_report_to_backend` endpoint is an upstream too so it increases the hits as well. It used to be `2` in the old test because that endpoint was accessed without going through APIcast.

eguzki (approved):

PR #1350, opened 2022-06-03
t/apicast-policy-chains.t: convert to APIcast::Blackbox

samugi:


t/apicast-policy-chains-crash.t
```diff
@@ -0,0 +1,34 @@
+use lib 't';
+use Test::APIcast::Blackbox 'no_plan';
+
+$ENV{APICAST_ACCESS_LOG_FILE} = "$Test::Nginx::Util::ErrLogFile";
+$ENV{APICAST_POLICY_LOAD_PATH} = 't/fixtures/policies';
+
+repeat_each(1);
```
samugi, line 7: new file created to avoid changing `repeat_each` on the other tests. This one test however requires to test with only one request (or to check the accum log) otherwise the check in the error logs would fail.
The reason is that the error is printed during the configuration load, which happens on boot by default with Blackbox. I think it's ok to repeat this request once because in fact we wouldn't even need a request for this particular test, but the framework doesn't allow specifying an empty `--- request`.

t/apicast-policy-chains-crash.t
```diff
@@ -0,0 +1,34 @@
...
+            ]
+        }
+      }
+    ]
+}
+--- request
+GET /test
+--- response_body
+GET /test HTTP/1.1
+--- error_code: 200
```
eguzki, line 32: the policy chain is corrupted and still returns 200 OK ? weird, isn't it?

samugi: the request returns 200 because the expected behaviour is for the failing policy to be just removed from the policy chain silently, leaving the rest of the policy chain in working conditions. See [here](https://github.com/3scale/APIcast/pull/1350/files#diff-083e627e453a3a925ace595f76e2aee23b78d31bfae85eac4e71692f7716158aL247-L280) the original test (just moved in this new file as explained above)

eguzki: thanks for the explanation

PR #1352, opened 2022-06-06
THREESCALE-8508 - /admin/api/account/proxy_configs endpoint for configuration loading

samugi:
This PR uses `/admin/api/account/proxy_configs` for the standard configuration loading process, as the default endpoint that APIcast will attempt to fetch the configuration from when the portal endpoint is provided without a path (admin portal endpoint).

With this proposed implementation, [APICAST_LOAD_SERVICES_WHEN_NEEDED](https://access.redhat.com/documentation/en-us/red_hat_3scale_api_management/2.11/html/administering_the_api_gateway/apicast_environment_variables#load-services-when-needed) is deprecated and the configuration is fetched "when needed" by default whenever possible, with the following logic:

* THREESCALE_PORTAL_ENDPOINT has no path
  * APICAST_CONFIGURATION_LOADER=boot
    * `http://${THREESCALE_PORTAL_ENDPOINT}/admin/api/account/proxy_configs/<env>.json?version=version` (i.e. all services from proxy_configs endpoint). Using the new endpoint now one request is enough, no need to loop through all services.
  * APICAST_CONFIGURATION_LOADER=lazy
    * `http://${THREESCALE_PORTAL_ENDPOINT}/admin/api/account/proxy_configs/<env>.json?host=host&version=version` (i.e. services that match `host` from proxy_configs endpoint). This matches the current behaviour when APICAST_LOAD_SERVICES_WHEN_NEEDED is enabled.
* THREESCALE_PORTAL_ENDPOINT has a path
  * APICAST_CONFIGURATION_LOADER=boot
    * `http://${THREESCALE_PORTAL_ENDPOINT}/<env>.json` (i.e. all services from master endpoint). Same as the current behaviour.
  * APICAST_CONFIGURATION_LOADER=lazy
    * `http://${THREESCALE_PORTAL_ENDPOINT}/<env>.json?host=host` (i.e. services that match `host` from master endpoint). Same as the current behaviour.

i.e. the path appended to ${THREESCALE_PORTAL_ENDPOINT} is:
|                      | APICAST_CONFIGURATION_LOADER=boot                           | APICAST_CONFIGURATION_LOADER=lazy                                     |
|----------------------|-------------------------------------------------------------|-----------------------------------------------------------------------|
| endpoint has no path | /admin/api/account/proxy_configs/${env}.json?version=version | /admin/api/account/proxy_configs/${env}.json?host=host&version=version |
| endpoint has a path  | /${env}.json                                                 | /${env}.json?host=host                                                 |

**Update** : When any of the following environment variables exist, the gateway will load services one by one. 
* `APICAST_SERVICE_%s_CONFIGURATION_VERSION`
* `APICAST_SERVICES_LIST`
* `APICAST_SERVICES_FILTER_BY_URL`

These environment variables are not compatible when pointing to master (when a custom path is provided). Added a check to log error when that incompatible scenario happens.


gateway/src/apicast/configuration_loader/remote_v2.lua
```diff
@@ -162,6 +146,18 @@ local function reset_env()
          resty_env.value('APICAST_CONFIGURATION_LOADER') == 'boot'
 end
 
+-- Returns a table that represents paths and query parameters for the current endpoint:
+-- http://${THREESCALE_PORTAL_ENDPOINT}/<env>.json?host=host
+-- http://${THREESCALE_PORTAL_ENDPOINT}/admin/api/account/proxy_configs/<env>.json?host=host&version=version
+local function configuration_endpoint_params(env, host, portal_endpoint_path)
+  local version = resty_env.get(
+      format('APICAST_SERVICE_%s_CONFIGURATION_VERSION', service_id)
```
eguzki, line 154: where is this `service_id` coming from?

eguzki:
Code looks ok. Also tests. 

Just to verify: APIcast no longer will request proxy configuration for a given service ID. When "lazy" configuration, the endpoint will have a query param for the given "host". Right?

Question: what if two services have the same host?

eguzki:
For managed APIcast deployment (on-prem releases), the THREESCALE_PORTAL_ENDPOINT env var has path configured to `/master/api/proxy/configs` 
https://github.com/3scale/3scale-operator/blob/69c1e2c41b2b9d308278e513d07d405b5d8196e6/pkg/3scale/amp/component/system_options.go#L114

This is still valid or should be changed?

gateway/src/apicast/configuration_loader/remote_v2.lua
samugi: apparently, that environment variable (for configuring a specific service version) only works when specific conditions are met:

* THREESCALE_PORTAL_ENDPOINT points to an admin portal, and
* APICAST_LOAD_SERVICES_WHEN_NEEDED is **not** configured

Because in order to fetch a specific version it's necessary to load services individually (so that the version can be specified for each service configuration URL). And this is not possible via master or the new services endpoint. This is currently the behavior we have also in master and even in previous stable versions.

For now, I updated the logic to fall back to the "old" endpoint and load services one by one when `APICAST_SERVICE_%s_CONFIGURATION_VERSION` is configured (and `THREESCALE_PORTAL_ENDPOINT` is an admin portal), and I added logs to notify when services will be loaded this way, individually. This means we would preserve the current behaviour when the version env var is set.

We need to verify and confirm that this is the best way to handle things.

samugi:
> Code looks ok. Also tests.
> 
> Just to verify: APIcast no longer will request proxy configuration for a given service ID. When "lazy" configuration, the endpoint will have a query param for the given "host". Right?
> 
> Question: what if two services have the same host?

That's fine @eguzki, since it points to master, this PR should have no effect on gateways configured that way.

secretsurf:
Is this lost in the ether?

kevprice83:
> Is this lost in the ether?

No @secretsurf sorry I am supposed to be reviewing and testing this but too many other burning issues I didn't manage to get to this yet.

eguzki:
One thing I concern about this is the paginated endpoint. From [the description in the issue](https://issues.redhat.com/browse/THREESCALE-8508)
```
Open Points:  

    Currently the endpoint /admin/api/account/proxy_configs/${ENV}

 is paginated, and APIcast does not handle pages, so we would need to 
make the pagination optional or to implement pagination on APIcast too before we can proceed with this
```

and I do not remember to see any pagination implementation. Maybe it was done after my initial review

kevprice83 (changes requested):
I think this PR is missing documentation because of how it behaves with different env vars. Without that documentation it will result in users run into issues as they won't understand what is the expected behaviour. Also a decision needs to be made whether we want to split the implementation into 2 iterations as this PR will not support >500 services but it does introduce support for >500 versions of a service's proxy config object.

gateway/src/apicast/configuration_loader/remote_v2.lua
eguzki: I added the same "exceptions" for `APICAST_SERVICES_LIST` and `APICAST_SERVICES_FILTER_BY_URL` env vars.

eguzki:
@pehala @thomasmaas this PR is ready for review

eguzki:
leaving pagination implementation for a (soon) coming PR

dfennessy (approved):
LGTM!

pehala (commented):
This seems like a breaking change, can you enable the old behaviour? Does enabling the old behaviour even make sense?

eguzki:
> This seems like a breaking change, can you enable the old behaviour? Does enabling the old behaviour even make sense?

Basically, we are making an opt-in behavior the default one. Opt-in is no longer needed, hence removal of env var.

pehala:
> > This seems like a breaking change, can you enable the old behaviour? Does enabling the old behaviour even make sense?
> 
> Basically, we are making an opt-in behavior the default one. Opt-in is no longer needed, hence removal of env var.

Right, and is there a option to opt-out? Does it even make sense to opt-out or is the new behaviour simply superior?

eguzki:
> > > This seems like a breaking change, can you enable the old behaviour? Does enabling the old behaviour even make sense?
> > 
> > 
> > Basically, we are making an opt-in behavior the default one. Opt-in is no longer needed, hence removal of env var.
> 
> Right, and is there a option to opt-out? Does it even make sense to opt-out or is the new behaviour simply superior?

For me it does not make sense to opt-out. But if for whatever reason there is a need to opt-out from default behavior (i.e. get all services and ask for proxy config for each service), there are ways. 
* Use `APICAST_SERVICE_${SERVICE_ID}_CONFIGURATION_VERSION` env var with a not existing ${SERVICE_ID}
* Use `APICAST_SERVICES_FILTER_BY_URL` with a filter that always match.

pehala (approved):
I do not have enough knowledge to do a full-on technical review. However, the proposed changes make sense to me.

PR #1353, opened 2022-06-06
remove container volume from docker-compose

eguzki:
### what

* `apicast-source` target not needed for cleaning up `clean-containers`
  * no need to have a container (running or stopped) to use `docker-compose.yaml` configuration
  * makes easier to support `podman`. But currently I am not sure podman can replace docker, I haven't tested. 
* `prove` service from `docker-compose.yaml` removed and created a dedicated `docker-compose.prove.yml`. That configuration is being used in `make prove-docker` command. Used to run integration tests without running the development image (`make development`)
* `clean-containers` makefile target cleans all docker compose environments there could be up and running. 

### Verification steps

Run integration tests in docker
```
make prove-docker
```

Clean running  environment(s) 

```
make clean-containers
```
check `apicast_build_0_prove` and `apicast_build_0_redis` have been cleaned.
```
docker ps -a | grep apicast_build_0
(empty)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,10 @@ DEVEL_DOCKER_COMPOSE_FILE ?= docker-compose-devel.yml
 DEVEL_DOCKER_COMPOSE_VOLMOUNT_MAC_FILE ?= docker-compose-devel-volmount-mac.yml
 DEVEL_DOCKER_COMPOSE_VOLMOUNT_DEFAULT_FILE ?= docker-compose-devel-volmount-default.yml
 
+PROVE_DOCKER_COMPOSE_FILE ?= docker-compose.prove.yml
+
+DOCKER_VOLUME_NAME ?= apicast-local-volume
+
 os = "$(shell uname -s)"
 
 # if running on Mac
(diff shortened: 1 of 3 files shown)


samugi:
the steps described work as expected. Anything specific to podman that can be tested here?

samugi:
> I would go over the makefile targets
> 
> `make development` then run busted and prove -> hits docker-compose-devel.yml `make prove-docker` -> hits docker-compose.prove.yml `make runtime-image` `make test-runtime-image` -> hits docker-compose.yaml `make dev-build`
> 
> If they work with podman, I would say that podman is supported.

so I reverted [this](https://github.com/3scale/APIcast/pull/1344/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L5) and then ran all the commands which executed just fine, however when I do `podman ps` I see no running containers, they populate the `docker ps` output, is that expected? I can see in the output of commands that it is actually using podman, e.g.:
```
$ make runtime-image                                                                                                                             
/usr/bin/podman  build -t apicast-runtime-image:latest
```

eguzki:
I would expect `podman ps` to return running containers :shrug: 

I would say, let's investigate podman support in other PR before enabling podman with 
```
DOCKER ?= $(shell which podman 2> /dev/null || which docker 2> /dev/null || echo "docker")
```

samugi (approved):
lgtm, we can investigate podman support separately

PR #1372, opened 2022-08-17
Latest releases

eguzki:
### what
* CHANGELOG.md: updated
* `gateway/src/apicast/version.lua` set to `latest`. It should be updated accordingly in release branches


kevprice83 (commented):
What is the advantage of specifying latest? Doesn't that mean every release will print latest? Having the specific versions is quite handy when troublshooting and we want to quickly refer to the version.

eguzki:
version should be updated accordingly in release branches.

kevprice83 (approved):

PR #1376, opened 2022-10-06
opentelemetry instrumentation THREESCALE-7735

eguzki:
### what

implements [THREESCALE-7735](https://issues.redhat.com/browse/THREESCALE-7735)

Add [Opentelemetry SDK] support for APIcast. Jaeger client (based on opentracing) support is still in place, but gets deprecated in favor of the Opentelemetry use.


eguzki:
follow up on #1379

PR #1377, opened 2022-10-12
Create a new workflow to create a release and tag in the repository

jjkiely:
##### This action will be manually triggerable (to create a tag in this repo alone) but can also by triggered by another action which can create tags in all upstream repos at once, similar to the old jenkins job.


jjkiely:
@eguzki I see you're a member of APIcast core can you review this one as well when you get the chance please?

eguzki (approved):

Issue #1375, opened 2022-09-30
Setup local env fails


imod:
Hi,

I'm trying to setup a local environment to develop a custom policy, but I fail to get APIcast up and running as described...

##### Version

`master` branch

##### Steps To Reproduce

1. `git clone git@github.com:3scale/APIcast.git`
2. `cd APIcast`
3. `make development`
4. `make dependencies` --> fails

##### Current Result

```
domi 🙈 APIcast $ make development
Running on Darwin
docker compose -f docker-compose-devel.yml -f docker-compose-devel-volmount-mac.yml up -d
[+] Running 2/2
 ⠿ Container apicast_build_0-redis-1        Running                                                                            0.0s
 ⠿ Container apicast_build_0-development-1  Started                                                                            1.5s
docker compose -f docker-compose-devel.yml -f docker-compose-devel-volmount-mac.yml exec -e COLUMNS="`tput cols`" -e LINES="`tput lines`" --user 501:20 development bash
bash-4.4$ make dependencies
/usr/local/openresty/luajit/bin/rover lock --roverfile=/opt/app-root/src/gateway/Roverfile
Resolving /opt/app-root/src/gateway/Roverfile
Cloning into 'lua-resty-repl'...
fatal: unable to connect to github.com:
github.com[0: 140.82.121.3]: errno=Connection refused

/usr/local/openresty/luajit/bin/luajit-2.1.0-beta3: /usr/share/lua/5.1/rover/lock.lua:169: attempt to index local 'rockspec' (a nil value)
stack traceback:
        /usr/share/lua/5.1/rover/lock.lua:169: in function 'expand_dependencies'
        /usr/share/lua/5.1/rover/lock.lua:199: in function 'resolve'
        /usr/share/lua/5.1/rover/cli/lock.lua:23: in function </usr/share/lua/5.1/rover/cli/lock.lua:11>
        /usr/share/lua/5.1/rover/cli.lua:33: in function </usr/share/lua/5.1/rover/cli.lua:27>
        /usr/lib64/luarocks/rocks/lua-rover/scm-1/bin/rover:8: in main chunk
        [C]: at 0x55ad138087a0
make: *** [Makefile:214: /opt/app-root/src/gateway/Roverfile.lock] Error 1
bash-4.4$ ^C
bash-4.4$ exit
```

##### Additional Information

* I tried to access github.com from within the running container, it works with `git` (ssh and https) and `curl` (https)

eguzki:
[...]

PR #1380, opened 2022-11-09
fix: use git https protocol

guicassolato:
Due to `lua_modules` make target invoking `rover lock` command before `translate_git_protocol` (see #1320), commonly used targets like `dependencies` often fail with a `Connection refused` error.

Reported in #1375.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ tmp/benchmark/
 .cpanm
 /vendor/cache
 /tmp/
+.gitconfig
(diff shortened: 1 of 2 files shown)


eguzki (approved):
I do not think that the `.gitconfig` is needed because the `translate_git_protocol` adds the config globally inside the container, but does not do any harm.

guicassolato:
> I do not think that the `.gitconfig` is needed because the `translate_git_protocol` adds the config globally inside the container, but does not do any harm.

Adding globally inside the container is exactly the problem here. Since the git context for `global` is the project directory itself, `git config --global` adds the `.gitconfig` file to the changes. Even after you leave the container, you still see that file uncommitted.

We could try .gitignoring `.gitconfig` instead if you like.

eguzki:
> nce the git conte

Ok I see, $HOME is the APIcast working dir. So, what I would do is to add the .gitconfig to the .gitignore and we do not need to commit it, as the content is duplicated in the Makefile. Not a big deal, as you want. The PR is already approved.

PR #1382, opened 2022-12-05
Adds matches operation to Conditional policy

jonathansimas:
It's already possible to use the 'matches' operator with APIcast Conditional Policy just editing the Product's policy chain yaml and using it instead of "==" and "!=", enabling the possibility of using regular expressions to match left and right.


tkan145:
Closing due to stale PR

PR #1389, opened 2023-02-16
THREESCALE-9009 fix OIDC jwt key verification

eguzki:
### what

Fixes [THREESCALE-9009](https://issues.redhat.com/browse/THREESCALE-9009) Token generated for a realm associated with a different 3scale Product is able to reach the upstream with no verification performed

When a oidc JWT belonging to realm A is used in a request targeting a 3scale service configured with OIDC authentication with clients configured in realm B, the first conflict APIcast finds is that the referenced public key (in the JWT) does not exist. This PR address this issue adding a check on the JWKey object. 

In the absence of this check, the lua code was working with a `nil` object and raised an error (runtime error: nil pointer dereference) captured only by the [policy chain manager](https://github.com/3scale/APIcast/blob/3scale-2.13.1-GA/gateway/src/apicast/policy_chain.lua#L215).

```lua
local status, return_val = pcall(self[i][phase_name], self[i], context)
```
When this error is raised, the current policy processing stage is halted and skipped, moving on to the next policy in the chain. The token is not even verified with the JWT library as it should be in the next step
```
local pubkey = jwk_obj.pem          -> Raises runtime error                                                
    
if jwk_obj.alg and jwk_obj.alg ~= jwt.header.alg then          -> Not executed                      
  return false, '[jwt] alg mismatch'                                                
end                                                                                 
                                                                                    
jwt = JWT:verify_jwt_obj(pubkey, jwt, self.jwt_claims)    ->Not executed                                      

```

 That causes the APIcast context to enter in a inconsistent state. Specifically, nginx `ctx.credentials` is not populated. Maybe more info but I have not gone through all of them because it is inconsistent, hence unpredictable behavior:
* When the routing policy is added, the request makes its way to upstream and APIcast sends a request to backend with a missing `app_id`. Backend responds with `404 Not Found` (as expected) but APICast ignores this and returns 200 OK to the downstream client.  
* When the routing policy is not in place, APIcast (under this incosisten state) is unable to find an upstream for the request and returns `404 Not Found` to the upstream client. Additionally, APIcast sends a request to backend with a missing `app_id`. Backend responds with `404 Not Found` (as expected).

With the check on the JWKey object, APIcast deals with a managed error (the token is invalid) and the behavior is expected, returning `403 Forbidden` to the downstream client as expected. In that case, backend is **not** called.

### Verification steps
* Setup Keycloak instance, realms, clients and users
* Install 3scale
* [Integrating 3scale with Red Hat Single Sign-On as the OpenID Connect identity provider](https://access.redhat.com/documentation/en-us/red_hat_3scale_api_management/2.13/html/administering_the_api_gateway/integrating-threescale-with-an-openid-connect-identity-provider#integrating-threescale-with-rhsso-as-the-openid-connect-identity-provider_oidc)
  * Configure a single Product A with an OpenID Provider + realm A
* Start dev environment
```
make development
make dependencies
```
* Run apicast locally
```
```
* Generate token from a realm B (not the same as configured for the Product A)
```
export ACCESS_TOKEN=$(curl -k -H "Content-Type: application/x-www-form-urlencoded" \
        -d 'grant_type=password' \
        -d 'client_id=my-client' \
        -d 'username=bob' \
        -d 'password=p' "https://myoidcprovider.example.com/auth/realms/B/protocol/openid-connect/token" | jq -r '.access_token')
```
* Run query with the invalid jwt
```
# capture apicast IP

curl -v -k -H "Host: example.com:443" -H "Accept: application/json" -H "Authorization: Bearer ${ACCESS_TOKEN}" http://${APICAST_IP}:8080/
```
* APIcast should notify the invalid token error in logs
```
2023/02/16 17:05:33 [debug] 100542#100542: *33 proxy.lua:287: rewrite(): oauth failed with [jwk] jwk not found, token might not belong to right realm, requestID=94a325d7d76eff70d6dd70d6a490a157
```
* The response should be `HTTP/1.1 403 Forbidden`
```
< HTTP/1.1 403 Forbidden
< Server: openresty
< Date: Thu, 16 Feb 2023 17:05:33 GMT
< Content-Type: text/plain; charset=us-ascii
< Transfer-Encoding: chunked
< Connection: keep-alive
< 
* Connection #0 to host 172.20.0.3 left intact
Authentication failed
```

### DIFF
--- a/gateway/src/apicast/oauth/oidc.lua
+++ b/gateway/src/apicast/oauth/oidc.lua
@@ -186,7 +186,8 @@ function _M:verify(jwt, cache_key)
   local jwk_obj = find_jwk(jwt, self.keys)
 
   if jwk_obj == nil then
-    return false, '[jwk] jwk not found, token might not belong to right realm'
+    ngx.log(ngx.ERR, "[jwt] failed verification for kid: ", jwt.header.kid)
+    return false, '[jwk] not found, token might belong to a different realm'
   end
 
   local pubkey = jwk_obj.pem
--- a/gateway/src/apicast/oauth/oidc.lua
+++ b/gateway/src/apicast/oauth/oidc.lua
@@ -105,7 +105,7 @@ end
 
 local function find_jwk(jwt, keys)
   local jwk = keys and keys[jwt.header.kid]
-  if jwk then return jwk end
+  return jwk
 end
 
 -- Parses the token - in this case we assume it's a JWT token
@@ -185,8 +185,12 @@ function _M:verify(jwt, cache_key)
   -- Find jwk with matching kid for current JWT in request
   local jwk_obj = find_jwk(jwt, self.keys)
 
+  if jwk_obj == nil then
+    return false, '[jwk] jwk not found, token might not belong to right realm'
+  end
+
   local pubkey = jwk_obj.pem
-  -- Check the jwk for the alg field and if not present skip the validation as it is 
+  -- Check the jwk for the alg field and if not present skip the validation as it is
   -- OPTIONAL according to https://www.rfc-editor.org/rfc/rfc7517#section-4.4
   if jwk_obj.alg and jwk_obj.alg ~= jwt.header.alg then
     return false, '[jwt] alg mismatch'
--- a/t/apicast-oidc.t
+++ b/t/apicast-oidc.t
@@ -389,17 +389,6 @@ to_json({
     keys => { somekid => { pem => $::public_key, alg => 'RS256' } },
   }]
 });
---- upstream
-  location /test {
-    echo "yes";
-  }
---- backend
-  location = /transactions/oauth_authrep.xml {
-    content_by_lua_block {
-      local expected = "provider_key=fookey&service_id=42&usage%5Bhits%5D=1&app_id=appid"
-      require('luassert').same(ngx.decode_args(expected), ngx.req.get_uri_args(0))
-    }
-  }
 --- request: GET /test
 --- error_code: 403
 --- more_headers eval
--- a/t/apicast-oidc.t
+++ b/t/apicast-oidc.t
@@ -313,7 +313,7 @@ my $jwt = encode_jwt(payload => {
 "Authorization: Bearer $jwt"
 --- no_error_log
 
-=== TEST 7: JWT verification fails when jwk.alg exists AND does not match jwt.header.alg 
+=== TEST 7: JWT verification fails when jwk.alg exists AND does not match jwt.header.alg
 (see THREESCALE-8249 for steps to generate tampered JWT. rsa.pub from fixtures used to sign)
 --- configuration env eval
 use JSON qw(to_json);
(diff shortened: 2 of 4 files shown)


gateway/src/apicast/oauth/oidc.lua
```diff
@@ -185,8 +185,12 @@ function _M:verify(jwt, cache_key)
   -- Find jwk with matching kid for current JWT in request
   local jwk_obj = find_jwk(jwt, self.keys)
 
+  if jwk_obj == nil then
+    return false, '[jwk] jwk not found, token might not belong to right realm'
```
kevprice83, line 189: I would suggest the following log message which would be more explicit and accurate:
```
ngx.log(ngx.ERR, "[jwt] failed verification for kid: ", jwt.header.kid)
return false, '[jwk] not found, token might belong to a different realm'
```

This would help customers identify the issue quicker and without having to enable debug log level.

t/apicast-oidc.t
```diff
@@ -362,3 +362,43 @@ my $jwt = 'eyJraWQiOiJzb21la2lkIiwiYWxnIjoiSFMyNTYifQ.'.
 "Authorization: Bearer $jwt"
 --- error_log
 [jwt] alg mismatch
+
+=== TEST 8: Key not available
+(Happens when the token belongs to a different realm)
```
kevprice83, line 367: I think the more accurate test description is that the "token was signed by a different key". The same key could be used accross multiple realms but with this fix the `iss` claim will be correctly validated against the `issuer` field in the proxy config. Equally if the `kid`s don't match then it won't even find the jwk.

In fact we can see that further down in this test, we have different realms but the logic never gets that far in the validation process because it fails before that with _"[jwk] jwk not found"_

t/apicast-oidc.t
```diff
@@ -362,3 +362,43 @@ my $jwt = 'eyJraWQiOiJzb21la2lkIiwiYWxnIjoiSFMyNTYifQ.'.
...
+use Crypt::JWT qw(encode_jwt);
+my $jwt = encode_jwt(payload => {
+  aud => 'something',
+  azp => 'appid',
+  sub => 'someone',
+  iss => 'https://example.com/auth/realms/b',
+  exp => time + 3600 }, key => \$::private_key, alg => 'RS256', extra_headers => { kid => 'otherkid' });
+"Authorization: Bearer $jwt"
+--- error_log
+[jwk] jwk not found
```
kevprice83, line 404: If you use the same error message suggested in my review then this should be updated to look for `[jwk] not found`, to save you from a failing test you would need to fix 😉

kevprice83 (changes requested):
just a change to the error message printed which would be super helpful in future troubleshooting and a minor change to the description of the integration test so we can check the scenarios tested as well in future. 

The rest looks good otherwise.

t/apicast-oidc.t
eguzki: added a test for the verification of the expected issuer

eguzki:
Thank for the review @kevprice83. All comments addressed. Ready for a new review

kevprice83 (commented):
I like that additional last test yes! Also it's using the [reason](https://github.com/SkyLothar/lua-resty-jwt/blob/ee1d024071f872e2b5a66eaaf9aeaf86c5bab3ed/lib/resty/jwt.lua#L707) field on the verification object so I think this makes the test super clear. Nice improvement :)

Everything else looks good 🥇

PR #1397, opened 2023-03-16
THREESCALE-8373 Pagination services and proxy config endpoints

eguzki:
### what
Load all services and proxy configuration pages

Fixes https://issues.redhat.com/browse/THREESCALE-8373

### Verification steps

* Create more than 500 services in a tenant
* Deploy APIcast using `THREESCALE_PORTAL_ENDPOINT=https://<token>@<provider domain>`
* Send a request to one of last services created
* Observe 200 OK response (also observe APIcast logs indicate that first 500 services were loaded in a first request and the remaining services in a second request)

### DIFF
--- a/doc/parameters.md
+++ b/doc/parameters.md
@@ -46,6 +46,7 @@ Defines how to load the configuration.
 In `boot` mode APIcast will request the configuration to the API manager when the gateway starts.
 In `lazy` mode APIcast will load the configuration on demand for each incoming request (to guarantee a complete refresh on each request `APICAST_CONFIGURATION_CACHE` should be set to `0`).
 
+
 ### `APICAST_CUSTOM_CONFIG`
 
 **Deprecated:** Use [policies](./policies.md) instead.
@@ -241,6 +242,8 @@ Replace `${ID}` with the actual Service ID. The value should be the configuratio
 
 Setting it to a particular version will prevent it from auto-updating and will always use that version.
 
+**Note**: This env var cannot be used with `THREESCALE_PORTAL_ENDPOINT` pointing to custom path (i.e. master path).
+
 ### `APICAST_UPSTREAM_RETRY_CASES`
 
 **Default**:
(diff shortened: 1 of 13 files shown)


eguzki:
@thomasmaas @pehala ready for review

cc @samugi @kevprice83

doc/parameters.md
```diff
@@ -241,6 +242,8 @@ Replace `${ID}` with the actual Service ID. The value should be the configuratio
 
 Setting it to a particular version will prevent it from auto-updating and will always use that version.
 
+**Note**: This env var cannot be used with `THREESCALE_PORTAL_ENDPOINT` pointing to custom path (i.e. master path).
```
pehala, line 245: Does that mean this variable won't work with basic staging/production APIcast created by 3scale operator? Did it work before and did it changed now or has it simply never worked before?

gateway/src/apicast/configuration_loader/remote_v2.lua
```diff
@@ -259,6 +243,83 @@ function _M:index(host)
...
 end
 
+-- Returns existing proxy configs in a single page
+-- @param http_client the http client object
+-- @param portal_endpoint 3scale API endpoint
+-- @param host proxy config filter based on request hostname. Optional, can be nil.
+-- @param env gateway environment
+-- @param page page in the paginated list. Defaults to 1 for the API, as the client will not send the page param.
+-- @param per_page number of results per page. Default and max is 500 for the API, as the client will not send the per_page param.
+local proxy_configs_per_page = function(http_client, portal_endpoint, host, env, page, per_page)
```
pehala, line 253: Why is it simply not a function? e.g. `function proxy_configs_per_page(..)...`

gateway/src/apicast/configuration_loader/remote_v2.lua
```diff
@@ -341,23 +422,32 @@ function _M:services()
...
   end
 
-  local url = services_index_endpoint(endpoint)
-  local res, err = http_client.get(url)
-
-  if not res and err then
-    ngx.log(ngx.DEBUG, 'services get error: ', err, ' url: ', url)
-    return nil, err
-  end
+  local SERVICES_PER_PAGE = 500
```
pehala, line 425: Is there any reason for this particular value? Do different values change performance in any reasonable way?

pehala (commented):
Nice job! Should a combination of this and #1352 improve the overall performance of APIcast?

doc/parameters.md
eguzki: Never worked before. The env var was silently ignored. Now (actually in the previous PR), I added a hard [check](https://github.com/3scale/APIcast/pull/1397/files#diff-45d2061c12ed47f3c6c589f0aa51e73640f5518ee3b97085a85d9822a9c9618fR340). The check will not kill the process, but will not process the request and will log error.

gateway/src/apicast/configuration_loader/remote_v2.lua
eguzki: fixed

gateway/src/apicast/configuration_loader/remote_v2.lua
eguzki: That's the default value used in the 3scale API. The gateway still needs to fix the value because it needs to iterate comparing the number of returned results with that value. If the gateway does not fix that value, a change in the 3scale API can break the gateway pagination feature. 

In the future, we can expose as env var

thomasmaas (approved):

PR #1401, opened 2023-05-02
Fix Transfer-Encoding header not removed for buffered requests

tkan145:
What

by default, APICast buffers the entire request before sending it upstream, however, if proxies are involved and the request is sent with a "Transfer-Encoding: chunked" header, the API will forward the request with the full buffered request body and "Transfer-Encoding: chunked" header and thus upset the upstream server.

This PR fixes https://issues.redhat.com/browse/THREESCALE-9542 by removing the "Transfer-Encoding: chunked".

### DIFF
--- a/gateway/src/apicast/http_proxy.lua
+++ b/gateway/src/apicast/http_proxy.lua
@@ -1,5 +1,7 @@
 local format = string.format
 
+local open = io.open
+local close = io.close
 local resty_url = require "resty.url"
 local resty_resolver = require 'resty.resolver'
 local round_robin = require 'resty.balancer.round_robin'
@@ -81,6 +83,20 @@ local function absolute_url(uri)
     )
 end
 
+function fsize (path)
+  local file, err = open(path)
+
+  if err or not file then
+    return nil, err
+  end
+
+  local current = file:seek()      -- get current position
+  local size = file:seek("end")    -- get file size
+  file:seek("set", current)        -- restore position
+  close(file)
+  return size
+end
+
 local function forward_https_request(proxy_uri, uri, skip_https_connect)
     -- This is needed to call ngx.req.get_body_data() below.
     ngx.req.read_body()
@@ -109,6 +125,12 @@ local function forward_https_request(proxy_uri, uri, skip_https_connect)
         ngx.log(ngx.INFO, "HTTPS Proxy: Request body is bigger than client_body_buffer_size, read the content from path='", temp_file_path, "'")
 
         if temp_file_path then
+          -- Set Content-Length only if none exist
+          if not request.headers["Content-Length"] then
+              local size, err = fsize(temp_file_path)
+              request.headers["Content-Length"] = size
+          end
+
           local body, err = file_reader(temp_file_path)
           if err then
             ngx.log(ngx.ERR, "HTTPS proxy: Failed to read temp body file, err: ", err)
--- a/gateway/src/apicast/http_proxy.lua
+++ b/gateway/src/apicast/http_proxy.lua
@@ -118,6 +118,11 @@ local function forward_https_request(proxy_uri, uri, skip_https_connect)
         end
     end
 
+    -- The whole request is buffered in the memory so remove the Transfer-Encoding: chunked
+    if request.headers["Transfer-Encoding"] == "chunked" then
+        request.headers["Transfer-Encoding"] = nil
+    end
+
     local httpc, err = http_proxy.new(request, skip_https_connect)
 
     if not httpc then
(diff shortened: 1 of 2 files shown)


gateway/src/apicast/http_proxy.lua
```diff
@@ -81,6 +83,20 @@ local function absolute_url(uri)
     )
 end
 
+function fsize (path)
+  local file, err = open(path)
```
kevprice83, line 87: This is I/O blocking, I don't think we want to do this on every request as it will really hurt performance.

gateway/src/apicast/http_proxy.lua
```diff
@@ -81,6 +83,20 @@ local function absolute_url(uri)
...
+
+  if err or not file then
+    return nil, err
+  end
+
+  local current = file:seek()      -- get current position
```
kevprice83, line 93: This seems unnecessary as we already have a [file_reader](https://github.com/3scale/APIcast/blob/fdf5b2ef36adaf884795e12ed91aa9ce8ffba2f3/gateway/src/resty/file.lua#L13) function which is implemented with lua coroutines and will be much safer in my opinion.

gateway/src/apicast/http_proxy.lua
```diff
@@ -109,6 +125,12 @@ local function forward_https_request(proxy_uri, uri, skip_https_connect)
         ngx.log(ngx.INFO, "HTTPS Proxy: Request body is bigger than client_body_buffer_size, read the content from path='", temp_file_path, "'")
 
         if temp_file_path then
+          -- Set Content-Length only if none exist
+          if not request.headers["Content-Length"] then
```
kevprice83, line 129: This check should not be needed. The RFC states that only one of `Transfer-Encoding` & `Content-Length` should be sent and in the case of chunked requests the latter should be omitted or ignored in which case we must always set this.

gateway/src/apicast/http_proxy.lua
```diff
@@ -109,6 +125,12 @@ local function forward_https_request(proxy_uri, uri, skip_https_connect)
...
+              local size, err = fsize(temp_file_path)
```
kevprice83, line 130: Better to move this logc to [file_reader](https://github.com/3scale/APIcast/blob/fdf5b2ef36adaf884795e12ed91aa9ce8ffba2f3/gateway/src/resty/file.lua#L13) function where it is safely read in chunks. I think what we can do here is push each chunk into a table, iterate over each chunk in the table to get the size, sum the sizes and then set that as the `Content-Length` value. I've not thought to much about the implementation yet but given each chunk should be of a [known size](https://github.com/3scale/APIcast/blob/fdf5b2ef36adaf884795e12ed91aa9ce8ffba2f3/gateway/src/resty/file.lua#L6) we might not even need to push each chunk into a table, simply do a count and multiply the size of each chunk by the length of the table - 1 and then we only ever need to calculate the size of the last chunk. Maybe @eguzki can comment on this proposal.

gateway/src/apicast/http_proxy.lua
```diff
@@ -118,6 +140,11 @@ local function forward_https_request(proxy_uri, uri, skip_https_connect)
         end
     end
 
+    -- The whole request is buffered in the memory so remove the Transfer-Encoding: chunked
+    if request.headers["Transfer-Encoding"] == "chunked" then
```
kevprice83, line 144: We should make it very clear here that this is a short-term fix and when we finally add support for chunked requests to then remove this logic.

kevprice83:
@eguzki let's try to discuss this soon. I have documented a patch here: https://access.redhat.com/solutions/7013421 and will share a gist or draft PR just as a reference but ultimately my feeling is that this is not something we want to add to the Product. I think it would be much better to solve this by adding support for chunked requests otherwise this patch is kind of anti-pattern in my opinion and customers won't realise their requests are being modified.

kevprice83:
@tkan145 did you review my comments yet? Did you try to implement that? In the end I have actually provided a bug fix for the customer which reported the issue, it's in the original support case. I still think the correct fix that we would want to ship though is full support for chunked requests and that we don't touch any request headers. @eguzki would you agree?

tkan145:
@kevprice83 yes I have an updated code for this PR and I also have a patch to support the native chunked request but I haven't fully tested it. Let me know what you guys want to do and I'll send another PR

kevprice83:
@tkan145 that should be submitted as a new PR in my opinion and this one closed.

tkan145:
Close this, as https://github.com/3scale/APIcast/pull/1400 has a better implementation

PR #1403, opened 2023-06-01
[THREESCALE-9542] Part 2: Add support to proxy request with Transfer-Encoding: chunked

tkan145:
### What:
Fix  https://issues.redhat.com/browse/THREESCALE-9542

This PR adds support to proxy the request with "Transfer-Encoding: chunked" when using with the proxy server.

### Note to reviewers

Please just review the last 2 commits. I will rebase once part 1 merged

### Verification steps:

* Checkout this branch

* Make runtime-image IMAGE_NAME=apicast-test
```shell
```

* Then run the gateway with the built image

```diff
diff --git a/dev-environments/https-proxy-upstream-tlsv1.3/apicast-config.json b/dev-environments/https-proxy-upstream-tlsv1.3/apicast-config.json
index 5227c5aa..24c45338 100644
--- a/dev-environments/https-proxy-upstream-tlsv1.3/apicast-config.json
+++ b/dev-environments/https-proxy-upstream-tlsv1.3/apicast-config.json
@@ -44,6 +44,11 @@
           "host": "backend"
         },
         "policy_chain": [
+          {
+              "name": "request_unbuffered",
+              "version": "builtin",
+              "configuration": {}
+          },
           {
             "name": "apicast.policy.http_proxy",
             "configuration": {
```

```shell
make certs
```

* Send chunked request with one chunk body

```shell
curl --resolve post.example.com:8080:127.0.0.1 -v -H "Transfer-Encoding: chunked"   -H "Content-Type: application/json"  -d @my-data.json "http://post.example.com:8080/?user_key=123"
```

The request should return 200 OK.  Note that upstream echo API is reporting that the request included `Transfer-Encoding: chunked` header and the expected body.

<details>

```shell
 ▲  curl --resolve post.example.com:8080:127.0.0.1 -v -H "Transfer-Encoding: chunked"   -H "Content-Type: application/json"  -d 'hello, world' "http://post.example.com:8080/?user_key=123"
* Added post.example.com:8080:127.0.0.1 to DNS cache
* Hostname post.example.com was found in DNS cache
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to post.example.com (127.0.0.1) port 8080 (#0)
> POST /?user_key=123 HTTP/1.1
> Host: post.example.com:8080
> User-Agent: curl/7.61.1
> Accept: */*
> Transfer-Encoding: chunked
> Content-Type: application/json
>
> c
* upload completely sent off: 19 out of 12 bytes
< HTTP/1.1 200 OK
< {
<   "args": {
<     "user_key": "123"
<   },
<   "data": "hello, world",
<   "files": {},
<   "form": {},
<   "headers": {
<     "Accept": "*/*",
<     "Content-Type": "application/json",
<     "Host": "example.com",
<     "Transfer-Encoding": "chunked",
<     "User-Agent": "curl/7.61.1"
<   },
<   "json": null,
<   "origin": "172.25.0.2",
<   "url": "http://example.com/post?user_key=123"
< }
* Connection #0 to host post.example.com left intact
```

</details>

* Send chunked request with few chunks in the body delayed in time. Python3 is required.

First get the APICast IPAddress

```
 ▲ docker inspect https-proxy-upstream-tlsv13-gateway-run-d76ff72726ec | grep IPAddress
```

```python
❯ cat <<EOF >chunked-request.py
import http.client
import time

def gen():
    yield bytes('hi', "utf-8")
    time.sleep(2)
    yield bytes('there', "utf-8")
    time.sleep(2)
    yield bytes('bye', "utf-8")

http.client.HTTPConnection.debuglevel = 1
conn = http.client.HTTPConnection('127.0.0.1', 8080)

headers = {'Content-type': 'application/octet-stream', 'Host': 'post.example.com'}

conn.request('POST', '/?user_key=foo', gen(), headers)

response = conn.getresponse()
print(response.read().decode())
EOF
```

Replace `127.0.0.1` with the IP of APIcast gateway above

```shell
> python3 chunked-request.py
send: b'POST /?user_key=foo HTTP/1.1\r\nAccept-Encoding: identity\r\nTransfer-Encoding: chunked\r\nContent-type: application/octet-stream\r\nHost: post.example.com\r\n\r\n'
send: b'2\r\nhi\r\n'                                                                                                                                                        
send: b'5\r\nthere\r\n'                                                                                                                                                     
send: b'3\r\nbye\r\n'                                                                                                                                                       
send: b'0\r\n\r\n'                                                                                                                                                          
reply: 'HTTP/1.1 200 OK\r\n'                                                                                                                                                
header: Access-Control-Allow-Credentials: true                                                                                                                              
header: Access-Control-Allow-Origin: *                                                                                                                                      
header: Date: Tue, 09 Jan 2024 03:40:18 GMT                                                                                                                                 
header: Content-Type: application/json                                                                                                                                      
header: Server: gunicorn/19.9.0                                                                                                                                             
{
  "args": {
    "user_key": "foo"
  }, 
  "data": "hitherebye", 
  "files": {}, 
  "form": {}, 
  "headers": {
    "Accept-Encoding": "identity", 
    "Content-Type": "application/octet-stream", 
    "Host": "example.com", 
    "Transfer-Encoding": "chunked",
    "User-Agent": "lua-resty-http/0.14 (Lua) ngx_lua/10019"
  }, 
  "json": null, 
  "origin": "172.18.0.4"
  "url": "http://example.com/post?user_key=foo"
}
```
* Note that the upstream service got transfer encoding chunked request and chunked encoding of the request body with the length bytes preceding each chunk.

<details>

```shell
> 2024/01/09 03:40:14.000960414  length=203 from=0 to=202 
POST /post?user_key=foo HTTP/1.1\r                        
User-Agent: lua-resty-http/0.14 (Lua) ngx_lua/10019\r     
Transfer-Encoding: chunked\r                              
Host: example.com\r                                       
Accept-Encoding: identity\r                               
Content-type: application/octet-stream\r                  
\r                                                        
> 2024/01/09 03:40:14.000960570  length=7 from=203 to=209 
2\r
hi\r
> 2024/01/09 03:40:16.000952052  length=10 from=210 to=219
5\r
there\r
> 2024/01/09 03:40:18.000954073  length=8 from=220 to=227
3\r
bye\r
> 2024/01/09 03:40:18.000954120  length=5 from=228 to=232
0\r
\r
< 2024/01/09 03:40:18.000954722  length=653 from=0 to=652
HTTP/1.1 200 OK\r
Server: gunicorn/19.9.0\r
Date: Tue, 09 Jan 2024 03:40:18 GMT\r
Connection: keep-alive\r
Content-Type: application/json\r
Content-Length: 423\r
Access-Control-Allow-Origin: *\r
Access-Control-Allow-Credentials: true\r
\r
{
  "args": {
    "user_key": "foo"
  }, 
  "data": "hitherebye", 
  "files": {}, 
  "form": {}, 
  "headers": {
    "Accept-Encoding": "identity", 
    "Content-Type": "application/octet-stream", 
    "Host": "example.com", 
    "Transfer-Encoding": "chunked", 
    "User-Agent": "lua-resty-http/0.14 (Lua) ngx_lua/10019"
  }, 
  "json": null, 
  "origin": "172.18.0.4", 
  "url": "http://example.com/post?user_key=foo"
}
```
</details>

* Send chunked request with expect 100-continue header.
```python
❯ cat <<EOF >chunked-request.py
import http.client
import time

def gen():
    yield bytes('hi', "utf-8")
    time.sleep(2)
    yield bytes('there', "utf-8")
    time.sleep(2)
    yield bytes('bye', "utf-8")

http.client.HTTPConnection.debuglevel = 1
conn = http.client.HTTPConnection('127.0.0.1', 8080)

headers = {'Content-type': 'application/octet-stream', 'Host': 'post.example.com', 'Expect': '100-continue'}

conn.request('POST', '/?user_key=foo', gen(), headers)

response = conn.getresponse()
print(response.read().decode())
EOF
```

* Note that the upstream service got transfer encoding chunked request and return 100 Continue
```
▲ python3 ./chunked-request.py

send: b'POST /?user_key=foo HTTP/1.1\r\nAccept-Encoding: identity\r\nTransfer-Encoding: chunked\r\nContent-type: application/octet-stream\r\nHost: post.example.com\r\nExpect: 100-continue\r\n\r\n'
send: b'2\r\nhi\r\n'                          
send: b'5\r\nthere\r\n'                       
send: b'3\r\nbye\r\n'                         
send: b'0\r\n\r\n'                            
reply: 'HTTP/1.1 100 Continue\r\n'            
headers: [b'\r\n']                            
reply: 'HTTP/1.1 200 OK\r\n'                  
header: Access-Control-Allow-Credentials: true
header: Access-Control-Allow-Origin: *        
header: Date: Tue, 09 Jan 2024 03:47:02 GMT   
header: Content-Type: application/json        
header: Server: gunicorn/19.9.0
{                                                          
  "args": {                                                
    "user_key": "foo"                                      
  },                                                       
  "data": "hitherebye",                                    
  "files": {},                                             
  "form": {},                                              
  "headers": {                                             
    "Accept-Encoding": "identity",                         
    "Content-Type": "application/octet-stream",            
    "Expect": "100-continue",                              
    "Host": "example.com",                                 
    "Transfer-Encoding": "chunked",                        
    "User-Agent": "lua-resty-http/0.14 (Lua) ngx_lua/10019"
  },                                                       
  "json": null,                                            
  "origin": "172.18.0.4",                                  
  "url": "http://example.com/post?user_key=foo"            
}
```

### DIFF
--- a/gateway/src/apicast/http_proxy.lua
+++ b/gateway/src/apicast/http_proxy.lua
@@ -1,5 +1,9 @@
 local format = string.format
 local tostring = tostring
+local ngx_flush = ngx.flush
+local ngx_get_method = ngx.req.get_method
+local ngx_http_version = ngx.req.http_version
+local ngx_send_headers = ngx.send_headers
 
 local resty_url = require "resty.url"
 local resty_resolver = require 'resty.resolver'
@@ -7,9 +11,19 @@ local round_robin = require 'resty.balancer.round_robin'
 local http_proxy = require 'resty.http.proxy'
 local file_reader = require("resty.file").file_reader
 local file_size = require("resty.file").file_size
+local client_body_reader = require("resty.http.request_reader").get_client_body_reader
+local send_response = require("resty.http.response_writer").send_response
 local concat = table.concat
 
 local _M = { }
+
+local http_methods_with_body = {
+  POST = true,
+  PUT = true,
+  PATCH = true
+}
+
+local DEFAULT_CHUNKSIZE = 32 * 1024
 
 function _M.reset()
     _M.balancer = round_robin.new()
@@ -84,52 +98,105 @@ local function absolute_url(uri)
     )
 end
 
-local function forward_https_request(proxy_uri, proxy_auth, uri, skip_https_connect)
-    -- This is needed to call ngx.req.get_body_data() below.
-    ngx.req.read_body()
-
-    -- We cannot use resty.http's .get_client_body_reader().
-    -- In POST requests with HTTPS, the result of that call is nil, and it
-    -- results in a time-out.
-    --
-    --
-    -- If ngx.req.get_body_data is nil, can be that the body is too big to
-    -- read and need to be cached in a local file. This request will return
-    -- nil, so after this we need to read the temp file.
-    -- https://github.com/openresty/lua-nginx-module#ngxreqget_body_data
-    local body = ngx.req.get_body_data()
-    local encoding = ngx.req.get_headers()["Transfer-Encoding"]
-
-    if not body then
-        local temp_file_path = ngx.req.get_body_file()
-        ngx.log(ngx.INFO, "HTTPS Proxy: Request body is bigger than client_body_buffer_size, read the content from path='", temp_file_path, "'")
-
-        if temp_file_path then
-          body, err = file_reader(temp_file_path)
-          if err then
-            ngx.log(ngx.ERR, "HTTPS proxy: Failed to read temp body file, err: ", err)
-            ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
-          end
-
-          if encoding == "chunked" then
-            -- If the body is smaller than "client_boby_buffer_size" the Content-Length header is
-            -- set based on the size of the buffer. However, when the body is rendered to a file,
-            -- we will need to calculate and manually set the Content-Length header based on the
-            -- file size
-            local contentLength, err = file_size(temp_file_path)
-            if err then
-                ngx.log(ngx.ERR, "HTTPS proxy: Failed to set content length, err: ", err)
-                ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
-            end
-
-            ngx.req.set_header("Content-Length", tostring(contentLength))
-          end
-        end
+local function handle_expect()
+    local expect = ngx.req.get_headers()["Expect"]
+    if type(expect) == "table" then
+        expect = expect[1]
     end
 
-    -- The whole request is buffered in the memory so remove the Transfer-Encoding: chunked
-    if ngx.var.http_transfer_encoding == "chunked" then
-        ngx.req.set_header("Transfer-Encoding", nil)
+    if expect and expect:lower() == "100-continue" then
+        ngx.status = 100
+        local ok, err = ngx_send_headers()
+
+        if not ok then
+            return nil, "failed to send response header: " .. (err or "unknown")
+        end
+
+        ok, err = ngx_flush(true)
+        if not ok then
+            return nil, "failed to flush response header: " .. (err or "unknown")
+        end
+    end
+end
+
+local function forward_https_request(proxy_uri, uri, proxy_opts)
+    local body, err
+    local sock
+    local opts = proxy_opts or {}
+    local req_method = ngx_get_method()
+    local encoding = ngx.req.get_headers()["Transfer-Encoding"]
+    local is_chunked = encoding and encoding:lower() == "chunked"
+
+    if http_methods_with_body[req_method] then
+      if opts.request_unbuffered and ngx_http_version() == 1.1 then
+        local _, err = handle_expect()
+        if err then
+            ngx.log(ngx.ERR, "failed to handle expect header, err: ", err)
+            return ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
+        end
+
+        if is_chunked then
+            -- The default ngx reader does not support chunked request
+            -- so we will need to get the raw request socket and manually
+            -- decode the chunked request
+            sock, err = ngx.req.socket(true)
+        else
+            sock, err = ngx.req.socket()
+        end
+
+        if not sock then
+            ngx.log(ngx.ERR, "unable to obtain request socket: ", err)
+            return ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
+        end
+
+        body = client_body_reader(sock, DEFAULT_CHUNKSIZE, is_chunked)
+      else
+        -- This is needed to call ngx.req.get_body_data() below.
+        ngx.req.read_body()
+
+        -- We cannot use resty.http's .get_client_body_reader().
+        -- In POST requests with HTTPS, the result of that call is nil, and it
+        -- results in a time-out.
+        --
+        --
+        -- If ngx.req.get_body_data is nil, can be that the body is too big to
+        -- read and need to be cached in a local file. This request will return
+        -- nil, so after this we need to read the temp file.
+        -- https://github.com/openresty/lua-nginx-module#ngxreqget_body_data
+        body = ngx.req.get_body_data()
+
+        if not body then
+            local temp_file_path = ngx.req.get_body_file()
+            ngx.log(ngx.INFO, "HTTPS Proxy: Request body is bigger than client_body_buffer_size, read the content from path='", temp_file_path, "'")
+
+            if temp_file_path then
+              body, err = file_reader(temp_file_path)
+              if err then
+                ngx.log(ngx.ERR, "HTTPS proxy: Failed to read temp body file, err: ", err)
+                ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
+              end
+
+              if is_chunked then
+                -- If the body is smaller than "client_boby_buffer_size" the Content-Length header is
+                -- set based on the size of the buffer. However, when the body is rendered to a file,
+                -- we will need to calculate and manually set the Content-Length header based on the
+                -- file size
+                local contentLength, err = file_size(temp_file_path)
+                if err then
+                    ngx.log(ngx.ERR, "HTTPS proxy: Failed to set content length, err: ", err)
+                    ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
+                end
+
+                ngx.req.set_header("Content-Length", tostring(contentLength))
+              end
+            end
+        end
+
+        -- The whole request is buffered in the memory so remove the Transfer-Encoding: chunked
+        if is_chunked then
+            ngx.req.set_header("Transfer-Encoding", nil)
+        end
+      end
     end
 
     local request = {
@@ -139,10 +206,10 @@ local function forward_https_request(proxy_uri, proxy_auth, uri, skip_https_conn
         path = format('%s%s%s', ngx.var.uri, ngx.var.is_args, ngx.var.query_string or ''),
         body = body,
         proxy_uri = proxy_uri,
-        proxy_auth = proxy_auth
+        proxy_auth = opts.proxy_auth
     }
 
-    local httpc, err = http_proxy.new(request, skip_https_connect)
+    local httpc, err = http_proxy.new(request, opts.skip_https_connect)
 
     if not httpc then
         ngx.log(ngx.ERR, 'could not connect to proxy: ',  proxy_uri, ' err: ', err)
@@ -154,8 +221,16 @@ local function forward_https_request(proxy_uri, proxy_auth, uri, skip_https_conn
     res, err = httpc:request(request)
 
     if res then
-        httpc:proxy_response(res)
-        httpc:set_keepalive()
+        if opts.request_unbuffered and is_chunked then
+            local bytes, err = send_response(sock, res, DEFAULT_CHUNKSIZE)
+            if not bytes then
+                ngx.log(ngx.ERR, "failed to send response: ", err)
+                return sock:send("HTTP/1.1 502 Bad Gateway")
+            end
+        else
+            httpc:proxy_response(res)
+            httpc:set_keepalive()
+        end
     else
         ngx.log(ngx.ERR, 'failed to proxy request to: ', proxy_uri, ' err : ', err)
         return ngx.exit(ngx.HTTP_BAD_GATEWAY)
@@ -208,7 +283,13 @@ function _M.request(upstream, proxy_uri)
         return
     elseif uri.scheme == 'https' then
         upstream:rewrite_request()
-        forward_https_request(proxy_uri, proxy_auth, uri, upstream.skip_https_connect)
+        local proxy_opts = {
+            proxy_auth = proxy_auth,
+            skip_https_connect = upstream.skip_https_connect,
+            request_unbuffered = upstream.request_unbuffered
+        }
+
+        forward_https_request(proxy_uri, uri, proxy_opts)
         return ngx.exit(ngx.OK) -- terminate phase
     else
         ngx.log(ngx.ERR, 'could not connect to proxy: ',  proxy_uri, ' err: ', 'invalid request scheme')
--- a/gateway/src/apicast/http_proxy.lua
+++ b/gateway/src/apicast/http_proxy.lua
@@ -170,7 +170,7 @@ local function forward_https_request(proxy_uri, uri, proxy_opts)
                 -- set by openresty based on the size of the buffer. However, when the body is rendered
                 -- to a file, we will need to calculate and manually set the Content-Length header based
                 -- on the file size
-                local contentLength, err = file_size(temp_file_path)
+                local contentLength, err = file_size(temp_file_path)()
                 if err then
                     ngx.log(ngx.ERR, "HTTPS proxy: Failed to set content length, err: ", err)
                     ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
--- a/gateway/src/apicast/http_proxy.lua
+++ b/gateway/src/apicast/http_proxy.lua
@@ -1,10 +1,12 @@
 local format = string.format
+local tostring = tostring
 
 local resty_url = require "resty.url"
 local resty_resolver = require 'resty.resolver'
 local round_robin = require 'resty.balancer.round_robin'
 local http_proxy = require 'resty.http.proxy'
 local file_reader = require("resty.file").file_reader
+local file_size = require("resty.file").file_size
 local concat = table.concat
 
 local _M = { }
@@ -86,39 +88,59 @@ local function forward_https_request(proxy_uri, proxy_auth, uri, skip_https_conn
     -- This is needed to call ngx.req.get_body_data() below.
     ngx.req.read_body()
 
+    -- We cannot use resty.http's .get_client_body_reader().
+    -- In POST requests with HTTPS, the result of that call is nil, and it
+    -- results in a time-out.
+    --
+    --
+    -- If ngx.req.get_body_data is nil, can be that the body is too big to
+    -- read and need to be cached in a local file. This request will return
+    -- nil, so after this we need to read the temp file.
+    -- https://github.com/openresty/lua-nginx-module#ngxreqget_body_data
+    local body = ngx.req.get_body_data()
+    local encoding = ngx.req.get_headers()["Transfer-Encoding"]
+
+    if not body then
+        local temp_file_path = ngx.req.get_body_file()
+        ngx.log(ngx.INFO, "HTTPS Proxy: Request body is bigger than client_body_buffer_size, read the content from path='", temp_file_path, "'")
+
+        if temp_file_path then
+          body, err = file_reader(temp_file_path)
+          if err then
+            ngx.log(ngx.ERR, "HTTPS proxy: Failed to read temp body file, err: ", err)
+            ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
+          end
+
+          if encoding == "chunked" then
+            -- If the body is smaller than "client_boby_buffer_size" the Content-Length header is
+            -- set based on the size of the buffer. However, when the body is rendered to a file,
+            -- we will need to calculate and manually set the Content-Length header based on the
+            -- file size
+            local contentLength, err = file_size(temp_file_path)
+            if err then
+                ngx.log(ngx.ERR, "HTTPS proxy: Failed to set content length, err: ", err)
+                ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
+            end
+
+            ngx.req.set_header("Content-Length", tostring(contentLength))
+          end
+        end
+    end
+
+    -- The whole request is buffered in the memory so remove the Transfer-Encoding: chunked
+    if ngx.var.http_transfer_encoding == "chunked" then
+        ngx.req.set_header("Transfer-Encoding", nil)
+    end
+
     local request = {
         uri = uri,
         method = ngx.req.get_method(),
         headers = ngx.req.get_headers(0, true),
         path = format('%s%s%s', ngx.var.uri, ngx.var.is_args, ngx.var.query_string or ''),
-
-        -- We cannot use resty.http's .get_client_body_reader().
-        -- In POST requests with HTTPS, the result of that call is nil, and it
-        -- results in a time-out.
-        --
-        --
-        -- If ngx.req.get_body_data is nil, can be that the body is too big to
-        -- read and need to be cached in a local file. This request will return
-        -- nil, so after this we need to read the temp file.
-        -- https://github.com/openresty/lua-nginx-module#ngxreqget_body_data
-        body = ngx.req.get_body_data(),
+        body = body,
         proxy_uri = proxy_uri,
         proxy_auth = proxy_auth
     }
-
-    if not request.body then
-        local temp_file_path = ngx.req.get_body_file()
-        ngx.log(ngx.INFO, "HTTPS Proxy: Request body is bigger than client_body_buffer_size, read the content from path='", temp_file_path, "'")
-
-        if temp_file_path then
-          local body, err = file_reader(temp_file_path)
-          if err then
-            ngx.log(ngx.ERR, "HTTPS proxy: Failed to read temp body file, err: ", err)
-            ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
-          end
-          request.body = body
-        end
-    end
 
     local httpc, err = http_proxy.new(request, skip_https_connect)
 
--- a/gateway/src/apicast/http_proxy.lua
+++ b/gateway/src/apicast/http_proxy.lua
@@ -98,27 +98,6 @@ local function absolute_url(uri)
     )
 end
 
-local function handle_expect()
-    local expect = ngx.req.get_headers()["Expect"]
-    if type(expect) == "table" then
-        expect = expect[1]
-    end
-
-    if expect and expect:lower() == "100-continue" then
-        ngx.status = 100
-        local ok, err = ngx_send_headers()
-
-        if not ok then
-            return nil, "failed to send response header: " .. (err or "unknown")
-        end
-
-        ok, err = ngx_flush(true)
-        if not ok then
-            return nil, "failed to flush response header: " .. (err or "unknown")
-        end
-    end
-end
-
 local function forward_https_request(proxy_uri, uri, proxy_opts)
     local body, err
     local sock
@@ -139,11 +118,6 @@ local function forward_https_request(proxy_uri, uri, proxy_opts)
       -- socket but hang on read in case of raw socket. Therefore, we only retrieve body from the
       -- socket if the content type is not "application/x-www-form-urlencoded"
       if opts.request_unbuffered and ngx_http_version() == 1.1 and not content_type_is_urlencoded then
-        local _, err = handle_expect()
-        if err then
-            ngx.log(ngx.ERR, "failed to handle expect header, err: ", err)
-            return ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
-        end
         if is_chunked then
             -- The default ngx reader does not support chunked request
             -- so we will need to get the raw request socket and manually
--- a/gateway/src/apicast/http_proxy.lua
+++ b/gateway/src/apicast/http_proxy.lua
@@ -126,20 +126,30 @@ local function forward_https_request(proxy_uri, uri, proxy_opts)
     local req_method = ngx_get_method()
     local encoding = ngx.req.get_headers()["Transfer-Encoding"]
     local is_chunked = encoding and encoding:lower() == "chunked"
+    local content_type = ngx.req.get_headers()["Content-Type"]
+    local content_type_is_urlencoded = content_type and content_type:lower() == "application/x-www-form-urlencoded"
+    local raw = false
 
     if http_methods_with_body[req_method] then
-      if opts.request_unbuffered and ngx_http_version() == 1.1 then
+
+      -- When the content type is "application/x-www-form-urlencoded" the body is always pre-read.
+      -- See: gateway/src/apicast/configuration/service.lua:214
+      --
+      -- Due to this, ngx.req.socket() will fail with "request body already exists" error or return
+      -- socket but hang on read in case of raw socket. Therefore, we only retrieve body from the
+      -- socket if the content type is not "application/x-www-form-urlencoded"
+      if opts.request_unbuffered and ngx_http_version() == 1.1 and not content_type_is_urlencoded then
         local _, err = handle_expect()
         if err then
             ngx.log(ngx.ERR, "failed to handle expect header, err: ", err)
             return ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
         end
-
         if is_chunked then
             -- The default ngx reader does not support chunked request
             -- so we will need to get the raw request socket and manually
             -- decode the chunked request
             sock, err = ngx.req.socket(true)
+            raw = true
         else
             sock, err = ngx.req.socket()
         end
@@ -151,6 +161,11 @@ local function forward_https_request(proxy_uri, uri, proxy_opts)
 
         body = client_body_reader(sock, DEFAULT_CHUNKSIZE, is_chunked)
       else
+        -- TODO: Due to ngx.req.read_body(). The current implementation will not work with grpc service
+        -- See: https://github.com/3scale/APIcast/pull/1419
+        -- Should we get the body from socket by default and only read buffered body if
+        -- "Content-Type: application/x-www-form-urlencoded"?
+        --
         -- This is needed to call ngx.req.get_body_data() below.
         ngx.req.read_body()
 
@@ -178,9 +193,9 @@ local function forward_https_request(proxy_uri, uri, proxy_opts)
 
               if is_chunked then
                 -- If the body is smaller than "client_boby_buffer_size" the Content-Length header is
-                -- set based on the size of the buffer. However, when the body is rendered to a file,
-                -- we will need to calculate and manually set the Content-Length header based on the
-                -- file size
+                -- set by openresty based on the size of the buffer. However, when the body is rendered
+                -- to a file, we will need to calculate and manually set the Content-Length header based
+                -- on the file size
                 local contentLength, err = file_size(temp_file_path)
                 if err then
                     ngx.log(ngx.ERR, "HTTPS proxy: Failed to set content length, err: ", err)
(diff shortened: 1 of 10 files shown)


gateway/src/resty/http/chunked.lua
```diff
@@ -0,0 +1,161 @@
+local co_wrap_iter = require("resty.coroutines").co_wrap_iter
```
kevprice83, line 1: Why are we reimplementing the library here? Surely we should leverage the functions available in [lua-resty-http](https://github.com/ledgetech/lua-resty-http/blob/master/lib/resty/http.lua#L971-L1001)?

kevprice83 (changes requested):
We need to add unit tests and integration tests for all the scenarios that reproduce the reported issue:

- http proxy env vars (over TLS)
- http_proxy policy (over TLS)
- camel_proxy policy (over TLS)

gateway/src/resty/http/chunked.lua
tkan145: @kevprice83 done with the unittest.

because the lua-resty-http is writtent as a decoder and only send the data body
https://github.com/3scale/APIcast/blob/master/gateway/src/apicast/http_proxy.lua#L168

t/http-proxy.t
```diff
@@ -15,7 +15,7 @@ $ENV{'LARGE_BODY'} = large_body();
 
 require("policies.pl");
 
-repeat_each(3);
+repeat_each(1);
```
tkan145, line 18: FIXME: Need to change this back to 3

gateway/src/resty/http/chunked.lua
eguzki: Sorry, I do not get that. https://github.com/3scale/APIcast/blob/master/gateway/src/apicast/http_proxy.lua#L168 is only about rewriting request host header.

gateway/src/apicast/http_proxy.lua
```diff
@@ -82,14 +88,58 @@ local function absolute_url(uri)
...
+        sock, err = ngx.req.socket(true)
+
+        if not sock then
+            if err == "no body" then
+                body = nil
+            else
+                return ngx.exit(ngx.HTTP_SERVICE_UNAVAILABLE)
+            end
+        else
+            body = chunked_reader(sock, chunksize)
```
eguzki, line 138: This method reads the entire body? In memory? What if body size is Gb length?

t/http-proxy.t
```diff
@@ -1227,3 +1227,539 @@ proxy request: CONNECT test-upstream.lvh.me:$TEST_NGINX_RANDOM_PORT HTTP/1.1
 --- no_error_log
 [error]
 --- user_files fixture=tls.pl eval
+
```
eguzki, line 1230: Tests with chunked requests with http_proxy policy are missing, right?

gateway/src/resty/http/chunked.lua
tkan145: Sorry I put wrong link there, I mean this one https://github.com/ledgetech/lua-resty-http/blob/master/lib/resty/http.lua#L480

gateway/src/apicast/http_proxy.lua
tkan145: it returns an iterator which can be used by lua-resty-http to read the body in chunk

eguzki: So, if it is an iterator, the request is not buffered. Then, APIcast will open a connection (or reuse existing one) as soon as there is some bytes to read (first chunk). Then, APIcast is not protecting upstream from Slowloris DDoS. Am I right?

t/http-proxy.t
eguzki: Those tests probably belong to `t/apicast-policy-http-proxy.t`

gateway/src/apicast/http_proxy.lua
```diff
@@ -130,11 +187,20 @@ local function forward_https_request(proxy_uri, uri, skip_https_connect)
     res, err = httpc:request(request)
 
     if res then
-        httpc:proxy_response(res)
+        -- if we are using raw socket we will need to send the response back with sock:send
+        if sock then
+            chunked_writer(sock, res ,chunksize)
```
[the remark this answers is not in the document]
tkan145: Good catch. I think I picked a bad name, all it does is get the body reader from the respond and send it back to the client. I will fix this in the next patch

gateway/src/apicast/http_proxy.lua
tkan145: Yes, you are right, maybe we need a better solution, any ideas?

gateway/src/resty/http/chunked.lua
eguzki: Ok, I see. I found this PR [Add support for chunked body to _send_body #285](https://github.com/ledgetech/lua-resty-http/pull/285) as alternative. I also like the idea of the our own wrapper while not in lua-resty-http

tkan145: Added tests for:
* http proxy env vars (over TLS)
* http_proxy policy (over TLS)
* camel_proxy policy (over TLS)

gateway/src/apicast/http_proxy.lua
```diff
@@ -132,8 +221,16 @@ local function forward_https_request(proxy_uri, proxy_auth, uri, skip_https_conn
     res, err = httpc:request(request)
 
     if res then
-        httpc:proxy_response(res)
-        httpc:set_keepalive()
+        if opts.request_unbuffered and is_chunked then
```
eguzki, line 224: Sorry, I do not get this. 

What is the relationship between request buffering option and response? Shouldn't the response be managed in the same way not matter existence of the unbuffered policy or being chunked request or not? i.e.

```
httpc:proxy_response(res)
httpc:set_keepalive()
```

gateway/src/resty/http/response_writer.lua
```diff
@@ -0,0 +1,58 @@
...
+        ngx.log(ngx.DEBUG, 'skipping sending nil')
+        return
+    end
+
+    return socket:send(data)
+end
+
+-- write_response writes response body reader to sock in the HTTP/1.x server response format,
+-- The connection is closed if send() fails or when returning a non-zero
+function _M.send_response(sock, response, chunksize)
```
eguzki, line 17: related to my previous question regarding handling of the response: what is this method doing?

eguzki (commented):
Looking good. 

However, this is complex and also hard to maintain. I want to try some other approach relying on lua-resty-http to 0.17.1 or some other library. If we cannot find a simpler (from the APIcast base code perspective), we can always use this code.

eguzki:
When using the python client, the response body is not shown. I wonder if APIcast is not handling the response correctly

eguzki:
I am going to try https://github.com/ledgetech/lua-resty-http#set_proxy_options looks promising

gateway/src/apicast/http_proxy.lua
```diff
@@ -82,15 +98,61 @@ local function absolute_url(uri)
...
+    local body, err
+    local sock
+    local opts = proxy_opts or {}
+    local req_method = ngx_get_method()
+    local encoding = ngx.req.get_headers()["Transfer-Encoding"]
+    local is_chunked = encoding and encoding:lower() == "chunked"
+
+    if http_methods_with_body[req_method] then
+      if opts.request_unbuffered and ngx_http_version() == 1.1 then
+        local _, err = handle_expect()
```
eguzki, line 132: I wonder if `Expect` needs to be handled for when buffering is enabled

eguzki: Actually, I do not think we should be doing this. The lua-resty-http lib is doing that for us. WDYT?

gateway/src/resty/http/request_reader.lua
```diff
@@ -0,0 +1,47 @@
...
+local cr_lf = "\r\n"
+
+-- chunked_reader return a body reader that translates the data read from
+-- lua-resty-http client_body_reader to HTTP "chunked" format before returning it
+--
+-- The chunked reader return nil when the final 0-length chunk is read
+local function chunked_reader(sock, chunksize)
+    chunksize = chunksize or 65536
+    local eof = false
+    local reader = httpc:get_client_body_reader(chunksize, sock)
```
eguzki, line 15: which method is `httpc:get_client_body_reader`?  I cannot find it

gateway/src/apicast/http_proxy.lua
```diff
@@ -101,26 +163,53 @@ local function forward_https_request(proxy_uri, proxy_auth, uri, skip_https_conn
...
+            if temp_file_path then
+              body, err = file_reader(temp_file_path)
+              if err then
+                ngx.log(ngx.ERR, "HTTPS proxy: Failed to read temp body file, err: ", err)
+                ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
+              end
+
+              if is_chunked then
+                -- If the body is smaller than "client_boby_buffer_size" the Content-Length header is
+                -- set based on the size of the buffer. However, when the body is rendered to a file,
```
eguzki, line 181: > If the body is smaller than "client_boby_buffer_size" the Content-Length header is set based on the size of the buffer

Who is doing that? In other words, when all the conditions meet:
* the request is chunked,
* buffering is enabled
* the request body is small

Who sets the `Content-Length` header?

gateway/src/apicast/http_proxy.lua
tkan145: lib lua-resty-http is a client library and it handles the Expect returned from the server, while we are acting as a server here and need to process the Expect header from the client.

I will run more tests

gateway/src/apicast/http_proxy.lua
tkan145: This is a bit tricky, after retrieve the raw socket, the above 2 methods will stop working, hence I need to send response back using the same raw socket

And we only use raw socket, when we have request_buffering option on

gateway/src/resty/http/request_reader.lua
tkan145: if I read the code correct, we overwrite the lua-resty-http connect method in resty/resolver/http.lua and set ```
local _M = setmetatable({}, { __index = resty_http })
```

so `httpc:get_client_body_reader` will return the method from lua-resty-http lib.

File: gateway/src/apicast/http_proxy.lua (hunk 10 above)
Comment (tkan145) [reply gateway/src/apicast/http_proxy.lua:181]: The lua-resty-http will set the `Content-Length` based on the body that we passed in. But good catch I should have put more details in the comment

**Comment by tkan145** [comment]
> I am going to try https://github.com/ledgetech/lua-resty-http#set_proxy_options looks promising

I tried it in #1434 in /http/proxy.lua

File: gateway/src/apicast/http_proxy.lua (hunk 8 above)
Comment (eguzki) [reply gateway/src/apicast/http_proxy.lua:132]: I think that when buffered, APIcast should protect upstream and should handle the `Expect: 100-Continue`. That is, it is the apicast who returns HTTP Response `100 Continue` and then consumes the body before opening the connection to upstream.

When unbuffered, the `Expect: 100-Continue` must be propagated to upstream and it is upstream who decides on responding HTTP Response `100 Continue`. And the lua resty http lib is handling that already, i.e. waiting for the `100 Continue` from upstream before (reading the request body) and sending the body.

I learned, if I am not wrong, that the current `master` branch is actually doing unbuffered approach when there is `Expected: 100 Continue` header. Because it opens connection to upstream and waits for `100 Continue` from upstream, and then it starts reading (from a body reader callback) the request body and sends it. Data is being sent to upstream as soon as it is available. Do you see this behavior as I do?

Comment (eguzki) [reply gateway/src/apicast/http_proxy.lua:132]: I heave re-written the message above. In case you have read it previosly, please re-read it again :pray:

Comment (tkan145) [reply gateway/src/apicast/http_proxy.lua:132]: I'm a bit confused here. I haven't read the openresty code but do you mean `ngx.req.read_body()` will send `100 Continue` downstream? Doesn't that also mean that APIcast returns `100 Continue` to the downstream application before establishing the upstream connection?

Regarding the 400, please correct me if I'm wrong, but I think the only case where the upstream server returns this error is if there is data in the request body. In my head the flow will be as follow

```
client -> Expect: 100-Continue -> upstream -> 100 Continue -> client
client -> start sending body -> upstream read body -> return 400
```

Comment (eguzki) [reply gateway/src/apicast/http_proxy.lua:132]: >  I haven't read the openresty code but do you mean ngx.req.read_body() will send 100 Continue downstream?

Yes!

> Doesn't that also mean that APIcast returns 100 Continue to the downstream application before establishing the upstream connection?

Exactly (when buffered mode is on)

Comment (tkan145) [reply gateway/src/apicast/http_proxy.lua:132]: From this nginx thread https://mailman.nginx.org/pipermail/nginx/2021-May/060643.html. I think nginx does not handle this well either

How about we send back error response, discard the body and close the socket?

Comment (eguzki) [reply gateway/src/apicast/http_proxy.lua:132]: > How about we send back error response, discard the body and close the socket?

It's aggressive, but can be a way out.

File: gateway/src/resty/http/request_reader.lua (hunk 9 above)
Comment (eguzki) [reply gateway/src/resty/http/request_reader.lua:15]: Why not require `local resty_http = require 'resty.http'` instead to make more explicit? the `get_client_body_reader` is a public exposed global function, not a method of any module table. Whatever you prefer.

File: gateway/src/apicast/http_proxy.lua (hunk 10 above)
Comment (eguzki) [reply gateway/src/apicast/http_proxy.lua:181]: Ok, I see,  It's because it is a string and the resty-http gets the length out of it. It happens [here](https://github.com/ledgetech/lua-resty-http/blob/v0.15/lib/resty/http.lua#L622-L627). I would make it explicit, but good enough.

File: gateway/src/apicast/http_proxy.lua (hunk 6 above)
Comment (eguzki) [reply gateway/src/apicast/http_proxy.lua:224]: the `proxy_response` method has been depreated. https://github.com/ledgetech/lua-resty-http#deprecated
So, we may need to implement this anyway. Maybe for the refactor to resty_http v0.17.0 PR.

**Comment by eguzki** [comment]
I am ok going with these changes for now. First https://github.com/3scale/APIcast/pull/1403#issuecomment-1856152114 needs to be fixed. 

However, in following up PR's, this (spaghetti) code needs some simplification. It is hard to maintain and understand. The `forward_https_request` method should implement the diagram below

```ascii
                                               ┌────────────────────────┐
                       YES                     │                        │                   NO
                                               │                        │
      ┌────────────────────────────────────────┤                        ├────────────────────────────────────────────┐
      │                                        │   Request Buffering?   │                                            │
      │                                        │                        │                                            │
      │                                        │                        │                                            │
      │                                        │                        │                                            │
      │                                        └────────────────────────┘                                            │
      │                                                                                                              │
      │                                                                          ┌───────────────────────────────────▼───────────────────────────────┐
      │                                                                          │                                                                   │
      │                                                                          │                                                                   │
      │                                                                          │                                                                   │
      ▼                                                                          │            Set up body reader from downstream socket              │
┌──────────────┐                                                                 │                                                                   │
│              │                                                                 │                                                                   │
│              │                                                                 │                                                                   │
│  Read body   │                                                                 │            Handle Expect: 100-Continue                            │
│              │                                                                 │                                                                   │
└────┬─────────┘                                                                 │            For Transfer-Encoding: chunked,                        │
     │                                                                           │                                                                   │
     │                                                                           │              set wrapper to encode (back) the request body        │
     │                                                                           │                                                                   │
     │                                                                           └──────────────────────────────────┬────────────────────────────────┘
┌────▼────────────────┐                                                                                             │
│                     │                                                                                             │
│ Set Content-Length  │                                                                                             │
│                     ├──────────────────────────────────┐                                                          │
│ Remove TE header    │                                  │                                                          │
│                     │                                  │                                                          │
└─────────────────────┘                                  │                                                          │
                                                         │                                                          │
                                                         ◄──────────────────────────────────────────────────────────┘
                                                         │
                                                         ▼
                                           ┌─────────────▼──────────────┐
                                           │                            │
                                           │   Connect  TLS Upstream    │
                                           │                            │
                                           │   via proxy                │
                                           │                            │
                                           │                            │
                                           └──────────────┬─────────────┘
                                                          │
                                                          │
                                                          │
                                           ┌──────────────▼──────────────┐
                                           │                             │
                                           │   Send Request              │
                                           │                             │
                                           │                             │
                                           └────────────────┬────────────┘
                                                            │
                                                            │
                                                            │
                                                            ▼
                                           ┌──────────────────────────────┐
                                           │                              │
                                           │     Write Downstream response│
                                           │                              │
                                           └──────────────────────────────┘
```

tkan145:
Fixed https://github.com/3scale/APIcast/pull/1403#issuecomment-1856152114

Also added 2 commits on top to handle `Expect: 100-continue` and `Content-Type: application/x-www-form-urlencoded`

eguzki (approved):
This was a hard one, wasn't it!

Great job :medal_military:

gateway/src/apicast/http_proxy.lua
kevprice83: Yeah I agree because this is something that will come up again in future when troubleshooting but it doesn't need to be done in this PR, can be added at a later date that `if headers["Content-Length"]=nil then headers["Content-Length"]=#body` (this will at least be a useful reference for now)

kevprice83 (changes requested):
I think I might need a summary on what is the expected behaviour with the different combinations here to understand how this works. I am not sure I get this to be honest which is probably expected but then I would expect the README to describe in more detail so customers and Support can understand better.

If the tests are indeed correct then only an update to the README is needed.

One note though; why are there no unit tests added/modified?

Final comment: I think we are introducing I/O blocking operations via the `file_size()` function and this is going to be executed on every request that meets the conditions. Executing things like `os.execute`, `io.open` etc would generally be safe in the `init` and `init_by_worker` phases because it's a one-time execution but in this scenario we are doing it for every request. Have we considered this already? How much is it harming performance as a result?

gateway/src/apicast/http_proxy.lua
```diff
@@ -101,26 +152,55 @@ local function forward_https_request(proxy_uri, proxy_auth, uri, skip_https_conn
...
+                ngx.log(ngx.ERR, "HTTPS proxy: Failed to read temp body file, err: ", err)
+                ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
+              end
+
+              if is_chunked then
+                -- If the body is smaller than "client_boby_buffer_size" the Content-Length header is
+                -- set by openresty based on the size of the buffer. However, when the body is rendered
+                -- to a file, we will need to calculate and manually set the Content-Length header based
+                -- on the file size
+                local contentLength, err = file_size(temp_file_path)
```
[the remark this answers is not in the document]
tkan145: Is it enough to wrap that functionality with a coroutine? I don't know how useful that would be since it would yield on the first call anyway. Also the [file_reader](https://github.com/3scale/APIcast/blob/master/gateway/src/resty/file.lua#L14) also call `io.open` all every request that has body buffered to file, so I guess we pay the price of calling `io.open` one more time?

But I totally agree with you that it is a I/O blocking function and should be avoided. 

Checking the module `lua-io-nginx` I can see that this module is currently considered experimental. And it seems like it runs the task on another thread. However, I'm not so sure about this because we have to pay for context switching, threads, locking, etc.

```
It's worth to mention that the cost time of a single I/O operation won't be reduced, it was just
transferred from the main thread (the one executes the event loop) to another exclusive thread.
Indeed, the overhead might be a little higher, because of the extra tasks transferring, lock waiting,
Lua coroutine resumption (and can only be resumed in the next event loop) and so forth. Nevertheless,
after the offloading, the main thread doesn't block due to the I/O operation, and this is the fundamental
advantage compared with the native Lua I/O library.
```

t/apicast-policy-camel.t
```diff
@@ -506,3 +516,709 @@ using proxy: http://foo:bar\@127.0.0.1:$Test::Nginx::Util::PROXY_SSL_PORT,
...
+[error]
+--- grep_error_log eval
+qr/a client request body is buffered to a temporary file/
+--- grep_error_log_out
+a client request body is buffered to a temporary file
+
+
+
+=== TEST 15: https_proxy with request_unbuffered policy, only upstream and proxy_pass will buffer
+the request
```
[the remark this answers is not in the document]
tkan145: it works like this

## Without request_buffering policy (default behavior)

```
HTTPS request ---> APIcast ---> [reach proxy code] ---> [call ngx.req.get_body_data()] ---> [ request is buffered to a file] ---> [construct a new request and a body reader from buffered file] ---> [ use lua-resty-http to perform handshake and send new request to camel server]  ---> Camel
```

## With request_buffering policy

```
HTTPS request ---> APIcast  ---> [reach proxy code] ---> [construct a new request and set up body reader from downstream socket (without reading the body first)] ---> [ use lua-resty-http to perform handshake and send new request to camel server]  ---> Camel
```

Upstream here is the `upstream` block in the test. It was a bit tricky to setup a test for this so I relied on `a client request body is buffered to a temporary file` in the log. 

* `proxy_pass` will buffer the request body by default
* `echo_read_request_body` will also buffer the request. `echo_request_body` used to echo the request body so we can check if the request was sent to the upstream server

But I will update README file with more details

gateway/src/apicast/http_proxy.lua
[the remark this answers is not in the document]
eguzki: No sure how expensive is 

```lua
function fsize (filename)
      local handle, err = open(filename)
      local current = handle:seek()      -- get current position
      local size = handle:seek("end")    -- get file size
      handle:seek("set", current)        -- restore position
      return size
    end
```

Theoretically any IO operation could block the thread. We could try coroutines or any other mean to make it non blocking. Reading [lua-nginx-module introduction](https://github.com/openresty/lua-nginx-module#introduction) it says:
```
Disk operations with relatively small amount of data can be done using the standard Lua io library but huge file reading and writing should be avoided wherever possible as they may block the Nginx process significantly. Delegating all network and disk I/O operations to Nginx's subrequests (via the [ngx.location.capture](https://github.com/openresty/lua-nginx-module#ngxlocationcapture) method and similar) is strongly recommended for maximum performance.
```
Not sure if we can follow that recommendation, though. @tkan145 we can try to discuss about this in a short call..

Anyway, AFAIK, we have never measured the capacity of APICast to handle traffic with request body big enough to be persisted in disk. All the tests performed where "simple" GET requests.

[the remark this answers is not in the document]
tkan145: Fixed and also updated README file. @kevprice83 can you help review the README file and let me know if I need to add anything else?

gateway/src/apicast/policy/request_unbuffered/README.md
```diff
@@ -1,19 +1,99 @@
...
 
 ## Caveats
 
-- Because APIcast allows defining mapping rules based on request content, POST requests with
-  `Content-type: application/x-www-form-urlencoded` will always be buffered regardless of the
+- Because APIcast allows defining mapping rules based on request content, ie `POST /some_path?a_param={a_value}`
+  will match a request like `POST "http://apicast_host:8080/some_path"` with a form URL-encoded body like: `a_param=abc`
+  , requests with `Content-type: application/x-www-form-urlencoded` will always be buffered regardless of the
   `request_unbuffered` policy is enabled or not.
+- For a request with "small" body that fits into [`client_body_buffer_size`](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_body_buffer_size) and with header "Transfer-Encoding: chunked", NGINX will always read and know the length of the body.
```
eguzki, line 96: This is nice to note, but it is not a caveat (a limitation), it is expected and correct. All about unbuffered request is the fact, as you correctly pointed out, that `the request body will be sent to the proxied server immediately as it received`. The transfer encoding is HOP-BY-HOP encoding and nothing prevents from changing it from one hop to the next one.

gateway/src/apicast/policy/request_unbuffered/README.md
```diff
@@ -1,19 +1,99 @@
...
+
+
+## Technical details
+
+By default, NGINX reads the entire request body into memory (or buffers large requests into disk) before proxying it to the upstream server. However, reading bodies can become expensive, especially when requests with large payloads are sent.
+
+For example, when the client sends 10GB, NGINX will buffer the entire 10GB to disk before sending anything to
+the upstream server.
+
+When `proxy_request_buffering` is in the chain, request buffering will be disabled and the request body will be sent to the proxied server immediately as it received. This can help minimize time spent sending data to a service and disk I/O for requests with big body. However, there are caveats and corner cases applied, [**Caveats**](#caveats)
```
eguzki, line 16: `proxy_request_buffering` is not the name of the policy, is it?

tkan145: oops

eguzki:
I kindly requested review to @3scale/documentation team

gateway/src/apicast/policy/request_unbuffered/README.md
```diff
@@ -1,14 +1,99 @@
...
-This policy allows to disable request buffering
+## Description
+
+When enable this policy will dymanically sets the [`proxy_request_buffering: off`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_request_buffering
+) directive per service.
+
+
+## Technical details
+
+By default, NGINX reads the entire request body into memory (or buffers large requests into disk) before proxying it to the upstream server. However, reading bodies can become expensive, especially when requests with large payloads are sent.
```
dfennessy, line 11:
```suggestion
By default, NGINX  reads the entire  request body into memory or buffers large requests to disk before forwarding them to the upstream server. Reading bodies, can become expensive, especially when sending requests containing large payloads.
```

gateway/src/apicast/policy/request_unbuffered/README.md
```diff
@@ -1,14 +1,99 @@
...
+
+For example, when the client sends 10GB, NGINX will buffer the entire 10GB to disk before sending anything to
+the upstream server.
+
+When `proxy_request_buffering` is in the chain, request buffering will be disabled and the request body will be sent to the proxied server immediately as it received. This can help minimize time spent sending data to a service and disk I/O for requests with big body. However, there are caveats and corner cases applied, [**Caveats**](#caveats)
```
dfennessy, line 16:
```suggestion
When the proxy_request_buffering is in the chain, it disables request buffering, sending the request body to the proxied server immediately upon receiving it. This can help minimize time spent sending data to a service and disk I/O for requests with big body. However, there are caveats and corner cases applied, [**Caveats**](#caveats)
```

gateway/src/apicast/policy/request_unbuffered/README.md
```diff
@@ -1,14 +1,99 @@
...
+- APIcast <> HTTP Proxy (policy) <> upstream TLS
+- APIcast <> HTTP Proxy (camel proxy) <> upstream TLS
+```
+
+## Why don't we also support disable response buffering?
+
+The response buffering is enabled by default in NGINX (the [`proxy_buffering: on`]() directive). It does
+this to shield the backend against slow clients ([slowloris attack](https://en.wikipedia.org/wiki/Slowloris_(computer_security))).
+
+If the `proxy_buffering` is disabled, the upstream server will be forced to keep the connection open until all data has been received by the client. Thereforce, NGINX [advises](https://www.nginx.com/blog/avoiding-top-10-nginx-configuration-mistakes/#proxy_buffering-off) against disabling `proxy_buffering` as it will potentially waste upstream server resources.
```
dfennessy, line 36:
```suggestion
If the `proxy_buffering` is disabled, the upstream server keeps the connection open until all data is received by the client. NGINX [advises](https://www.nginx.com/blog/avoiding-top-10-nginx-configuration-mistakes/#proxy_buffering-off) against disabling `proxy_buffering` as it will potentially waste upstream server resources.
```

gateway/src/apicast/policy/request_unbuffered/README.md
```diff
@@ -1,14 +1,99 @@
...
+          "https_proxy": "http://192.168.15.103:8443/",
+          "all_proxy": "http://192.168.15.103:8080/"
+      }
+    }
+]
+```
+
+## Caveats
+
+- Because APIcast allows defining mapping rules based on request content, ie `POST /some_path?a_param={a_value}`
```
dfennessy, line 92:
```suggestion
- APIcast allows defining of mapping rules based on request content. For example, `POST /some_path?a_param={a_value}`
```

dfennessy (changes requested):
I've added a few suggestions.

eguzki:
After running again the verification steps, The first  request with small body (chunked TE), reaches upstream still with the TE chunked. And I think this is expected. As for this use case (TE chunked and TLS upstream through proxy) APIcast generates a socket reader that decodes the chunk encoding and re-encodes it back to be forwarded. So the request reaching upstream has to be TE chunked. 

```
❯ curl --resolve post.example.com:8080:127.0.0.1 -v -H "Transfer-Encoding: chunked"   -H "Content-Type: application/json"  -d @my-data.json "http://post.example.com:8080/?user_key=123"
Warning: Couldn't read data from file "my-data.json", this makes an empty 
Warning: POST.
* Added post.example.com:8080:127.0.0.1 to DNS cache
* Hostname post.example.com was found in DNS cache
*   Trying 127.0.0.1:8080...
* Connected to post.example.com (127.0.0.1) port 8080 (#0)
> POST /?user_key=123 HTTP/1.1
> Host: post.example.com:8080
> User-Agent: curl/7.81.0
> Accept: */*
> Transfer-Encoding: chunked
> Content-Type: application/json
> 
* upload completely sent off: 5 out of 0 bytes
< HTTP/1.1 200 OK
< Date: Mon, 22 Jan 2024 09:14:35 GMT
< Server: gunicorn/19.9.0
* no chunk, no close, no size. Assume close to signal end
< 
{
  "args": {
    "user_key": "123"
  }, 
  "data": "", 
  "files": {}, 
  "form": {}, 
  "headers": {
    "Accept": "*/*", 
    "Content-Type": "application/json", 
    "Host": "example.com", 
    "Transfer-Encoding": "chunked", 
    "User-Agent": "curl/7.81.0"
  }, 
  "json": null, 
  "origin": "172.18.0.4", 
  "url": "http://example.com/post?user_key=123"
}
* Closing connection 0
```
You might want to update verification steps.

gateway/src/apicast/policy/request_unbuffered/README.md
```diff
@@ -1,14 +1,97 @@
...
+
+## Why don't we also support disable response buffering?
+
+The response buffering is enabled by default in NGINX (the [`proxy_buffering: on`]() directive). It does this to shield the backend against slow clients ([slowloris attack](https://en.wikipedia.org/wiki/Slowloris_(computer_security))).
+
+If the `proxy_buffering` is disabled, the upstream server keeps the connection open until all data is received by the client. NGINX [advises](https://www.nginx.com/blog/avoiding-top-10-nginx-configuration-mistakes/#proxy_buffering-off) against disabling `proxy_buffering` as it will potentially waste upstream server resources.
+
+## Why does upstream receive a "Content-Length" header when the original request is sent with "Transfer-Encoding: chunked"
+
+For a request with "small" body that fits into [`client_body_buffer_size`](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_body_buffer_size) and with header "Transfer-Encoding: chunked", NGINX will always read and know the length of the body. Then it will send the request to upstream with the "Content-Length" header.
```
eguzki, line 37: This is not true for the use case this PR is taking care of. Only when nginx handles the proxying task. 

Actually, for a future enhancement, we could do the same. Read one buffer from the socket with size (S configurable). If all the full body has been read, send upstream with content-length. If the body full body is not read, proxy the request with TE: chunked.

eguzki (approved):
For me this is mergeable. 

I left a couple of comments that are just nitpicks.

kevprice83 (approved):
This looks really good to me now especially with the new README. I'd still like to see the unit tests in there but it's not urgent.

Another question I had to ensure my understanding:

So when an HTTPS proxy is used and TE: chunked and body size is greater than the body_buffer_size then APIcast will buffer, decode then re-encode the request so that the upstream will still receive the originally TE: chunked request format right?

eguzki:
> So when an HTTPS proxy is used and TE: chunked and body size is greater than the body_buffer_size then APIcast will buffer, decode then re-encode the request so that the upstream will still receive the originally TE: chunked request format right?

No need to twist it too much. It is much simpler. When an HTTPS proxy is used and TE;
* with `unbuffered_request` policy in place (i.e. `proxy_request_buffering` off): connection with upstream will be created upon request headers are received. Request headers and transfer encoding will be propagated as it is regardless of the body size.
>**NOTE**:  Internally, but the customers do not need to know this, for TE chunked, APIcast willl decode the encoding and re-encode again to keep the same encoding. And this is done in streaming mode, chunk by chunk as they are read from the socket. 
* with missing `unbuffered_request` policy (i.e. `proxy_request_buffering` on): The request body will be read (buffered) by APIcast. TE header will be replaced by the "Content-Length" header for the upstream connection. Only when the entire body has been read by APIcast, APIcast will initiate the connection with upstream sending the request with Content-Length header and removing the chunked encoding. 
>**NOTE**:  Internally, when the body is bigger than the `client_boby_buffer_size`, APIcast willl write the body in a file after removing the chunked encoding and the upstream http client will receive a body reader which is an iterator reading from the file.

eguzki:
@dfennessy your approval is required as you requested changes.

dfennessy (approved):
LGTM!

gateway/src/apicast/policy/request_unbuffered/README.md
tkan145: Updated verification steps. I will address this is the future PR.

PR #1409, opened 2023-08-31
[THREESCALE-4393] Add support to use Basic Authentication with the forward proxy

tkan145:
What

Fixes https://issues.redhat.com/browse/THREESCALE-4393

This PR support the use of Basic Auth with forward proxy in the following scenarios:
* `APIcast <> forward HTTP proxy with Basic Auth <> HTTP upstream`
* `APIcast <> forward HTTP proxy with Basic Auth <> HTTPS upstream`

## Verification steps 

* Add the following line to `examples/forward-proxy/tinyproxy.conf`

```
BasicAuth foo bar
```

### APIcast <> forward HTTP proxy with Basic Auth <> HTTP upstream

* Update `examples/forward-proxy/apicast-config.json` with the following:

```
{                                                             
    "services": [                                             
        {                                                     
            "backend_version": "1",                           
            "proxy": {                                        
                "hosts": ["one"],                             
                "api_backend": "http://postman-echo.com",       
                "backend": {                                  
                    "host": "backend"                         
                },                                            
                "policy_chain": [                             
                    {                                         
                    },                                        
                    {                                         
                        "configuration": {                    
                            "http_proxy": "http://foo:bar@proxy:443/"
                        }                                     
                    }                                         
                ],                                            
                "proxy_rules": [                              
                    {                                         
                        "http_method": "GET",                 
                        "pattern": "/",                       
                        "metric_system_name": "hits",         
                        "delta": 1,                           
                        "parameters": [],                     
                        "querystring_parameters": {}          
                    }                                         
                ]                                             
            }                                                 
        }                                                     
    ]                                                         
}                                                             
```

* run `docker-compose.forward-proxy.yml`

```
make forward-proxy-gateway
```

* Get APIcast IP

```
❯ docker ps
CONTAINER ID   IMAGE                   COMMAND                  CREATED          STATUS         PORTS                NAMES
bab24ff74d69   apicast-test            "container-entrypoin…"   9 seconds ago    Up 9 seconds   8080/tcp, 8090/tcp   apicast_build_0-gateway-run-a32123385dce
dec323e010f9   nginx:1.23.4            "/docker-entrypoint.…"   10 seconds ago   Up 9 seconds   80/tcp, 443/tcp      apicast_build_0-upstream-1
a0b1ffe5b514   apicast_build_0-proxy   "/usr/bin/tinyproxy …"   10 seconds ago   Up 9 seconds   0/tcp                apicast_build_0-proxy-1

❯ APICAST_IP=$(docker inspect apicast_build_0-gateway-run-a32123385dce | yq e -P '.[0].NetworkSettings.Networks.apicast_build_0_default.IPAddress' -)
```

* Send request to APICast

```
❯ curl -v -H "Host: one" http://${APICAST_IP}:8080/get?user_key=foo
*   Trying 127.0.0.1...                                                                                                         
* TCP_NODELAY set                                                                                                               
* Connected to localhost (127.0.0.1) port 8080 (#0)                                                                             
> GET /get?user_key= HTTP/1.1                                                                                                   
> Host: one                                                                                                                     
> User-Agent: curl/7.61.1                                                                                                       
> Accept: */*                                                                                                                   
>                                                                                                                               
< HTTP/1.1 200 OK                                                                                                               
< Server: openresty                                                                                                             
< Date: Thu, 31 Aug 2023 06:12:41 GMT                                                                                           
< Content-Type: application/json; charset=utf-8                                                                                 
< Content-Length: 370                                                                                                           
< Connection: keep-alive                                                                                                        
< Via: 1.1 tinyproxy (tinyproxy/1.11.1)                                                                                         
< ETag: W/"172-1BPDY6s8yDG5TJHwqYbKsiphJMI"                                                                                     
< set-cookie: sails.sid=s%3AnwD0yJG_9FLjXIev7zU9n_6U3CFFN5KD.QOtPZ4Jy84tU46c8%2FiYS%2FCsaolhffNCqFLeGPbcH%2FcM; Path=/; HttpOnly
<                                                                                                                               
{                                                                                                                               
  "args": {                                                                                                                     
    "user_key": ""                                                                                                              
  },                                                                                                                            
  "headers": {                                                                                                                  
    "x-forwarded-proto": "http",                                                                                                
    "x-forwarded-port": "80",                                                                                                   
    "host": "postman-echo.com",                                                                                                 
    "x-amzn-trace-id": "Root=1-64f02f59-455a8e1e1cf53ad137ba4e7a",                                                              
    "via": "1.1 tinyproxy (tinyproxy/1.11.1)",                                                                                  
    "user-agent": "curl/7.61.1",                                                                                                
    "accept": "*/*"                                                                                                             
  },                                                                                                                            
  "url": "http://postman-echo.com/get?user_key="                                                                                
* Connection #0 to host localhost left intact                                                                                   
                                                                                                 
```

We can see that the the request is going through the proxy

### APIcast <> forward HTTP proxy with Basic Auth <> HTTPS upstream

* Update `examples/forward-proxy/apicast-config.json` with the following:

```
{                                                             
    "services": [                                             
        {                                                     
            "backend_version": "1",                           
            "proxy": {                                        
                "hosts": ["one"],                             
                "api_backend": "https://postman-echo.com",       
                "backend": {                                  
                    "host": "backend"                         
                },                                            
                "policy_chain": [                             
                    {                                         
                    },                                        
                    {                                         
                        "configuration": {                    
                            "https_proxy": "http://foo:bar@proxy:443/"
                        }                                     
                    }                                         
                ],                                            
                "proxy_rules": [                              
                    {                                         
                        "http_method": "GET",                 
                        "pattern": "/",                       
                        "metric_system_name": "hits",         
                        "delta": 1,                           
                        "parameters": [],                     
                        "querystring_parameters": {}          
                    }                                         
                ]                                             
            }                                                 
        }                                                     
    ]                                                         
}                                                             
```

* run `docker-compose.forward-proxy.yml`

```
make forward-proxy-gateway
```

* Get APIcast IP

```
❯ docker ps
CONTAINER ID   IMAGE                   COMMAND                  CREATED          STATUS         PORTS                NAMES
f0b18fa16dda   apicast-test            "container-entrypoin…"   9 seconds ago    Up 9 seconds   8080/tcp, 8090/tcp   apicast_build_0-gateway-run-a32123385dce
7b3107db8aa4   nginx:1.23.4            "/docker-entrypoint.…"   10 seconds ago   Up 9 seconds   80/tcp, 443/tcp      apicast_build_0-upstream-1
f0b18fa16dda   apicast_build_0-proxy   "/usr/bin/tinyproxy …"   10 seconds ago   Up 9 seconds   0/tcp                apicast_build_0-proxy-1

❯ APICAST_IP=$(docker inspect apicast_build_0-gateway-run-a32123385dce | yq e -P '.[0].NetworkSettings.Networks.apicast_build_0_default.IPAddress' -)
```

* Send request to APICast

```
❯ curl -v -H "Host: one" http://${APICAST_IP}:8080/get?user_key=foo
*   Trying 127.0.0.1...                                                                                                            
* TCP_NODELAY set                                                                                                                  
* Connected to localhost (127.0.0.1) port 8080 (#0)                                                                                
> GET /get?user_key= HTTP/1.1                                                                                                      
> Host: one                                                                                                                        
> User-Agent: curl/7.61.1                                                                                                          
> Accept: */*                                                                                                                      
>                                                                                                                                  
< HTTP/1.1 200 OK                                                                                                                  
< Server: openresty                                                                                                                
< Content-Type: application/json; charset=utf-8                                                                                    
< Transfer-Encoding: chunked                                                                                                       
< Connection: keep-alive                                                                                                           
< set-cookie: sails.sid=s%3ALAz2UZyLXVxBZ8VY2XjC90y0N54Mk_JV.xJ0meVF%2F1%2FEP1m7BY0SZ9jNQ%2FLaIl3jaoHHsxH%2FPIhQ; Path=/; HttpOnly 
< Date: Thu, 31 Aug 2023 06:11:54 GMT                                                                                              
< ETag: W/"146-Gpco5rr4Zu8qMK21l8EJcRbPeqI"                                                                                        
<                                                                                                                                  
{                                                                                                                                  
  "args": {                                                                                                                        
    "user_key": ""                                                                                                                 
  },                                                                                                                               
  "headers": {                                                                                                                     
    "x-forwarded-proto": "https",                                                                                                  
    "x-forwarded-port": "443",                                                                                                     
    "host": "postman-echo.com",                                                                                                    
    "x-amzn-trace-id": "Root=1-64f02f2a-2403dd5e6c8931eb1878f63e",                                                                 
    "user-agent": "curl/7.61.1",                                                                                                   
    "accept": "*/*"                                                                                                                
  },                                                                                                                               
  "url": "https://postman-echo.com/get?user_key="                                                                                  
* Connection #0 to host localhost left intact                                                                                      

```

We can see that the the request is going through the proxy

### DIFF
--- a/gateway/src/apicast/http_proxy.lua
+++ b/gateway/src/apicast/http_proxy.lua
@@ -82,7 +82,7 @@ local function absolute_url(uri)
     )
 end
 
-local function forward_https_request(proxy_uri, uri, skip_https_connect)
+local function forward_https_request(proxy_uri, proxy_auth, uri, skip_https_connect)
     -- This is needed to call ngx.req.get_body_data() below.
     ngx.req.read_body()
 
@@ -102,7 +102,8 @@ local function forward_https_request(proxy_uri, uri, skip_https_connect)
         -- nil, so after this we need to read the temp file.
         -- https://github.com/openresty/lua-nginx-module#ngxreqget_body_data
         body = ngx.req.get_body_data(),
-        proxy_uri = proxy_uri
+        proxy_uri = proxy_uri,
+        proxy_auth = proxy_auth
     }
 
     if not request.body then
@@ -156,15 +157,20 @@ end
 
 function _M.request(upstream, proxy_uri)
     local uri = upstream.uri
+    local proxy_auth
 
-    if not ngx.var.http_proxy_authorization then
-        if proxy_uri.user or proxy_uri.password then
-            local proxy_auth = "Basic " .. ngx.encode_base64(concat({ proxy_uri.user or '', proxy_uri.password or '' }, ':'))
-            ngx.req.set_header("Proxy-Authorization", proxy_auth)
-        end
+    if proxy_uri.user or proxy_uri.password then
+        proxy_auth = "Basic " .. ngx.encode_base64(concat({ proxy_uri.user or '', proxy_uri.password or '' }, ':'))
     end
 
     if uri.scheme == 'http' then -- rewrite the request to use http_proxy
+        -- Only set "Proxy-Authorization" when sending HTTP request. When sent over HTTPS,
+        -- the `Proxy-Authorization` header must be sent in the CONNECT request as the proxy has
+        -- no visibility into the tunneled request.
+        if not ngx.var.http_proxy_authorization and proxy_auth then
+            ngx.req.set_header("Proxy-Authorization", proxy_auth)
+        end
+
         local err
         local host = upstream:set_host_header()
         upstream:use_host_header(host)
@@ -177,7 +183,7 @@ function _M.request(upstream, proxy_uri)
         return
     elseif uri.scheme == 'https' then
         upstream:rewrite_request()
-        forward_https_request(proxy_uri, uri, upstream.skip_https_connect)
+        forward_https_request(proxy_uri, proxy_auth, uri, upstream.skip_https_connect)
         return ngx.exit(ngx.OK) -- terminate phase
     else
         ngx.log(ngx.ERR, 'could not connect to proxy: ',  proxy_uri, ' err: ', 'invalid request scheme')
--- a/gateway/src/apicast/http_proxy.lua
+++ b/gateway/src/apicast/http_proxy.lua
@@ -5,6 +5,7 @@ local resty_resolver = require 'resty.resolver'
 local round_robin = require 'resty.balancer.round_robin'
 local http_proxy = require 'resty.http.proxy'
 local file_reader = require("resty.file").file_reader
+local concat = table.concat
 
 local _M = { }
 
@@ -155,6 +156,13 @@ end
 
 function _M.request(upstream, proxy_uri)
     local uri = upstream.uri
+
+    if not ngx.var.proxy_authorization then
+        if proxy_uri.user or proxy_uri.password then
+            local proxy_auth = "Basic " .. ngx.encode_base64(concat({ proxy_uri.user or '', proxy_uri.password or '' }, ':'))
+            ngx.req.set_header("Proxy-Authorization", proxy_auth)
+        end
+    end
 
     if uri.scheme == 'http' then -- rewrite the request to use http_proxy
         local err
--- a/gateway/src/apicast/http_proxy.lua
+++ b/gateway/src/apicast/http_proxy.lua
@@ -167,7 +167,10 @@ function _M.request(upstream, proxy_uri)
         -- Only set "Proxy-Authorization" when sending HTTP request. When sent over HTTPS,
         -- the `Proxy-Authorization` header must be sent in the CONNECT request as the proxy has
         -- no visibility into the tunneled request.
-        if not ngx.var.http_proxy_authorization and proxy_auth then
+        --
+        -- Also DO NOT set the header if using the camel proxy to avoid unintended leak of
+        -- Proxy-Authorization header in requests
+        if not ngx.var.http_proxy_authorization and proxy_auth and not upstream.skip_https_connect then
             ngx.req.set_header("Proxy-Authorization", proxy_auth)
         end
 
--- a/gateway/src/apicast/http_proxy.lua
+++ b/gateway/src/apicast/http_proxy.lua
@@ -157,7 +157,7 @@ end
 function _M.request(upstream, proxy_uri)
     local uri = upstream.uri
 
-    if not ngx.var.proxy_authorization then
+    if not ngx.var.http_proxy_authorization then
         if proxy_uri.user or proxy_uri.password then
             local proxy_auth = "Basic " .. ngx.encode_base64(concat({ proxy_uri.user or '', proxy_uri.password or '' }, ':'))
             ngx.req.set_header("Proxy-Authorization", proxy_auth)
--- a/gateway/src/apicast/policy/http_proxy/Readme.md
+++ b/gateway/src/apicast/policy/http_proxy/Readme.md
@@ -54,8 +54,8 @@ The policy expect the URLS following the `http://[<username>[:<passwd>]@]<host>[
       "name": "apicast.policy.http_proxy",
       "configuration": {
           "all_proxy": "http://foo:bar@192.168.15.103:8888/",
-          "https_proxy": "https://foo:bar@192.168.15.103:8888/",
-          "http_proxy": "https://foo:bar@192.168.15.103:8888/"
+          "https_proxy": "http://foo:bar@192.168.15.103:8888/",
+          "http_proxy": "http://foo:bar@192.168.15.103:8888/"
       }
     }
 ]
--- a/gateway/src/apicast/policy/http_proxy/Readme.md
+++ b/gateway/src/apicast/policy/http_proxy/Readme.md
@@ -43,6 +43,8 @@ used.
 
 ## Configuration
 
+The policy expect the URLS following the `http://[<username>[:<passwd>]@]<host>[:<port>]` format, e.g.:
+
 ```
 "policy_chain": [
     {
@@ -51,15 +53,17 @@ used.
     {
       "name": "apicast.policy.http_proxy",
       "configuration": {
-          "all_proxy": "http://192.168.15.103:8888/",
-          "https_proxy": "https://192.168.15.103:8888/",
-          "http_proxy": "https://192.168.15.103:8888/"
+          "all_proxy": "http://foo:bar@192.168.15.103:8888/",
+          "https_proxy": "https://foo:bar@192.168.15.103:8888/",
+          "http_proxy": "https://foo:bar@192.168.15.103:8888/"
       }
     }
 ]
 ```
 
-- If http_proxy or https_proxy is not defined the all_proxy will be taken. 
+- If http_proxy or https_proxy is not defined the all_proxy will be taken.
+- The policy supports for proxy authentication via the `<username>` and `<passwd>` options.
+- The `<username>` and `<passwd>` are optional, all other components are required.
 
 ## Caveats
 
@@ -67,7 +71,7 @@ used.
   always send to the proxy. 
 - In case of HTTP_PROXY, HTTPS_PROXY or ALL_PROXY parameters are defined, this
   policy will overwrite those values. 
-- Proxy connection does not support authentication.
+- 3scale currently does not support connecting to an HTTP proxy via TLS. For this reason, the scheme of the HTTPS_PROXY value is restricted to http.
 
 
 ## Example Use case
--- a/t/apicast-policy-camel.t
+++ b/t/apicast-policy-camel.t
@@ -318,6 +318,7 @@ EOF
 
 
 === TEST 5: API backend connection uses http proxy with Basic Auth
+Check that the Proxy Authorization header is not sent
 --- configuration
 {
   "services": [
@@ -358,7 +359,7 @@ EOF
     access_by_lua_block {
       assert = require('luassert')
       local proxy_auth = ngx.req.get_headers()['Proxy-Authorization']
-      assert.equals(proxy_auth, "Basic Zm9vOmJhcg==")
+      assert.falsy(proxy_auth)
       ngx.say("yay, api backend")
     }
   }
@@ -371,6 +372,7 @@ yay, api backend
 using proxy: http://foo:bar@127.0.0.1:$TEST_NGINX_HTTP_PROXY_PORT
 
 === TEST 6: API backend using all_proxy with Basic Auth
+Check that the Proxy Authorization header is not sent
 --- configuration
 {
   "services": [
@@ -411,7 +413,7 @@ using proxy: http://foo:bar@127.0.0.1:$TEST_NGINX_HTTP_PROXY_PORT
     access_by_lua_block {
       assert = require('luassert')
       local proxy_auth = ngx.req.get_headers()['Proxy-Authorization']
-      assert.equals(proxy_auth, "Basic Zm9vOmJhcg==")
+      assert.falsy(proxy_auth)
       ngx.say("yay, api backend")
     }
   }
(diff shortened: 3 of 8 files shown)


gateway/src/apicast/http_proxy.lua
```diff
@@ -156,6 +157,13 @@ end
 function _M.request(upstream, proxy_uri)
     local uri = upstream.uri
 
+    if not ngx.var.proxy_authorization then
```
eguzki, line 160: where does this var come from?

t/apicast-policy-http-proxy.t
```diff
@@ -194,3 +194,169 @@ proxy request: CONNECT test-upstream.lvh.me:$TEST_NGINX_RANDOM_PORT HTTP/1.1
...
+      local assert = require('luassert')
+      local proxy_auth = ngx.req.get_headers()['Proxy-Authorization']
+      assert.equals(proxy_auth, "Basic Zm9vOmJhcg==")
+    }
+  }
+--- request
+GET /?user_key=value
+--- error_code: 200
+--- error_log env
+proxy request: CONNECT test-upstream.lvh.me:$TEST_NGINX_RANDOM_PORT HTTP/1.1
```
eguzki, line 247: `CONNECT` with plain upstream?

t/apicast-policy-http-proxy.t
```diff
@@ -194,3 +194,169 @@ proxy request: CONNECT test-upstream.lvh.me:$TEST_NGINX_RANDOM_PORT HTTP/1.1
...
```
eguzki, line 300: `CONNECT` with plain upstream?

tkan145: Fixed

gateway/src/apicast/http_proxy.lua
tkan145: My bad, should be http_proxy_authorization. Fixed in the newest patch

tkan145:
> HAve you considered tests with HTTP(S)_PROXY env var?

Added tests for HTTP(S)_PROXY env var.

### To discuss:
For HTTP connections sent through the proxy, the set Proxy-Authorization in the request itself. However when sent over HTTPS, the Proxy-Authorization header must be sent in the CONNECT request as the proxy has no visibility into further tunneled requests.

The camel proxy policy is interesting here as we sent over HTTPS but skip CONNECT, so I wonder what is the correct approach here
* Set Proxy-Authorization header in the original request
* Don't set the header and update the documentation saying that camel proxy policy is incompatible with proxy authentication

t/apicast-policy-http-proxy.t
```diff
@@ -194,3 +194,165 @@ proxy request: CONNECT test-upstream.lvh.me:$TEST_NGINX_RANDOM_PORT HTTP/1.1
...
+      ngx.exit(ngx.OK)
+    }
+  }
+--- upstream
+server_name test-upstream.lvh.me;
+  location / {
+    access_by_lua_block {
+      local assert = require('luassert')
+      local proxy_auth = ngx.req.get_headers()['Proxy-Authorization']
+      assert.equals(proxy_auth, "Basic Zm9vOmJhcg==")
```
eguzki, line 240: Not a big deal, but according to [rfc7235#section-4.4](https://datatracker.ietf.org/doc/html/rfc7235#section-4.4),
```
A proxy MAY relay
   the credentials from the client request to the next proxy if that is
   the mechanism by which the proxies cooperatively authenticate a given
   request.
```
So asserting in the upstream the existence of the `Proxy-Authorization` maybe not be accurate.

t/apicast-policy-camel.t
```diff
@@ -315,3 +315,193 @@ ETag: foobar
 <<EOF
 using proxy: http://127.0.0.1:$Test::Nginx::Util::PROXY_SSL_PORT,
 EOF
+
+
+=== TEST 5: API backend connection uses http proxy with Basic Auth
```
eguzki, line 320: I wonder how this works with the camel proxy when upstream is TLS. 

You are adding the `Proxy-Authorization` header to the method `_connect_proxy_https`. But precisely, for the camel proxies, that method is not being used. Instead, `_connect_tls_direct` is being used. Correct me if I am wrong.

Does it even make sense to add authentication support on camel proxy when upstream is TLS?

eguzki (changes requested):
Looking super good!

Missing documentation.

At least, I would update the readme in the policies folder.

t/apicast-policy-camel.t
eguzki: Now I see your comment

```
The camel proxy policy is interesting here as we sent over HTTPS but skip CONNECT, so I wonder what is the correct approach here

    Set Proxy-Authorization header in the original request
    Don't set the header and update the documentation saying that camel proxy policy is incompatible with proxy authentication
```

Regarding `Set Proxy-Authorization header in the original request`, IMO it does not make sense to me. The header would not even be read by camel proxy and would end up in the upstream. 

Regarding `Don't set the header and update the documentation saying that camel proxy policy is incompatible with proxy authentication`, we may need to investigate further how camel proxy works and if they define a use case for proxy auth on TLS connections. Maybe here? https://github.com/zregvart/camel-netty-proxy They say it is only an example.

eguzki: In anyway, the integration tests for the TLS upstream does not make sense as it is for now.

t/apicast-policy-camel.t
```diff
@@ -315,3 +315,193 @@ ETag: foobar
...
+--- request
+GET /?user_key=value
+--- response_body
+yay, api backend
+--- error_code: 200
+--- error_log env
+using proxy: http://foo:bar@127.0.0.1:$TEST_NGINX_HTTP_PROXY_PORT
+
+
+=== TEST 7: using HTTPS proxy for backend with Basic Auth.
```
eguzki, line 427: I wonder how this works with the camel proxy when upstream is TLS. 

You are adding the `Proxy-Authorization` header to the method `_connect_proxy_https`. But precisely, for the camel proxies, that method is not being used. Instead, `_connect_tls_direct` is being used. Correct me if I am wrong.

Does it even make sense to add authentication support on camel proxy when upstream is TLS?

eguzki: Now I see your comment

```
The camel proxy policy is interesting here as we sent over HTTPS but skip CONNECT, so I wonder what is the correct approach here

    Set Proxy-Authorization header in the original request
    Don't set the header and update the documentation saying that camel proxy policy is incompatible with proxy authentication
```

Regarding `Set Proxy-Authorization header in the original request`, IMO it does not make sense to me. The header would not even be read by camel proxy and would end up in the upstream. 

Regarding `Don't set the header and update the documentation saying that camel proxy policy is incompatible with proxy authentication`, we may need to investigate further how camel proxy works and if they define a use case for proxy auth on TLS connections. Maybe here? https://github.com/zregvart/camel-netty-proxy They say it is only an example.

eguzki: Anyway, the integration tests for the TLS upstream does not make sense as it is for now.

t/apicast-policy-http-proxy.t
eguzki: Ideally, the test should check that the header reaches the proxy instead. Can we add assertions at the proxy?

tkan145: Added documentation for http_proxy policy.

I tried extending /t/fixture/proxy.lua to enable proxy authentication checks for each test but couldn't figure out the best way to do so, I ended up printing the header in the debug log and assert the log, quick and easy.

gateway/src/apicast/policy/http_proxy/Readme.md
```diff
@@ -51,23 +53,25 @@ used.
     {
       "name": "apicast.policy.http_proxy",
       "configuration": {
-          "all_proxy": "http://192.168.15.103:8888/",
-          "https_proxy": "https://192.168.15.103:8888/",
-          "http_proxy": "https://192.168.15.103:8888/"
+          "all_proxy": "http://foo:bar@192.168.15.103:8888/",
+          "https_proxy": "https://foo:bar@192.168.15.103:8888/",
```
eguzki, line 57: proxy urls should all be `http://`, `https` not supported

t/apicast-policy-camel.t
tkan145: so I talked to few friends from the FuseSource team and no one really knows what's called a camel proxy.

The example you provided is built using Camel netty-http. And checking the [source code of netty-http](https://github.com/apache/camel/tree/main/components/camel-netty-http/src/main/java/org/apache/camel/component/netty/http) it looks like it doesn't support proxy authentication and CONNECT out of the box. 

With that said, I believe the camel proxy's behavior will depend on the implementation.

tkan145: The whole point of camel proxy is to proxy the request without CONNECT, otherwise they can just use the normal proxy. 

I would vote to not support authentication with camel proxy. Let me know what you think and I will push a new patch

gateway/src/apicast/policy/http_proxy/Readme.md
tkan145: fixed

t/apicast-policy-camel.t
eguzki: It is being supported feature https://access.redhat.com/documentation/en-us/red_hat_3scale_api_management/2.13/html/administering_the_api_gateway/apicast-policies#camel-service_standard-policies

IDK how many customers are using it, tho

What I would do is to add doc (a note somewhere) saying that camel proxying does not support basic client auth.

gateway/src/apicast/policy/http_proxy/Readme.md
```diff
@@ -51,23 +53,25 @@ used.
...
+- The `<username>` and `<passwd>` are optional, all other components are required.
 
 ## Caveats
 
 - This policy will disable all load-balancing policies and traffic will be
   always send to the proxy. 
 - In case of HTTP_PROXY, HTTPS_PROXY or ALL_PROXY parameters are defined, this
   policy will overwrite those values. 
-- Proxy connection does not support authentication.
+- 3scale currently does not support connecting to an HTTP proxy via TLS. For this reason, the scheme of the HTTPS_PROXY value is restricted to http.
```
eguzki, line 74: > 3scale currently does not support connecting to an HTTP proxy via TLS. For this reason, the scheme of the HTTPS_PROXY value is restricted to http.

IMO not very accurate

 > 3scale currently does not support connecting to an HTTP proxy via TLS

Correct.

> For this reason, the scheme of the HTTPS_PROXY value is restricted to http.

All proxy URLs should be `http` based, never `https` based. 

If upstream is http -> then HTTP_PROXY is used by apicast

if upstream is https -> then HTTPS_PROXY is used by apicast

t/apicast-policy-camel.t
tkan145: It's already in the camel proxy Readme file 
https://github.com/3scale/APIcast/blob/master/gateway/src/apicast/policy/camel/Readme.md?plain=1#L65

gateway/src/apicast/policy/http_proxy/Readme.md
tkan145: > All proxy URLs should be http based, never https based.

That's exactly what I meant when I said `For this reason, the HTTPS_PROXY value scheme is restricted to http.` Maybe that causes more confusion, maybe we should delete that sentence?

t/apicast-policy-camel.t
tkan145: Also fixed the camel proxy integration tests

eguzki:
Changes looks good to me. 

I cannot approve because there are git conflicts and some integration tests are failing

tkan145:
Ack, let me rebase and review the test log. It seems like a bunch of unrelated tests also failed

tkan145:
Conflict resolved but I'm still not sure why the tests are failing, it seems very random to me

eguzki:
> Conflict resolved but I'm still not sure why the tests are failing, it seems very random to me

Yeah, there are few flaky tests we should be fixing sooner or later. 

After few tries, no more than 5, they should all pass

eguzki (approved):

PR #1416, opened 2023-10-23
balancer.t conversion to blackbox

odkq:
Runs OK within docker environment, returning the expected 1 1 in sequence

Not sure if all the components are in place before the test, though. In the original test require('resty.balancer.round_robin').cache_size = 1 was set globally and here it is set on each request.

### DIFF
--- a/t/balancer.t
+++ b/t/balancer.t
@@ -1,27 +1,35 @@
 use lib 't';
-use Test::APIcast 'no_plan';
+use Test::APIcast::Blackbox 'no_plan';
 
-$ENV{TEST_NGINX_HTTP_CONFIG} = "$Test::APIcast::path/http.d/*.conf";
-$ENV{RESOLVER} = '127.0.1.1:5353';
-
-env_to_nginx(
-    'RESOLVER'
-);
-master_on();
+repeat_each(1);
 run_tests();
 
 __DATA__
 
 === TEST 1: round robin does not leak memory
 Balancing different hosts does not leak memory.
---- http_config
-  lua_package_path "$TEST_NGINX_LUA_PATH";
-  init_by_lua_block {
-    require('resty.balancer.round_robin').cache_size = 1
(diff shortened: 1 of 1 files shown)


t/balancer.t
```diff
@@ -1,27 +1,35 @@
...
-  lua_package_path "$TEST_NGINX_LUA_PATH";
-  init_by_lua_block {
-    require('resty.balancer.round_robin').cache_size = 1
+--- configuration
+  {
+      "services": [
+      {
+          "proxy": {
+            "policy_chain": [
+            { "name": "apicast.policy.upstream",
```
eguzki, line 17: instead of `upstream` poilicy, maybe better `"api_backend": "http://test:$TEST_NGINX_SERVER_PORT/",` ?

eguzki (commented):
Looking at the test, that it does not test the functionality on top of apicast, would it make more sense to move this test to unittests in `spec/resty/balancer/round_robin_spec.lua` ??

t/balancer.t
odkq: I added it, with authentication as I did not find a straightforward way to use "api_backend" w/o authentication in other blackbox enabled tests. Is there a way to disable authentication as it does not seem to be needed for this test?

eguzki: > Is there a way to disable authentication as it does not seem to be needed for this test?

You can add a policy in the policy chain called [default_credentials](https://github.com/3scale/APIcast/blob/master/gateway/src/apicast/policy/default_credentials/apicast-policy.json). However, it is not a big deal to add `?user_key=value` as you did. I think it is ok like that

odkq: I see, having unauthenticated requests is not a common use case. Thank you @eguzki !

eguzki:
any update? this is still in draft

tkan145:
Closing due to stale PR

PR #1419, opened 2023-10-31
THREESCALE-9976 Fix GRPC

eguzki:
### What

Fixes https://issues.redhat.com/browse/THREESCALE-9976

GRPC leverages the POST HTTP2 method ([GRPC Spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)) . APIcast [reads the body](https://github.com/3scale/APIcast/pull/1419/files#diff-b4a0139162c96d858c01c0872560f495c15fee4e54df8ea52acda64e429c96cfR72) looking for request params for POST based requests using [ngx.req.read_body](https://openresty-reference.readthedocs.io/en/latest/Lua_Nginx_API/#ngxreqread_body). However, [ngx.req.read_body](https://openresty-reference.readthedocs.io/en/latest/Lua_Nginx_API/#ngxreqread_body) does not support HTTP2 and it blocks the request. The clients receive connection timeout errors.

Added development environment for GRPC scenario

### Verification steps

```shell
```
* Run grpc dev environment

```shell
cd dev-environments/grpc
make gateway-certs
make upstream-certs
```

* Download GRPC client

```shell
make grpcurl
```

* Run request to echo endpoint

```shell
bin/grpcurl -vv -insecure -H "app_id: abc123" -H "app_key: abc123" -authority gateway.example.com 127.0.0.1:8443 main.HelloWorld/Greeting
```

The command should return the echo response

```
Resolved method descriptor:
rpc Greeting ( .main.GreetingMessage ) returns ( .main.GreetingReply );

Request metadata to send:
app_id: abc123
app_key: abc123

Response headers received:
content-type: application/grpc
date: Fri, 03 Nov 2023 11:41:16 GMT
server: openresty

Estimated response size: 32 bytes

Response contents:
{
  "hostname": "d7728ab86df0",
  "clientAddress": "172.19.0.3:53564"
}

Response trailers received:
(empty)
Sent 0 requests and received 1 response
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Fix GRPC on HTTP2 POST method [PR #1419](https://github.com/3scale/apicast/pull/1419) [THREESCALE-10224](https://issues.redhat.com/browse/THREESCALE-9976)
+
 - Fixed CVE-2023-44487 (HTTP/2 Rapid Reset) [PR #1417](https://github.com/3scale/apicast/pull/1417) [THREESCALE-10224](https://issues.redhat.com/browse/THREESCALE-10224)
 
 ### Added
@@ -22,11 +24,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - In boot mode on `init_worker` check configuration expiration [PR #1399](https://github.com/3scale/APIcast/pull/1399) [THREESCALE-9003](https://issues.redhat.com/browse/THREESCALE-9003)
 - Removes the warning message at the bootstrap [PR #1398](https://github.com/3scale/APIcast/pull/1398) [THREESCALE-7942](https://issues.redhat.com/browse/THREESCALE-7942)
 - Set NGiNX variable variables_hash_max_size to 2048 to avoid startup warning [PR #1395](https://github.com/3scale/APIcast/pull/1395) [THREESCALE-7941](https://issues.redhat.com/browse/THREESCALE-7941)
-- Dev environment on aarch64 host [PR #1381](https://github.com/3scale/APIcast/pull/1381) 
+- Dev environment on aarch64 host [PR #1381](https://github.com/3scale/APIcast/pull/1381)
 
 ### Added
 
-- Doc: Policy Development Tutorial [PR #1384](https://github.com/3scale/APIcast/pull/1384) 
+- Doc: Policy Development Tutorial [PR #1384](https://github.com/3scale/APIcast/pull/1384)
 - Opentelemetry support. Opentracing is now deprecated [PR #1379](https://github.com/3scale/APIcast/pull/1379) [THREESCALE-7735](https://issues.redhat.com/browse/THREESCALE-7735)
 - `/admin/api/account/proxy_configs` endpoint for configuration loading [PR #1352](https://github.com/3scale/APIcast/pull/1352) [THREESCALE-8508](https://issues.redhat.com/browse/THREESCALE-8508)
 - Pagination of services and proxy config endpoints [PR #1397](https://github.com/3scale/APIcast/pull/1397) [THREESCALE-8373](https://issues.redhat.com/browse/THREESCALE-8373)
--- a/dev-environments/grpc/Makefile
+++ b/dev-environments/grpc/Makefile
@@ -1,0 +1,66 @@
+SHELL = /usr/bin/env bash -o pipefail
+.SHELLFLAGS = -ec
+.DEFAULT_GOAL := gateway
+MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
+WORKDIR := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
+DOCKER ?= $(shell which docker 2> /dev/null || echo "docker")
+
+gateway: ## run gateway configured to access upstream powered with TLS
+	$(DOCKER) compose -f docker-compose.yml run --service-ports gateway
+
+clean:
+	$(DOCKER) compose down --volumes --remove-orphans
+	$(DOCKER) compose -f docker-compose.yml down --volumes --remove-orphans
+	- rm -rf gateway-cert
+	- rm -rf upstream-cert
+	- rm -rf bin
+
+ca:
+	openssl genrsa -out rootCA.key 2048
+	openssl req -batch -new -x509 -nodes -key rootCA.key -sha256 -days 1024 -out rootCA.pem
+
+clientcerts:
+	openssl req -subj '/CN=$(DOMAIN)'  -newkey rsa:4096 -nodes \
+			-sha256 \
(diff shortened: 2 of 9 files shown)


gateway/src/apicast/configuration/service.lua
```diff
@@ -30,13 +33,42 @@ local http_methods_with_body = {
   PATCH = true
 }
 
+local function is_http2_request()
+  return ngx.req.http_version == '2.0'
+end
+
+local function content_type_is_urlencoded(headers)
+    local ct = headers["Content-Type"]
```
tkan145, line 41: As I remember when running on Openshift, headers are all lower case. so perhaps use `ngx.var.http_content_type` instead?

dev-environments/grpc/Makefile
```diff
@@ -0,0 +1,66 @@
...
+# go-install-tool will 'go install' any package $2 and install it to $1.
+define go-install-tool
+@[ -f $(1) ] || { \
+set -e ;\
+TMP_DIR=$$(mktemp -d) ;\
+cd $$TMP_DIR ;\
+go mod init tmp ;\
+echo "Downloading $(2)" ;\
+GOBIN=$(WORKDIR)/bin go install $(2) ;\
+rm -rf $$TMP_DIR ;\
```
tkan145, line 64: What if user don't have Go installed on the machine? Perhaps we should put a prerequisite in the Readme file?

gateway/src/apicast/configuration/service.lua
```diff
@@ -30,13 +33,42 @@ local http_methods_with_body = {
...
+    end
+
+    return str_find(lower(ct), "application/x-www-form-urlencoded", 1, true) ~= nil
+end
+
+
 local function read_body_args(...)
   local method = ngx.req.get_method()
 
+  ngx.log(ngx.WARN, "is_http2_request: ", is_http2_request())
```
tkan145, line 59: I would move this log inside the if statement, otherwise this will print every time this function is called.

gateway/src/apicast/configuration/service.lua
```diff
@@ -174,9 +206,13 @@ end
 local function get_request_params(method)
   local params = ngx.req.get_uri_args()
 
-  if method == "GET" then
+  ngx.log(ngx.WARN, "is_http2_request: ", is_http2_request())
```
tkan145, line 209: Same as above

tkan145:
Not sure why GH print the review twice :confused:

gateway/src/apicast/configuration/service.lua
eguzki: good catch! I forgot to remove it

gateway/src/apicast/configuration/service.lua
eguzki: From https://github.com/openresty/lua-nginx-module/tree/v0.10.19#ngxreqhttp_version

```
Also, by default, an __index metamethod is added to the resulting Lua table and will normalize the keys to a pure lowercase form with all underscores converted to dashes in case of a lookup miss. For example, if a request header My-Foo-Header is present, then the following invocations will all pick up the value of this header correctly:
 ngx.say(headers.my_foo_header)
 ngx.say(headers["My-Foo-Header"])
 ngx.say(headers["my-foo-header"])
```

CHANGELOG.md
```diff
@@ -22,11 +24,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - In boot mode on `init_worker` check configuration expiration [PR #1399](https://github.com/3scale/APIcast/pull/1399) [THREESCALE-9003](https://issues.redhat.com/browse/THREESCALE-9003)
 - Removes the warning message at the bootstrap [PR #1398](https://github.com/3scale/APIcast/pull/1398) [THREESCALE-7942](https://issues.redhat.com/browse/THREESCALE-7942)
 - Set NGiNX variable variables_hash_max_size to 2048 to avoid startup warning [PR #1395](https://github.com/3scale/APIcast/pull/1395) [THREESCALE-7941](https://issues.redhat.com/browse/THREESCALE-7941)
-- Dev environment on aarch64 host [PR #1381](https://github.com/3scale/APIcast/pull/1381)
```
tkan145, line 25: It's just a small thing but I think this is an unrelated change?

CHANGELOG.md
```diff
@@ -22,11 +24,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - In boot mode on `init_worker` check configuration expiration [PR #1399](https://github.com/3scale/APIcast/pull/1399) [THREESCALE-9003](https://issues.redhat.com/browse/THREESCALE-9003)
 - Removes the warning message at the bootstrap [PR #1398](https://github.com/3scale/APIcast/pull/1398) [THREESCALE-7942](https://issues.redhat.com/browse/THREESCALE-7942)
 - Set NGiNX variable variables_hash_max_size to 2048 to avoid startup warning [PR #1395](https://github.com/3scale/APIcast/pull/1395) [THREESCALE-7941](https://issues.redhat.com/browse/THREESCALE-7941)
-- Dev environment on aarch64 host [PR #1381](https://github.com/3scale/APIcast/pull/1381) 
+- Dev environment on aarch64 host [PR #1381](https://github.com/3scale/APIcast/pull/1381)
 
 ### Added
 
-- Doc: Policy Development Tutorial [PR #1384](https://github.com/3scale/APIcast/pull/1384)
```
tkan145, line 29: Same as above

tkan145:
> However, [ngx.req.read_body](https://openresty-reference.readthedocs.io/en/latest/Lua_Nginx_API/#ngxreqread_body) does not support HTTP2 and it blocks the request.

Do you mean ngx.req.get_post_args? As the line `client request body saved to a temporary file` can be seen from the log, so I think ngx.req.read_body works with http2

CHANGELOG.md
eguzki: It's just the IDE tool removing trailing spaces

CHANGELOG.md
eguzki: It's just the IDE tool removing trailing spaces

tkan145 (approved):

eguzki:
> > However, [ngx.req.read_body](https://openresty-reference.readthedocs.io/en/latest/Lua_Nginx_API/#ngxreqread_body) does not support HTTP2 and it blocks the request.
> 
> ~Do you mean ngx.req.get_post_args? As the line `client request body saved to a temporary file` can be seen from the log, so I think ngx.req.read_body works with http2~
> 
> Ignore my previous comment. just found this one [openresty/lua-nginx-module#2172](https://github.com/openresty/lua-nginx-module/issues/2172)

From https://github.com/openresty/lua-nginx-module#ngxreqread_body

```
Due to the stream processing feature of HTTP/2 or HTTP/3, this api could potentially block the entire request. Therefore, this api is effective only when HTTP/2 or HTTP/3 requests send content-length header. For requests with versions lower than HTTP/2, this api can still be used without any problems.
```

PR #1442, opened 2024-02-05
docker compose up instead of docker compose run

eguzki:
### WHAT

Dev environments use `docker compose up` instead of `docker compose run`. 

I was confused when using `docker compose run` that when listing the containers running in the project, apicast was not there. I learned it is because we were using `docker compose run`. Then I learned that `run` command is meant fro running "one-off" task. More appropriate for running tests. Furthermore, it does not respect some service configuration like command or ports.


tkan145:
LGTM!

tkan145 (approved):

PR #1443, opened 2024-02-05
THREESCALE-10582 fix integration of upstream connection policy with camel policy

eguzki:
### What

Fixes: https://issues.redhat.com/browse/THREESCALE-10582

Upstream timeouts don't work with Camel Service. Actually upstream connection policy is not working when `https_proxy` is being used (either with env vars or policy).

This PR fixes the integration of upstream connection with any use case for `https_proxy`. 

It was considered adding connection options to the proxy policy and camel policy. Mainly because "Upstream connection" policy is referring to "upstream", which can be confusing when a proxy is being used, as the connection to upstream backend is no longer made by APIcast. Instead, APIcast creates a connection to the proxy. So the Upstream connection opts should, ideally, only apply to connections to backend "upstream". 

We decided to apply upstream connection policy to any "upstream" connection the APIcast make initiate, either a proxy or a actual upstream backend. That way, implementation wise it is easier. No need to add extended connection parameters to the proxy policies. Furthermore, if users are using Upstream connection policy together with `http_proxy`, the configuration still applies. With a new connection parameters in the proxy policies, this last use case would be broken. Additionally, if connection opts are added to the policies as optional params, we would need to add new env vars as well for the use case where proxies are configured via env vars. Too complex just to stick "upstream" concept to the actual backend (`api_backend` in the service configuration). Instead, upstram connection policy applies to any "upstream" connection APIcast does, regardless of being it a proxy or backend upstream. 

#### Upstream connection integration with `https_proxy` camel proxy 

```shell
```

* Camel proxy dev environment setup

```shell
cd dev-environments/camel-proxy
make certs
```

* Testing `https_proxy` use case: APIcast --> camel proxy --> upstream (TLS)

This env uses as `api_backend` the real env `https://echo-api.3scale.net:443`. My roundtrip latency is ~400ms. 

```
❯ curl -i -w "tcp:%{time_total}\n" https://echo-api.3scale.net:443 2>/dev/null
HTTP/1.1 200 OK
content-type: application/json
x-3scale-echo-api: echo-api/1.0.3
vary: Origin
x-content-type-options: nosniff
content-length: 524
x-envoy-upstream-service-time: 0
date: Mon, 05 Feb 2024 14:50:56 GMT
server: envoy

{
  "method": "GET",
  "path": "/",
  "args": "",
  "body": "",
  "headers": {
    "HTTP_VERSION": "HTTP/1.1",
    "HTTP_HOST": "echo-api.3scale.net",
    "HTTP_USER_AGENT": "curl/7.81.0",
    "HTTP_ACCEPT": "*/*",
    "HTTP_X_FORWARDED_FOR": "81.61.128.254",
    "HTTP_X_FORWARDED_PROTO": "https",
    "HTTP_X_ENVOY_EXTERNAL_ADDRESS": "81.61.128.254",
    "HTTP_X_REQUEST_ID": "a78c1edb-b0bf-42a2-8c44-e1ae9d4dba49",
    "HTTP_X_ENVOY_EXPECTED_RQ_TIMEOUT_MS": "15000"
  },
  "uuid": "b48cdf4d-1419-4aaf-871e-10cd9c67f68e"
}tcp:0.398028
``` 
Let's start with timeouts set to 1 sec and the request should be accepted.
```patch
patch <<EOF
diff --git a/dev-environments/camel-proxy/apicast-config.json b/dev-environments/camel-proxy/apicast-config.json
index 91201afa..8f92f029 100644
--- a/dev-environments/camel-proxy/apicast-config.json
+++ b/dev-environments/camel-proxy/apicast-config.json
@@ -44,6 +44,14 @@
           "host": "backend"
         },
         "policy_chain": [
+          {
+            "name": "apicast.policy.upstream_connection",
+            "configuration": {
+              "connect_timeout": 1,
+              "send_timeout": 1,
+              "read_timeout": 1
+            }
+          },
           {
             "name": "apicast.policy.camel",
             "configuration": {
EOF
```

Run environment

```shell
```

The request should be accepted (`200 OK`), as the connection timeouts should not be exceeded. 

```sh
curl --resolve https-proxy.example.com:8080:127.0.0.1 -v "http://https-proxy.example.com:8080/?user_key=123"
```

Now, let's lower the timeouts threshold to something like 100ms that should be exceeded because the upstream is far away.

Stop the gateway
```
CTRL-C
```
Restore `apicast-config.json` file
```
git checkout apicast-config.json
```

Apply 100ms timeouts

```patch
patch <<EOF
diff --git a/dev-environments/camel-proxy/apicast-config.json b/dev-environments/camel-proxy/apicast-config.json
index 91201afa..8f92f029 100644
--- a/dev-environments/camel-proxy/apicast-config.json
+++ b/dev-environments/camel-proxy/apicast-config.json
@@ -44,6 +44,14 @@
           "host": "backend"
         },
         "policy_chain": [
+          {
+            "name": "apicast.policy.upstream_connection",
+            "configuration": {
+              "connect_timeout": 0.1,
+              "send_timeout": 0.1,
+              "read_timeout": 0.1
+            }
+          },
           {
             "name": "apicast.policy.camel",
             "configuration": {
EOF
```

Run environment

```shell
```

The request should fail (`502 Bad Gateway`), as the connection timeouts should be exceeded.

```sh
curl --resolve https-proxy.example.com:8080:127.0.0.1 -v "http://https-proxy.example.com:8080/?user_key=123"
```

The logs should show the following line

```
[error] 19#19: *2 lua tcp socket read timed out
```

Clean the env before starting next step

```
make clean
```

#### Upstream connection integration with `https_proxy` with proxy policy (tinyproxy)

```shell
cd ${APICAST_PROJECT} && git checkout THREESCALE-10582-upstream-policy-with-camel-policy

```

* `https_proxy` dev environment setup

```shell
make certs
```
* Testing `https_proxy` use case: APIcast --> tiny  proxy --> upstream (TLS)

Timeouts set to 0.1 sec
```patch
patch <<EOF
index 5227c5aa..34a2ed23 100644
@@ -11,6 +11,14 @@
           "host": "backend"
         },
         "policy_chain": [
+          {
+            "name": "apicast.policy.upstream_connection",
+            "configuration": {
+              "connect_timeout": 0.1,
+              "send_timeout": 0.1,
+              "read_timeout": 0.1
+            }
+          },
           {
             "configuration": {
EOF
```
Run environment

```shell
```
The request should be accepted (200 OK), as the connection timeouts should not be exceeded.
```sh                                                                                        
curl --resolve get.example.com:8080:127.0.0.1 -i "http://get.example.com:8080/?user_key=123" 
```                                                                         

Now, let's simulate some network latency using  docker containers with Traffic control. We will add 200ms of latency to the container running `socat` between the proxy and the backend upstream. It is called `example.com` service.

Stop the gateway
```
CTRL-C
```

We are going to modify network-related stuff, so the NET_ADMIN capability is needed.
```patch
patch <<EOF
diff --git a/dev-environments/https-proxy-upstream-tlsv1.3/docker-compose.yml b/dev-environments/https-proxy-upstream-tlsv1.3/docker-compose.yml
index 9fa735f7..0f802e8b 100644
--- a/dev-environments/https-proxy-upstream-tlsv1.3/docker-compose.yml
+++ b/dev-environments/https-proxy-upstream-tlsv1.3/docker-compose.yml
@@ -39,6 +39,8 @@ services:
     restart: unless-stopped
     volumes:
       - ./cert/example.com.pem:/etc/pki/example.com.pem
+    cap_add:
+      - NET_ADMIN
   two.upstream:
     image: kennethreitz/httpbin
     expose:
EOF
```

Run environment with the new config

```shell
```

install the `tc` (traffic control) command

```
docker compose exec example.com apk add iproute2-tc
```

Add 200ms latency to the outbound traffic of `example.com` service.
```
docker compose exec example.com tc qdisc add dev eth0 root netem delay 200ms
```
The request should be rejected (`503 Service Temporarily Unavailable`), as the connection timeouts should not be exceeded.
```sh                                                                                        
curl --resolve get.example.com:8080:127.0.0.1 -i "http://get.example.com:8080/?user_key=123" 
```      
The logs should show the following line

```
[error] 19#19: *2 lua tcp socket read timed out
```

### DIFF
--- a/gateway/src/resty/resolver/http.lua
+++ b/gateway/src/resty/resolver/http.lua
@@ -8,8 +8,14 @@ local _M = setmetatable({}, { __index = resty_http })
 
 local mt = { __index = _M }
 
-function _M.new()
+function _M.new(opts)
+  opts = opts or { }
   local http = resty_http:new()
+
+  local timeouts = opts.timeouts
+  if timeouts then
+    http:set_timeouts(timeouts.connect_timeout, timeouts.send_timeout, timeouts.read_timeout)
+  end
 
   http.resolver = resty_resolver:instance()
   http.balancer = round_robin.new()
--- a/gateway/src/resty/resolver/http.lua
+++ b/gateway/src/resty/resolver/http.lua
@@ -8,22 +8,8 @@ local _M = setmetatable({}, { __index = resty_http })
 
 local mt = { __index = _M }
 
-function _M.new(opts)
-  opts = opts or { }
+function _M.new()
   local http = resty_http:new()
-
-  local timeouts = opts.timeouts
-  if timeouts then
-    ngx.log(ngx.DEBUG, 'setting timeouts (secs), connect_timeout: ', timeouts.connect_timeout,
-      ' send_timeout: ', timeouts.send_timeout, ' read_timeout: ', timeouts.read_timeout)
-    -- lua-resty-http uses nginx API for lua sockets
-    -- in milliseconds
-    -- https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#tcpsocksettimeouts
-    local connect_timeout = timeouts.connect_timeout and timeouts.connect_timeout * 1000
-    local send_timeout = timeouts.send_timeout and timeouts.send_timeout * 1000
-    local read_timeout = timeouts.read_timeout and timeouts.read_timeout * 1000
-    http:set_timeouts(connect_timeout, send_timeout, read_timeout)
-  end
 
   http.resolver = resty_resolver:instance()
   http.balancer = round_robin.new()
--- a/gateway/src/resty/resolver/http.lua
+++ b/gateway/src/resty/resolver/http.lua
@@ -14,7 +14,15 @@ function _M.new(opts)
 
   local timeouts = opts.timeouts
   if timeouts then
-    http:set_timeouts(timeouts.connect_timeout, timeouts.send_timeout, timeouts.read_timeout)
+    ngx.log(ngx.DEBUG, 'setting timeouts (secs), connect_timeout: ', timeouts.connect_timeout,
+      ' send_timeout: ', timeouts.send_timeout, ' read_timeout: ', timeouts.read_timeout)
+    -- lua-resty-http uses nginx API for lua sockets
+    -- in milliseconds
+    -- https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#tcpsocksettimeouts
+    local connect_timeout = timeouts.connect_timeout and timeouts.connect_timeout * 1000
+    local send_timeout = timeouts.send_timeout and timeouts.send_timeout * 1000
+    local read_timeout = timeouts.read_timeout and timeouts.read_timeout * 1000
+    http:set_timeouts(connect_timeout, send_timeout, read_timeout)
   end
 
   http.resolver = resty_resolver:instance()
--- a/t/apicast-policy-http-proxy.t
+++ b/t/apicast-policy-http-proxy.t
@@ -202,10 +202,8 @@ ETag: foobar
 --- error_code: 200
 --- error_log env
 proxy request: CONNECT test-upstream.lvh.me:$TEST_NGINX_RANDOM_PORT HTTP/1.1
---- user_files fixture=tls.pl eval
---- error_log env
 using proxy: $TEST_NGINX_HTTPS_PROXY
-
+--- user_files fixture=tls.pl eval
 
 === TEST 4: using HTTP proxy with Basic Auth
 --- configuration
--- a/t/apicast-policy-upstream-connection.t
+++ b/t/apicast-policy-upstream-connection.t
@@ -144,3 +144,84 @@ proxy request: CONNECT test-upstream.lvh.me:$TEST_NGINX_RANDOM_PORT HTTP/1.1
 using proxy: $TEST_NGINX_HTTPS_PROXY
 proxy_response(): timeout
 --- user_files fixture=tls.pl eval
+
+=== TEST 3: Set timeouts using HTTPS proxy for backend using HTTPS_PROXY env var
+In this test we set some timeouts to 1s. To force a read timeout, the upstream
+returns part of the response, then waits 3s (more than the timeout defined),
+and after that, it returns the rest of the response. Backend is configured with https_proxy
+This test uses the "ignore_response" section, because we know that the response
+is not going to be complete and that makes the Test::Nginx framework raise an
+error. With "ignore_response" that error is ignored.
+--- env eval
+(
+  "https_proxy" => $ENV{TEST_NGINX_HTTPS_PROXY},
+)
+--- configuration random_port env
+{
+  "services": [
+    {
+      "backend_version":  1,
+      "proxy": {
+        "api_backend": "https://test-upstream.lvh.me:$TEST_NGINX_RANDOM_PORT",
+        "proxy_rules": [
+          { "pattern": "/test", "http_method": "GET", "metric_system_name": "hits", "delta": 1 }
+        ],
+        "policy_chain": [
+          {
+            "name": "apicast.policy.upstream_connection",
+            "configuration": {
+              "connect_timeout": 1,
+              "send_timeout": 1,
+              "read_timeout": 1
+            }
+          },
+          {
+            "name": "apicast.policy.apicast"
+          }
+        ]
+      }
+    }
+  ]
+}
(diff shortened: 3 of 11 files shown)


t/apicast-policy-http-proxy.t
```diff
@@ -202,10 +202,8 @@ ETag: foobar
 --- error_code: 200
 --- error_log env
 proxy request: CONNECT test-upstream.lvh.me:$TEST_NGINX_RANDOM_PORT HTTP/1.1
---- user_files fixture=tls.pl eval
---- error_log env
```
eguzki, line 206: there cannot be two `--- error_log` directives. Only one of them is being applied.

t/apicast-policy-upstream-connection.t
```diff
@@ -1,6 +1,8 @@
 use lib 't';
 use Test::APIcast::Blackbox 'no_plan';
 
+require("http_proxy.pl");
```
eguzki, line 4: one comment not related to this PR: I am thinking about running one http proxy (tinyproxy for example) in the docker compose env we run for the e2e tests and get rid of this custom, perl implemented, proxy we are using. Currently, in addition to the apicast dev container, there is a redis one. 

Same can be done for the camel proxy. 

For backend, upstream, it is very handy to do some asserts so, "outsourcing" upstream may not be convenient. 

wdyt?

gateway/src/resty/resolver/http.lua
```diff
@@ -8,9 +8,23 @@ local _M = setmetatable({}, { __index = resty_http })
...
+  if timeouts then
+    ngx.log(ngx.DEBUG, 'setting timeouts (secs), connect_timeout: ', timeouts.connect_timeout,
+      ' send_timeout: ', timeouts.send_timeout, ' read_timeout: ', timeouts.read_timeout)
+    -- lua-resty-http uses nginx API for lua sockets
+    -- in milliseconds
+    -- https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#tcpsocksettimeouts
+    local connect_timeout = timeouts.connect_timeout and timeouts.connect_timeout * 1000
+    local send_timeout = timeouts.send_timeout and timeouts.send_timeout * 1000
+    local read_timeout = timeouts.read_timeout and timeouts.read_timeout * 1000
+    http:set_timeouts(connect_timeout, send_timeout, read_timeout)
```
tkan145, line 25: Can we handle this inside `gateway/src/resty/http/proxy.lua` connect() method instead? This `new()` return a lua-resty-http instance, so you should be able to call `set_timeouts()` there.

t/apicast-policy-upstream-connection.t
```diff
@@ -59,3 +61,88 @@ GET /
 --- error_log
 upstream timed out
 --- error_code:
+
+=== TEST 2: Set timeouts using HTTPS proxy for backend
```
tkan145, line 65: Can we also add a test case when using `upstream_connection` with camel proxy and with proxy env variables?

t/apicast-policy-upstream-connection.t
```diff
@@ -59,3 +61,88 @@ GET /
...
+      assert.equals(result, "test-upstream.lvh.me:")
+    }
+}
+--- request
+GET /test?user_key=test3
+--- ignore_response
+--- more_headers
+User-Agent: Test::APIcast::Blackbox
+ETag: foobar
+--- error_code:
```
tkan145, line 143: unused error_code

tkan145 (changes requested):

t/apicast-policy-upstream-connection.t
tkan145: I think there are a few tests that watch the output of this proxy module, ie headers.  So we will need to rewrite a few tests, but on the flip side we get more reliable test (may be), ie avoid the case where e2e tests run before the proxy module is ready 

We will need at least 4 containers for both `dev` and `ci` image. 
  *  Normal proxy
  * A camel proxy
  * A camel proxy support HTTPS
  * A proxy that accept BasicAuth

I had a similar problem with #1414 where I need 4 more containers just to be able to test redis sentinel.

As long as it works well and is easy to maintain, I really don't mind adding additional pods.

eguzki: Looks like we have a TODO list. We can implement step by step and remove the perl based proxy when it is fully being replaced by other proxies. 

PS: `dev` and`ci` image are currently the same one

eguzki:
all comments addressed

tkan145:
LGTM!

tkan145 (approved):

PR #1445, opened 2024-02-08
Replace luafilesystem-ffi to fix build issue on aarch64

tkan145:
This is a ~~experimental PR~~, replace the lfs-ffi https://github.com/3scale/luafilesystem-ffi/blob/master/lfs_ffi.lua#L921  with https://github.com/lunarmodules/luafilesystem to ~~hopefully~~ fix the  issue on arm64 https://issues.redhat.com/browse/THREESCALE-10662.

### Verification Steps

The lfs module is mainly used during startup, so the criteria is to have the APIcast start without errors.

```
```
* HTTP dev environment setup
```
cd dev-environments/plain-http-upstream
```

The gateway should be running without error

*  The request should be accepted (200 OK)
```
curl --resolve get.example.com:8080:127.0.0.1 -v "http://get.example.com:8080/?user_key=123"
```


eguzki:
it would be good to have verification steps

eguzki:
Is there an easy way to build ARM docker image on x86 host? 

Is there an easy way to run ARM docker image on x86 host?

tkan145:
Docker supports building arm images on x86 with `-platform linux/arm64`. I haven't tried running the arm image on x86 but I don't think it will work. The easiest way is to run the image inside a VM.

https://docs.docker.com/build/building/multi-platform/

eguzki:
> The easiest way is to run the image inside a VM.

Can you elaborate more on that?

Is this PR still in progress? (the title says "do not merge"). Then, what is missing?

tkan145:
I'm using this on my x86 laptop https://github.com/multiarch/qemu-user-static with QEMU. 

```
$ docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
$ docker run --rm -t arm64v8/ubuntu uname -m
```

This is more like a test PR so @gsaslis can trigger a build for arm64 using code in this branch.  I have provided 2 patches, 1 adds a new structure to `luafilesystem-ffi` and this patch replaces luafilesystem-ffi with another lib.

If the first patch works then great, everyone happy
If that one doesn't work but this one does then we may need to analyze whether this new lib has any impact on performance.

gsaslis:
@tkan145 it looks like some CI checks are still failing. 

Please ping me here once those are taken care of, so I can adapt the product build accordingly. ;)

tkan145:
@gsaslis all the tests passed now :smile:

eguzki (approved):

PR #1453, opened 2024-04-02
[THREESCALE-10934] [3scale_batcher] Update regex to match key with special chars

tkan145:
## What

Fix https://issues.redhat.com/browse/THREESCALE-10934

## Dev notes
In my opinion there are 2 ways to fix this:
1. Replace the current regrex with the regrex from porta repo.  The allowed key formats are describe in the linked JIRA. I don't like this approach because we always have to make sure that the APIcast regular expression matches the one from the porta. Also, the current code allows `set` keys of any format but restricting the format when `get` is also strange to me.
2. Allow everything except spaces and let porta handle the validation during the authorization call. I think this is a better option but would be happy to discuss further.

## Verification Steps 

* Configure a single Product A with user_key 
```
aGVsbG93b3JsZAo=
```
* Configure batcher policy with the following:
```
{ "name" : "apicast.policy.3scale_batcher", "configuration" : { "batch_report_seconds" : 1 } }
```
* Start dev environment
```
make development
make dependencies
```
* Run apicast locally
```
```
* Run query with the valid user_key
```
# capture apicast IP

curl -v -k -H "Host: example.com:443" "http://${APICAST_IP}:8080/?user_key=aGVsbG93b3JsZAo="
```
* Check that `credentials not found` error does not appear in the log. For example:
```
reports_batcher.lua:99: get_all(): failed to get report for key service_id:12,user_key:aGVsbG93b3JsZAo=,metric:Hits err: credentials not found, context: ngx.timer, client: 10.10.10.1, server: 0.0.0.0:8080 
```


eguzki (approved):
Verification steps working

PR #1458, opened 2024-04-18
Update readme

dagbay-rh:
updating readme... pointless update just to kick off a build of main to pull in the latest version of ubi8-minimal into the latest image


dagbay-rh:
pointed to wrong target branch

PR #1490, opened 2024-08-23
remove unused backend client

eguzki:


eguzki:
Broken link issue fixed in https://github.com/3scale/APIcast/pull/1492

eguzki:
@tkan145 is this `____http_call` location being used? maybe for backend?

tkan145:
It's being used [here](https://github.com/3scale/APIcast/blob/master/gateway/src/resty/http_ng/backend/ngx.lua#L26). However, I don't think we are currently using ngx backend, [default](https://github.com/3scale/APIcast/blob/master/gateway/src/resty/http_ng.lua#L247) is resty backend.

tkan145 (approved):
LGTM

PR #1520, opened 2025-01-16
[THREESCALE-11603] Remove opentracing support

tkan145:
## What
https://issues.redhat.com/browse/THREESCALE-11603


.gitmodules
```diff
@@ -139,10 +136,7 @@
 [submodule "dependencies/git/ngx_http_redis"]
 	path = dependencies/git/ngx_http_redis
 	url = https://github.com/osokin/ngx_http_redis
-[submodule "dependencies/git/nginx-opentracing"]
-	path = dependencies/git/nginx-opentracing
-	url = https://github.com/opentracing-contrib/nginx-opentracing
-[submodule "dependencies/git/apicast-nginx-module"]
+[submodule "dependencies/git/apicas-nginx-module"]
```
gsaslis, line 139: (I can't believe I noticed this, but) we have a typo here, right?

```suggestion
[submodule "dependencies/git/apicast-nginx-module"]
```

gsaslis (approved):
Looks Good To Me (LGTM)

.gitmodules
tkan145: yes typo, I haven't checked any Konflux build but the build should fail right :thinking: ?

tkan145: Oh I see, it came from my commit, I was confused by the two block quotes. I need more coffee :coffee:

tkan145: fixed

PR #1521, opened 2025-01-20
Build sync 2.16 branch

tkan145:
## What

Sync current Konflux build with midstream alpha branch


gsaslis:
wow, lots of updates here and I'm not sure I can follow all of them. 

As long as CI is green, I'll say let's go ahead with merging, so we can also verify the downstream build.

gsaslis (approved):

PR #1522, opened 2025-01-20
Konflux build 2.15

tkan145:
## What
Port #1514 so we can build 3scale 2.15 on Konflux. This does not include the jaeger-client-cpp rpm so we still need to figure out what is the best way to build it


gsaslis (approved):
@tkan145 One concern with this PR is whether merging this will affect the 2.15 build on the Continuous Productization as a Service (CPaaS). 

It looks like only new files are being added here, so the build will _probably not break_, as we're not changing files that the product build depends on. 

On the other hand, we'll now have these files in two places - one for CPaaS (in the midstream repo) and one for Konflux (here). This is something to keep in mind, if we need to make changes in either place, as there's no _single source of truth_ and we'll have to port the changes across these 2 streams until we know we can ship 2.15 on konflux.

In all, the changes themselves look good to merge. And after merge, we should verify that both the CPaaS product build still works (taking care to _not_ include it) in 2.15.2. 

What I would recommend is that we change the target branch, so we have one for cpaas and one for konflux. I'll leave the branch name up to you and all we'd then need is to point the konflux build to the new branch. Once all is done, we can retire this other branch (or the cpaas one).

Does that make sense?

tkan145:
@gsaslis yes I understand your concern. Updated to use new branch

PR #1530, opened 2025-01-27
[THREESCALE-6116] Conditional policy: disable UI field

jlledom:
[THREESCALE-6116](https://issues.redhat.com/browse/THREESCALE-6116) is assigned to porta but I think it belongs to here.

I don't think we'll be able to edit the `conditional` policy from UI because this is a corner case where we need to recursively add a jsonschema form inside another. I'm not sure that's possible but in any case it would take much more resources than the issue deserves IMO.

After a discussion in the comments, it was agreed to just add a notice about this policy being managed by operator. In my opinion, the right place to do this is the policy schema in Apicast.

After taking a look, I think we don't actually need to disable the whole policy from UI, only the recursive field `policy_chain`. I tested the `condition` field and it works fine and can be edited from UI without problems.

Since I already made some changes in the policy locally, I thought it would be good to share them.

This is how the form looks now:

image

### DIFF
--- a/doc/policies_list/3.15.0/policies.json
+++ b/doc/policies_list/3.15.0/policies.json
@@ -2726,7 +2726,7 @@
     ],
     "conditional": [
       {
-        "name": "Conditional Policy [Tech preview]",
+        "name": "Conditional Policy",
         "$schema": "http://apicast.io/policy-v1.1/schema#manifest#",
         "configuration": {
           "required": [
@@ -2734,11 +2734,11 @@
           ],
           "properties": {
             "policy_chain": {
-              "items": {
-                "type": "object"
-              },
-              "type": "array",
-              "description": "The policy chain to execute when the condition is true"
+              "type": "object",
+              "description": [
+                "The policy chain to execute when the condition is true. ",
+                "This field is managed by the operator and can't be edited."
+              ]
             },
             "condition": {
               "$ref": "#/definitions/condition"


tkan145:
Hi @jlledom, thanks for working on this. 

`policies_list/x.x.x/policies.json` is something that we will generate right before each release. The actual configure for the policy is [here](https://github.com/3scale/APIcast/blob/master/gateway/src/apicast/policy/conditional/conditional.lua)

> After taking a look, I think we don't actually need to disable the whole policy from UI, only the recursive field policy_chain. I tested the condition field and it works fine and can be edited from UI without problems.

Yes you can modify other fields of the policy, but it is unusable without the nested policy chain, so I think we have a few options:

1. Remove this policy from the admin portal completely (remove apicast-policy.json file from APIcast) and update the documentation to mention that this policy is only configurable via the operator's CR
2. Remove all fields from the jsonschema and add  a notice (update apicast-policy.json file) that this policy is only configurable via operator's CR

Once we agree on how we want to move forward. I will create a new `policies_list/x.x.x/policies.json` for the 2.16 release and that will reflect in the admin portal without any code changes on your part. WDYT?

jlledom:
> `policies_list/x.x.x/policies.json` is something that we will generate right before each release. The actual configure for the policy is [here](https://github.com/3scale/APIcast/blob/master/gateway/src/apicast/policy/conditional/conditional.lua)

Good to know.

> Yes you can modify other fields of the policy, but it is unusable without the nested policy chain, so I think we have a few options:

Yes, the user would need to update policy_chain from the API, but still could modify other fields from UI. Though I understand it can be confusing for the user.

>     1. Remove this policy from the admin portal completely (remove apicast-policy.json file from APIcast) and update the documentation to mention that this policy is only configurable via the operator's CR
> 

I'm not sure who has to decide, but I vote for this.

dfennessy (approved):
/lgtm

tkan145:
thanks @dfennessy 

@jlledom let go with option 1. Do you might to update this PR?  revert this current change and make a new commit deleting the file [here](https://github.com/3scale/APIcast/blob/master/gateway/src/apicast/policy/conditional/apicast-policy.json). Or would you prefer to close this PR and let me handle it?

jlledom:
@tkan145 I'll close this and let you handle it. Thanks!

PR #1572, opened 2026-02-17
THREESCALE-12258 Stream response back when using proxy

tkan145:
## What
[THREESCALE-12258](https://issues.redhat.com/browse/THREESCALE-12258)

In the proxy code, we call `httpc:proxy_response(res)`, lua-resty-http then call `sock:receive(max_chunk_size)`. When `Content-Length` header exist but no "max_chunk_size" passes in, the function will try to read the chunk with the size equal to value of Content-Lenght header (see [here](https://github.com/ledgetech/lua-resty-http/blob/master/lib/resty/http.lua#L525)). Every iteration of the loop, `proxy_response` will then allocates a string with the size of "Content-Length" on the Lua/LuaJIT GC heap.  
                                                                                  
These strings are short-lived (passed to ngx_print and then discarded), but LuaJIT's GC doesn't immediately free them (read more [here](https://github.com/tarantool/tarantool/wiki/LuaJIT-3.0-new-Garbage-Collector#rationale)). With a large response body, the dead strings pile up. LuaJIT's GC is incremental and may not keep up, causing peak memory to be much higher than the actual response body size.

### Before

Version | Memory - 130Mb payload | After request is finished | CPU
-- | -- | -- | --
2.16 | 0.563Gib | 0.298Gib | 5.23m
2.15.3 | 0.718Gib | 0.433Gib | 9.36m
2.14 | 0.714Gib | 0.432Gib | 6m
2.13 | 0.715GiB | 0.440Gib | 7m

### After

Version | Memory - 130Mb payload | After request is finished | CPU
-- | -- | -- | --
2.16 | 0.173Gib | 0.042Gib | 5.23m
2.15 | 0.173Gib | 0.043Gib | 9.36m

**NOTE:** we can see even after this patch, the response is still fully buffered in memory. 

When running on a local machine, I found that responses were sent almost instantly and memory usage remained stable at around ~60MB throughout the request. 

However, when running in a OCP cluster and sending a request from outside the cluster, the response appeared to be buffered and only sent after the entire response had been downloaded.

request with 30M payload
apicast-memory

Requests to the same gateway via a different pod don't seem to cache the response, so I suspect there's something funny with the OCP routing/ingress.

We can call `ngx.flush(true)` immediately following `nginx.print` to flush the buffer. This essentially achieves the same result as setting [proxy_buffering](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffering) to `off`. However, I remain uncertain about the merits of this idea.

The response via our Lua code using proxy policies will have the header `Transfer-Encoding: chunked` (we strip the `Content-Length` header [here](https://github.com/tkan145/APIcast/blob/bfdef91574966fd4e23702215f9cb58d7febec33/gateway/src/resty/http/response_writer.lua#L23))

via nginx

```
< HTTP/1.1 200 OK                        
< Server: openresty                      
< Date: Wed, 11 Mar 2026 02:18:10 GMT    
< Content-Type: application/octet-stream 
< Content-Length: 100000000              
< Connection: keep-alive                 
```

via proxy policies

```
< HTTP/1.1 200 OK                        
< Server: openresty                      
< Content-Type: application/octet-stream 
< Transfer-Encoding: chunked             
< Connection: keep-alive                 
< Date: Wed, 11 Mar 2026 02:16:52 GMT    
```

## Verification steps
* Checkout this branch
* Build a new runtime-image
```
```
* Get inside dev-env
```
```
Modify docker-compose.yaml as follow

```diff
diff --git a/dev-environments/https-proxy-upstream-tlsv1.3/docker-compose.yml b/dev-environments/https-proxy-upstream-tlsv1.3/docker-compose.yml
index 25a49c52..e583db50 100644                                                                                                                 
--- a/dev-environments/https-proxy-upstream-tlsv1.3/docker-compose.yml                                                                          
+++ b/dev-environments/https-proxy-upstream-tlsv1.3/docker-compose.yml                                                                          
@@ -42,6 +42,6 @@ services:                                                                                                                     
     volumes:                                                                                                                                   
       - ./cert/example.com.pem:/etc/pki/example.com.pem                                                                                        
   two.upstream:                                                                                                                                
-    image: quay.io/kuadrant/authorino-examples:talker-api                                                                                      
+    image: quay.io/an_tran/httpbingo:latest                                                                                                    
     expose:                                                                                                                                    
       - "8080"                                                                                                                                 
```
* Generate certs
```
make certs
```
* First run with the latest image
```
make gateway IMAGE_NAME=quay.io/3scale/apicast:latest
```
* In another terminal run
```
docker stats
```
* Send a request
```
curl --resolve get.example.com:8080:127.0.0.1 -v "http://get.example.com:8080/bytes/100000000?user_key="
?user_key=123"
```
Take note of the memory usage
* Stop the gateway
```
CTRL-C
```
* Start the gateway again but with new image
```
```
* Send the request again and check the memory usage, you should see lower memory consumption this time

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Correct FAPI header to `x-fapi-interaction-id` [PR #1557](https://github.com/3scale/APIcast/pull/1557) [THREESCALE-11957](https://issues.redhat.com/browse/THREESCALE-11957)
 - Only validate oidc setting if authentication method is set to oidc [PR #1568](https://github.com/3scale/APIcast/pull/1568) [THREESCALE-11441](https://issues.redhat.com/browse/THREESCALE-11441)
+- Reduce memory consumption when returning large response that has been routed through a proxy server. [PR #1572](https://github.com/3scale/APIcast/pull/1572) [THREESCALE-12258](https://issues.redhat.com/browse/THREESCALE-12258)
 
 ### Added
 - Update APIcast schema manifest [PR #1550](https://github.com/3scale/APIcast/pull/1550)
--- a/gateway/src/apicast/http_proxy.lua
+++ b/gateway/src/apicast/http_proxy.lua
@@ -1,7 +1,9 @@
 local format = string.format
 local tostring = tostring
+local ngx = ngx
 local ngx_get_method = ngx.req.get_method
 local ngx_http_version = ngx.req.http_version
+local ngx_req_get_headers = ngx.req.get_headers
 
 local resty_url = require "resty.url"
 local url_helper = require('resty.url_helper')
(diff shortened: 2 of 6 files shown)


eguzki (approved):
LGTM 🎖️

PR #1575, opened 2026-04-21
THREESCALE-12258 Stream response back when using proxy  (2.16)

tkan145:
## What
Backport #1572 to 2.16


urbanikb (approved):
There is a linter error for 404 document - it's not important to back-port that.

eguzki (approved):

PR #1588, opened 2026-06-08
THREESCALE-10151 fix proxy policy doesn't send headers set by APIcast to the API Backend

tkan145:
## What 

Fix https://redhat.atlassian.net/browse/THREESCALE-10151

## Verification steps:
* Checkout this branch
* Build new runtime-image
```
```
* Get inside dev-envinroment 
```
cd dev-environments/https-proxy-upstream-tlsv1.3/
```
* Update apicast-config.json as follow
```diff
diff --git a/dev-environments/https-proxy-upstream-tlsv1.3/apicast-config.json b/dev-environments/https-proxy-upstream-tlsv1.3/apicast-config.json
index 5227c5aa..bf662f8a 100644
--- a/dev-environments/https-proxy-upstream-tlsv1.3/apicast-config.json
+++ b/dev-environments/https-proxy-upstream-tlsv1.3/apicast-config.json
@@ -4,6 +4,7 @@
       "id": "1",
       "backend_version": "1",
       "proxy": {
+        "secret_token": "token",
         "hosts": ["get.example.com"],
         "api_backend": "https://example.com/get",
         "backend": {
```
* Start the gateway
```
```
* Send a request
```
curl --resolve get.example.com:8080:127.0.0.1 -v "http://get.example.com:8080/?user_key=123"
```
* You should see the secret token in the response
```shell
* Request completely sent off
< HTTP/1.1 200 OK
< Transfer-Encoding: chunked
< Connection: keep-alive
< Date: Mon, 08 Jun 2026 03:19:35 GMT
< Server: WEBrick/1.6.1 (Ruby/2.7.4/2021-07-07)
<
{
  "method": "GET",
  "path": "/get",
  "query_string": "user_key=123",
  "body": "",
  "headers": {
    "X-3scale-Proxy-Secret-Token": "token",
    "User-Agent": "curl/8.15.0",
    "X-3scale-Debug": "",
    "Accept": "*/*",
    "Host": "example.com",
    "X-Real-Ip": "172.20.0.1",
    "Version": "HTTP/1.1"
  },
  "uuid": "c199b276-0856-4929-a0f4-3766d8c97b3b"
```

### DIFF
--- a/gateway/src/apicast/http_proxy.lua
+++ b/gateway/src/apicast/http_proxy.lua
@@ -138,10 +138,15 @@ local function forward_https_request(proxy_uri, uri, proxy_opts)
       end
     end
 
+    local headers = ngx_req_get_headers(0, true)
+    headers["X-Real-IP"] = ngx.var.remote_addr
+    headers["X-3scale-debug"] = nil
+    headers["X-3scale-proxy-secret-token"] = ngx.var.secret_token
+
     local request = {
         uri = uri,
         method = req_method,
-        headers = ngx_req_get_headers(0, true),
+        headers = headers,
         path = (ngx.var.uri or '') .. (ngx.var.is_args or '') .. (ngx.var.query_string or ''),
         body = body,
         proxy_uri = proxy_uri,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Correct FAPI header to `x-fapi-interaction-id` [PR #1557](https://github.com/3scale/APIcast/pull/1557) [THREESCALE-11957](https://issues.redhat.com/browse/THREESCALE-11957)
 - Only validate oidc setting if authentication method is set to oidc [PR #1568](https://github.com/3scale/APIcast/pull/1568) [THREESCALE-11441](https://issues.redhat.com/browse/THREESCALE-11441)
 - Reduce memory consumption when returning large response that has been routed through a proxy server. [PR #1572](https://github.com/3scale/APIcast/pull/1572) [THREESCALE-12258](https://issues.redhat.com/browse/THREESCALE-12258)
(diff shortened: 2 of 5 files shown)


gateway/src/apicast/http_proxy.lua
```diff
@@ -138,10 +138,15 @@ local function forward_https_request(proxy_uri, uri, proxy_opts)
       end
     end
 
+    local headers = ngx_req_get_headers(0, true)
+    headers["X-Real-IP"] = ngx.var.remote_addr
+    headers["X-3scale-debug"] = ""
```
mayorova:
```suggestion
    headers["X-3scale-debug"] = nil
```

The corresponding `proxy_set_header X-3scale-debug "";` does actually **remove** the header if it is present, but in Lua we need to set it to `nil` to remove.
Otherwise, an empty `X-3scale-debug` header will be **added** to the request, which, I understand, is not what we want (because the header is only intended for APIcast itself).

gateway/src/apicast/http_proxy.lua
```diff
@@ -159,6 +164,8 @@ local function forward_https_request(proxy_uri, uri, proxy_opts)
     local res
     res, err = httpc:request(request)
 
+    local inspect = require 'inspect'
```
mayorova: I guess these lines need to be removed as they were probably just added for debugging.

gateway/src/apicast/http_proxy.lua
tkan145: Fixed, thanks

gateway/src/apicast/http_proxy.lua
tkan145: good catch, removed.

gateway/src/apicast/http_proxy.lua
```diff
@@ -138,10 +138,15 @@ local function forward_https_request(proxy_uri, uri, proxy_opts)
       end
     end
 
+    local headers = ngx_req_get_headers(0, true)
+    headers["X-Real-IP"] = ngx.var.remote_addr
+    headers["X-3scale-debug"] = nil
```
mayorova, line 143: I would include another test verifying that when the client request sends `X-3scale-debug: some-value`, the upstream API doesn't receive it.

mayorova (approved):
👍

PR #1593, opened 2026-07-02
Delete openshift template directory

tkan145:
## What

We no longer use template to deploy APIcast so delete the template directory


eguzki (approved):
These PR's are my favorite ones :) 🔥

